PR #3, opened 2019-12-10
Hotfix master

mlandis:
Fix linker error


bjoelle:
@bredelings it's the covarion test - the covarion likelihood is different on master and development because it was fixed on development, but testiphy only has the new version of the test

bjoelle:
at some point, we may want to look at putting the expected values of the tests into the revbayes repo rather than the testiphy one, to avoid these kind of issues ?

bredelings:
If only the covarion test is failing, then the recommended way to handle that is to add a line deleting the test directory:
```
   cd testphy
  rm -rf tests/likelihood/covarion/
```

bredelings:
Hmm... As far as I know, everybody agrees that the covarion model on the master branch is incorrect.  In those circumstances, I don't think you want to the test to succeed on the master branch, I think you want to XFAIL it -- that is, the test fails, but that is OK, so testing succeeds.

This maybe highlights three practical issues:
1. revbayes tests may serve a different function that testiphy tests.  For example, revbayes tests that MCMC takes a certain path are not really testing for correctness, they are just checking that nothing has unexpectedly changed.  testiphy tests, on the other hand, are trying to check things that are correct in a way that is comparable between different programs.
2. in practice, we want a way to avoid tests failing because of testiphy on master when the likelihood is only fixed on development.  To me, this sounds like we want branch-specific XFAILs then.
3. it is possible that different programs would have different ideas as to what the "right" answer is.  This has not happened yet, except for different implementations of gamma binning, but it probably will.

So, the simple way to have branch-specific XFAILs is to put branch-specific lines that delete certain tests into the `.travis.yml`.  This is not beautiful, but it does work.

We could put of list of XFAILED tests in a file like .testiphy.xfail, and delete all the  directories in there.

willpett:
> We could put of list of XFAILED tests in a file like .testiphy.xfail, and delete all the directories in there.

This makes sense to me as a solution.

PR #17, opened 2020-03-10
Generate help database when building with meson

bredelings:


willpett:
Are you actively maintaining/using the meson build system? I wonder how important it is for us to maintain two separate build systems.

bredelings:
I am actively maintaining it.  It is maybe not important.
I feel like the current cmake build system is kind of hacky, so I don't mind maintaining the meson build system so that I don't have to use it :-P.

bredelings:
Additionally, adding additional libraries like beagle to the meson build definitions is quite a bit easier than using cmake.  If at some point when the benefits are clear people still don't switch, then maybe I will give up.

PR #18, opened 2020-03-10
Add some help entries.

bredelings:


help/md/fnHKY.md
```diff
@@ -1,9 +1,24 @@
...
+
 ## description
+The HKY85 model.
+
 ## details
+
 ## authors
+
 ## see_also
+
```
wadedismukes: Could be worth putting in `fnGTR` and `fnHKY` here as well.

help/md/fnGTR.md
```diff
@@ -1,9 +1,20 @@
...
 ## title
+The General Time-Reversible rate matrix
 ## description
+The GTR rate matrix.
 ## details
+The general time-reversible rate matrix:
+  Q(i,j) = exchangeRates(i,j) * baseFrequencies[j]
+The exchangeRates are symmetric.
 ## authors
 ## see_also
```
wadedismukes: Could be worth referencing the other rate matrices in `see_also`.

bredelings:
Thanks for the review!. I wasn't sure which other rate matrices to see_also, and which ones to exclude.  
For nucleotides, we've got fnF81, fnK80, fnK81, fnT92, fnTrN, fnHKY, fnGTR, fnJC, etc.
Do you have an idea for which ones I should mention?
One idea would be for fnHKY to reference fnTrN and vice versa because one specializes the other...

wadedismukes:
> Thanks for the review!. I wasn't sure which other rate matrices to see_also, and which ones to exclude.
> For nucleotides, we've got fnF81, fnK80, fnK81, fnT92, fnTrN, fnHKY, fnGTR, fnJC, etc.
> Do you have an idea for which ones I should mention?
> One idea would be for fnHKY to reference fnTrN and vice versa because one specializes the other...

I think that's a good idea. So we would have for the ones you've done: fnHKY referencing fnTrN (and vice-versa), fnGTR referencing both of these and vice-versa. Or should GTR only reference fnTrN since it is the next level of specialization down? Then we would have:
GTR references TrN (or have GTR reference all the other ones)
TrN references GTR and HKY, K91
HKY references TrN, F81, K80
F81 references HKY, JC
K80 references HKY
Does that make sense?

PR #23, opened 2020-03-11
added doxygen comments to BimodalNormalDistribution.(h/cpp)

spydr073:
Added doxygen style comments to the BimodalNormalDistribution class, and removed tags for author/copywrite/since as outlined in the last MadPhylo video conference.


wadedismukes:
👍

PR #24, opened 2020-03-11
Help

wadedismukes:
Added more help entries


bredelings:
I think maybe you should regenerate the help database as well, instead of just updating the markdown files.  Otherwise, looks good.

PR #26, opened 2020-03-11
Killian dev lognormal comments

spydr073:
Added doxygen style comments to lognormal distribution. Also removed the redundant tags author/license/since etc.


bjoelle:
it looks like this pull request includes changes from your other pull request as well - can you fix that ? otherwise, we have to review the same changes twice
feel free to Slack me if you need further instructions

spydr073:
Yep, just noticed that I forgot to switch to the correct branch before adding... o_O

Would it be easier to kill this pull request and I resubmit a pull request for the clean 'BimodalLognormalDistribution'?

bjoelle:
that would work, yes !

PR #32, opened 2020-03-12
Disable mac travis tests [ci skip]

bredelings:


bredelings:
Unfortunately, disabling the mac checks doesn't pass travis because ... the mac tests won't run.
And [skip ci] doesn't give you a free pass, which is probably good.
Maybe the project owners can circumvent travis here?

PR #41, opened 2020-03-16
Modified the demography functions

rbillenstein:
Added doxygen comments, deleted commented-out code, changed a few comments within the code.
Checking correctness of the doxygen descriptions would be good.
In ExponentialDemographicFunction.cpp and LinearDemographicFunction.cpp one condition was used twice in the if clauses, I added a "@todo" there, so that these conditions can be either deleted or adjusted if they should be a different condition.


rbillenstein:
I opened the issue and will now merge the branch.

PR #42, opened 2020-03-16
Stabilize matrix exponential for RateMatrix_FreeK

bredelings:
Enforce that row probabilities of exp(Q) sum to 1.0 for RateMatrix_FreeK.

This is mostly important for unrealistically large branch lengths.  If the row probabilities for these weird branch lengths sum to 3, then we will get very favorable likelihoods for them, and not reject them.  So, this basically prevents inaccurate matrix exponentials from breaking MCMC.

This seems to be necessary for covarion models with a lot of states.  I think this is kind of orthogonal from making the exponential more accurate, since even more accurate versions can get weird results if the branch length is large enough.

### DIFF
--- a/src/core/datatypes/phylogenetics/ratematrix/RateMatrix_FreeK.cpp
+++ b/src/core/datatypes/phylogenetics/ratematrix/RateMatrix_FreeK.cpp
@@ -235,7 +235,27 @@ void RateMatrix_FreeK::calculateTransitionProbabilities(double startAge, double 
             tiProbsComplexEigens(t, P);
         }
     }
-    
+
+    for(int i=0; i < num_states; i++)
+    {
+        double total = 0;
+        for(int j=0; j < num_states; j++)
+        {
+#ifdef DEBUG_MATRIX_EXPONENTIAL
+            if (P[i][j] < -1.0e-6)
+                std::cerr<<my_method<<" exp(M)["<<i<<","<<j<<"] = "<<P[i][j]<<"\n";
+#endif
+            P[i][j] = std::max(0.0, P[i][j]);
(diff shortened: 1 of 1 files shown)


bredelings:
I could implement some checks that are performed in debug mode.  But, does revbayes have a debug mode?

bredelings:
Hmm... we could add `-DNDEBUG` to the cmake flags in `regenerate.sh` for if `-debug` is not given.

bjoelle:
@bredelings There are debug preprocessor switches in `RbOptions.h` but I'm not sure if they can be set via CMake

bredelings:
For the record, sometimes errors occur only for very large branch lengths that should be rejected.  If these branch lengths cause errors, then MCMC won't work.

bredelings:
@bjoelle I don't think that header really solves the problem.  RevBayes also has some runtime options that can affect the number of digits of precision, for example.  Is there a debug mode there?

bredelings:
See #51

meyerx:
It was just a suggestion for the error check.
If there isn't an error check at runtime and no easy solution to make it happen, so be it.

bredelings:
OK, let's merge this now, and maybe make an issue to remind us to add some logging.  Can someone approve this?

bredelings:
For future reference, I think small negative values happen a lot.  Often then are about -10^-12, so rounding them to zero is not a problem.  I think that the matrix rotation adds positive and negative numbers, so if the true value is near zero, you can't avoid roundoff errors leading to negative numbers.

bredelings:
Its possible to fix the roundoff errors for small values of t by computing exp(Qt)-I using expm1(eigenvalue).

PR #43, opened 2020-03-16
Changed the call for matrix exponentiation method to the one 'owned' …

meyerx:
I encountered stability issues with the stochastic mapping implementation that would result in segmentation fault (for "large FnFreeK matrix").

This small fix changes the matrix exponentiation method from exponentiateMatrixByScalingAndSquaring(...) to calculateTransitionProbabilities(...):

> calculateTransitionProbabilities(startAge, endAge, rate, P);
>//    exponentiateMatrixByScalingAndSquaring(branch_length * rate, P);

The forced use of the scaling and squaring implementation (i.e., exponentiateMatrixByScalingAndSquaring(...)) was unstable for some matrices, while the RateMatrix default method (e.g., tiProbsEigens(..) or tiProbsScalingAndSquaring(...) called by calculateTransitionProbabilities(...)) behaves well.

My understanding is that if the default exponentiation method (the one called by calculateTransitionProbabilities) is safe for likelihood evaluation, it should also be safe for stochastic mapping.

### DIFF
--- a/src/core/datatypes/phylogenetics/ratematrix/AbstractRateMatrix.cpp
+++ b/src/core/datatypes/phylogenetics/ratematrix/AbstractRateMatrix.cpp
@@ -364,8 +364,8 @@ bool AbstractRateMatrix::simulateStochasticMapping(double startAge, double endAg
 
     // transition probabilities
     TransitionProbabilityMatrix P(num_states);
-//    calculateTransitionProbabilities(startAge, endAge, rate, P);
-    exponentiateMatrixByScalingAndSquaring(branch_length * rate, P);
+    calculateTransitionProbabilities(startAge, endAge, rate, P);
+//    exponentiateMatrixByScalingAndSquaring(branch_length * rate, P);
     stochastic_matrix = std::vector<MatrixReal>();
 
     // dominating rate


bredelings:
This might be a good idea, but technically if there's a segfault, then this is hiding another bug.
The program should never segfault even if the matrix exponential is unstable.
Do you know where the segfault is happening?

shiranab:
I did checkout from this branch. When I run it, calculateTransitionProbabilities() goes to the implementation in RateMatrix_Chromosomes.cpp which calls the same exponentiateMatrixByScalingAndSquaring() (in RateMatrix_Chromosomes and not in the RateMatrix).

meyerx:
@bredelings I spent some time to figure out where this bug was coming from and Mike May helped me to do so.

1. It was for a DEC model using the FreeK rate matrix ("large state space") with a _eigen_ has method for matrix exponentiation.
2. The matrix exponentiation using **calculateTransitionProbabilities** would fail. That is the resulting  **P** matrix would filled with junk values (e.g., P[i][j] > 1.E5).
3. The evaluation of the dominating rate would be therefore also non nonsensical.
4. This would lead to the test at line 443 to fail at the first iteration of the loop. I understand this test as a safeguard for the uniformization approach.
> num_events > max_events
5. After the failure of this test, the function (i.e., **AbstractRateMatrix::simulateStochasticMapping**) therefore return with empty/untouched **transition_states** and **transition_times** vectors .
6. The calling function (**simulateStochasticMapping** in class **RateGenerator_Epoch**) would then insert nothing in its local **transition_states** vector (line 285) and go on with its life until reaching line 290. 
7. At line 290, the function expects to have at least one state (the initial one) resulting in a seg. fault.
> std::vector<size_t> save_states(1, transition_states[0]);

I have some trouble wrapping up my head around the test of point 4 and figuring out if sometime it would return false without leading to a seg. fault.
Additionally, I know that @shiranab has spotted some potential "bugs" in the downstream management of *false* returned at this same location.

meyerx:
@shiranab The idea is to delegate the choice of the matrix exponentiation to the one 1) defined by the user or 2) forced by the AbstractRateMatrix implementation/instance.

Are you in your specific case using a **RateMatrix_Chromosomes** ? In which case, as expected, the choice of the matrix exponentiation method is delegated to it and it seems to default to **exponentiateMatrixByScalingAndSquaring** from your message. But that the choice of the model developer or user so it behaves as one would expect. Maybe that calling the PADE version of the squaring and scaling would be safer though.

The line that I modified would force the use of **exponentiateMatrixByScalingAndSquaring** all the time regardless of what users or model developers have specified (if I'm correct).

shiranab:
Okay, you're right and I agree. I also ran several different examples (matrices of different types and data) and I agree that this is the rational thing to do.

PR #44, opened 2020-03-16
Test gcc-5, gcc-6 and gcc-7

bredelings:
Since we have the computing power available, lets test older compilers.

The really interesting ones are probably 4.9 and maybe 4.8, which are installed on clusters, but for those I think I need to build boost with the old ABI.  This is the first step.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,7 +12,10 @@ jobs:
   build:
     strategy:
       matrix:
-        name: [ubuntu-gcc-8,
+        name: [ubuntu-gcc-5,
+               ubuntu-gcc-6,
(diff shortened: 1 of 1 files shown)


bredelings:
Yeah, I think we can, but it is a little more complicated.
* I tried testing gcc-4.9 on ubuntu-16.04, but the boost version was 1.58, and our requirement is 1.60.
* I tried adding a PPA, and that seemed to work, but linking failed because the PPA was compiled using the ABI from gcc-5.0+
So, I think we can add gcc-4.9 by building boost, I just haven't done it yet.
We can either merge this, and then add gcc-4.9, or wait until I add the gcc-4.9 tests to this branch and merge it then.

PR #53, opened 2020-03-19
Allow passing -Darg to cmake.

bredelings:
This removes arguments that start with `-D` from the current argument handling, and passes them to  cmake.  It also prints how and where cmake is called.

Passing args to cmake allows doing things like 
* `./build.sh -DBoost_DEBUG=ON`
* `./build.sh -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang`
and many other things.  We then have the option of making future options to `./build.sh` be cmake options instead of options to generate different scripts.

### DIFF
--- a/projects/cmake/build.sh
+++ b/projects/cmake/build.sh
@@ -4,6 +4,7 @@ set -e
 all_args="$@"
 mpi="false"
 travis="false"
+cmake_args=""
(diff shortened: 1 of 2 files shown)


wrightaprilm:
Works on Ubuntu 19.04.

PR #55, opened 2020-03-19
fixes to how fossil tips are initialized in CDBDP

mikeryanmay:
There appears to be a problem with how the initial probability for fossil tips are initialized in the character-dependent birth-death models. My fix is in line with how this initialization is done for constant-rate models, as in Stadler (2010) JTB, g_e(s_1) [page 399, bottom right], and for state-dependent models, as in Kuhnert et al. (2016) MBE, g_{i,k}^e(t_e) [page 2015, top].

I've attached a script to demonstrate the problem by comparing likelihoods between the constant-rate FBD and the CDBDP (with rates that are the same for each state). Running this script with the current implementation will show that the likelihoods are different, and with the proposed fix that the likelihoods are the same. (I've changed the file extensions so that GitHub will allow me to attach them to this pull request!)

[comparison.txt](https://github.com/revbayes/revbayes/files/4357285/comparison.txt)
[tree.txt](https://github.com/revbayes/revbayes/files/4357286/tree.txt)


hoehna:
Looks good to me.

Issue #50, opened 2020-03-17
Paths cannot be copied from the tutorials in Rev


bjoelle:
due to path separators being different between Unix OSes and Windows

wrightaprilm:
As a note: If you're on windows and launch via GitBash or a similar terminal emulator, *NIX style paths will work. Except the `~`. If you click the executable on Windows to start RevBayes, you should be able to used mac-style relative paths once your path is set. For example, this morning, I set my path like so:

```
setwd('c:\\Users\\april\\revbayes\\')
```

And afterwards could do:

```
source("scripts/mcmc_example.Rev")
``` 

And this behavior varies between Windows machines, as well. Some will allow NIX-style separators, some will not. It's very frustrating in my systematics class, which has 4 different Windows OSes present. Given that relative paths seem to work NIX-style once the working directory is set, perhaps pointing users to the [setup tutorial](https://revbayes.github.io/tutorials/tutorial_structure/) is the way to go?

davidcerny:
This also seems to have been resolved; for v1.2.4 on Windows 10 Pro for Workstations (v22H2), I can not only use relative paths with Unix separators (like `source("scripts/mcmc_example.Rev")`), but also this:

```
setwd("C:/Users/Graham/Downloads/")
```

I didn't have to install any extra software, just double-click the executable. @bjoelle Would you be okay with closing this?

bjoelle:
yes, this is resolved I think

PR #62, opened 2020-04-10
Build and test Windows using meson cross compile and wine

njbooher:
It builds. Caching is currently off. And tests are failing for some reason.

Making the PR to request @bredelings advice on cleaning it up.


bjoelle:
The test failures could be tied to the path being different in Windows ? - see issue #50

njbooher:
Ah. That would make sense. Will worry about it next week.

bredelings:
Woah, cool stuff!  I haven't looked at the cross-build stuff in a while, I hope the `make_winroot.sh` script was useful.  I think meson may have gotten a bit more proficient in handling pkg-config for cross-compiles, but I haven't had a chance to look at that yet.

I was going to look into running a build _on windows_, but didn't get too far.  There are a few  issues about this:
https://github.com/actions/virtual-environments/issues/30
https://github.com/actions/starter-workflows/issues/95

There is also this "action":
https://github.com/numworks/setup-msys2

njbooher:
The windows version apparently defaults the working directory to the path the executable lives in rather than the directory you run it from in the shell.

https://github.com/revbayes/revbayes/blob/9a1fcdf469613c494041a2d7e38a583c050f75a4/src/core/utils/RbSettings.cpp

njbooher:
Well, at least the tests run now.

bredelings:
I feel like I remember problems making my laptop run tests on other tools with wine due to DLL issues.

njbooher:
I worked around the default working directory difference by copying the executable to the test directory for each test. The DLL issues mentioned were just me forgetting that executable depended on dlls in the directory I copied it from, so now it's copying those too. It's a massive hack, but works for now.

The tests run now. Initially they failed due to line endings differences in the output (\r\n vs \n), but that's now fixed. Now several of them pass. The current failures seem to be due to differences in floating point formatting. (e.g. 6.96098e-005 vs 6.96098e-05). I changed the actions config to upload the output of the tests when they fail as an artifact. You can view it if you want by going to [the job log page](https://github.com/revbayes/revbayes/pull/62/checks?check_run_id=586429813), clicking 'Artifacts', then 'test-results' in the dropdown.

njbooher:
Tests seem to be passing now.

PR #63, opened 2020-04-10
Dev huelsenbeck

jhuelsenbeck:
Added a high-quality header-only JSON library that we use in the Mac GUI, but that could also be quite useful elsewhere in the program. Currently, we use JSON in one part of the core code, using boost to parse the JSON. Using this library reduces our dependency on boost.


bredelings:
This PR currently contains two commits, one which adds the json library and one of which adds the file `src/revlanguage/Info.plist`.  Is commit with Info.plist a mistake?  At least it doesn't seem like it should be part of adding the json library.

bredelings:
I was going to suggest putting the library in `src/libs/nlohmann/nlohmann/json.hpp`.  This way you could add `src/libs/nlohmann` to the search path for include files, and include "nlohmann/json.hpp".

However, since revbayes currently puts _every_ subdirectory into the search path, you _could_ put this library anywhere and then include just "json.hpp".  I'm curious if anyone else has an opinion?

jhuelsenbeck:
Could somebody approve this so I can get on with my life? Sheesh.

bjoelle:
@jhuelsenbeck can you remove the `src/revlanguage/Info.plist` file ? As Ben said it appears to be an IDE-generated file and not part of the JSON library.
You can add more commits to this branch and they will be added directly to the PR.

PR #66, opened 2020-05-05
Issue and PR templates

wrightaprilm:
I added 3 new issue and 3 new PR templates. They're a bit verbose, happy to edit.


bjoelle:
@wrightaprilm I think these look good. 
The one thing I'm not sure about is whether to keep the computer information as part of the pull request templates, since pull requests have to go through automated tests on all systems anyway. What do you think ?

wrightaprilm:
Honestly, I don't know enough about how the build system works to say either way. If you think it's unnecessary, that seems good.

PR #80, opened 2020-06-16
Dev boussau

Boussau:
I have implemented a few new functions in RlTree that are based on corresponding functions in Tree. In particular we can renumber the nodes of a tree as in a reference tree, and we can reroot a tree by making a bifurcation at the root. I have tested the functions through the revlanguage and as far as I can tell they seem to work!


Boussau:
Hi Will, Thanks for taking the time to look at my code! 
My hope was that the existing functions in the revlanguage would do what I needed, which is to get a tree rooted and bifurcated at the root.
But that did not work.
The reason I implemented this function is that in revbayes one can have a branch length tree with a trifurcation at the root node. There is a root function: it moves the "root" to another node, but won't make the tree bifurcating at the root. 
Once there is a trifurcation (A, B,C), there are several ways to resolve it into a bifurcation: (A, (B,C)) or ((A,B),C)... I wanted a specific way, which means the make_bifurcated needs to basically know which outgroup one wants to use. So in the end, that would mean 2 functions with the same arguments: first the rooting function, then the make bifurcated function. That's why I thought it better to have one function do it all at once, and keep the pre-existing functions, which do their own things and may be useful to someone else...
I hope this is clear...

willpett:
I see, okay that makes sense. So does the existing `make_bifurcating` function just resolve the trifurcation arbitrarily? It seems to me that if you modify `make_bifurcating` so it can accept an optional clade argument specifying how the trifurcation is resolved, then you could use the existing `root` method followed by this modified `make_bifurcating' method. In this way a new combined method is not needed, or am I missing something? I just seems confusing to me to have two `make_bifurcating` functions that may not resolve trifurcations the same way.

Boussau:
make_bifurcating is a strange function: it actually only handles nodes of degree 1 (and makes fossils with them, if I remember well), not nodes of degree 3 or above. So if you give it a tree with a trifurcation, it will return a tree with a trifurcation. Again, I did not want to mess up someone else's code, so I did not change this function.

PR #81, opened 2020-06-23
fixed memory leak in homeolog phasing

wf8:
simple fix. Thank you!


bjoelle:
@wf8 The actual fix looks good to me but your pull request also contains a lot of changes that revert recent changes to `development` - can you fix that ?

wf8:
thank you @bjoelle for catching my mistake, should be fixed now.

PR #82, opened 2020-07-30
Fix boost location hints for cmake.

bredelings:
This branch fixes the fact that when using `build.sh` with `-boost_root` or `-boost_lib`, any previous options got erased.

It also:
* uses BOOST_LIBRARYDIR instead of BOOST_LIBRARY, which is not correct according to the docs
* cleans up and removes the variables LOCAL_BOOST_ROOT/LOCAL_BOOST_LIBRARY
* adds more examples to `build.sh -h` to show passing cmake variables directly.
* clarifies BOOST diagnostic messages printed by cmake.


bredelings:
@hoehna I changed that in the refactor-cmake branch back in March.  That particular cmake file is not autogenerated, although others are.

Issue #83, opened 2020-07-31
Majority-rule consensus of unrooted trees produced a rooted tree


zhangchicool:
For the discrete morphological matrix attached (data/50discrete.nex), and i believe for nonclock analyses in general, the consensusTree() function in RevBayes produced a rooted tree, with clade probability 1.0 for t1 (output_nonclock/disc_Mkv.maj.tre). It seems the consensusTree() function treated t1 as a hard outgroup constraint, but I did not specify any outgroup.  The majority-rule consensus tree should be unrooted in this nonclock analysis. 
In comparison, MrBayes produced an unrooted consensus tree as expected (mb-nonclock/*).

I compiled the code from source in the master branch (I tried the pre-compiled version 1.1.0 too).  My system is macOS Catalina (10.15.6) with Xcode 11.6 and Apple clang version 11.0.3. I installed cmake and boost using the latest homebrew. I hope the attached files are sufficient to  reproduce the described phenomena. If you need anything more let me know. Thanks.

[tests-cons.zip](https://github.com/revbayes/revbayes/files/5004521/tests-cons.zip)

willpett:
It seems that this issue is solved by #80

PR #87, opened 2020-08-04
Removed default outgroup from consensus tree

bjoelle:
Addresses issue #83 by removing the default outgroup from consensus tree building.
Default outgroup is still used to summarize clades, but the final tree is unrooted.


willpett:
After some checking, it seems that the changes in #80 solve #83, so I will close this pull request and merge #80

Issue #86, opened 2020-07-31
setNumStatesPartition function partitions alignment based on the maximum state, not the total number


bjoelle:
which leads to confusing numbers being reported to users, see issue #84

wrightaprilm:
This should be easy to fix. I'll take a look at it Weds.

willpett:
Did we decide what the appropriate behavior should be here? I am inclined to leave it as is, since I think it is fair to expect the user to recode the data if the maximum state does not reflect the size of the state space.

The only weird thing is when you have a constant character with all 0's. Those get partitioned as unary characters... which I don't think is ever useful?

bjoelle:
@willpett I thought we had settled on using the character description if available, and otherwise use the max number of states but print a warning ?

willpett:
I don't know how I feel about printing a warning. I would just put it in the documentation. Is there another software with a similar function that behaves a different way? It's not obvious to me that the user should expect one or the other.

bjoelle:
the issue is that there is currently no help files for internal functions, so the current behaviour is leading to confusion for the user
we could remove the warning once the help has been updated ?

willpett:
Yeah that makes sense.

wrightaprilm:
I think the behavior is rational.  I'm working on modifying my PR to only give the warning if the issue is actually noted. So, if you've done something like code 4 and 5, but not 0, 1, 2 or 3. But today is my heavy teaching day, so I'm unlikely to get to it until Friday. Anyone else is welcome to take a stab at moving the warning if they have the time.

willpett:
Yeah I can do that.

dwbapst:
Just chiming in here from the paleontology/R-side, warnings are a good thing exactly for the situation April mentions.

I think one would find there's lots of datasets where binary characters are probably coded 1-2, rather than 0-1, and so a worker is likely to forget to convert all of them. Might want to think about a specific warning to just remind users to start the numbering at 0 when 0 is missing.

Another scenario, one I see quite often, is where a character is coded (1,2,3,6), where the worker skipped zero *and* defined a multistate character and then realized when they were done that two of the states (4,5) which they defined never applied to any of the taxa they ended up using...

willpett:
All fair points! I'll close this issue for now.

PR #88, opened 2020-08-07
Update user messaging regarding setNumStates*

wrightaprilm:
In response to #86, we now have a message print to screen noting the behavior, with an example.


wrightaprilm:
Noted that I hadn't done a merge on my branch. Will fix the long commit history on mine and reopen.

PR #89, opened 2020-08-07
add user message to setNumStates*

wrightaprilm:
Pursuant to #86, explain behavior of the `setNumStates*` methods so users aren't caught off-guard.


willpett:
Ok, until we get the help pages for member methods working, I think the warning is a good idea. I made it a little more brief (also using RBOUT), and I hope it's okay that I renamed your method to `getNumStatesVector`, since I think that's what you meant?

wrightaprilm:
Can we leave the name as `setNumStatesVector()` for continuity between the two?

willpett:
Well, it is a getter function not a setter.

Issue #85, opened 2020-07-31
Large positive log likelihoods running continuous character matrix


zhangchicool:
For the continuous traits matrix (data/50traits.nex) and a nonclock analysis, RevBayes produced large positive log likelihoods, and the likelihoods kept increase as the MCMC continued (output_nonclock/cont_uBM.log). I would expect negative values of log likelihoods to be reasonable. 

Together with issue #84 , it prevents me from running a combined data analysis, as the combined analysis (scripts/mcmc_ncl_combine.rev) produced inconsistent likelihoods between runs (output_nonclock/comb_Mkv_uBM*). I believe the results are not correct.

I compiled the code from source in the master branch (I tried the pre-compiled version 1.1.0 too).  My system is macOS Catalina (10.15.6) with Xcode 11.6 and Apple clang version 11.0.3. I installed cmake and boost using the latest homebrew. I hope the attached files are sufficient to  reproduce the described phenomena. If you need anything more let me know. Thanks.

[test-cont.zip](https://github.com/revbayes/revbayes/files/5004596/test-cont.zip)

PR #94, opened 2020-08-22
Fix likelihood bug in PhyloBrownianProcessREML

willpett:
A bug in `PhyloBrownianProcessREML` erroneously updated the root mean estimate before it was used in the computation of the root contrast, causing the likelihood to be computed incorrectly for unrooted trees. This solves #85.


hoehna (dismissed):
I think the check for zero branch length was necessary for some situations before. Simply removing might be dangerous.

Good catch for the update of the root mean. Since we are fixing this specific bug, we should also fix the exact same issue in:
- PhyloBrownianProcessMultiSampleREML
- PhyloMultiSampleOrnsteinUhlenbeckProcessREML
- PhyloOrnsteinUhlenbeckREML
possibly also for
- PhyloMultivariateBrownianProcessREML
- PhyloMultivariateBrownianProcessMultiSampleREML

willpett:
Thanks for taking a look!
- If a check for zero-length branch lengths is necessary, I think it should be done at the likelihood computation level (i.e. return neginf) instead of arbitrarily setting the branch length to 1e-16.
- I will go ahead and fix the same bug in other models.

willpett:
This doesn't seem to be an issue for the OU REML classes.

willpett:
I checked all the occurrences of `computeBranchTime` and I don't see any obvious reason why a zero-length branch should cause an issue. If there is an instance where it causes a problem, I think implementing a check in the affected model is the proper way to handle it, which is beyond the scope of this pull request. More integration tests for continuous character models are needed.

willpett (commented):
Looks ok to me.

bredelings:
> I checked all the occurrences of `computeBranchTime` and I don't see any obvious reason why a zero-length branch should cause an issue. If there is an instance where it causes a problem, I think implementing a check in the affected model and either throwing an exception or returning `-inf` is the proper way to handle it, which is beyond the scope of this pull request. In the meantime, any hypothetical problem will be much easier to identify if it manifests as a crash/error rather than an incorrectly sampled posterior, so I think removing the check is appropriate.

That seems reasonable to me...

bjoelle (approved):

PR #98, opened 2020-08-24
wadedismukes/issue96

wadedismukes:
---
name: Bug Fix
about: Have you fixed a bug in RevBayes?
title: "Added exception message to Monitor Screen when generations = 0"
labels: ''
assignees: ''

---

Thank you for contributing a bug fix to RevBayes!

**Describe the bug**
When number of generations is set to 0 for monitor screen the system throws the following error: `Floating point exception: 8`. This is non-informative .

### DIFF
--- a/src/revlanguage/datatypes/basic/Integer.cpp
+++ b/src/revlanguage/datatypes/basic/Integer.cpp
@@ -2,6 +2,7 @@
 #include <string>
 #include <vector>
 
+#include "IntegerPos.h"
 #include "RlBoolean.h"
 #include "Integer.h"
 #include "Natural.h"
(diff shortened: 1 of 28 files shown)


willpett:
I thought a way to solve this might be to define a new datatype `IntegerPos` or something and make that the type of the argument. That way you'd get the generic argument mismatch error instead of having to write a custom error.

wadedismukes:
> I thought a way to solve this might be to define a new datatype `IntegerPos` or something and make that the type of the argument. That way you'd get the generic argument mismatch error instead of having to write a custom error.

Yeah that is definitely a more elegant solution. I will get to work on that. I was actually surprised to see that natural numbers include 0 in Revbayes! I learned that 0 is not a natural number. But yeah I can do that Will!

willpett:
Well I don't know. Your solution is definitely easier. It was just something that occurred to me. Are there other situations where `IntegerPos` would be useful? I suppose there are probably some distributions with that as their domain?

wadedismukes:
Yeah maybe like a truncated poisson? I can't think of any others but can't hurt to have more options. This is likely not the only place this shows up.

bredelings:
I would just comment that the other solution is to eliminate RealPos as a type.  In my opinion, range-checking like this doesn't work at the type level.  Carry on.

willpett:
Probably want to update other monitor types to use `IntegerPos` also.

PR #100, opened 2020-08-28
Copy emit_letters in RateMatrix_FreeK copy construction

willpett:
The member variable `emit_letters` was not copied during copy construction of `RateMatrix_FreeK`, causing a crash with printing the value of the rate matrix.


mlandis:
Looks appropriate

Issue #14, opened 2020-03-04
Segfault with deterministic vector / Yang & Rannala Dirichlet prior


bredelings:
Hi,
I found a segfault trying to generate branch lengths using the Yang & Rannala Dirichlet prior.
```
n_branches = 24

total ~ dnExponential(1.0)
frac ~ dnDirichlet( rep(1.0, n_branches) )
for (i in 1:n_branches)
{
    bl[i] := total * frac[i]
}
frac
sum(frac)
bl
sum(bl)
```

This segfaults with the last release, and also with fix-tutorials (ca77f2) from Fri Feb 21.

-BenRI

willpett:
Here's a reduced version that produces the same error:
```
frac ~ dnDirichlet( rep(1.0, 2) )
bl := frac[1]
sum(frac)
bl
```
Currently tracking down the segfault...

willpett:
Apparently, this is related to the fact that RevBayes resolves function names differently depending on whether you call them with an assignment operator.

Specifically, `sum(frac)` executes without an error, but `a:=sum(frac)` gives an error that `sum` is ambiguous because it isn't defined for a `Simplex` argument.

I suspect that the fact such a simplex sum function doesn't exists causes `sum(frac)` to corrupt `frac` somehow, such that referring to it later causes a segfault.

My bandaid solution was to define a simplex sum function. However, it may be worth investigating how RevBayes resolves function names in these cases to avoid further bugs in the future.

bredelings:
I took at look at this in gdb, and it looks to me like it might result from a weird dynamic_cast?

Specifically, the crash happens in this function:

``` C++
template <class memberObjectType, class valueType>                                                                                                         
void RevBayesCore::MemberFunction<memberObjectType,valueType>::update( void )                                                                               
{                                                                                                                                                                                                                                                                                                                    
    const MemberObject<valueType>& the_member_object = dynamic_cast<const MemberObject<valueType>& >( the_member_variable->getValue() );                    │
     the_member_object.executeMethod(method_name,args,*this->value);                                                                                         }
```
Here we have
* `this->dag_node->name` = `frac[1]`
* `the_member_variable` points to `frac`
* `method_name` = "[]"
* `*this->value` =  `(double&) 0.22987656677307056`
So, it looks like its trying to call the subscript method on `frac` to get `frac[1]`.

The SEGV occurs when trying to call `the_member_object.executeMethod(...)`.  Why is that?
Well `the_member_variable->getValue()` yields

[...]

PR #102, opened 2020-08-29
Make ModelVector and Simplex use [] the same way

willpett:
`ModelVector` implements the `[]` method as a `MemberProcedure`, but `Simplex` implements it as a `MemberFunction`, executing the method on the core side via the `MemberObject` interface of `RevBayesCore::Simplex`. The existence of two different implementations caused a crash when converting variables between the two types (see #14).

- Implements `[]` as a `MemberProcedure` in `Simplex`
- Removes `Func_sumSimplex` which was put in place as a stop gap when the original issue popped up.


bredelings:
This seems like a better fix that previously, but I still don't completely understand it.  Is it the case that when the type of a Simplex is converted to a ModelVector, that the value is then dynamically casted to a C++ RbVector even though it was created as a C++ Simplex object?

And then the goal of this patch is to make various operations from ModelVector work on a Simplex object that is interpreted as a ModelVector?

willpett:
The goal is to specifically make `[]` work on a `RevLanguage::Simplex` object interpreted as a `RevLanguage::ModelVector`. Other methods should be unaffected. When the core value was cast to a `RevBayesCore::RbVector`, it was losing the `RevBayesCore::MemberObject` implementation of `RevBayesCore::Simplex` where `[]` was defined.

bjoelle (approved):

Issue #93, opened 2020-08-21
Rate matrix functions allow non-simplex stationary frequencies


afmagee:
One can construct rate matrices from stationary frequencies that do not sum to 1. The resulting Q matrices appear to use the incorrect values, and the resulting transition probabilities are therefore different. The "simplex" base frequencies seem to be allowed to be any values in [0,1].

The following code demonstrates the problem using F81, but it has been observed in WAG, F81, HKY, and GTR. 

```
# RevBayes version (1.1.0)
# Build from dev-hoehna-lab (rapture-539-g6b163d) on Mon Aug 17 18:09:39 PDT 2020

### Rate matrices

fnF81(rep(0.25,4))
# produces this (correct) rate matrix:
#   [ [ -1.0000, 0.3333, 0.3333, 0.3333 ] ,
#     [ 0.3333, -1.0000, 0.3333, 0.3333 ] ,
#     [ 0.3333, 0.3333, -1.0000, 0.3333 ] ,
#     [ 0.3333, 0.3333, 0.3333, -1.0000 ] ]

fnF81(rep(1.0,4))
# should throw error but instead produces this rate matrix:
#   [ [ -0.2500, 0.0833, 0.0833, 0.0833 ] ,
#     [ 0.0833, -0.2500, 0.0833, 0.0833 ] ,
#     [ 0.0833, 0.0833, -0.2500, 0.0833 ] ,
#     [ 0.0833, 0.0833, 0.0833, -0.2500 ] ]

fnF81(rep(0.0,4)):
# should throw error but instead produces this rate matrix:
#   [ [ -nan, -nan, -nan, -nan ] ,
#     [ -nan, -nan, -nan, -nan ] ,
#     [ -nan, -nan, -nan, -nan ] ,
#     [ -nan, -nan, -nan, -nan ] ]

### Transition probability matrices

fnF81(rep(0.25,4)).getTransitionProbabilities(0.1)
# produces this (correct) output:
# [ [ 0.906, 0.031, 0.031, 0.031 ], [ 0.031, 0.906, 0.031, 0.031 ], [ 0.031, 0.031, 0.906, 0.031 ], [ 0.031, 0.031, 0.031, 0.906 ] ]

fnF81(rep(1.0,4)).getTransitionProbabilities(0.1)
# produces this output:
[...]

Issue #101, opened 2020-08-29
Static type conversion disconnects variables from the DAG


willpett:
A variable whose type doesn't match a function argument exactly may sometimes be automatically converted to the appropriate type. The result of this type conversion is a new constant node with the same value as the original variable. When passing arguments by value, this works fine. However, when passing arguments by reference, this causes a problem, because replacing a variable with a constant will disconnect the variable from its original DAG node.

As an example, a `Natural` may always be converted to a `Real`. Therefore, the following code executes without error:
```
mu ~ dnPoisson(1)
x ~ dnNormal(mu, 1)
```
However, if you examine `mu`, which was a `Natural<stochastic>`, you'll see it has been replaced with a `Real<constant>`:
```
> str(mu)

   _variable     = mu
   _RevType      = Real
   _value        = 1
   _dagType      = Constant node
   _children     = [ x ]
```

In general, the rules for type conversion are fairly strict, so this doesn't come up very often.

However, when a function is called in a "static" context (either as part of a constant node assignment, or as a bare function call), the rules for type conversion are much more relaxed. This can cause all sorts of strange behavior.

For example, the following throws an error:
```
x ~ dnNormal(0,1)
sqrt(x)
mvSlide(x)
```
Here, `x` is a `Real<stochastic>` but has been converted to a `RealPos<constant>` by `sqrt`.

I'm not sure what the solution to this problem is. One idea is to simply hand off the DAG node to the new variable during conversion. But then the DAG node's value must also be converted, which might cause problems (e.g. with `Simplex` to `RealPos[]`, see #14). And anyway, the variable's type is permanently converted, which could cause other downstream side effects.

Maybe something along the lines of what @bredelings suggested in #99 would work, where all type conversion happens through deterministic conversion nodes? Or possibly relying on dedicated internal functions to do the domain checking.

bredelings:
This issue seems helpful.  Regardless what solution is eventually chosen, it might be worth writing a document that 
* clearly specifies how the type system affects execution, and is affected by it.
* clearly specifies what the type system is supposed to allow and prohibit (e.g. `x ~ dnNormal(0,1) ; y ~ dnGamma(x,x)` would presumably be prohibited)

PR #103, opened 2020-08-29
Make sure ModelVector is normalized when converting to Simplex

willpett:
When testing whether a `ModelVector` can be converted a `Simplex`, the only check in place was that each element is convertible to a `Probability`. This implements a check to make sure the vector is also normalized. This solves #93.


willpett:
Looks like this is failing in `testiphy` because apparently the vector `freqs = [0.4,0.3,0.2,0.1]` doesn't sum to 1.0. Unfortunately, due to floating point representation, there no guarantee that a vector will sum to 1.0 internally... So a vector may fail to convert to `Simplex` even if it sums to 1.0 in the Rev script.

bredelings:
So, whereas #102 is designed to make conversions from ModelVector to Simplex work by deriving Simplex from ModelVector only, this PR is designed to make conversions from ModelVector to Simplex work by checking that the ModelVector has entries that sum to 1?

bredelings:
Is the ModelVector value actually replaced with a Simplex value?  Or is ModelVector dynamic_cast-ed to Simplex?

willpett:
Whether a variable is actually replaced with its converted value depends on the context (see #101). This PR just allows ModelVector to be used when a Simplex is required under the condition that its values are normalized.

bjoelle (approved):

Issue #106, opened 2020-09-04
Debug build problem in Ubuntu


brpetrucci:
In attempting to set up RevBayes with debugging for CLion, I stumbled upon a problem. After cloning from the repo as described in [the Downloads page](https://revbayes.github.io/compile-linux), I attempted to run 
```
./build.sh -debug true
```
(alternatively, I also tried modifying `build.sh` to have `true` as the default for debug, and got the same result). This led to an error in the build. I'll attach the entire build log for reference - the error occurs around 51%. This error was reproduced for Ubuntu 20.04, Ubuntu 18.04 and Red Hat Enterprise Linux 7.6. It appears the debug build does not run well in Linux. The release build (i.e. `./build.sh` with no extra arguments) builds correctly.
[build_error.txt](https://github.com/revbayes/revbayes/files/5177253/build_error.txt)

bredelings:
I don't have time to look further right now, but it looks like there is a compilation error:

```
In file included from /home/petrucci/revbayes/src/core/distributions/phylogenetics/substitution/PhyloCTMCSiteHomogeneousNucleotide.h:51,
                 from /home/petrucci/revbayes/src/revlanguage/distributions/phylogenetics/character/Dist_phyloCTMC.cpp:9:
/usr/lib/gcc/x86_64-linux-gnu/9/include/pmmintrin.h: In member function ‘void RevBayesCore::PhyloCTMCSiteHomogeneousNucleotide<charType>::computeInternalNodeLikelihood(const RevBayesCore::TopologyNode&, size_t, size_t, size_t, size_t) [with charType = RevBayesCore::DnaState]’:
/usr/lib/gcc/x86_64-linux-gnu/9/include/pmmintrin.h:86:1: error: inlining failed in call to always_inline ‘__m128d _mm_hadd_pd(__m128d, __m128d)’: target specific option mismatch
   86 | _mm_hadd_pd (__m128d __X, __m128d __Y)
      | ^~~~~~~~~~~
In file included from /home/petrucci/revbayes/src/revlanguage/distributions/phylogenetics/character/Dist_phyloCTMC.cpp:9:
/home/petrucci/revbayes/src/core/distributions/phylogenetics/substitution/PhyloCTMCSiteHomogeneousNucleotide.h:483:37: note: called from here
  483 |             __m128d gt = _mm_hadd_pd(g_acgt,t_acgt);
```

brpetrucci:
[...]

PR #107, opened 2020-09-05
Enable sse3 flags even on debug builds to avoid compile failure.

bredelings:


bredelings:
This is supposed to fix  #106 
In my machine (with gcc-10) I am not getting an error even without the fix, maybe because the default -march has sse3 support enabled?

willpett (approved):
Worth a try!

PR #109, opened 2020-10-12
Added changelog

bjoelle:
to make releases easier and provide a record of improvements made

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,0 +1,11 @@
+### Current changes
+
+#### New models/analyses
+
+#### New features
+
+#### Bug fixes
+
+ * fixed an error in the likelihood calculation for discrete characters when using ambiguous characters and partitioning standard data by state number
+ * fixed a crash when using ordered rate matrices
+ * fixed a crash when printing the value of a rate matrix


wadedismukes (commented):
I think we should add a release action to workflow that adds this to the release notes. If you disagree it looks good otherwise.

wadedismukes:
> I think we should add a release action to workflow that adds this to the release notes. If you disagree it looks good otherwise.
See this action below:
https://github.com/actions/create-release

Basically we just make it happen whenever we merge development into master. (rather than on a schedule or on push or something.

bjoelle:
@wadedismukes adding it to the release notes is a good idea for sure !
as I remember the release process was mostly manual last time, but we can take another look at automating it for the next one
I was also planning to link to that file from the website once it's accepted

wadedismukes (approved):
Other than the above comments looks good (I will add a tab in Projects for the stuff we talked about as it is not pressing).

wadedismukes:
> @wadedismukes adding it to the release notes is a good idea for sure !
> as I remember the release process was mostly manual last time, but we can take another look at automating it for the next one
> I was also planning to link to that file from the website once it's accepted

Oh that sounds good. I like the idea of having it on the website. Not sure why I didn't merge this before.

PR #110, opened 2020-10-13
Notify Slack on result of github-actions

bjoelle:
Will send a notification to the #github channel (one notification per commit)


wadedismukes (commented):
Looks good.

wadedismukes (approved):

PR #112, opened 2020-10-15
removing deprecated set-env

bjoelle:
set-env is deprecated and will be removed eventually (see https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/)


wadedismukes (approved):
This looks good. Glad they deprecated the `::set-env` its a little cryptic.

PR #113, opened 2020-10-21
Fixed mac build on github actions

bjoelle:
Temporary workaround (see issue https://github.com/actions/virtual-environments/issues/1811)


wadedismukes (approved):
Glad this got fixed. Maybe we should keep an issue to keep an eye on the temporary workaround to change it when it is fixed completely? Otherwise good to go.

PR #122, opened 2020-12-17
Add a Doublet alphabet

bredelings:
Initial stab at a doublet alphabet 
[mcmc_GTR.Rev.txt](https://github.com/revbayes/revbayes/files/5707272/mcmc_GTR.Rev.txt)

### DIFF
--- a/src/core/datatypes/phylogenetics/character/CharacterTranslator.cpp
+++ b/src/core/datatypes/phylogenetics/character/CharacterTranslator.cpp
@@ -88,7 +88,7 @@ RevBayesCore::AbstractDiscreteTaxonData* CharacterTranslator::translateCharacter
         }
         else
         {
-            trans_taxon_data = translateToCodon( d );
+            trans_taxon_data = translateToDoublet( d );
         }
     }
     else
--- a/src/core/datatypes/phylogenetics/character/CharacterTranslator.cpp
+++ b/src/core/datatypes/phylogenetics/character/CharacterTranslator.cpp
@@ -8,6 +8,7 @@
 #include "AminoAcidState.h"
 #include "Cloneable.h"
 #include "CodonState.h"
+#include "DoubletState.h"
 #include "DiscreteCharacterState.h"
 #include "DnaState.h"
 #include "NaturalNumbersState.h"
@@ -75,6 +76,21 @@ RevBayesCore::AbstractDiscreteTaxonData* CharacterTranslator::translateCharacter
             trans_taxon_data = translateToCodon( d );
         }
     }
+    else if ( type == "Doublet")
+    {
+        if ( d.getCharacter(0).getDataType() == "DNA" )
+        {
+            trans_taxon_data = translateToDoubletFromDna( dynamic_cast< const DiscreteTaxonData<DnaState>& >(d) );
+        }
+        else if ( d.getCharacter(0).getDataType() == "RNA" )
+        {
+            trans_taxon_data = translateToDoubletFromRna( dynamic_cast< const DiscreteTaxonData<RnaState>& >(d) );
+        }
+        else
+        {
+            trans_taxon_data = translateToCodon( d );
+        }
+    }
     else
     {
         throw RbException("Cannot translate character data object of type \""+ d.getCharacter(0).getDataType() +"\" into type \"" + type + "\"" );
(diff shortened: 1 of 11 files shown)


bredelings:
I'm getting at issue where, when using actual data, the likelihood becomes positive:

```
Iter        |      Posterior   |     Likelihood   |          Prior   |             TL   |    elapsed   |        ETA   |
0           |       -214.583   |       -641.796   |        427.213   |       11.01846   |   00:00:00   |   --:--:--   |
100         |        7568.24   |        7259.99   |         308.25   |       22.91475   |   00:00:07   |   --:--:--   |
```
Does anyone have an idea what could be the problem here?

bjoelle:
@bredelings is this the same issue as #85 ? if so, PR #94 fixes it

bredelings:
@bjoelle Thanks for looking at this.

Here the problem seems to be that the matrix exponentiation for GTR was production transition probabilities > 100000.  I'm not sure why, since the branch length was not large.  But it could be that one of the frequencies got really small, which I think can make that code go crazy.

I fixed it by normalizing each row of the P[i][j] matrix to sum to 1.0.  I think I did this for a different matrix-exponentiaion routine already...

src/core/datatypes/phylogenetics/character/CharacterTranslator.cpp
```diff
@@ -75,6 +76,21 @@ RevBayesCore::AbstractDiscreteTaxonData* CharacterTranslator::translateCharacter
...
+        {
+            trans_taxon_data = translateToDoubletFromDna( dynamic_cast< const DiscreteTaxonData<DnaState>& >(d) );
+        }
+        else if ( d.getCharacter(0).getDataType() == "RNA" )
+        {
+            trans_taxon_data = translateToDoubletFromRna( dynamic_cast< const DiscreteTaxonData<RnaState>& >(d) );
+        }
+        else
+        {
+            trans_taxon_data = translateToCodon( d );
```
afmagee, line 91: @bredelings Should this be `trans_taxon_data = translateToDoublet( d );`?

bredelings: Yes!  Thanks for catching that.   That routine throws an exception, so we would have gotten the wrong error message.

afmagee:
I took a look and things seem sane as far as I can tell. A few other comments/questions. I thought Rev style was to generally avoid `auto`? Normalizing the transition probability matrix seems like it could cause a massive headache with the integration tests. Have you checked? If so, maybe it's best to do that separately? 

I can see why for now we don't want to deal with ambiguities in translation, as that is kind of a massive headache for translating to 1s/0s in a PLV for a tip. In the mean time, if we're going to require either -/?/N or fully specified, we might contemplate something to help users prepare their dataset by automatically making partial ambiguities into full ambiguities. That might be painful in Rev, but it should be easy enough in RevGadgets. This isn't really a priority for the pull request, just a thought for things we will want to do before we make a tutorial on doublet models.

bredelings:
> I thought Rev style was to generally avoid `auto`?

I either didn't know this, or forgot this.  Is this written down anywhere?

I'll add a patch removing 'auto', except where the type is already explicit.
(Hmm... see https://google.github.io/styleguide/cppguide.html#Type_deduction)

> Normalizing the transition probability matrix seems like it could cause a massive headache with the integration tests. Have you checked? If so, maybe it's best to do that separately?

I forgot that integrations tests aren't normally run.  However, I just ran them, and after waiting for revbayes.github.io to download (!?), they all passed.  So maybe its fine?  Perhaps these cases are either using fnFreeK or are already normalized...  the point of the normalization is to handle edge cases.

> I can see why for now we don't want to deal with ambiguities in translation, as that is kind of a massive headache for translating to 1s/0s in a PLV for a tip.

I think I figured out how to do the translation.  I'm more concerned that we don't silently change the interpretation of the data.

> In the mean time, if we're going to require either -/?/N or fully specified, we might contemplate something to help users prepare their dataset by automatically making partial ambiguities into full ambiguities.

If the user explicitly requests treating `A?` as `AN`, then that's fine.  I'm not totally sure where we would do this in Rev...  Possibly as an extra argument to translateCharacters( )?


> That might be painful in Rev, but it should be easy enough in RevGadgets. This isn't really a priority for the pull request, just a thought for things we will want to do before we make a tutorial on doublet models.

Yeah, that's a good point.

For a FASTA file, you might be able to get away with 's/?/N/g', assuming there are no '?' characters in the sequence names or comments.  I think at one point I wrote a script to apply regular expressions either (i) only to sequence names or (ii) only to character data.

bredelings:
@bjoelle @willpett @meyerx @hoehna 
Any other comments?

1. I'm curious if anybody else has an opinion on whether we should treat A- and A? as AN.  I could implement it.

2. My recollection is that Sebastian didn't like 'auto x = ...' because the type is not apparent from immediate inspection.  Google's coding style has a similar concern, but allows use of auto in situations where the type is duplicated, so `T* x = new T` could be replaced with `auto x = new T`.  I still have left one situation with `auto` because the type name is super-long.  At some level, I just need to know which way to do this.

bjoelle:
@bredelings no other comments from me
I don't have an opinion on point 1. For point 2 I don't think RevBayes has an official style guide (maybe it should ?) and I don't remember any discussion specifically about `auto`, so I would say it's fine to use.
I will let @afmagee approve the PR since he had more in-depth comments

afmagee (approved):
There appears to be an issue with the python build on OSX 10.13 (but not 10.14, for some reason) but other than that this looks good to me!

PR #124, opened 2021-01-11
add option for number of cores to build script

willpett:


bjoelle (changes requested):
can you add the new option to the help message (l26-33) ? thanks

bjoelle (approved):

PR #126, opened 2021-01-13
Fixed checkpointing bug partially. Changed printing functions, particularly vector printing.

brpetrucci:
dev_checkpointing is intended to fix the checkpointing bug whereby after recovering a run from checkpoint, the posterior probability goes to infinity. I realized this happened due to the automatic rounding of doubles to a file when checkpointing, which might lead to wrong probability calculations when recovering from checkpoint. 

To fix this, I changed the `printForComplexStoring` function to print with maximum precision, and avoid rounding. This is used in checkpointing, while `printForSimpleStoring` is used everywhere else. This also required writing a specific version of `printForComplexStoring` for trees, to override the `<<` operator which does not take arguments and therefore cannot change behavior depending on the type of logging. This fixed checkpointing partially, but there are still issues when fossil ages are used. Still, I figured it was worth including this into development while I figure out how to fix the rest of the problems.

This led to errors in the GitHub tests due to the vector printing functions, since `printForComplexStoring` was previously being used to print unflattened vectors (i.e. logging a vector as `[a, b, c]` as opposed to `a  b  c`). To solve this, considering unflattened printing is only used in the `StochasticVariableMonitor`, I created a new variable of the `AbstractVariableMonitor` class, `flatten`, with default `true`, and set it to `false` for the former monitor. Then, I added `flatten` (with the same default) as an argument to the `printFor` functions, where in the case of the vector version it decided whether to print flattened or not. This solved the errors, and the latest version of dev_checkpointing passed all tests while maintaining the no-rounding of state checkpoint files.


hoehna (changes requested):
1) src/core/analysis/mcmc/Mcmc.cpp
Why does this file now always request to flatten vectors? I'm not sure either which way is better, but this seems to me a change that may or may not be expected. If this is not intentional, then perhaps better change it back. If it intentional, then please comment in the code.

2) Printer.h
It would be better to derive Tree from printable then making a special case for printFor CompleteStoring. So please remove this special case function from Printer.h, and derive Tree from Printable and implement printForUser(), printForSimpleStoring(), printForComplexStoring() in Tree.

3) Printer.h
If you are already at this, perhaps you could update the documentation? It looks like I simply copied the Cloner.h class, but didn't change the comments. If you need help, then please let me know.

4) ContinuousCharacterData.cpp
This seems like placeholder code. Maybe at least add a todo comment for this, or better, add a "throw RbException" so that the users knows that this won't work.

5) TopologyNode.cpp
So does this mean a precision of 10^-6 didn't work, but 10^-10 does? Shouldn't there be an even higher precision?

brpetrucci:
Hi Dr. Höhna, thank you for reviewing.

1. Checkpointing was already always flattening by default before. It seemed to me like flatten in Mcmc.cpp was not being used in the same way as the vector printing functions, which is part of why I thought it'd be good to make everything consistent now. As it stands, checkpointing always flattens like before, and the only moment a vector is not flattened is in the StochasticVariable monitor. I can surely include a comment there about this, yes.
2. I will try this soon.
3. For sure! I will let you know if I need help, thank you.
4. It's not placeholder, it is just so that the signature matches the parent method signature. So I don't think an RbException is necessary, though I could certainly throw one in. It seems to me from scanning the code that there is no way for a user to call these print functions without flatten already previously having a value, so that this will work. I could be wrong, though.
5. The command `std::numeric_limits<double>::digits10` is used to call the maximum possible precision. I believe "digits10" is meant to get it in decimal, as opposed to another base, and not to mean that 10^-10 is the precision. If I misunderstood your point, though, let me know.

In summary I believe the `flatten = true` line in Mcmc.cpp did not change the way vectors are printed in checkpointing, but simply make it consistent with the new setup of printing functions; the character data printing functions are not placeholders and should be working fine, if my investigation of the functions calling them are correct; and the maximum precision is already being used for checkpointing. So I will implement the change you requested for the trees, and add comments on the flatten point. Let me know if you disagree with anything I said or want me to make other changes!

brpetrucci:
Oh, sorry! I did not think to alter that since the body of the function was already implemented and I was just changing the signature, but you're right, it's probably better to change that and throw an error. Will add it to the to-do list

brpetrucci:
Added comments to Printer.h, albeit simple ones; let me know if you want more details. Deleted the Tree printing in Printer.h and defined them instead in the Tree files, deriving Tree from Printable. Also added the error for `printForComplexPrinting` on the character data file. Let me know if you want me to do anything else before merging!

hoehna (approved):
Looks good to me.

Issue #5, opened 2019-12-11
Conflicting monophyletic clade constraints


NickCrouch:
Hi,
There appears to be a problem with designating nested clade constraints. I encounter this problem when analyzing the exemplar bears data. In the tutorial a single clade constraint is defined:
`clade_ursinae = clade("Helarctos_malayanus", "Melursus_ursinus", "Ursus_abstrusus", "Ursus_americanus", "Ursus_arctos", "Ursus_maritimus",   "Ursus_thibetanus")`
which works find. If a second constraint is specified and then enforced, then the error is returned
`clade_ursus = clade("Ursus_abstrusus","Ursus_thibetanus")`
`constraints = v(clade_ursinae, clade_ursus)`

> Error: Cannot simulate tree: conflicting monophyletic clade constraints. Check that all clade constraints are properly nested.

This error derives from the error checks `found_some` and `found_all` in "TopologyConstrainedTreeDistribution.cpp" but I'm not sure how. 

I replicated this issue using the recently pushed "oh, rapture" update.

ziad-khouri:
Can confirm that this is still an issue as of commit 11f057bb037e5dc54a246688f36c5a9e880a0daa development branch.

bredelings:
OK, this issue is easy to reproduce by adding the ` clade_ursus = clade("Ursus_abstrusus","Ursus_thibetanus")` constraint in `model_FBDRP.Rev`, as mentioned above.

The issue occurs when simulating the initial tree.  This happens in`TopologyConstrainedTreeDistribution::simulateTree( void )` in `/src/core/distributions/phylogenetics/tree/TopologyConstrainedTreeDistribution.cpp`.  

Interestingly, if you switch the order of the two clades from ` constraints = v(clade_ursinae, clade_ursus)` to `constraints = v(clade_ursus, clade_ursinae)`, then we get:
```
> source("mcmc_CEFBDRP_Ranges.Rev")
   Processing file "mcmc_CEFBDRP_Ranges.Rev"
   Successfully read one character matrix from file 'data/bears_cytb.nex'
   Successfully read one character matrix from file 'data/bears_morphology.nex'
   Processing file "scripts/model_FBDRP.Rev"
   Processing of file "scripts/model_FBDRP.Rev" completed
   Processing file "scripts/model_UExp.Rev"
   Processing of file "scripts/model_UExp.Rev" completed
   Processing file "scripts/model_GTRG.Rev"
[...]

PR #130, opened 2021-01-26
Fix nested clade constraints

bredelings:
This PR fixes the problem by sorting the clade constraints so that clades come before other clades that they are nested within.  Previously, the sort was only by the clade age.

The Clade data structure is trying to be three or four different things at once (a set of taxa, a rooted split, a dated node - plus the tip taxa also have dates).  The include group for the clade-treated-as-a-rooted split is specified, but the exclude group isn't really.  Any include/exclude bitmaps must be manually set by the user.  This should maybe be cleaned up in the future, but I didn't change the Clade object here.

I defined some new functions for checking clade conflict, overlap, and nesting.  These implicitly assume that two clades are being treated as rooted splits with the same set of full taxa.


bredelings:
Fixes #5

willpett (approved):
Nice catch on the sorting. Good solution.

PR #136, opened 2021-02-02
Sebastian's fixes to the RNG

bjoelle:
and adjustments to the tests


willpett (approved):
Good deal. It will be good to sort out a solution to the diverging RNG between boost versions at some point.

PR #137, opened 2021-02-08
bug fix for segfault under DEC + stochastic mapping

mlandis:
Some time recently the stochastic mapping code was changed to always use calculateTransitionProbabilities(). This was apparently not an issue for most rate matrices, but it caused stochastic mapping under DEC to segfault (invalid index access).

Context: DEC has an option to condition on survival (i.e. not entering the "null range") which is used when computing calculateTransitionProbabilities(). Using method for stochastic mapping causes the algorithm to fail, because the algorithm requires P(t)=exp(Qt) to work. Stochastic mapping under uniformization fails when using P that conditions on survival because P(t|survival) != exp(Qt).

To restore the original functionality, I added a virtual function for calculateTransitionProbabilitiesForStochasticMapping() that by default calls calculateTransitionProbabilities(). RateMatrix_DECRateMatrix::calculateTransitionProbabilitiesForStochasticMapping() computes the unconditional form of P(t)=exp(Qt).

### DIFF
--- a/src/core/datatypes/phylogenetics/ratematrix/AbstractRateMatrix.cpp
+++ b/src/core/datatypes/phylogenetics/ratematrix/AbstractRateMatrix.cpp
@@ -199,6 +199,16 @@ std::vector<double> AbstractRateMatrix::calculateStationaryFrequencies(void) con
     return pi;
 }
 
+/** This function computes the transition probability matrix that is used for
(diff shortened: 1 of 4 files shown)


mikeryanmay (approved):
This all checks out! Compiles just fine and fixes the segfault.

PR #146, opened 2021-02-12
Dev hoehna lab

hoehna:
This is the long overdue pull request to merge the branch from my group into the main development branch. I think that this branch has diverged for >2 years, and there are many, many new features and bug fixes. These new features and bugfixes are described in the following papers:
- https://www.biorxiv.org/content/10.1101/2021.01.14.426715v1.abstract
- https://www.biorxiv.org/content/10.1101/2020.12.19.423498v2.abstract
- https://journals.plos.org/ploscompbiol/article?rev=2&id=10.1371/journal.pcbi.1007999
- https://www.biorxiv.org/content/10.1101/2020.11.11.376459v1.abstract
- https://www.biorxiv.org/content/10.1101/2020.10.17.343889v3.abstract
- https://academic.oup.com/genetics/article/213/2/581/5930581?login=true

There are several more unpublished features which are ready for publication/use.

I've tested the integration tests and they run fine, except 5 expected issues:
1. ChromoSSE test gives different result because root is not considered as speciation event.
2. JC_unrooted test gives different results because the (unrooted) NNI move has changed.
3. Mk test gives different results because the (unrooted) NNI move has changed.
4. Partition test gives different results because the (unrooted) NNI move has changed and I am not allowing unresolved sites to be treated as invariant anymore.
5. steppingstone test gives different results because the (unrooted) NNI move has changed and I am not allowing unresolved sites to be treated as invariant anymore.


bredelings:
Right now the windows build test is failing because the `tests` directory does not exist.

hoehna:
Does anyone know how to fix this?

bredelings:
I made a branch called dhl-fix-1 that puts back the tests submodule. All the tests pass except the 5 that @hoehna mentioned.

@bjoelle Does that seem like a good solution to you?

bjoelle:
@bredelings thanks ! I think the easiest would be to add the commits that re-adds the tests directly to this PR
@hoehna for the tests that are failing, I have added information on how to update the tests in the README of the tests repo (https://github.com/revbayes/tests/blob/master/README.md), please let me know if anything there is unclear

bredelings:
I merged in the commit that puts back the tests submodule.  (BTW, the submodules are defined in `.gitmodules` file)

To pull that, I suspect people need to first delete the tests directory, then pull, and then maybe do `git submodule init; git submodule update` or something like that.

1. So, I think the idea is now to update the 5 tests in the tests repo so that the dev-hoehna-lab branch passes the tests.  @hoehna or @afmagee do you have time to do that?

2. As Joelle said, submodules refer to a specific commit of the sub-repo.  So after we update the `tests` repo, we need to add a commit to this PR that updates the submodule to refer to the new commit in the `tests` repo.  I could do this, if somebody else does not do it first.

bredelings:
OK, so now the revbayes tests pass.  However, 4 of the testiphy tests are failing.

```
  Running revbayes test: tests/likelihood/gtr_inv/2  ... FAIL! ['likelihood'] 
       likelihood error: absolute=-35.93836039237249, relative=0.003635849420787554 (Got -9920.3852208423723 but expected -9884.446860450)
  
  Running revbayes test: tests/likelihood/gtr_inv/1  ... FAIL! ['likelihood'] 
       likelihood error: absolute=-1.4279818158547641, relative=0.00021529269199617175 (Got -6634.1743264068546 but expected -6632.7463445910)

  Running revbayes test: tests/likelihood/gtr_gamma_inv/2  ... FAIL! ['likelihood'] 
       likelihood error: absolute=-10.611420887626991, relative=0.001370681638806056 (Got -7752.322250887627 but expected -7741.71083)
  
  Running revbayes test: tests/likelihood/gtr_gamma_inv/1  ... FAIL! ['likelihood'] 
       likelihood error: absolute=-0.5883496161450239, relative=8.938339338128717e-05 (Got -6582.9029596161454 but expected -6582.31461)
```
For example, this test is failing: https://gitlab.com/testiphy/testiphy/-/blob/master/tests/likelihood/gtr_inv/1/rb-command.Rev

Also it looks like there are some new merge conflicts.

-BenRI

hoehna:
Hi Ben,

I guess all these difference to the testiphy results stem from my change that sites/columns that contain a missing character are not considered as missing. I did these changes because I analyzed a highly fragmented UCE dataset and realized that >50% of the loci had almost all sides declared as invariant. There was definitely something off with these analyses.

Is there any theory/reference how missing states contribute to invariant sites?

hoehna:
To follow up, the commit that is causing the issue is probably: commit 88697e36ecc14753dd8bfae044b9a82eed131eb4
I can look tomorrow for a dataset where this gave completely bonkers results.

bredelings:
Hi Sebastian,

Hmm... that's interesting.  So you are saying that a site is not "invariant" if it contains any missing data?  I think the traditional approach doesn't explicitly classify sites as "invariant" -- but it would assign a probability of pi[A] to a column like AAAANNAAA  pi[A] under the invariant model with an invariant fraction of 1.0.  That is also what you would get of you have a rate mixture and one component has a rate of 0.

In theory, having lots of missing data shouldn't lead to inferring a high proportion of invariable sites unless the prior supports that.  If all sites are missing you should recover the prior.  So, it seems like having a lot of missing data should be fine with the traditional likelihood for +INV.

Yeah, if you could look for a dataset that shows the problem, that would be interesting.  Feel free to send me a copy -- I would try and take a look.


-BenRI

bredelings:
@hoehna Hmm... what do you think about removing commit 88697e3 from dev-hoehna-lab and putting it on its own branch?  Then we could review/discuss it separately, which might make it easier to review dev-hoehna-lab.  No pressure though.

bredelings:
@hoehna I updated testiphy with the reduced test-case.

I'm still not sure the code find to invariant columns correctly if a column contains only ambiguous characters.  For example, if a column contains R and Y, then it can't be invariant, but if it contains R and K, then it could be invariant.

bredelings:
This loop here seems like it will start setting N  characters to gaps if treatUnknownAsGap, but then stop in the middle of the matrix if it finds an ambiguous character.

```
    // find the unique site patterns and compute their respective frequencies
    std::vector<TopologyNode*> nodes = tau->getValue().getNodes();
    for (size_t site = 0; site < num_sites; ++site)
    {

        for (std::vector<TopologyNode*>::iterator it = nodes.begin(); it != nodes.end(); ++it)
        {
            if ( (*it)->isTip() )
            {
                AbstractDiscreteTaxonData& taxon = value->getTaxonData( (*it)->getName() );
                DiscreteCharacterState &c = taxon.getCharacter(site_indices[site]);

                // if we treat unknown characters as gaps and this is an unknown character then we change it
                // because we might then have a pattern more
                if ( treatAmbiguousAsGaps && (c.isAmbiguous() || c.isMissingState()) )
                {
                    c.setGapState( true );
                }
                else if ( treatUnknownAsGap && (c.getNumberOfStates() == c.getNumberObservedStates() || c.isMissingState()) )
                {
                    c.setGapState( true );
                }
                else if ( !c.isGapState() && (c.isAmbiguous() || c.isMissingState()) )
                {
                    ambiguousCharacters = true;
                    break;
                }
            }
        }

        // break the loop if there was an ambiguous character
        if ( ambiguousCharacters )
        {
            break;
        }
    }

```

willpett:
`RevBayesCore::PhyloCTMCSiteHomogeneousConditional<charType>::isSitePatternCompatible` may be of some use in this situation. This function determines whether a site pattern is compatible with a particular ascertainment bias coding, e.g. constant characters excluded, parsimony informative excluded. It could be used more generally to identify (potentially) invariant sites.

bredelings:
OK, the issue that I mentioned above is not fixed on development.

With the additional fixes that @willpett did in commit 834b3544d69b, maybe we should avoid setting `-` to have all its bits set for DNA in commit d0e2c44?

willpett:
Yeah, with the new treatment of ambiguous invariant site patterns in `834b354`, I think the changes in `d0e2c44` no longer apply? I reverted them to resolve the conflicts from merging development into this branch.

willpett:
This means the test output for `test_Partition` and `test_steppingstone` will need to be updated again.

src/core/datatypes/phylogenetics/character/DnaState.cpp
```diff
@@ -408,6 +408,10 @@ RbBitSet DnaState::getState(void) const
     switch ( state )
     {
         case '-':
+            bs.set(0);
```
bredelings, line 411: This change should probably be reverted now.

src/core/datatypes/phylogenetics/character/DoubletState.cpp
```diff
@@ -79,11 +79,11 @@ std::string DoubletState::getStringValue(void) const
     RbBitSet letters1(4);
     RbBitSet letters2(4);
 
-    for(int i=0;i<state.size();i++)
+    for (int i=0;i<state.size();i++)
     {
         if (state[i])
         {
-            auto& doublet = DOUBLETS[i];
+            const std::string& doublet = DOUBLETS[i];
```
bredelings, line 86: Changing code that someone else wrote to suit your style preference seems a bit arbitrary.

.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "tests"]
 	path = tests
-	url = https://github.com/revbayes/tests
+	url = https://github.com/revbayes/tests.git
```
bredelings, line 3: Hmm... what's going on here.  Do these both work?

src/core/distributions/phylogenetics/tree/MultispeciesCoalescentUniformPrior.cpp
```diff
@@ -34,87 +36,186 @@ MultispeciesCoalescentUniformPrior::~MultispeciesCoalescentUniformPrior()
...
+        // std::cout << "current time: " << current_time << std::endl;
+
         double a = times[i] - current_time;
-        
+        current_time = times[i];
+
         // get the number j of individuals we had before the current coalescence
         size_t j = k - i;
-        double n_pairs = j * (j-1.0) / 2.0;
+        double n_pairs = j * (j-1.0);
```
bredelings, line 71: Why is the `/2.0` getting removed?

src/core/distributions/phylogenetics/tree/MultispeciesCoalescentUniformPrior.cpp
```diff
@@ -34,87 +36,186 @@ MultispeciesCoalescentUniformPrior::~MultispeciesCoalescentUniformPrior()
...
     }
-    
+
     // compute the probability of no coalescent event in the final part of the branch
     // only do this if the branch is not the root branch
     if ( add_final_interval == true )
     {
         double final_interval = end_age - current_time;
         size_t j = k - times.size();
-        double n_pairs = j * (j-1.0) / 2.0;
```
bredelings, line 72: Why is the `/2.0` getting removed?

src/core/distributions/phylogenetics/tree/TopologyConstrainedTreeDistribution.cpp
```diff
@@ -54,14 +53,16 @@ TopologyConstrainedTreeDistribution::TopologyConstrainedTreeDistribution(TypedDi
...
+    starting_tree( t ),
+    rooting_known( false ),
+    is_rooted( true )
 {
-    AbstractRootedTreeDistribution* tree_base_distribution = dynamic_cast<AbstractRootedTreeDistribution*>(base_distribution);
-    if (tree_base_distribution == NULL)
-    {
-        throw(RbException("Can only constrain tree distributions of type AbstractRootedTreeDistribution"));
-    }
+//    AbstractRootedTreeDistribution* tree_base_distribution = dynamic_cast<AbstractRootedTreeDistribution*>(base_distribution);
```
bredelings, line 61: Why is this commented out?

src/core/distributions/phylogenetics/tree/birthdeath/BirthDeathBurstProcess.cpp
```diff
@@ -433,8 +451,8 @@ double BirthDeathBurstProcess::lnQ(double t) const
         E_previous /= (2*birth);
 
         B = ((1.0 - 2.0*((1.0-burst)*E_previous+burst*E_previous*E_previous)) * birth + death ) /  A;
+//        B = ((1.0 - 2.0*E_previous) * birth + death ) /  A;
```
bredelings, line 454: This this the remnant of an experiment?

src/core/distributions/phylogenetics/tree/birthdeath/BirthDeathBurstProcess.cpp
```diff
@@ -444,7 +462,7 @@ double BirthDeathBurstProcess::lnQ(double t) const
         lnD_previous -= 2 * log(1 + B_previous + exp(-A * t_b) * (1 - B_previous));
         
         double this_lnD = lnD_previous;
-        this_lnD += log(1-burst+2*burst*E_previous);
+//        this_lnD += log(1-burst+2*burst*E_previous);
```
bredelings, line 465: Are we keeping this line because we are not sure which version is right?

src/core/distributions/phylogenetics/tree/birthdeath/StateDependentSpeciationExtinctionProcess.cpp
```diff
@@ -1440,10 +1490,10 @@ RevLanguage::RevPtr<RevLanguage::RevVariable> StateDependentSpeciationExtinction
         static_cast<TreeDiscreteCharacterData*>(this->value)->setCharacterData( v.clone() );
    
         // simulate character history over the tree conditioned on the new tip data
-        size_t num_nodes = value->getNumberOfNodes();
-        std::vector<std::string> character_histories(num_nodes);
-        drawStochasticCharacterMap(character_histories);
-        static_cast<TreeDiscreteCharacterData*>(this->value)->setTimeInStates(time_in_states);
+//        size_t num_nodes = value->getNumberOfNodes();
```
bredelings, line 1493: Should this be commented out?

src/core/distributions/phylogenetics/tree/birthdeath/StateDependentSpeciationExtinctionProcess.cpp
```diff
@@ -1919,61 +2000,13 @@ void StateDependentSpeciationExtinctionProcess::setValue(Tree *v, bool f )
...
-        {
-            if (i != j)
-            {
-                total_rates[i] += rate_matrix->getRate(i, j, 0.0, getEventRate());
-            } 
-        }
-    }
-
-    return total_rates;
+//    size_t num_nodes = value->getNumberOfNodes();
```
bredelings, line 2003: Should this be commented out -- maybe delete it along with the comment about simulating, if we are not simulating here?

src/core/moves/MetropolisHastingsMove.cpp
```diff
@@ -259,10 +260,90 @@ void MetropolisHastingsMove::performMcmcMove( double prHeat, double lHeat, doubl
...
+    
     // Propose a new value
     proposal->prepareProposal();
-    double ln_hastings_ratio = proposal->doProposal();
-    
+    double ln_hastings_ratio = RbConstants::Double::neginf;
+    try {
+        ln_hastings_ratio = proposal->doProposal();
+    }
+    catch (RbException e)
```
bredelings, line 339: I don't think catching exceptions by value is safe.  If you do 
``` c++
catch (const RbException& e)
```
I think that is safe.

src/core/moves/MetropolisHastingsMove.cpp
```diff
@@ -300,11 +381,34 @@ void MetropolisHastingsMove::performMcmcMove( double prHeat, double lHeat, doubl
 
             if ( the_node->isClamped() )
             {
-                ln_likelihood_ratio += the_node->getLnProbabilityRatio();
+                try {
+                    ln_likelihood_ratio += the_node->getLnProbabilityRatio();
+                }
+                catch (RbException e)
```
bredelings, line 387: Catching exception by value.

src/core/moves/MetropolisHastingsMove.cpp
```diff
@@ -300,11 +381,34 @@ void MetropolisHastingsMove::performMcmcMove( double prHeat, double lHeat, doubl
...
+                {
+                    ln_likelihood_ratio = RbConstants::Double::neginf;
+
+                    if ( e.getExceptionType() != RbException::MATH_ERROR )
+                    {
+                        throw RbException(e);
```
bredelings, line 393: This should probably just be `throw;` -- that rethrows the exception that is currently being handled.

src/core/moves/MetropolisHastingsMove.cpp
```diff
@@ -300,11 +381,34 @@ void MetropolisHastingsMove::performMcmcMove( double prHeat, double lHeat, doubl
...
+                }
             }
             else
             {
-                ln_prior_ratio += the_node->getLnProbabilityRatio();
+                try
+                {
+                    ln_prior_ratio += the_node->getLnProbabilityRatio();
+                }
+                catch (RbException e)
```
bredelings, line 403: Catching exception by value.

src/core/moves/MetropolisHastingsMove.cpp
```diff
@@ -300,11 +381,34 @@ void MetropolisHastingsMove::performMcmcMove( double prHeat, double lHeat, doubl
...
+                {
+                    ln_prior_ratio = RbConstants::Double::neginf;
+
+                    if ( e.getExceptionType() != RbException::MATH_ERROR )
+                    {
+                        throw RbException(e);
```
bredelings, line 409: throw;

src/core/moves/MetropolisHastingsMove.cpp
```diff
@@ -320,11 +424,34 @@ void MetropolisHastingsMove::performMcmcMove( double prHeat, double lHeat, doubl
         {
             if ( the_node->isClamped() )
             {
-                ln_likelihood_ratio += the_node->getLnProbabilityRatio();
+                try
+                {
+                    ln_likelihood_ratio += the_node->getLnProbabilityRatio();
+                }
+                catch (RbException e)
```
bredelings, line 431: catching exception by value.

src/core/moves/MetropolisHastingsMove.cpp
```diff
@@ -320,11 +424,34 @@ void MetropolisHastingsMove::performMcmcMove( double prHeat, double lHeat, doubl
...
+                {
+                    ln_likelihood_ratio = RbConstants::Double::neginf;
+
+                    if ( e.getExceptionType() != RbException::MATH_ERROR )
+                    {
+                        throw RbException(e);
```
bredelings, line 437: throw;

src/core/moves/MetropolisHastingsMove.cpp
```diff
@@ -320,11 +424,34 @@ void MetropolisHastingsMove::performMcmcMove( double prHeat, double lHeat, doubl
...
+                    }
+                }
             }
             else
             {
-                ln_prior_ratio += the_node->getLnProbabilityRatio();
+                try {
+                    ln_prior_ratio += the_node->getLnProbabilityRatio();
+                }
+                catch (RbException e)
```
bredelings, line 446: Catching exception by value.

src/core/moves/MetropolisHastingsMove.cpp
```diff
@@ -320,11 +424,34 @@ void MetropolisHastingsMove::performMcmcMove( double prHeat, double lHeat, doubl
...
+                {
+                    ln_prior_ratio = RbConstants::Double::neginf;
+
+                    if ( e.getExceptionType() != RbException::MATH_ERROR )
+                    {
+                        throw RbException(e);
```
bredelings, line 452: throw;

bredelings (changes requested):
I was able to quickly review 289 / 394 changed files.  A lot of these changes are changing `DagNode*` to `const DagNode*`, but there are still a fair number of other changes.

 I've made a few comments with some things that should perhaps be clarified or fixed.

projects/cmake/build.sh
```diff
@@ -153,12 +152,12 @@ fi
...
 
     echo "Running './regenerate.sh $(pwd)/$BUILD_DIR"
-    ./regenerate.sh $(pwd)/$BUILD_DIR
+    bash regenerate.sh $(pwd)/$BUILD_DIR
     cd ${BUILD_DIR}
     echo
     echo
     echo "Running 'cmake ../../../src $cmake_args' in $(pwd)"
-    cmake ../../../src $cmake_args
+    cmake -D Boost_NO_BOOST_CMAKE:BOOL=ON ../../../src $cmake_args
```
bredelings, line 160: what's the rationale here?  Do we want to do this for everybody?

projects/cmake/build.sh
```diff
@@ -153,12 +152,12 @@ fi
     fi
 
     echo "Running './regenerate.sh $(pwd)/$BUILD_DIR"
-    ./regenerate.sh $(pwd)/$BUILD_DIR
+    bash regenerate.sh $(pwd)/$BUILD_DIR
```
bredelings, line 155: What's the rationale here?

Should we change regenerate.sh to start with `#!/bin/bash`?  What about it is bash-specific?

src/core/datatypes/trees/Clade.cpp
```diff
@@ -125,6 +127,11 @@ bool Clade::operator==(const Clade &c) const
         return false;
     }
     
+    if ( c.clade_name != clade_name )
+    {
+        return false;
+    }
+
```
bredelings, line 134: Hmm... do we want clades with the same members but different names to be unequal?

src/core/datatypes/trees/Clade.cpp
```diff
@@ -168,6 +175,11 @@ bool Clade::operator<(const Clade &c) const
         return false;
     }
     
+    if ( c.clade_name != clade_name )
+    {
+        return c.clade_name < clade_name;
+    }
+    
     for (size_t i = 0; i < taxa.size(); ++i)
```
bredelings, line 183: do we want clades with the same members but different names to be unequal?

bredelings (changes requested):

src/revlanguage/workspace/RbRegister_Dist.cpp
```diff
@@ -162,11 +163,14 @@
 #include "Dist_bdp_complete.h"
 #include "Dist_BDSTP.h"
 #include "Dist_BirthDeathBurstProcess.h"
+#include "Dist_BranchRateTree.h"
+// #include "Dist_BDSTP.h"
```
wadedismukes, line 167: Line 167 should be deleted.

src/revlanguage/moves/tree/Move_TreeScale.cpp
```diff
@@ -52,8 +49,24 @@ void Move_TreeScale::constructInternalObject( void )
     delete value;
     
     // now allocate a new tree scale move
-    RevBayesCore::TypedDagNode<RevBayesCore::Tree> *tmp = static_cast<const TimeTree &>( tree->getRevObject() ).getDagNode();
-    RevBayesCore::StochasticNode<RevBayesCore::Tree> *t = static_cast<RevBayesCore::StochasticNode<RevBayesCore::Tree> *>( tmp );
+    
+    // get the tree(s) variable
+    RevBayesCore::StochasticNode<RevBayesCore::Tree> *t = NULL;
+    RevBayesCore::StochasticNode< RevBayesCore::RbVector<RevBayesCore::Tree> > *vec_t = NULL;
+    if ( tree->getRevObject().isType( TimeTree::getClassTypeSpec() ) )
```
wadedismukes, line 56: A comment describing what this ifelse is doing would be nice.

src/core/distributions/phylogenetics/tree/birthdeath/BirthDeathBurstProcess.cpp
```diff
@@ -135,9 +135,9 @@ double BirthDeathBurstProcess::computeLnProbabilityTimes( void )
         if ( n.isRoot() == false )
         {
             double parent_age = n.getParent().getAge();
-            if ( fabs(node_age - burst_time) > 1E-10 &&
+            if ( fabs(node_age - burst_time) > 1E-4 &&
```
willpett, line 138: Again, in my opinion these comparison thresholds should not be hardcoded in place. I thought that's what functions in `RbMathLogic.h` were for? In general, though, I also question whether these precision-based equality comparisons are really always necessary.

src/core/distributions/phylogenetics/tree/birthdeath/StateDependentSpeciationExtinctionProcess.cpp
```diff
@@ -97,7 +97,7 @@ StateDependentSpeciationExtinctionProcess::StateDependentSpeciationExtinctionPro
     process_age( age ),
     mu( ext ),
     lambda(NULL),
-    psi( NULL),
+    phi( NULL),
```
willpett, line 100: Why did `psi` got renamed to `phi`?

src/core/functions/phylogenetics/ratematrix/CovarionFunction.cpp
```diff
@@ -14,12 +14,12 @@ namespace RevBayesCore { class DagNode; }
...
 
 CovarionFunction::CovarionFunction(bool r) : TypedFunction<RateGenerator>( NULL ),
-rescale( r ),
-//num_observed_states( 0 ),
-//num_hidden_states( 0 ),
-rate_matrices( NULL ),
-rate_scalars( NULL ),
-switch_rates( NULL )
+    rescale( r ),
+    //num_observed_states( 0 ),
```
willpett, line 18: Delete these?

src/core/utils/BirthDeathForwardSimulator.cpp
```diff
@@ -32,20 +34,42 @@ bool BirthDeathForwardSimulator::checkParameters( void ) const
...
+   }
+
+   if ( !(Phi.size() == NUM_TIME_INTERVALS || Phi.size() == 1) )
+   {
+       throw RbException("You need to provide the same number of time intervals as fossilization probabilities.");
+   }
+
+   if ( !(R.size() == NUM_TIME_INTERVALS || R.size() == 1) )
+   {
+       std::cout << "R.size() == " << R.size() << std::endl;
```
willpett, line 69: These debug messages should be in RBOUT or deleted.

src/revlanguage/distributions/phylogenetics/character/Dist_CTMC.h
```diff
@@ -0,0 +1,72 @@
+#ifndef Dist_CTMC_H
```
willpett, line 1: Definitely add a help page for this. It's not clear to me how this is different from PhyloCTMC.

src/revlanguage/distributions/phylogenetics/tree/Dist_BDSTP.cpp
```diff
@@ -289,12 +296,16 @@ const MemberRules& Dist_BDSTP::getParameterRules(void) const
         dist_member_rules.push_back( new ArgumentRule( "lambda",  paramTypes, "The birth rate(s).", ArgumentRule::BY_CONSTANT_REFERENCE, ArgumentRule::ANY ) );
         dist_member_rules.push_back( new ArgumentRule( "mu",      paramTypes, "The death rate(s).", ArgumentRule::BY_CONSTANT_REFERENCE, ArgumentRule::ANY ) );
         dist_member_rules.push_back( new ArgumentRule( "phi",     paramTypes, "The serial sampling rate(s).", ArgumentRule::BY_CONSTANT_REFERENCE, ArgumentRule::ANY ) );
```
willpett, line 298: Need to accept `psi` as a valid label for the serial sampling rates for compatibility with FBDP literature.

src/revlanguage/distributions/phylogenetics/tree/Dist_FBDP.cpp
```diff
@@ -283,7 +291,11 @@ const MemberRules& Dist_FBDP::getParameterRules(void) const
         paramTypes.push_back( ModelVector<RealPos>::getClassTypeSpec() );
         dist_member_rules.push_back( new ArgumentRule( "lambda",  paramTypes, "The birth rate(s).", ArgumentRule::BY_CONSTANT_REFERENCE, ArgumentRule::ANY ) );
         dist_member_rules.push_back( new ArgumentRule( "mu",      paramTypes, "The death rate(s).", ArgumentRule::BY_CONSTANT_REFERENCE, ArgumentRule::ANY ) );
-        dist_member_rules.push_back( new ArgumentRule( "psi",     paramTypes, "The serial sampling rate(s).", ArgumentRule::BY_CONSTANT_REFERENCE, ArgumentRule::ANY ) );
+        
+        std::vector<std::string> aliases_serial_sampling;
+        aliases_serial_sampling.push_back("phi");
+        aliases_serial_sampling.push_back("psi");
+        dist_member_rules.push_back( new ArgumentRule( aliases_serial_sampling,     paramTypes, "The serial sampling rate(s).", ArgumentRule::BY_CONSTANT_REFERENCE, ArgumentRule::ANY ) );
```
willpett, line 298: Like this :)

src/revlanguage/distributions/phylogenetics/tree/Dist_FBDP.cpp
```diff
@@ -97,7 +97,7 @@ RevBayesCore::AbstractBirthDeathProcess* Dist_FBDP::createDistribution( void ) c
     // death rate
     RevBayesCore::DagNode* d_s = mu->getRevObject().getDagNode();
     // serial sampling rate
-    RevBayesCore::DagNode* s_s = psi->getRevObject().getDagNode();
+    RevBayesCore::DagNode* s_s = phi->getRevObject().getDagNode();
```
willpett, line 100: I'm perplexed by the psi -> phi renaming.

src/revlanguage/distributions/phylogenetics/tree/Dist_episodicBirthDeath.cpp
```diff
@@ -161,8 +164,15 @@ RevBayesCore::EpisodicBirthDeathProcess* Dist_episodicBirthDeath::createDistribu
...
+    // tree for initialization
+    RevBayesCore::Tree* init = NULL;
+    if ( initial_tree->getRevObject() != RevNullObject::getInstance() )
+    {
+        init = static_cast<const TimeTree &>( initial_tree->getRevObject() ).getDagNode()->getValue().clone();
+    }
+    
     // create the internal distribution object
-    RevBayesCore::EpisodicBirthDeathProcess* d = new RevBayesCore::EpisodicBirthDeathProcess(ra, sr, st, er, et, r, strategy, inc_clades, cond, t);
+    RevBayesCore::EpisodicBirthDeathProcess* d = new RevBayesCore::EpisodicBirthDeathProcess(ra, sr, st, er, et, r, mp, strategy, inc_clades, cond, t, init);
```
willpett, line 175: Why can't you just use `setValue` here instead of passing the initial tree as an argument?

src/revlanguage/main.cpp
```diff
@@ -191,10 +191,13 @@ int main(int argc, char* argv[]) {
     {
         source_files = args["file"].as<std::vector<std::string> >();
     }
+//    source_files.push_back("scripts/MCMC_dating_ex4.Rev");
     
-    if ( args.count("args") and args.count("cmd"))
+    if ( args.count("args") && args.count("cmd"))
```
willpett, line 196: This doesn't seem necessary, and is slightly less performant (probably).

willpett (changes requested):

src/core/distributions/phylogenetics/tree/birthdeath/BirthDeathProcess.cpp
```diff
@@ -126,40 +127,69 @@ double BirthDeathProcess::computeLnProbabilityTimes( void )
...
+        ln_prob_times -= log( p_0_T )*num_taxa;
+
+        
+//        double p_s = pSurvival(start, end, r);
+//        double rate = rateIntegral(start, end) - log(r);
+//        double e = p_s * exp(rate);
+        // log(n-1) + 4*log(p_s) + 2*rate + log( 1 - e) * (n-2);
+//        ln_prob_times += lnProbNumTaxa( total_species, 0, present_time, true );
+//        ln_prob_times -= lnProbNumTaxa( num_taxa, 0, present_time, true );
+
```
bjoelle, line 168: commented out code should just be deleted (same for l117-118, 131 and 147 above)

src/core/distributions/phylogenetics/tree/birthdeath/EpisodicBirthDeathProcess.h
```diff
@@ -13,21 +13,28 @@ namespace RevBayesCore {
...
-//        double                                              computeLnProbabilityTimes(void) const;                                                              //!< Compute the log-transformed probability of the current value.
-        double                                              lnProbNumTaxa(size_t n, double start, double end, bool MRCA) const { throw RbException("Cannot compute P(nTaxa)."); }
-        void                                                prepareProbComputation(void);
+        void                                                swapParameterInternal(const DagNode *oldP, const DagNode *newP);                //!< Swap a parameter
+//        double                                              computeLnProbabilityTimes(void) const;                                         //!< Compute the log-transformed probability of the current value.
+//        double                                              lnProbNumTaxa(size_t n, double start, double end, bool MRCA) const { throw RbException("Cannot compute P(nTaxa)."); }
```
bjoelle, line 36: commented out functions can just be removed

bjoelle (changes requested):

.gitmodules
bredelings: The tests run, this is fine!

projects/cmake/build.sh
bredelings: Currently regenerate has `#!/bin/sh`, which doesn't include some features that bash has...

So, maybe remove this change and modify regenerate.sh, and then if we get a problem with that, fix it then.

src/core/datatypes/trees/Clade.cpp
bredelings: My intuition says "no"...

src/core/datatypes/phylogenetics/character/DnaState.cpp
hoehna: Why should this be reverted? The current code treats the missing states the same as '?', and 'N'. Should we treat these differently?

src/core/analysis/mcmc/output/TreeSummary.cpp
```diff
@@ -507,10 +507,21 @@ void TreeSummary::mapContinuous(Tree &tree, const std::string &n, size_t paramIn
                     {
 
                         root_checked = true;
-
-                        size_t sample_clade_index = sample_clade_indices[ summary_newick[j] ];
-
-                        const TopologyNode &sample_node = sample_tree.getNode( sample_clade_index );
+
```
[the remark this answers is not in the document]
hoehna: OK, I've removed the commented out code, but tbh I don't remember why this was commented out, nor why it was there in the first place.

src/core/datatypes/RbConstIteratorImpl.h
```diff
@@ -38,6 +38,7 @@ namespace RevBayesCore {
 //        RbConstIteratorImpl(const RbConstIteratorImpl<valueType,indicator> &v);
         
         // public (stl-like) vector functions
+//        RbConstIteratorImpl<valueType,indicator>            operator+(size_t i) const { return RbConstIteratorImpl<valueType,indicator>(it+i); }                                                                        //!< Increment index (prefix)
```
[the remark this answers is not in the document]
hoehna: These were not used and I didn't want to create orphan code. So I commented these functions out at the time. They are not deleted.

src/core/datatypes/RbConstIteratorImpl.h
```diff
@@ -68,6 +69,7 @@ namespace RevBayesCore {
 //        RbConstIteratorImpl(const RbConstIteratorImpl<valueType,1> &v);
         
         // public (stl-like) vector functions
+//        RbConstIteratorImpl<valueType,1>                    operator+(size_t i) const { return RbConstIteratorImpl<valueType,1>(it+i); }                                                                        //!< Increment index (prefix)
```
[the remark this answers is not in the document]
hoehna: See above (likewise) ;)

src/core/datatypes/RbVectorImpl.h
```diff
@@ -63,18 +64,31 @@ namespace RevBayesCore {
...
+        RbIterator<valueType>                               erase(RbIterator<valueType> pos) { return this->std::vector<valueType>::erase( pos.getStlIterator() ); }
+        RbConstIterator<valueType>                          erase(RbConstIterator<valueType> pos) { return this->std::vector<valueType>::erase( pos ); }
 
 //        valueType&                                          operator[](size_t i) { return values[i]; }
 //        const valueType&                                    operator[](size_t i) const { return values[i]; }
 //        void                                                clear(void) { values.clear(); }
 //        void                                                insert(size_t i, const valueType &v) { values[i] = v; }
 //        void                                                push_back(const valueType &v) { values.push_back( v ); }
+//        void                                                insert(size_t i, const valueType &v) { delete values[i]; values[i] = Cloner<valueType, IsDerivedFrom<valueType, Cloneable>::Is >::createClone( v ); }
```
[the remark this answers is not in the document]
hoehna: See above

src/core/moves/MetropolisHastingsMove.cpp
hoehna: Does a simple "throw" without giving the exception work here? If I add "throw e;", then I get a warning about throwing a local variable.

src/core/moves/MetropolisHastingsMove.cpp
hoehna: Same as above; does "throw;" actually propagate the error message?

src/core/moves/MetropolisHastingsMove.cpp
hoehna: Same as above, what do you by "by value"?

src/core/moves/MetropolisHastingsMove.cpp
hoehna: Same as before ...

projects/cmake/build.sh
hoehna: Not quite sure but this was necessary for me and a few other people to keep the compile working smoothless without any changes in how the compile is called. This might need a bit more discussion ...

projects/cmake/build.sh
hoehna: OK, for now I've reverted this change and we see when it pops up again ...

src/core/datatypes/trees/Clade.cpp
hoehna: well, it really depends in which setting this is used. But I guess we never have two differently named but otherwise identical clades which we wanted to compare.

I've commented out the code (with a comment)...

src/core/datatypes/trees/Clade.cpp
hoehna: Same as above. Probably not necessary and therefore commented out now.

src/revlanguage/workspace/RbRegister_Dist.cpp
hoehna: Done.

src/core/distributions/phylogenetics/characterhistory/AbstractTreeHistoryCtmcSiteIID.h
```diff
@@ -68,9 +68,9 @@
 //        double                                                              computeBranchRate(size_t index);
```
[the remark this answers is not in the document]
hoehna: I deleted the files, although we should check with Michael Landis about this.

src/core/distributions/phylogenetics/characterhistory/GeneralTreeHistoryCtmc.h
```diff
@@ -66,9 +66,9 @@ namespace RevBayesCore {
         virtual double                                      computeRootLikelihood(const TopologyNode &n);
         virtual double                                      computeInternalNodeLikelihood(const TopologyNode &n);
         virtual double                                      computeTipLikelihood(const TopologyNode &node);
-        // (not needed)        void                         keepSpecialization(DagNode* affecter);
-        // (not needed)        void                         restoreSpecialization(DagNode *restorer);
-        virtual void                                        touchSpecialization(DagNode *toucher, bool touchAll);
+        // (not needed)        void                         keepSpecialization(const DagNode* affecter);
```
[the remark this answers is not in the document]
hoehna: Not sure what should be deleted. Anyways, my change didn't affect any comments, so maybe we should rather go to the source of the problem.

src/core/distributions/phylogenetics/characterhistory/GeneralTreeHistoryCtmcSiteIID.h
```diff
@@ -54,9 +54,9 @@ namespace RevBayesCore {
         virtual double                                      computeRootLikelihood(const TopologyNode &n);
         virtual double                                      computeInternalNodeLikelihood(const TopologyNode &n);
         virtual double                                      computeTipLikelihood(const TopologyNode &node);
-        // (not needed)        void                         keepSpecialization(DagNode* affecter);
-        // (not needed)        void                         restoreSpecialization(DagNode *restorer);
-        virtual void                                        touchSpecialization(DagNode *toucher, bool touchAll);
+        // (not needed)        void                         keepSpecialization(const DagNode* affecter);
```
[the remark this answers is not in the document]
hoehna: OK, I've deleted these although I'm not sure what the primary author had in mind.

src/core/distributions/phylogenetics/continuousCharacter/PhyloBrownianProcessREML.cpp
```diff
@@ -159,12 +140,12 @@ void PhyloBrownianProcessREML::recursiveComputeLnProbability( const TopologyNode
 {
 
     // check for recomputation
-    if ( node.isTip() == false && dirty_nodes[node_index] )
+    if ( node.isTip() == false ) // && dirty_nodes[node_index] == true
```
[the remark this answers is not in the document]
hoehna: It is now used again and was only deleted as debug.

src/core/datatypes/math/MatrixReal.cpp
```diff
@@ -364,12 +419,6 @@ double MatrixReal::getLogDet() const
         
 //        if (std::isnan(tot))
```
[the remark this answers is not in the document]
hoehna: yes, that was for debugging and is now deleted

src/core/datatypes/math/MatrixReal.cpp
```diff
@@ -364,12 +419,6 @@ double MatrixReal::getLogDet() const
...
 //        {
-//            std::cerr << "in MatrixReal::getLogDet(): nan\n";
-//            std::cerr << "eigen values:\n";
-//            for (size_t i=0; i<nRows; i++)
-//            {
-//                std::cerr << eigenval[i] << '\n';
-//            }
 //            RbException("Problem when computing log-determinant.");
```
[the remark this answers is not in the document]
hoehna: Code is gone ...

src/core/moves/MetropolisHastingsMove.cpp
bredelings: By value is like `catch (RbException e)`.
By reference is like `catch (RbException& e)`.
Sorry, I was referring to a comment above, but it was still kind of unclear.

src/core/datatypes/phylogenetics/Taxon.cpp
```diff
@@ -198,8 +198,17 @@ const std::string& Taxon::getSpeciesName( void ) const
...
+    
+    if ( age_range.getMax() < a )
+    {
+        age_range.setMax(a);
+        age_range.setMin(a);
+    }
+    else
+    {
+        age_range.setMin(a);
+//        age_range.setMax(a);
```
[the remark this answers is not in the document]
hoehna: Yes, I've added a comment here.

src/core/distributions/UniformPartitioningDistribution.h
```diff
@@ -38,7 +38,7 @@ namespace RevBayesCore {
         UniformPartitioningDistribution*                    clone(void) const;                                                                      //!< Create an independent clone
         double                                              computeLnProbability(void);
         void                                                executeMethod(const std::string &n, const std::vector<const DagNode*> &args, long &rv) const;     //!< Map the member methods to internal function calls
-        const std::vector<valueType>&                       getParameterValues(void) const;
+//        const std::vector<valueType>&                       getParameterValues(void) const;
```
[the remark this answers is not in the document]
hoehna: Done

src/core/distributions/UniformPartitioningDistribution.h
```diff
@@ -217,11 +217,11 @@ int RevBayesCore::UniformPartitioningDistribution<valueType>::getCurrentIndex( v
 }
 
 
-template <class valueType>
-const std::vector<valueType>& RevBayesCore::UniformPartitioningDistribution<valueType>::getParameterValues( void ) const
-{
-    return parameter_values->getValue();
-}
+//template <class valueType>
```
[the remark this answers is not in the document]
hoehna: Done

src/core/distributions/phylogenetics/continuousCharacter/PhyloBranchRatesBM.cpp
```diff
@@ -75,8 +79,9 @@ double PhyloBranchRatesBM::recursiveLnProb( const TopologyNode& node, std::vecto
         }
         double node_value = log(ln_node_value);
         double stand_dev = sigma->getValue() * sqrt(node.getBranchLength());
-        double mean = parent_value + drift->getValue() * node.getBranchLength();
-        ln_prob += RbStatistics::Normal::lnPdf(node_value, stand_dev, mean);
+//        double mean = parent_value + drift->getValue() * node.getBranchLength();
```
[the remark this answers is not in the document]
hoehna: good point. I don't know but since drift seems to be used otherwise I've added it back in.

src/core/distributions/phylogenetics/tree/TopologyConstrainedTreeDistribution.cpp
hoehna: because I added code to work on unrooted tree too. I removed the check.

src/core/distributions/phylogenetics/tree/birthdeath/BirthDeathBurstProcess.cpp
hoehna: yes, we found bugs in the theory and tested these. Last I remember is that this now works.

src/core/moves/MetropolisHastingsMove.cpp
bredelings: Surprisingly it does! 
It re-throws the exception object that was caught, without making a copy.

src/core/distributions/phylogenetics/tree/birthdeath/BirthDeathBurstProcess.cpp
hoehna: New version works.

src/core/distributions/phylogenetics/tree/birthdeath/StateDependentSpeciationExtinctionProcess.cpp
hoehna: I've added an explanation.

src/core/distributions/phylogenetics/tree/birthdeath/StateDependentSpeciationExtinctionProcess.cpp
hoehna: I'ld rather keep it if we need to revert ...

src/core/analysis/mcmc/output/TraceNumeric.cpp
```diff
@@ -46,14 +46,14 @@ TraceNumeric* TraceNumeric::clone() const
...
 
-    // test stationarity within chain
-    size_t nBlocks = 10;
-    StationarityTest testS = StationarityTest(nBlocks, 0.01);
-    passedStationarityTest = testS.assessConvergence(*this);
-
-    // Geweke's test for convergence within a chain
-    GewekeTest testG = GewekeTest(0.01);
-    passedGewekeTest = testG.assessConvergence(*this);
+//    // test stationarity within chain
```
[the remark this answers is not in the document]
bredelings: If nobody is using it, then maybe remove it?

Or, if somebody might fix it, then may be add a comment like.  Maybe some thing like:
"Sebastian: This code doesn't work any more and caused some runs to crash.
                    I'm commenting it out in case somebody wants to try and fix it."

Up to you.

src/core/moves/MetropolisHastingsMove.cpp
hoehna: Hm, I'm not aware of this being an issue. If so, then should we check the entire code base?

bredelings: This is my example of catching exceptions by reference, instead of by value.

src/core/moves/MetropolisHastingsMove.cpp
bredelings: This is my explanation of `throw;` with no argument.

src/core/distributions/phylogenetics/substitution/CTMCProcess.h
```diff
@@ -0,0 +1,1285 @@
+#ifndef CTMCProcess_H
```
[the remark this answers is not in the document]
hoehna: No, this is a process that works for a single population. It also is used more for teaching.

src/core/distributions/phylogenetics/substitution/PopGenInfinitesSites.h
```diff
@@ -42,9 +42,9 @@ namespace RevBayesCore {
         
         
         // virtual methods that may be overwritten, but then the derived class should call this methods
-//        void                                                                keepSpecialization(DagNode* affecter);
-//        void                                                                restoreSpecialization(DagNode *restorer);
-//        void                                                                touchSpecialization(DagNode *toucher, bool touchAll);
+//        void                                                                keepSpecialization(const DagNode* affecter);
```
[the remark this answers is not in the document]
hoehna: Done.

src/core/distributions/phylogenetics/tree/AbstractRootedTreeDistribution.cpp
```diff
@@ -113,6 +176,8 @@ double AbstractRootedTreeDistribution::computeLnProbability( void )
     // proceed as long as derived classes validate a non-zero likeilhood
     if ( isLnProbabilityNonZero() == false )
     {
+        std::cerr << "Prob zero" << std::endl;
```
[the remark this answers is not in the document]
hoehna: No, this was actually debug code

src/core/distributions/phylogenetics/tree/AbstractRootedTreeDistribution.cpp
```diff
@@ -146,6 +211,8 @@ double AbstractRootedTreeDistribution::computeLnProbability( void )
             }
             else if( the_node.getAge() - the_node.getParent().getAge() > 0 )
             {
+                std::cerr << "node younger than parent" << std::endl;
```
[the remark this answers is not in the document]
hoehna: Nope, I removed this because it was for debugging.

src/core/distributions/phylogenetics/tree/AbstractRootedTreeDistribution.cpp
```diff
@@ -167,6 +234,8 @@ double AbstractRootedTreeDistribution::computeLnProbability( void )
     
     if ( ra > getOriginAge() || ra != getRootAge() )
     {
+        std::cerr << "root not root age" << std::endl;
```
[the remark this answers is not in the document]
hoehna: Nope, I removed this because it was for debugging.

src/core/distributions/phylogenetics/tree/AbstractRootedTreeDistribution.cpp
```diff
@@ -176,6 +245,8 @@ double AbstractRootedTreeDistribution::computeLnProbability( void )
     {
         if ( ra < (*it)->getAge() )
         {
+            std::cerr << "root younger than node" << std::endl;
```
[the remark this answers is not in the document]
hoehna: Nope, I removed this because it was for debugging.

src/core/distributions/phylogenetics/tree/birthdeath/AbstractBirthDeathProcess.cpp
```diff
@@ -28,9 +28,10 @@ using namespace RevBayesCore;
  * \param    cdt       The condition of the process (time/survival/nTaxa)
  * \param    tn        Taxon names used during initialization.
  * \param    uo        If true ra is the origin time otherwise the root age of the process.
+ * \param    t         The starting tree if we want to avoid simulating trees.
```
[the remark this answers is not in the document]
hoehna: Yes, that's the idea.

projects/cmake/build.sh
bredelings: Yeah, more discussion might be good.  We should definitely fix the scripts so that it works smoothly for you and the few other people -- I'm worried that this would maybe break it for people that it is currently working for.

Let's figure out why the `-D Boost_NO_BOOST_CMAKE:BOOL=ON` was necessary.  It is possible that this might depend on the version of boost -- i.e. with the latest version this might not be needed.

bredelings:
@hoehna I think if you merge development into here that will fix the tests passing.  @bjoelle has a fix to the URL for boost.

src/revlanguage/distributions/phylogenetics/tree/Dist_FBDP.cpp
hoehna: Thanks!

src/revlanguage/distributions/phylogenetics/tree/Dist_FBDP.cpp
hoehna: This was to avoid double usage of the variable psi. There were situations were internally a new local variable psi was used for the tree, e.g., when simulating trees. I think that it might be dangerous to use local variables with the same name, so we changed it.

src/revlanguage/main.cpp
hoehna: Not sure about the performance, but I thought it might be nicer to use the same coding style throughout the codebase (although I might fighting a lost cause here).

projects/cmake/build.sh
willpett: Yeah, I thought this was the purpose of `$cmake_args`? It allows for user-specific build arguments that may depend on their local environment (i.e. different Boost versions, install locations, CMake versions, etc).

src/core/datatypes/phylogenetics/Taxon.cpp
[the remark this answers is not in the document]
willpett: Why does this only set min if max > a?

src/revlanguage/distributions/phylogenetics/tree/Dist_episodicBirthDeath.cpp
willpett: Should every distribution have a starting value argument?

willpett:
@hoehna Any progress on this? What is the current status of this pull request?

projects/cmake/build.sh
willpett: It's probably safe to use module mode by default though, so if we want to keep this, I'd suggest just putting `set(Boost_NO_BOOST_CMAKE ON)` into `CMakeLists.txt`

willpett:
I tried running the tests locally on a couple different systems and reproduced the same set of failures as GitHub actions (`ChromoSSE` and `FBD`). I am not getting other failures. What type of system are you running @hoehna?

willpett:
Also, after looking at the `mk` validation tests, I'm assuming the reason they are failing is because the `phyloCTMC` is not being simulated from the prior, but is instead clamped using data from `bears.nex`.

hoehna:
Here are the tests results for me:
#### Checking output from tests... 
#### Test passed: tutorials/ctmc
#### Test passed: tutorials/dating
#### Test passed: tutorials/fbd
#### Test passed: BDP
#### Test passed: BDP_sampling
#### Test passed: BDP_survival
#### Test passed: BM
#### Test passed: BM_MVN
#### Test passed: BP
>>>> Test failed: CBDSP
	mismatch: primates_CBDSP.trees
	mismatch: primates_CBDSP_MAP.tree
>>>> Test failed: ChromoSSE
	mismatch: mcmc_ChromoSSE.out
#### Test passed: Coal_const
#### Test passed: Coal_skyline
#### Test passed: DEC
#### Test passed: EBD
>>>> Test failed: FBD
	mismatch: FBDP.trees
#### Test passed: HSMRF_BDP
>>>> Test failed: JC_BDP
	mismatch: mcmc_JC_BDP.out
>>>> Test failed: JC_EBD
	mismatch: mcmc_JC_EBD.out
>>>> Test failed: JC_unrooted
	mismatch: mcmc_JC_unrooted_uniform.out
#### Test passed: MVN_move
>>>> Test failed: Mk
	mismatch: Mk.log
	mismatch: Mk.trees
	mismatch: Mki.log
	mismatch: Mki.trees
	mismatch: Mkv.log
	mismatch: Mkv.trees
#### Test passed: OU
>>>> Test failed: Partition
	mismatch: PS_codon_map.tre
	mismatch: PS_codon_run_1.log
	mismatch: PS_codon_run_1.trees
	mismatch: PS_codon_run_2.log
	mismatch: PS_codon_run_2.trees
	mismatch: PS_gene_map.tre
	mismatch: PS_gene_run_1.log
	mismatch: PS_gene_run_1.trees
	mismatch: PS_gene_run_2.log
	mismatch: PS_gene_run_2.trees
	mismatch: PS_uniform_map.tre
	mismatch: PS_uniform_run_1.log
	mismatch: PS_uniform_run_1.trees
	mismatch: PS_uniform_run_2.log
	mismatch: PS_uniform_run_2.trees
>>>> Test failed: UCLD_noncentered
	mismatch: UCLD_noncentered.trees
>>>> Test failed: basics
	mismatch: Test_basic.txt
>>>> Test failed: multiple_mcmc
	mismatch: mcmc_HKY.out
	mismatch: mcmc_HKY_run_1.out
	mismatch: mcmc_HKY_run_2.out
	mismatch: mcmc_JC.out
	mismatch: mcmc_JC_run_1.out
	mismatch: mcmc_JC_run_2.out
#### Test passed: mvBM
#### Test passed: skygrid
>>>> Test failed: steppingstone
	mismatch: PS_uniform_marginal_likelihood.out
	mismatch: PS_uniform_stone_0.log
	mismatch: PS_uniform_stone_0.trees
	mismatch: PS_uniform_stone_1.log
	mismatch: PS_uniform_stone_1.trees
	mismatch: PS_uniform_stone_10.log
	mismatch: PS_uniform_stone_10.trees
	mismatch: PS_uniform_stone_2.log
	mismatch: PS_uniform_stone_2.trees
	mismatch: PS_uniform_stone_3.log
	mismatch: PS_uniform_stone_3.trees
	mismatch: PS_uniform_stone_4.log
	mismatch: PS_uniform_stone_4.trees
	mismatch: PS_uniform_stone_5.log
	mismatch: PS_uniform_stone_5.trees
	mismatch: PS_uniform_stone_6.log
	mismatch: PS_uniform_stone_6.trees
	mismatch: PS_uniform_stone_7.log
	mismatch: PS_uniform_stone_7.trees
	mismatch: PS_uniform_stone_8.log
	mismatch: PS_uniform_stone_8.trees
	mismatch: PS_uniform_stone_9.log
	mismatch: PS_uniform_stone_9.trees
	mismatch: model_uniform.out
	mismatch: model_uniform_stone_0.out
	mismatch: model_uniform_stone_1.out
	mismatch: model_uniform_stone_10.out
	mismatch: model_uniform_stone_2.out
	mismatch: model_uniform_stone_3.out
	mismatch: model_uniform_stone_4.out
	mismatch: model_uniform_stone_5.out
	mismatch: model_uniform_stone_6.out
	mismatch: model_uniform_stone_7.out
	mismatch: model_uniform_stone_8.out
	mismatch: model_uniform_stone_9.out

src/core/distributions/phylogenetics/tree/birthdeath/BirthDeathProcess.cpp
hoehna: Should be cleaned up now. See commit 23890f2093ad82dbac200152daef01659a467f7b

src/core/distributions/phylogenetics/tree/birthdeath/EpisodicBirthDeathProcess.h
hoehna: Done, see commit 068ea9105f4dfc96cb18b414809dadaa8acdf2b9

projects/cmake/build.sh
hoehna: OK, this has been reverted and pushed to a separate branch.

src/core/analysis/Model.cpp
```diff
@@ -321,13 +318,16 @@ void Model::getOrderedStochasticNodes(const DagNode* dagNode, std::vector<DagNod
     }
 
     // Then I can add myself to the nodes visited, and to the ordered vector of stochastic nodes
-    if ( dagNode->isStochastic() ) //if the node is stochastic
```
bredelings, line 324: Wow, cool change.

I wonder if this fixes the issue with
```for (i in 1:98) {

  dx[i] ~ dnNormal(0,1)

  x[i+2] := dx[i] - x[i] + 2 * x[i+1]

  # # Use below print statement to watch how time to compute probability slows as model is built

  print(x[1].lnProbability())
}
```

src/core/analysis/Model.cpp
```diff
@@ -321,13 +318,16 @@ void Model::getOrderedStochasticNodes(const DagNode* dagNode, std::vector<DagNod
...
+    if ( the_dag_node->isStochastic() && (visitedNodes.find(the_dag_node) == visitedNodes.end()) ) //if the node is stochastic
```
bredelings, line 321: Wow, cool change!  I wonder if this fixes the problem with:
```
for (i in 1:98) {

  dx[i] ~ dnNormal(0,1)

  x[i+2] := dx[i] - x[i] + 2 * x[i+1]

  # # Use below print statement to watch how time to compute probability slows as model is built

  print(x[1].lnProbability())
}
```

src/core/distributions/phylogenetics/tree/MultispeciesCoalescent.cpp
```diff
@@ -32,66 +34,68 @@ MultispeciesCoalescent::~MultispeciesCoalescent()
...
+
         // compute the number of pairs: pairs = C(j choose 2) = j * (j-1.0) / 2.0
         double n_pairs = j * (j-1.0) / 2.0;
         double lambda = n_pairs * theta;
-        
+
         // add the density for this coalescent event
         //lnProbCoal += log( lambda ) - lambda * a;
         //Corrected version:
         ln_prob_coal += log( theta ) - lambda * a;
```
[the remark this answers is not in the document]
hoehna: I'm deferring this request to the branch `hsiang_dev`. Allison has been working on fixing several of the multispecies coalescent implementations, and the fixes will be made when we merge her branch.

src/core/distributions/phylogenetics/tree/MultispeciesCoalescentUniformPrior.cpp
hoehna: This implementation is not working and I'm deferring to the corrected implementation in the branch `hsiang_dev`.

src/core/distributions/phylogenetics/tree/MultispeciesCoalescentUniformPrior.cpp
hoehna: This implementation does not work and I'm deferring to the correcting version in the `hsiang_dev` branch.

src/core/moves/MetropolisHastingsMove.cpp
hoehna: OK, I changed these in commit a4f5cf43.

src/core/moves/MetropolisHastingsMove.cpp
hoehna: OK, I changed to catch by reference in commit a4f5cf43

src/core/moves/MetropolisHastingsMove.cpp
hoehna: I've changed this to `throw e`. For me that is more readable than `throw`. Should be the same, right? Now changed in commit 95e107f0.

src/core/moves/MetropolisHastingsMove.cpp
hoehna: Also changed in previous commit

src/core/moves/MetropolisHastingsMove.cpp
hoehna: Changed in previous commit.

src/core/moves/MetropolisHastingsMove.cpp
hoehna: Changed with the previous commit.

src/core/moves/MetropolisHastingsMove.cpp
hoehna: Changed with previous commit.

src/core/moves/MetropolisHastingsMove.cpp
hoehna: Changed.

src/core/moves/MetropolisHastingsMove.cpp
hoehna: Changed

src/core/distributions/phylogenetics/tree/MultispeciesCoalescentUniformPrior.cpp
```diff
@@ -34,87 +36,186 @@ MultispeciesCoalescentUniformPrior::~MultispeciesCoalescentUniformPrior()
...
+    {
+        double ngc = getNumberOfGeneCopies();
+        double integral_limit = 2 * fn / theta_max;
+
+        // std::cout << "fn total: " << fn << std::endl;
+        // std::cout << "ngc: " << ngc << std::endl;
+        // std::cout << "integral limit: " << integral_limit << std::endl;
+
+        double upper_incomplete_gamma = 0.0;
+        if ( ngc <= 2 )
```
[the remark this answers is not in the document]
hoehna: This implementation does not work. I defer to the new implementation in the branch `hsiang_dev`.

src/core/distributions/phylogenetics/tree/birthdeath/AbstractPiecewiseConstantFossilizedRangeProcess.cpp
```diff
@@ -267,7 +267,9 @@ double AbstractPiecewiseConstantFossilizedRangeProcess::computeLnProbabilityRang
                 return RbConstants::Double::neginf;
             }
         }
-        else if ( !( b > o && o >= y && y >= d && d >= 0.0 ) )
+        // else if ( !( b > o && o >= y && y >= d && d >= 0.0 ) )
+        // Sebastian: This code is not safe if the values (y and d) are not exactly equal but differ due to machine rounding
```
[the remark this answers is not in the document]
hoehna: I was reading in trees where this condition was not matched, and therefore I could never get a valid likelihood. We do have some epsilon in the RbSettings called `tolerance`.

src/core/distributions/phylogenetics/tree/birthdeath/BirthDeathBurstProcess.cpp
hoehna: My tests showed that if a tree is read in from a file, there are a lot of rounding errors for branch lengths and thus node ages. This means that the events for a simulated tree written to a file didn't match when read in again. The problem with a global threshold is that unfortunately there is not one threshold that works in all situations, because the rounding has different impacts in different situations.

src/core/functions/phylogenetics/ratematrix/CovarionFunction.cpp
hoehna: Sure, I cleaned up the code in commit `0e001e47`

src/core/utils/BirthDeathForwardSimulator.cpp
hoehna: Sure, I removed this debug output in commit `5f590d1b`.

src/revlanguage/distributions/phylogenetics/character/Dist_CTMC.h
hoehna: It's for a single population/branch, so no tree is used. I'm working on something to use this process to model demographic processes. I've added a bit of description in commit `47d45a38`.

src/revlanguage/distributions/phylogenetics/tree/Dist_BDSTP.cpp
hoehna: has been added

src/revlanguage/distributions/phylogenetics/tree/Dist_episodicBirthDeath.cpp
hoehna: Probably yes. I'm leaving this for other work.

src/core/analysis/Model.cpp
hoehna: No, that still seems to be a problem.

bjoelle (approved):

src/core/moves/MetropolisHastingsMove.cpp
bredelings: It is actually not the same.  `throw e` constructs a new RbException on the stack, whereas `throw` rethrows the one that was caught, without constructing a new one.

hoehna: Hmm, I can change it to simply use throw, although I find that less readable. Does it make a difference?

bredelings (approved):
I have a few comments, but it looks good.

willpett (approved):

Issue #135, opened 2021-02-02
rbinomial() has two arguments named n


jembrown:
`rbinomial()` uses `n` for both the number of values to draw and the number of trials, making it impossible to provide argument names for both values. R uses `size` for the number of trials to avoid this.

PR #148, opened 2021-02-17
Rename rbinomial, rnbinomial and rmultinomial argument names to size

willpett:
This solves issue #135 and brings the Rev syntax inline with R.


bjoelle (approved):

willpett:
I realized the `size` argument in R is actually the number of trials, not the number of random draws. So I instead changed a few distributions to reflect this.

hoehna (approved):
beatiful

PR #150, opened 2021-02-19
Generalize the incomplete gamma function

willpett:
The internal RevBayes function for the incomplete gamma integral was implemented using [this algorithm](https://people.sc.fsu.edu/~jburkardt/cpp_src/asa032/asa032.html) for the regularized lower incomplete gamma function. I've generalized the function so now you can choose whether it should be regularized, and whether to use the upper or lower limits of integration.

The previous function had an extra argument `scale` that determined the scale over which the function should be regularized.
I didn't see any need for this since it is always just &Gamma;(&alpha;). The only reason I can think this was included was for some small optimization from eliminating repeated calls to the gamma function, or simply for hidden testing purposes. I went ahead and removed it.


willpett:
Now this is failing the testiphy `missing` test because I don't have the changes from `dev-hoehna-lab`... or did I do something wrong?

bredelings (approved):
Nice simplification.

PR #151, opened 2021-02-19
Fix weird ambiguous char issue

bredelings:
Factor out code that modifies or scans the data matrix into separate functions.
    
Previously, three of these functions were in a single loop, which could result in some of the modification stopping on the first ambiguous character.

### DIFF
--- a/src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
+++ b/src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
@@ -503,7 +503,88 @@ void RevBayesCore::AbstractPhyloCTMCSiteHomogeneous<charType>::bootstrap( void )
     pattern_counts = bootstrapped_pattern_counts;
 
 }
+namespace RevBayesCore
+{
+inline void mark_ambiguous_and_missing_as_gap(AbstractHomologousDiscreteCharacterData& data, const vector<size_t>& site_indices, std::vector<TopologyNode*> nodes)
+{
+    for (auto& node: nodes)
+    {
+        if ( node->isTip() )
+        {
+            AbstractDiscreteTaxonData& taxon_data = data.getTaxonData( node->getName() );
+            for (auto site_index: site_indices)
+            {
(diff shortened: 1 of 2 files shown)


willpett (approved):
It looks like the issue was that each site was only allowed to account for a single invariant site pattern, but more than one may be possible at sites with ambiguous characters, as @bredelings pointed out.

willpett:
There is still the question of whether `allow_ambiguous_as_invariant` should be `true` or `false`. I would argue that a site with gaps cannot be invariant since it implies some indels have occurred. So maybe invariant sites should be marked after marking gaps, but before marking ambiguous/unknown characters as gaps, i.e. when using `treatAmbiguousAsGaps` or `treatUnknownAsGap`

bredelings:
@willpett Thanks for the review!

I think the traditional approach is to assume that:

1. the +INV model adds a mixture component with rate=0 -- it doesn't explicitly label any columns as "invariant" or not.  That is unique to revbayes.

2. all substitution models, including +INV model, treat `-` as `N`.

They are therefore assuming that `allow_gaps_as_invariant` should be `true`.

It would be nice to find an original paper describing the +INV model.  Maybe an alternative would to be see what other software programs do?

bredelings:
@willpett Oops, I didn't notice that you had added a patch!  However, it looks like your patch handles (i) a column where all characters are (say) `R` and also (ii) handles gaps generically without needing to set their bitvector to 1111111... in every alphabet.  Nice :-)

willpett:
> 1. the +INV model adds a mixture component with rate=0 -- it doesn't explicitly label any columns as "invariant" or not. That is unique to revbayes.

RevBayes doesn't label columns as invariant, it just labels columns that could potentially be invariant. Then it computes the likelihood for those sites assuming a mixture component with rate=0 like you said. Variable sites cannot be invariant so the likelihood is computed without the extra rate component for those.

> 2. all substitution models, including +INV model, treat - as N

Yeah I guess it is fair to say the substitution rate at an otherwise invariant site with gaps could be zero with some probability... because indels aren't "substitutions" in the sense that this model is treating them.

PR #152, opened 2021-02-20
small fix to incomplete gamma function

willpett:
Fixing a small error I missed in #150.

### DIFF
--- a/src/core/math/RbMathFunctions.cpp
+++ b/src/core/math/RbMathFunctions.cpp
@@ -701,7 +701,7 @@ double RbMath::incompleteGamma(double x, double alpha, bool regularized, bool lo
                 pn3 /= overflow;
             }
         } while (true);
-	gin = 1.0 - factor * gin;
+        gin = 1.0 - factor * gin;
     } else {
         // series expansion
         gin = 1;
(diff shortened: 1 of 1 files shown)


src/core/math/RbMathFunctions.cpp
```diff
@@ -716,12 +716,10 @@ double RbMath::incompleteGamma(double x, double alpha, bool regularized, bool lo
...
     }
 
-    if ( lower = false )
-    {
-        gin = 1.0 - gin;
-    }
-
-    return regularized ? gin : exp(log(gin) + scale);
+    gin = lower ? gin : 1.0 - gin;
+    gin = regularized ? gin : exp(log(gin) + scale);
```
bredelings, line 720: Is there a reason here to do `exp(log(gin)+scale)` versus (say) `gin*exp(scale)`?

willpett: It's just a one-liner log-sum-exp. It's to avoid under/overflow the event that gin is very small and/or scale is very large.

willpett: I suppose I could add a catch for overflow.

bredelings (approved):

PR #169, opened 2021-04-16
Update slice sampling

bredelings:
Hi,

This updates slice sampling to use the "doubling" strategy instead of the "stepping out" strategy to find slice boundaries.

It also introduces a new argument to mvSlice to select between the two strategies -- by the new "doubling" strategy is used by default.

I have done some basic testing, and the "doubling" strategy can definitely speed up burnin, in cases where you want a large step size initially, but not generally.

-BenRI

### DIFF
--- a/src/core/moves/SliceSamplingMove.cpp
+++ b/src/core/moves/SliceSamplingMove.cpp
@@ -5,6 +5,9 @@
 #include <iostream>
 #include <utility>
 #include <vector>
+#include <tuple>
+
+#include <boost/optional.hpp>
 
 #include "DagNode.h"
 #include "SliceSamplingMove.h"
@@ -14,8 +17,14 @@
 #include "RbOrderedSet.h"
 #include "StochasticNode.h"
 
+using boost::optional;
+
 using namespace RevBayesCore;
 
+const double log_0 = RbConstants::Double::min;
+
+// How can we make a run-time adjustable log level?
+int log_level = 0;
 
 /** 
  * Constructor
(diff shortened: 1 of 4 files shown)


src/core/moves/SliceSamplingMove.cpp
```diff
@@ -14,8 +17,14 @@
...
 #include "StochasticNode.h"
 
+using boost::optional;
+
 using namespace RevBayesCore;
 
+const double log_0 = RbConstants::Double::min;
+
+// How can we make a run-time adjustable log level?
+int log_level = 0;
```
bjoelle, line 27: I think these declarations would be better with the other members in the h file

bjoelle, line 27: should the log level be a static member ?

bredelings: Actually I was thinking that revbayes should have an option for a logging level, and this is a placeholder for it.  Right now revbayes has options for `linewidth`, 'outputPrecision', and a few other things (see `core/utils/RbSettings.cpp`).

bredelings: I think you need an `extern` declaration in the header file, and another declaration in the `.cpp` file.  Otherwise the symbol will be defined in multiple `*.o` files and the linker will give an error.

For the moment I wasn't planning to export the `log_0` symbol.  However, if other `.cpp` files wanted to use `log_0`, then should it defined somewhere more central than SliceSamplingMove.h?

Long-term, I was thinking of adding a type called log_double_t that stores the log of a number, but supports all the normal operators, like *, +, etc.  something like...

```
class log_double_t {
    /// Natural log of the number.
    double value;
public:

    /// Access the log of the number.
    constexpr double  log() const {return value;}
    /// Access the log of the number.
    constexpr double& log()       {return value;}

    log_double_t& operator +=(const log_double_t& y) {loginc(value,y.log()); return *this;}
    log_double_t& operator -=(const log_double_t& y) {value = logdiff(log(),y.log()); return *this;}
    constexpr log_double_t& operator *=(const log_double_t& y) {value += y.log(); return *this;}
    constexpr log_double_t& operator /=(const log_double_t& y) {value -= y.log(); return *this;}

    operator double() const {return exp(value);}

    constexpr log_double_t& operator=(double y) {
        assert(y >= 0);
        if (y == 0)
            value = log_0;
        else if (y == 1)
            value = 0;
        else
            value = ::log(y);
        return *this;
    }

    constexpr log_double_t():value(log_0) {}

    constexpr log_double_t(double x):value(0) {
        operator=(x);
    }
};

...
```

bjoelle: I didn't get that these were intended to be general, sorry ! I thought they were members of the `SliceSamplingMove` class
since log_0 is directly referencing from RbConstants here I'm not sure if it's worth making a new symbol for it ? but if you think it is clearer that way I would think putting it in RbConstants as well is the way to go

I like the log_double idea but I think overloading the assignment operator would be confusing, I would use a different name to make it clear that there is a transformation happening (`setToLogOf(y)` ?)

bjoelle: I think it would fine to add the log level directly to RbSettings 
and we can make a note that this exists so more people can use it in their own code - maybe file it as a "feature request" in the issues ?

willpett (approved):

PR #172, opened 2021-05-18
fixing a bug in mnSiteMixtureAllocation

mikeryanmay:
There was a bug in mnSiteMixtureAllocation that transposed the site-specific allocations for site-rates and site matrices. I fixed it!

### DIFF
--- a/src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
+++ b/src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
@@ -1297,33 +1297,8 @@ void RevBayesCore::AbstractPhyloCTMCSiteHomogeneous<charType>::drawStochasticCha
         std::string simmap_string = "{" + end_states[root_index][site].getStringValue() + "," + StringUtilities::toString( root.getBranchLength() ) + "}";
         character_histories[root_index] = simmap_string;
 
-        // the mixture components are in a vector that is a flattened version of a
-        // matrix with rate components in columns and matrix components in rows.
-        // the matrix is in row-major order.
-        // here, we compute the row and column indices from the vector index.
-
-        size_t mixture_component_index = this->sampled_site_mixtures[site];
-
-        // get the number of rate categories
-        size_t num_site_rates = 1;
-        if (this->site_rates != NULL)
-        {
-        	num_site_rates = this->site_rates->getValue().size();
-        }
-
-        // determine the rate (column index)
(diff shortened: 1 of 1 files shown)


mikeryanmay:
Very confused about the pull request failing all of the checks! I created this branch from development within the last 15 minutes, so it's unclear to me what the problem is. Any assistance would be greatly appreciated.

bjoelle:
@mikeryanmay it's not RevBayes, there is an issue with the Boost installation

bredelings:
Ah, this is trying to download boost from bintray, but bintray is going away.

bredelings:
"Instead of: `https://dl.bintray.com/boostorg/release`/ you should use `https://boostorg.jfrog.io/artifactory/main/release/` to retrieve boost releases."

bjoelle:
@bredelings I've updated the link on the branch `actions_fix`, will PR if it passes

bjoelle:
builds are fixed in PR #173

bredelings:
Is this also broken in `drawStochasticCharacterMap` (same file)?

Perhaps the common code can be factored out into an inline function?

mikeryanmay:
Yes, it's also broken there. Thanks for catching that. 

An easy solution is to have `drawStochasticCharacterMap` call `getSampledMixtureComponents`. Are you suggesting turning `getSampledMixtureComponents` into an inline member function (the common code is effectively the entire body of `getSampledMixtureComponents`)?

bredelings:
Hmm... I guess I was actually thinking of an inline function that decomposed an arbitrary index into its `rate_component` part and its `matrix_component` part.

The logical thing is to make it return a pair<int,int>, but probably the current method of returning the results through parameters that are passed by reference would be better for now.  Until c++17, we cannot do e.g. `auto [x,y] = function_that_returns_a_pair();`.

Anyway, I think whatever way you implement it is fine as long as the code isn't duplicated.  So, if you want to just call getSampledMixtureComponents from drawStochasticCharacterMap, then I'm fine with that, although that wasn't what I was originally thinking.

mikeryanmay:
Yes, I see the logic in such a function. I think I'd rather fix this straightforwardly, then return later for a more serious overhaul (there are other performance issues with this code that need to be cleaned up).

bredelings:
OK, then perhaps just do something (anything) to fix `drawStochasticCharacterMap` and make sure that the comment there is correct, or remove it?

mikeryanmay:
Done! (I think)

bredelings (approved):

bredelings:
Ah, right, I didn't notice.  Great!

PR #176, opened 2021-06-02
allowing site mixture allocation for codons; moving site mixture allo…

mikeryanmay:
A minor fix to allow the site-mixture-allocation monitor to work for codon datatypes.


bredelings:
@bjoelle It looks like this is failing to download boost again.  The URL seems right, but the `tar -zxf` command after the download is complaining that the tarball is not in gzip format.

bjoelle:
@bredelings is that up to date ? it's complaining of conflicts with `development` from what I can see

mikeryanmay:
Oh this must be my mistake! I didn't update my local development before the pull request. My mistake.

bredelings:
Can you write a few words about what is happening here?
It looks like the patch has three parts -- the part that checks "Codons" seems fairly simple, but the other parts are a bit harder to follow from just the diff.

For example, I see that there is a new function called `drawSiteMixtureAllocations()` that replaces `drawJointConditionalAncestralStates(startStates, endStates);`

What is the difference?

mikeryanmay:
I've added some (hopefully clarifying) comments to the code as well as the commit message. Let me know if anything is still unclear.

bredelings:
I think I see now.

My concern is that the new `drawSiteMixtureAllocations()` duplicates a ton of code from `drawJointConditionalAncestralStates()`.  The code duplication seems bad, so I'm not sure that change is a net improvement, as it is.

It's possible to split out the change to work on "Codons" as a separate patch...

One possible way to get around the code duplication would be to make a single function to sample both the mixture components and the ancestral states for the root node.  You could make this function pass back a pair of vectors (vec1=mixture components, vec2=root states).  Then you could call this function from both `drawSiteMixtureAllocations()` and  `drawJointConditionalAncestralStates()`.  Then each caller can decide whether to set `sampled_site_mixtures=vec1` and `startStates[root.getIndex()]=vec2`.

willpett:
@bredelings are you requesting changes here?

bredelings:
Well, I did request that the amount of code duplication be decreased... but its not a correctness issue.  So, if you want to approve and merge this, its fine by me.

willpett (approved):

Issue #139, opened 2021-02-09
RNG discrepancies resulting in test failures


bjoelle:
There is an issue where tests pass on a local machine but then fail on the github-actions. Seen in PR #136 last time.

It seems related to changes to the RNG in Boost which leads to different versions producing different outputs. I'm not sure why the FBD test was the only affected, it could be due to it being longer and/or using more operators.

I'm opening this issue mostly as a record in case people run into it again.

willpett:
I think this is actually caused by different implementations of `libm` on different platforms, leading to slightly different results from standard math functions like `exp`, `log`, `sin` etc. Switching to a common implementation across all platforms (`openlibm`) has solved this issue on the `fbd` branch.

PR #183, opened 2021-09-02
Use openlibm for GitHub Actions builds

willpett:
Sometimes, the FBD integration tests were failing on some platforms but not others, leading to speculation that different versions of the boost RNG were to blame (see #139). Actually, the discrepancies were caused by different implementations of the standard math library `libm`, causing minute differences in the output of `std::exp`. This modifies the GitHub Actions build so that all platforms use a common implementation, [openlibm](https://openlibm.org/).

I also added a couple other compiler flags (`-mfpmath=sse -ffp-contract=off`) to the integration tests to help ensure more deterministic floating point math across platforms (especially 32-bit systems).

Finally, a few small fixes to the build script make sure that all platform-specific artifacts are properly uploaded.


.github/workflows/build.yml
```diff
@@ -117,15 +117,15 @@ jobs:
...
-          brew install pkg-config pandoc boost ccache coreutils
+          brew install pkg-config pandoc boost ccache coreutils openlibm
           sudo xcode-select -switch /Applications/Xcode_${{ matrix.version }}.app
           ccache --set-config=cache_dir=$HOME/.ccache
 
     - name: Install (Windows meson cross compile)
       if: matrix.name == 'windows'
       run: |
           sudo apt-get update
-          sudo apt-get install -y pandoc libcairo2-dev ccache dos2unix zstd
```
bredelings, line 128: I would think that we don't need this for a cross-compile?

bredelings (approved):
This looks good!

.github/workflows/build.yml
willpett: I don't think there's an alternative package without them.

willpett: oh wait nevermind, I see what you're saying.

PR #186, opened 2021-10-02
Consolidate file monitor arguments

willpett:
Collect the arguments relevant for monitors that print to a file in a separate base class. This makes the names and descriptions consistent across monitors (e.g. the use of the word delimiter vs. separator).


wadedismukes (approved):
Looks good!

PR #187, opened 2021-10-02
Improved reading of delimited files

willpett:
This pull request does several things:

1. Consolidate parsing of strings with delimiters into `StringUtilities::stringSplit`. This was to avoid multiple implementations of the same functionality.
2. Change the behavior of `StringUtilities::stringSplit` to allow an arbitrary amount of whitespace to serve as a delimiter. This was to allow files delimited with whitespace to be formatted in a more natural, human-readable way.
3. Change the default delimiter to whitespace on nearly all delimited file readers. The default in most cases was already `\t` and so this change is backwards compatible.
4. Fix inconsistent use of `separator` vs. `delimiter` in various file reading function arguments. Some functions used `separator`, others used `delimiter`.
5. Rename some functions to have more natural-sounding names. For example, I think `readDelimitedDataFile` is more grammatical than `readDataDelimitedFile`.
6. Various other small fixes.

### DIFF
--- a/src/core/datatypes/phylogenetics/TreeDiscreteCharacterData.cpp
+++ b/src/core/datatypes/phylogenetics/TreeDiscreteCharacterData.cpp
@@ -149,7 +149,7 @@ void TreeDiscreteCharacterData::writeToFile(const std::string &dir, const std::s
         // tab-delimited file
         fm = RbFileManager(dir, fn + ".tsv");
         RevBayesCore::DelimitedCharacterDataWriter writer; 
-        writer.writeData(fm.getFullFileName(), *character_data, "\t"[0]);
+        writer.writeData(fm.getFullFileName(), *character_data);
 
         // write the character history's time spent in each state
         fm = RbFileManager(dir, fn + "_time_in_states.tsv");
--- a/src/core/datatypes/phylogenetics/characterdata/AbstractHomologousDiscreteCharacterData.cpp
+++ b/src/core/datatypes/phylogenetics/characterdata/AbstractHomologousDiscreteCharacterData.cpp
@@ -17,7 +17,7 @@ void AbstractHomologousDiscreteCharacterData::writeToFile(const std::string &dir
         // NEXUS does not support NaturalNumbers so write tab delimited file
         RbFileManager fm = RbFileManager(dir, fn + ".tsv");
         RevBayesCore::DelimitedCharacterDataWriter writer; 
-        writer.writeData(fm.getFullFileName(), *this, "\t"[0]);
+        writer.writeData(fm.getFullFileName(), *this);
     }
     else
     {
--- a/src/core/io/BranchLengthDistributionReader.cpp
+++ b/src/core/io/BranchLengthDistributionReader.cpp
@@ -7,10 +7,8 @@
 using namespace RevBayesCore;
 
 
-BranchLengthDistributionReader::BranchLengthDistributionReader(const std::string &fn, char d, size_t numSkipped) : DelimitedDataReader(fn, d, numSkipped)
+BranchLengthDistributionReader::BranchLengthDistributionReader(const std::string &fn, std::string d, size_t numSkipped) : DelimitedDataReader(fn, d, numSkipped)
 {
-	
-	filename = fn;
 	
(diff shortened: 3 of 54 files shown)


bredelings:
This touches a lot of files!

Interesting points so far:
* the empty string is used to indicate the default of whitespace delimited files.  In a lot of call sites, this meaning of "" isn't obvious.
* stripping leading and trailing whitespace is an interesting change.  Mostly because this could change the meaning of TSV files with empty fields.  I guess you could use "\t" as the delimiter in that case.
* some function names were changed -- did you add aliases to avoid breaking old code that used them?
* adding "separator" and "delimiter" as synonymous arguments seems cool.  But "separator/delimiter"?  Not so sure.

willpett:
> * the empty string is used to indicate the default of whitespace delimited files.  In a lot of call sites, this meaning of "" isn't obvious.

That's true. I could add a note to argument descriptions to explain that?

> * stripping leading and trailing whitespace is an interesting change.  Mostly because this could change the meaning of TSV files with empty fields.  I guess you could use "\t" as the delimiter in that case.

Yeah I figured if there are explicitly empty fields delimited by some whitespace character, then the user should provide the delimiter.

> * some function names were changed -- did you add aliases to avoid breaking old code that used them?

Yes, I added the appropriate aliases. None of the test scripts were changed.

> * adding "separator" and "delimiter" as synonymous arguments seems cool.  But "separator/delimiter"?  Not so sure.

This is necessary in case the user doesn't specify a value for the argument. When an argument has multiple possible labels, but the user doesn't specify a value (i.e. it just uses the default value), its label becomes a concatenation of all possible labels joined by slashes. Maybe this isn't how it should work, but that's how argument labels are handled currently.

willpett:
> > * the empty string is used to indicate the default of whitespace delimited files.  In a lot of call sites, this meaning of "" isn't obvious.
> 
> That's true. I could add a note to argument descriptions to explain that?

Alternatively, the default could be a more informative string, like `"\s*"` or something.

bredelings:
> Alternatively, the default could be a more informative string, like "\s*" or something.
Hmm.... well maybe "" is a better default, since (if I understand correctly) you aren't actually implementing regexes, and the magic version implies the whitespace trimming.

That's interesting though, because the delimiter api isn't expressive enough to describe the whitespace-trimming behavior.  Do you think it would make sense to make the api a bit more powerful?  For example
* allow actual regexes for the delimiter -- you could use boost regex
* add a boolean flag to enable/disable trimming leading/trailing whitespace?

Another issue that may come up is handling quotes in CSV files.  If a delimiter occurs inside a quoted string, we aren't supposed to split on it, and if you have `""` inside a quoted string, it represents a literal `"`.
https://techterms.com/definition/csv

willpett:
> Hmm.... well maybe "" is a better default...

I thought this would be a good choice because it's the same convention used by R. In `read.table` an empty string delimiter means delimit on whitespace `\s+`. I could even just copy past the argument description from R.

> * allow actual regexes for the delimiter -- you could use boost regex

That could be very cool, although I'm not sure how often a use case will actually come up with RevBayes.

> * add a boolean flag to enable/disable trimming leading/trailing whitespace?

I actually already added a boolean flag argument `trim` to `StringUtilities::stringSplit` that tells it to remove leading/trailing whitespace from each field after splitting them by the delimiter. This was already the default behavior of all delimited data readers (e.g. `readDataDelimitedFile`), I just migrated it into `stringSplit`. One idea I had was to use this argument to determine when to strip leading/trailing whitespace from the whole line, i.e. when `trim=true` and `delimiter=""`, then you assume that the first and last fields cannot be empty, whereas with `trim=false` and `delimiter=""` you would return empty strings for the first and last fields. In the end I didn't go this route because I figured it was too confusing to explain, and if people are delimiting on `\s+` then they probably  aren't worrying about empty fields anyway. So I chose to just strip them by default.

> Another issue that may come up is handling quotes in CSV files. If a delimiter occurs inside a quoted string, we aren't supposed to split on it, and if you have `""` inside a quoted string, it represents a literal `"`. https://techterms.com/definition/csv That could be an incremental change though, I suspect.

Very true! I hadn't thought of that, but it seems very important. I suppose this behavior was never handled properly by RevBayes. I'll definitely implement it.

PR #191, opened 2021-10-11
Dev pomo

mrborges23:
1. PoMo rate matrices were rescaled so they are all in generation time.
2. Two count reader functions were added: both use the weighted-sampled method but one feeds on count files and the other on Fasta alignments: i.e., convertCountFileToNaturalNumbers() and convertFastaFileToNaturalNumbers().


willpett:
Hi Rui,

Thanks for the pull request! You should merge into `development`, though, instead of `master`. It looks like the integration tests are not passing for your branch due to a bad Mac OS configuration on GitHub actions, which hopefully will be resolved when you get up to date with `development`.

PR #201, opened 2021-11-18
Bump the minimum cmake version to 2.8.12

bredelings:
I'm getting a deprecation warning for setting the minimum version to 2.6.  (I have cmake 3.21.4).

```
    CMake Deprecation Warning at CMakeLists.txt:1 (cmake_minimum_required):
    
    Compatibility with CMake < 2.8.12 will be removed from a future version of CMake.
```

Since RHEL 6 and 7 have 2.8.12, this should be safe for most old clusters.

https://gitlab.kitware.com/cmake/community/-/wikis/CMake-Versions-on-Linux-Distros

Does anybody have a cluster with cmake older than 2.8.12?  Often you can use a newer cmake than the default system cmake by typing something like `module add cmake`.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.6)
+cmake_minimum_required(VERSION 2.8.12)
 project(RevBayes)
 
 # Consider:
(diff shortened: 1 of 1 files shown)


bredelings:
It looks like we can actually go to 3.5.1

PR #202, opened 2021-11-18
Fix boost cmake

bredelings:
This defaults to disabling `BoostConfig.cmake` since it may not work with cmake < 3.3

However, this version
* allows the user to override.
* prints a message explaining what is happening.


bredelings:
If we can require cmake > 3.3 then we might not need this one.
We talked about requiring some early v3 version of cmake today in the meeting (I'm thinking version 3.5.1), so lets not merge this yet.

bredelings:
It looks like instead of this we can require cmake > 3.3

PR #205, opened 2021-11-25
Allow creating new functions much more quickly

bredelings:
Currently, adding a new function often requires creating 3 level of objects.  For example, for HKY, we create
* `RateMatrix_HKY` in `core/datatypes/phylogenetics/ratematrix/`
* `HkyRateMatrixFunction` in `core/functions/phylogenetics/ratematrix/`
* `Func_hky` in `revlanguage/functions/phylogenetics/ratematrix/`

The `RateMatrix_HKY` isn't computed from scratch when its parameters  change, but has its own `update()` function.  

However, in some cases -- such as the sqrt function -- we don't need this level complexity.  The sqrt function simply computes its argument from scratch and doesn't cache anything.  So, it can get away with just
* `SqrtFunction` in `core/functions/math/`
* `FuncSqrt` in `revlanguage/functions/math/`

Interestingly, the `SqrtFunction` is largely boilerplate.  Creating files like `core/functions/math/SqrtFunction.{h,cpp}` takes time every time a developer wants to add a new function.  It turns out that we can eliminate this boilerplate using templates.

This PR adds a function template `GenericFunction<ReturnType,Arg1Type,Arg2Type, ..,ArgNType>(function,arg1,arg2,...argn)`.  We can use GenericFunction to create something like `SqrtFunction` in 1 line, and without having to create any new files.  All we need is a normal C++ function that computes the result from a series of input arguments.

As an example application, this patch does:
```
     RevBayesCore::TypedDagNode<double>* arg = static_cast<const Real &>( this->args[0].getVariable()->getRevObject() ).getDagNode();
-    RevBayesCore::SqrtFunction* f = new RevBayesCore::SqrtFunction( arg );
+    auto f = new RevBayesCore::GenericFunction<double, double>(sqrt, arg );
```
Here the `sqrt` function already exists, so we don't have to write anything.  We supply template parameters `<double, double>` because the `sqrt` function returns a `double` (1st parameter) and takes a `double` (2nd parameter).

We can then remove the `core/functions/math/SqrtFunction.{h,cpp}` which are no longer needed.  (If people don't want to actually remove SqrtFunction, thats fine.  That was mostly a proof-of-concept demonstration.)  Then the only file we need to implent `sqrt` is just
*  `FuncSqrt` in `revlanguage/functions/math/`

Thus, we go from 3 levels of indirection, down to 1!

My codon models branch is waiting on this -- it should should substantially decrease the amount of boilerplate needed to add new functions when the returned object type doesn't need an update() method.


bredelings:
The main issue with this PR, I think, is that it requires Boost 1.74.

Hopefully we can talk about this at our next monthly meeting.  It might be fine to require Boost 1.74, since we are suggesting people download and compile Boost 1.71 already on the platforms where it matters.

bredelings:
This PR should be able to go in.
Alternatively, I could wait after until the 
* boost 1.74
* gcc > 6
* c++14
PRs go in, and submit a slightly cleaned up version of this at that time.

bredelings:
Closing in favor of #211

PR #207, opened 2021-12-07
Require Boost version 1.74

bredelings:
I think that linux compiles already need to download and compile boost, with the version in the examples being 1.71
This PR bumps that to 1.71 and makes it a requirement.
The latest version is 1.77.
I don't think this would affect mac or windows.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,42 +28,49 @@ jobs:
             os: ubuntu-18.04
             compiler: gcc
             version: "6"
+            install_boost: 1.74
             mpi: false
 
           - name: ubuntu-gcc-7
             os: ubuntu-latest
             compiler: gcc
             version: "7"
+            install_boost: 1.74
             mpi: false
 
           - name: ubuntu-gcc-8
             os: ubuntu-latest
             compiler: gcc
             version: "8"
+            install_boost: 1.74
             mpi: false
 
           - name: ubuntu-gcc-9
             os: ubuntu-latest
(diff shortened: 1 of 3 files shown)


bredelings:
Per the meeting today, we should have consensus on this.

davidcerny (approved):

willpett:
Sorry I was kind of out of the loop on this when it got discussed. Linux builds definitely don't need to compile boost. The version in the current Ubuntu repositories is 1.71, which is found and used just fine when building RevBayes. This update makes it necessary to install/compile a separate version. Why do we need 1.74? If the need isn't super urgent, I'd say we should roll this back.

bredelings:
This is needed for PR #211.

Hmm... needing to compile a newer version of boost by hand is indeed annoying.  Are you talking about CI, or do you have a machine where ubuntu only has 1.71?

bredelings:
mmmm.... so we needthe `boost::mp11:tuple_transform` function, which was added in version 1.74.

Possibly we could just steal the definition from the boost:

```
template<class F, class... Tp, std::size_t... J>
BOOST_MP11_CONSTEXPR auto tuple_transform_impl( integer_sequence<std::size_t, J...>, F const& f, Tp&&..>
    -> decltype( tp_forward_v( tuple_apply( f, tp_extract<J>( std::forward<Tp>(tp)... ) )... ) )
{
    return tp_forward_v( tuple_apply( f, tp_extract<J>( std::forward<Tp>(tp)... ) )... );
}

```

PR #208, opened 2021-12-07
Change the standard to c++14.

bredelings:
C++14 adds few features, and fixes some defects in C++11.
This patch doesn't actually doesn't change any code -- it just allows compiling C++14 code.
This patch is needed for a small code cleanup on a different branch.


bredelings:
From the meeting today, I think this is OK.

willpett (approved):

PR #210, opened 2021-12-07
Bump minimum gcc version to 6, add gcc-10 to tests.

bredelings:
Based on our discussion today, bump the minimum gcc version to 6.
(On Jeremy Brown's cluster, for example, you can add gcc-8 with `module add gcc-8`)
So it seems that we no longer need to support very old versions because of the default compiler on RHEL6/7 is old.

This patch allows us to avoid the need to work around a big in gcc-5 on a different branch.


bredelings:
Hmm... the gcc-5 check will never run, because it was removed.  So, testing should be regarded as complete.

davidcerny (approved):

PR #211, opened 2021-12-12
Allow creating new functions much more quickly

bredelings:
Currently, adding a new function often requires creating 3 level of objects. For example, for HKY, we create

    RateMatrix_HKY in core/datatypes/phylogenetics/ratematrix/
    HkyRateMatrixFunction in core/functions/phylogenetics/ratematrix/
    Func_hky in revlanguage/functions/phylogenetics/ratematrix/

The RateMatrix_HKY isn't computed from scratch when its parameters change, but has its own update() function.

However, in some cases -- such as the sqrt function -- we don't need this level complexity. The sqrt function simply computes its argument from scratch and doesn't cache anything. So, it can get away with just

    SqrtFunction in core/functions/math/
    FuncSqrt in revlanguage/functions/math/

Interestingly, the SqrtFunction is largely boilerplate. Creating files like core/functions/math/SqrtFunction.{h,cpp} takes time every time a developer wants to add a new function. It turns out that we can eliminate this boilerplate using templates.

This PR adds a function template GenericFunction<ReturnType,Arg1Type,Arg2Type, ..,ArgNType>(function,arg1,arg2,...argn). We can use GenericFunction to create something like SqrtFunction in 1 line, and without having to create any new files. All we need is a normal C++ function that computes the result from a series of input arguments.

As an example application, this patch does:

     RevBayesCore::TypedDagNode<double>* arg = static_cast<const Real &>( this->args[0].getVariable()->getRevObject() ).getDagNode();
-    RevBayesCore::SqrtFunction* f = new RevBayesCore::SqrtFunction( arg );
+    auto f = new RevBayesCore::GenericFunction<double, double>(sqrt, arg );

Here the sqrt function already exists, so we don't have to write anything. We supply template parameters <double, double> because the sqrt function returns a double (1st parameter) and takes a double (2nd parameter).

We can then remove the core/functions/math/SqrtFunction.{h,cpp} which are no longer needed. (If people don't want to actually remove SqrtFunction, thats fine. That was mostly a proof-of-concept demonstration.) Then the only file we need to implent sqrt is just

    FuncSqrt in revlanguage/functions/math/

Thus, we go from 3 levels of indirection, down to 1!

My codon models branch is waiting on this -- it should should substantially decrease the amount of boilerplate needed to add new functions when the returned object type doesn't need an update() method.


bredelings:
I wrote some explanation of the code, in hopes that it is helpful for people.
I also cleaned up the code.
We can either merge this version as is, or merge #207 and #208, and the merge this one separately.

willpett:
Sorry I was kind of out of the loop when the boost version update was discussed. Separate compilation of boost is not required on linux. For example, the version in the current Ubuntu repositories is 1.71, which is found and used just fine when building RevBayes. Requiring 1.74 makes it necessary to install a non-default version, so you need to know a specific ppa etc. Is 1.74 really needed for this PR?

bredelings:
mmmm.... so we need the boost::mp11:tuple_transform function, which was added in version 1.74.

Possibly we could just steal the definition from the boost until 1.74 becomes more common, but it is a bit complex.

willpett:
Is there a way to implement this without `boost::mp11::tuple_transform`?

bredelings:
Not obviously.  We basically need to change a tuple of dag nodes to a tuple of values from the respective dag nodes.
I bet I can extract the tuple_apply thing from boost...

Do you want to address the boost issue immediately, or can I think about this for a day or two?

willpett:
Sure sure, it's not urgent. I was just reviewing PR's and noticed this issue. I can work with 1.74 for now.

bredelings:
@willpett I've figured out how to extract boost::mp11::tuple_apply from boost 1.74.  So, my code will no longer break if we just require boost 1.71.  Do you still want to set the requirement to boost 1.71?  I can submit a PR.

bredelings:
Drat.  Well, I've _almost_ figured it out.

bredelings:
OK, it is figured out now.

PR #217, opened 2022-02-09
Add move constructors and move-assignment operators for MatrixReal, Simplex, RbVector, and TransitionProbabilityMatrix

bredelings:
This adds move-constructors and move-assignment operators for RbVector, Simplex, MatrixReal, and TransitionProbabilityMatrix.  In cases like
```
   MatrixReal m = compute_matrix(....);
```
this basically allows `m` to avoid allocating new memory and copying the temporary variable.  Instead it is able to steal the resources of the temporary variable and use them.  So, it becomes feasible to return a RealMatrix by value.  It should also speed up existing code that returns these objects by value.

This is an alternative to code that takes the output value as a function argument:
```
virtual void calculateTransitionProbabilities(double startAge, double endAge, double rate, TransitionProbabilityMatrix& P) const = 0;   //!< Calculate the transition matrix
```
The TransitionProbabilityMatrix is really an output parameter.  Passing it as a parameter avoids 
* allocating memory when we first construct P
* allocating memory and copying the matrix when we copy-construct P
The move-constructor / move-assigment operator eliminate the _second_ problem in the case where we want to return a TransitionProbabilityMatrix by value.


bredelings:
Also, note that this PR is relative to misc-fixes-1 at the moment.

willpett (approved):

PR #219, opened 2022-02-11
Pre-release checklist

willpett:
Release v1.2 Lagerstätte


bredelings:
I'd like to get the #220 merged before release.

bredelings:
Hey Will, perhaps the NEWS file should continue to contain info about past releases?

willpett:
Hm, I guess I was thinking the news file just contains changes since the last release. People can just go check the news file for each release to see the changes. Otherwise maybe we call it the Change Log?

bredelings:
I think that NEWS files traditionally look like this:

https://fossies.org/linux/gdb/gdb/NEWS

So, they contain information on the last few releases, and are aimed at users instead of developers.  I think some projects contain Changelog files in addition to a NEWS file, but they are more granular and aimed at developers.  Now that you can run `git log`, I suspect many projects have dropped Changelog files.

bredelings (approved):

PR #221, opened 2022-03-01
Update make_winroot.sh, describe cross-builds in README.md, and fix RevStudio build

bredelings:
This updates the `make_winroot.sh` script and allows building RevStudio again:
* We no longer need to generate a `pkg-config` wrapper with modern versions of meson.
* `c_args` and `cpp_args` should be in section [built-in options] now.
* give helpful error message if `zstd` cannot be found to decompress the packages
* adds an option to not download packages if you just want to regenerate win64-cross.txt
* avoid redownloading MINGW packages that have already been downloaded
* download packages MINGW packages for GTK if asked (needed for RevStudio)
* switch from `x86_64-w64-mingw32-gcc` to `x86_64-w64-mingw32-gcc-posix` to use POSIX threads (needed for RevStudio)
* general cleanups

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -194,8 +194,9 @@ jobs:
       run: |
         /usr/local/bin/meson build ${GITHUB_WORKSPACE} --prefix=${GITHUB_WORKSPACE}/wincrossbuild --cross-file=win64-cross.txt
         /usr/local/bin/ninja -C build install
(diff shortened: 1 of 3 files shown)


bredelings:
This PR switches the windows cross-build for github actions to use `x86_64-w64-mingw32-gcc-posix`, which uses POSIX threads instead of Win32 threads.  This doesn't really affect anything, except that we need to use the posix DLLs for libstdc++6.dll and libgcc_s_seh.dll instead of the win32 ones.

PR #224, opened 2022-03-23
Remove build-ninja.sh in favor of -ninja true argument to build.sh

bredelings:


davidcerny (approved):
I had no trouble building RevBayes from this commit using the `-ninja true` flag on the cluster I usually use (Scientific Linux 7.4, gcc 10.1.0, boost 1.71.0, cmake 3.11.0), so I'm approving this PR.

PR #225, opened 2022-03-29
Change .getIncludedCharacterIndices() to use 1-based indexing

davidcerny:
A simple change that makes the indices outputted by the `getIncludedCharacterIndices` method of class `HomologousCharacterData` consistent with indices accepted by other methods of the same class, such as `includeCharacter` and `excludeCharacter`. Respects the convention that 0-based indexing is only used internally, while 1-based indexing is used for user-facing applications.


bredelings (approved):

davidcerny:
Hi Ben, that's correct – consistent indexing should make it easier to check which characters are included in a given partition after multiple applications of `.includeCharacter()` and `.excludeCharacter()`, which has been part of the general workflow for several of my recent analyses.

PR #227, opened 2022-04-03
Change readTreeTrace() to avoid creation of 0-length traces

davidcerny:
Currently, `readTreeTrace()` allows the user to specify a burnin argument such that the resulting trace is of length 0 (either by passing a Probability of 1, or an Integer equal to the number of trees in the trace). This results in the failure of several member methods of class `TraceTree` (e.g., `computePairwiseRFDistances`), and of functions that take a tree trace as an argument (e.g., `mccTree()`, `mapTree()`). Since traces of length 0 are presumably never desirable, it seems easier to prevent their creation in the first place. Here, this is done by checking for the two cases above and (1) reinterpreting 1 as a number of samples (rather than a fraction) in the former case, or (2) throwing an error in the latter case.

### DIFF
--- a/src/revlanguage/functions/io/Func_readTreeTrace.cpp
+++ b/src/revlanguage/functions/io/Func_readTreeTrace.cpp
@@ -207,18 +207,38 @@ RevPtr<RevVariable> Func_readTreeTrace::execute( void )
 
         for(size_t i = 0; i < rv->getValue().size(); i++)
         {
-            (*rv)[i].getValue().setBurnin(burnin);
+            if (burnin >= long( (*rv)[i].getValue().size() ) )
+            {
+                throw RbException("Burnin is greater than or equal to the number of trees in the trace. Please specify a smaller burnin instead.");
+            }
+            else
+            {
+                (*rv)[i].getValue().setBurnin(burnin);
+            }
         }
     }
     else
     {
         double burninFrac = static_cast<const Probability &>(b).getValue();
-
-        for(size_t i = 0; i < rv->getValue().size(); i++)
+            
+        // A probability of 1 will be re-interpreted as an integer to avoid creating a trace of length 0
+        if (burninFrac == 1.0)
         {
-            burnin = long( floor( rv->getValue()[i].getValue().size()*burninFrac ) );
+            burnin = 1;
 
-            (*rv)[i].getValue().setBurnin(burnin);
+            for(size_t i = 0; i < rv->getValue().size(); i++)
+            {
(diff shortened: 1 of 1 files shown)


bredelings:
Hmm.... if the trace has 1 tree only, do all these functions work?

davidcerny:
Hmm, good point. The member methods are doing okay (e.g., `computePairwiseRFDistances` returns an empty vector), but `mapTree()` and `mccTree()` give me a segfault.

bredelings:
Interesting -- I would have assumed that mapTree() and mccTree() would in effect return the single tree in the trace.  But apparently not.  I wonder how many samples they need in order to work.

davidcerny:
Two trees are the current minimum, apparently, but I agree that returning the single tree in the trace would be the expected (and desirable) behavior in the 1-tree case. I'll look into this and change them accordingly. Should I close this PR for now?

bredelings:
Maybe.  It also makes me wonder if there is a conceptually more satisfying solution.  But probably the user should get a warning of some kind about a 0-length trace.  If course, its possible that the warning should come from the functions like mccTree() themselves, instead of at the creation of the trace.

Special-casing a probability of 1.0 to treat it as an integer also seems a bit ugly.

davidcerny:
That makes a lot of sense. I guess part of the reason why I put the warning directly in `readTreeTrace()` was that 0-length traces cause the failure of not just other functions that take trace arguments, but also of several member methods (`summarize()` and `computePairwiseRFDistances()` cause RevBayes to enter an infinite loop; `computeEntropy()` returns a nonzero number as usual, which also seems like a bug). Would you recommend addressing these using method-specific warnings, too?

bredelings:
Hmm... I don't have strong feeling about that.  Maybe avoiding 0-length traces is the way to go, along with making mapTree() and mccTree() handle the case where there is only 1 tree in the trace.

Perhaps the question is something like (i) is it an improvement and (ii) relatively simple.  It could always be further improved in the future.

About changing Probability(1.0) to an integer...  it seems to me like if the user writes "1" instead of "1.0" then it should be an integer always.  But the RevBayes approach to types is pretty crazy.  Another alternative would be to use different argument names for burnin_fraction and burnin_iterations.

bredelings:
What if you just change `Probability|Integer<any> burnin` to `Integer<any>|Probability burnin`?
If the "Integer" comes first, then maybe it will treat "1" as an integer?

davidcerny:
Will try that out – thanks so much!

PR #230, opened 2022-04-16
Codon model argument order

bredelings:
This PR reorders arguments in the newly-introduced codon model functions to place submodels as the first argument.

This is necessary to allow the use of the R 4.1 pipe operator `|>` for easy substitution model composition, if and when the the pipe operator is added to revbayes.  However, we need to do this before the 1.2 release so that we don't break users scripts by changing the argument order later.

### DIFF
--- a/help/md/fnFMutSel.md
+++ b/help/md/fnFMutSel.md
@@ -38,10 +38,10 @@ fnCodonGY94, fnCodonMG94, fnFMutSel0, fnMutSel
         F ~ dnIID(61, dnNormal(0,1))
         omega ~ dnUniform(0,1)
         # The FMutSel model from Yang and Nielsen (2008)        
-        Q1 := fnFMutSel(F, omega, fnGTR(er, nuc_pi))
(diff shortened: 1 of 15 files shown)


davidcerny (approved):
The code changes all look good to me. The alterations to `AbstractPhyloCTMCSiteHomogeneous.h` seem like they were not required for the purposes of this PR (correct me if I'm wrong), but they do make the code a little clearer.

I tested this on an example Ben posted to Slack, namely the creation of a deterministic node containing a codon matrix:

```
seed(12345)
kappa <- 2
nuc_pi ~ dnDirichlet([1, 1, 1, 1])
omega <- 0.5
for (i in 1:61) {
  fitnesses[i] ~ dnNormal(0, 1)
}
Q1 := fnMutSel(fndNdS(fnX3(fnHKY(kappa,nuc_pi)),omega),fitnesses)
```

The output was identical to that generated by a recent `development` build under the current syntax, as expected. On a related note, I can definitely see now how multiple nested function calls like this naturally lend themselves to the use of a pipe operator.

PR #231, opened 2022-04-16
Pipe operator

bredelings:
This PR contains the changes to the grammar to add the pipe operator `|>`.  This pipe operator was added to R 4.1.  It is similar to the `%>%` pipe operator in the magrittr package and used by tidyverse packages, but a bit simpler.
```
rnorm(100) %>% matrix(ncol = 2) %>% plot()   # pipe operator
rnorm(100) |>  matrix(ncol = 2) |>  plot()   # new builtin pipe operator
```
The new pipe operator transforms `x |> f(y)` to `f(x,y)` inside the parser.

**Update Sep 2022**

R 4.2 adds a new "pipe placeholder" for piping to arguments that are not the first argument.
```
    mtcars |> subset(cyl == 4) |> lm(mpg ~ disp, data = _)
```
I've updated the patch to support placeholders.

**Substitution model stacking**

Pipes can also be used to write a substitution model `A + B + C` (i.e. GTR+GAMMA+I) as `A |> B |> C`.  For example:
```
# GTR + X3 + MutSel + dNdS
fnGTR(er,pi) |> fnX3() |> fnMutSel(fitnesses) |> fndNdS(omega)   # with pipes
fndNdS(fnMutSel(fnX3(fnGTR(er, pi)), fitnesses), omega)          # without pipes
```
When mixture model objects are introduced, it could be used to write a GTR+G+I model as:
```
fnGTR(er,pi) |> fnGamma(alpha) |> fnInv(p)
```

**Based on feedback at the Apr 2022 monthly meeting:**
*  it sounds like the pipe idea is too new to go into the upcoming (1.2) release, but might go into the next release after that.  We need more time for people to get an idea if it is sufficiently useful to include.
* It seems most useful for substitution models, so people who are interested in substitution models will talk amongst ourselves to see what kind of consensus we develop.

**Overview of pipes in R**
Here is an overview of why people should use pipes in R: https://uc-r.github.io/pipe
Overview of `%>%` pipes in R: https://r4ds.had.co.nz/pipes.html
Differences between `%>%` and the new builtin `|>` pipe: https://www.infoworld.com/article/3621369/use-the-new-r-pipe-built-into-r-41.html


tests/test_basics/output/pipes.Rev.out
hoehna, line 1: I don't think this file should be on the GitHub repo. If this is the expected output, then shouldn't it be in output_expected?

hoehna (changes requested):

tests/test_basics/output/pipes.Rev.out
bredelings: Yes, fixed.

src/grammar/grammar.y
```diff
@@ -100,7 +103,6 @@ Parser& parser = Parser::getParser();
 %type <boolValue> FALSE TRUE
 %type <string> RBTAB
 %type <string> identifier typeSpec optDims dimList 
-%type <syntaxElement> variable
```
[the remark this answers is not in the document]
bredelings: No worries, this is not a nitpick!

Initially there were some increment expressions that SHOULD work that did not parse.  When I changed the grammar to parse them, they executed correctly.

However, I think that should not be part of this PR.
I will remove them, and maybe submit them later with more rationale/justification.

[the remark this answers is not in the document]
bredelings: Removed.

src/grammar/grammar.y
```diff
@@ -364,10 +370,10 @@ expression  :   constant                    { $$ = $1; }
...
             |   AND expression %prec UAND   { $$ = new SyntaxUnaryExpr(SyntaxUnaryExpr::UAnd, $2); }
 
-            |   DECREMENT variable          { $$ = new SyntaxDecrement( $2, false ); }
-            |   variable DECREMENT          { $$ = new SyntaxDecrement( $1, true ); }
-            |   INCREMENT variable          { $$ = new SyntaxIncrement( $2, false ); }
-            |   variable INCREMENT          { $$ = new SyntaxIncrement( $1, true ); }
+            |   DECREMENT expression        { $$ = new SyntaxDecrement( $2, false ); }
+            |   expression DECREMENT        { $$ = new SyntaxDecrement( $1, true ); }
+            |   INCREMENT expression        { $$ = new SyntaxIncrement( $2, false ); }
+            |   expression INCREMENT        { $$ = new SyntaxIncrement( $1, true ); }
```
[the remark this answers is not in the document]
bredelings: Removed.

src/grammar/grammar.y
```diff
@@ -500,42 +510,46 @@ variable    :   identifier optElements
...
+                    for (auto& element: *$2)
                     {
-                        $$ = new SyntaxIndexOperation($$,*it);
+                        $$ = new SyntaxIndexOperation($$, element);
                     }
-                    delete $5;
+                    delete $2;
                 }
-            |   '(' expression ')' '[' expression ']' optElements
+            |   '(' expression ')' elementList
```
hoehna, line 525: I couldn't directly see where elementList was defined. Is that now a more generic way of stitching together multiple "calls" within the same expression? Would 'rBDP(10,rootAge=1,lambda=1,taxa=[taxon("A"),taxon("B"),taxon("C"),taxon("D")])[3].rootAge()' work? (I couldn't come up with any better test.)

hoehna (approved):

src/grammar/grammar.y
bredelings: elementList isn't new - its defined like this:

```
elementList :   '[' expression ']'              { $$ = new std::list<SyntaxElement*>(1, $2); }
            |   '[' ']'                         { $$ = new std::list<SyntaxElement*>(); }
            |   elementList '[' expression ']'  { $1->push_back($3); $$ = $1; }
            |   elementList '[' ']'             { $1->push_back( NULL ); $$ = $1; }
            ;
```

And optElements is defined like this:

```
optElements :   /* empty */                     { $$ = new std::list<SyntaxElement*>(); }
            |   elementList                     { $$ = $1; }
            ;
```

So, if we replace `'[' expression ']' optElements` with `elementList` there is _almost_ no change.  The only change is that now something like `f(x)[]` will parse.  Previously `f(x)[3][]` would parse but `f(x)[]` would not parse.

Possibly we should make a non_empty_elementList that looks like this:
```
nonEmptyElementList :   '[' expression ']'              { $$ = new std::list<SyntaxElement*>(1, $2); }
            |   elementList '[' expression ']'  { $1->push_back($3); $$ = $1; }
            ;
```
but I would leave that for a future cleanup.

bredelings: BTW, `rBDP(10,rootAge=1,lambda=1,taxa=[taxon("A"),taxon("B"),taxon("C"),taxon("D")])[3].rootAge()` does work.

PR #234, opened 2022-04-19
Dev boost fix path

spydr073:
Changed `CMakLists.txt` and `build.sh` to help cmake find non-standard boost locations as well as problems from multiple boost installations on the same system. The build script also now provides 2 ways to specify boost locations: 1) for standard installations you can specify `boost_root`, and 2) for non-standard installations where the include and lib folders are not located in the same directory you can specify `boost_include` and `boost_root`.


bredelings (changes requested):
Hmm... the idea here is to handle the case where ${boost_root} expands to a path with a space in it.  Without quotes, a path like "A/B C/D" will be treated as two different words.  This patch removes the functionality.

Can you clarify what was broken that the patch is fixing?  I can kind of see how this might not work, given how complicated shell variable expansion can get when there are multiple stages, but I'm not quite seeing it...

projects/cmake/build.sh
```diff
@@ -31,6 +32,7 @@ while echo $1 | grep ^- > /dev/null; do
 -help           <true|false>    : update the help database and build the YAML help generator. Defaults to false.
 -boost_root     string          : specify directory containing Boost headers (e.g. `/usr/include`). Defaults to unset.
 -boost_lib      string          : specify directory containing Boost libraries. (e.g. `/usr/lib`). Defaults to unset.
+-boost_include  string          : specify directory containing Boost includes. (e.g. `/usr/include`). Defaults to unset.
 -static_boost	<true|false>    : link using static Boost libraries. Defaults to false.
```
bredelings, line 36: I'm inclined to think that we should replace our own `-boost_root` , `-boost_lib`, and `-boost_include` flags by just recommending that people use the cmake flags (i.e. -DBOOST_ROOT=path`) directly.

This would make requests for help on e.g. stackoverflow directly applicable to our use case.  It also removes an (unnecessary?) layer of indirection between the user and cmake, making what is going on more opaque.

projects/cmake/build.sh
```diff
@@ -108,6 +110,10 @@ if [ -n "$boost_lib" ] ; then
     cmake_args="-DBOOST_LIBRARYDIR=\"${boost_lib}\" $cmake_args"
 fi
 
+if [ -n "$boost_include" ] ; then
+    cmake_args="-DBOOST_INCLUDEDIR=${boost_include} $cmake_args"
+fi
+
 if [ "$static_boost" = "true" ] ; then
```
bredelings, line 117: In addition to describing the path to the include files, the user might want to specify 
`-DBoost_DEBUG=ON` and `-DBoost_VERBOSE=ON`.

src/CMakeLists.txt
```diff
@@ -69,17 +69,38 @@ if ("${JUPYTER}" STREQUAL "ON")
...
+#endif()
+#if (BOOST_INCLUDEDIR)
+#  MESSAGE("  BOOST_INCLUDEDIR: ${BOOST_INCLUDEDIR}")
+#endif()
+#if (BOOST_LIBRARYDIR)
+#  MESSAGE("  BOOST_LIBRARYDIR: ${BOOST_LIBRARYDIR}")
+#endif()
 MESSAGE("Boost location hints:")
+if (BOOST_ROOT OR BOOST_INCLUDEDIR OR BOOST_LIBRARYDIR) 
+  set(Boost_NO_SYSTEM_PATHS ON)
```
bredelings, line 86: What is the benefit of this?  Does it affect linking, or only include paths?

src/CMakeLists.txt
```diff
@@ -69,17 +69,38 @@ if ("${JUPYTER}" STREQUAL "ON")
...
+  set(Boost_NO_BOOST_CMAKE ON)
```
bredelings, line 87: What is the benefit of this?

src/CMakeLists.txt
```diff
@@ -69,17 +69,38 @@ if ("${JUPYTER}" STREQUAL "ON")
...
+  set(Boost_NO_WARN_NEW_VERSION ON)
```
bredelings, line 88: I presume that these warnings exist only to let users know that there might be new BOOST libraries available in this version of boost.  Since this isn't relevant to us, I can see turning these warnings off...

projects/cmake/build.sh
```diff
@@ -108,6 +110,10 @@ if [ -n "$boost_lib" ] ; then
     cmake_args="-DBOOST_LIBRARYDIR=\"${boost_lib}\" $cmake_args"
 fi
 
+if [ -n "$boost_include" ] ; then
+    cmake_args="-DBOOST_INCLUDEDIR=${boost_include} $cmake_args"
```
bredelings, line 114: As a separate, but related, point to the one above, I think we should set `Boost_INCLUDE_DIR` directly, instead of setting BOOST_INCLUDEDIR, and then setting `Boost_INCLUDE_DIR` later.

src/CMakeLists.txt
```diff
@@ -69,17 +69,38 @@ if ("${JUPYTER}" STREQUAL "ON")
...
+endif()
 if (BOOST_ROOT)
   MESSAGE("  BOOST_ROOT: ${BOOST_ROOT}")
-else()
-  MESSAGE("  BOOST_ROOT not set")
+  set(Boost_INCLUDE_DIR ${BOOST_ROOT}/include)
```
bredelings, line 92: I don't think we should set `Boost_INCLUDE_DIR` directly.
BOOST_ROOT is supposed to take care of this.
Also, if the user hasn't specified Boost_INCLUDE_DIR, then they can still do so manually.

src/CMakeLists.txt
```diff
@@ -69,17 +69,38 @@ if ("${JUPYTER}" STREQUAL "ON")
...
+  set(Boost_LIBRARY_DIR ${BOOST_ROOT}/lib)
```
bredelings, line 93: Similarly for this -- I'm not sure we should do this.
Also, I think that `${BOOST_ROOT}/lib` isn't always the right location for the libraries.

src/CMakeLists.txt
```diff
@@ -69,17 +69,38 @@ if ("${JUPYTER}" STREQUAL "ON")
...
+  INCLUDE_DIRECTORIES(${Boost_INCLUDE_DIR})
 endif()
 if (BOOST_INCLUDEDIR)
   MESSAGE("  BOOST_INCLUDEDIR: ${BOOST_INCLUDEDIR}")
+  set(Boost_INCLUDE_DIR ${BOOST_ROOT})
+  INCLUDE_DIRECTORIES(${Boost_INCLUDE_DIR})
```
bredelings, line 99: Why do we need to add an `INCLUDE_DIRECTORIES(...)` flag directly?
Doesn't cmake do this already?

src/CMakeLists.txt
```diff
@@ -69,17 +69,38 @@ if ("${JUPYTER}" STREQUAL "ON")
...
 endif()
 if (BOOST_LIBRARYDIR)
   MESSAGE("  BOOST_LIBRARYDIR: ${BOOST_LIBRARYDIR}")
+  set(Boost_LIBRARY_DIR ${BOOST_LIBRARYDIR})
```
bredelings, line 103: Here, I just think that we should allow users to set (or not set) `-DBoost_LIBRARY_DIR` directly.

bredelings (changes requested):
Hey Killian,
Thanks for this patch!
Is there a specific case that was broken before that these patches fix?
If so, can you describe it?
That would make it easier to see what the goals are, and what is being fixed.
-BenRI

projects/cmake/build.sh
bredelings, line 36: I think it would also help to have `-debug_boost   <true|false>` or `-boost_debug   <true|false>`.  This could add the cmake argument `-DBoost_DEBUG=ON` and maybe `-DBoost_VERBOSE=ON`.

bredelings (changes requested):

src/CMakeLists.txt
bredelings: OK, I guess BOOST_ROOT is ignored unless we add Boost_NO_BOOST_CMAKE

projects/cmake/build.sh
bredelings: Argh... this seems a little bit complicated: https://android.googlesource.com/platform/prebuilts/cmake/windows-x86/+/refs/heads/studio-1.3-dev/share/cmake-3.2/Modules/FindBoost.cmake

src/CMakeLists.txt
spydr073: Yeah, this one is not necessary, but more of a convenience. When there are multiple versions of boost there can be an overwhelming number of warning messages (a few lines per boost component).

src/CMakeLists.txt
spydr073: Im not 100% sure what this is doing behind the scenes, but I believe that it is ignoring the environment variables that are related to boost (like things that are exported in bashrc).

projects/cmake/build.sh
spydr073: I guess cmake has its own versions of the normal boost variables. I was trying to be consistent and extra safe to make sure that the intended paths would be used in the build process.

src/CMakeLists.txt
bredelings: Maybe add a comment line about like # Ignore any environment variable hints about boost paths ?

src/CMakeLists.txt
spydr073: Yes, the BOOST_ROOT should handle setting this. This was just being extra safe, and could be removed.

src/CMakeLists.txt
bredelings: Maybe add an comment line like this?
# Suppress excessive warnings -- they aren't relevant for us

src/CMakeLists.txt
spydr073: Same as above

src/CMakeLists.txt
spydr073: I do not quite remember the rationale for this. I will work with it tomorrow to figure it out.

src/CMakeLists.txt
spydr073: We could try that, but I am not confident in the behavior of cmake when its special boost variables are not set. Maybe this would be fine when using with Boost_NO_SYSTEM_PATHS and Boost_NO_BOOST_CMAKE.

projects/cmake/build.sh
spydr073: Yep good idea, that would be easy to add. Will get on that tomorrow.

src/CMakeLists.txt
spydr073: Just tried removing this and tested on a couple of machines. On my laptop (nixos) it does not seem to be needed. However, on a debian cloud instance removing it results in the boost components not being found. I think maybe disabling the CMake specific boost settings changes how the includes are handled internally. 

I don't have access to a machine that has boost installed by the system in a normal location (i.e. an install that allows for just using `./build.sh` without any flags), so I can not test this theory though.

spydr073 (commented):
Made some changes based on your comments. If you have access to a machine that has boost installed in the normal location (i.e. you can compile using `./build.sh` without any flags), it would be nice to double check that the build script still works in this case. All of the machines I have access to have non-standard install paths...

bredelings:
I found some things:
* if Boost_NO_BOOST_CMAKE is off (so we are doing things the _new_ way), then BOOST_ROOT *is* respected.  However, I think that BOOST_INCLUDEDIR and BOOST_LIBRARY_DIR are not.
* if boost is manually compiled with `./boostrap.sh --prefix=installed; ./b2 ../installed_boost` then it makes a standard boost installation with `include` and `lib` as sister directories.  Then its possible to use the new way and just specify `BOOST_ROOT`

The old way (with Boost_NO_BOOST_CMAKE=ON) seems extremely complicated.  Since its also the old way, I suspect it will be deprecated eventually.  So its possible that we should try to aim for the new way.

When finding BOOST the old way, I also realized that both BOOST_INCLUDEDIR and Boost_INCLUDE_DIR are relevant.
* BOOST_INCLUDEDIR says where to look for boost and can be set by an environment variable.
* Boost_INCLUDE_DIR records the result of finding the boost include directory.
So, both are relevant, but setting the latter one gives more control, I think, because it describes the _output_ of the search process, whereas the former describes an _input_ to the search process.

bredelings:
On another note, I think that maybe we should consider fixing problems incrementally, instead of trying a fix-the-world approach.

It seems like there are a few obvious bugs -- for example the bug with quotes -- that could be merged into `development` first.

After that fix, we could then add incremental fixes, while checking what each one improves.

bredelings:
This PR was really.helpful in identitying some of the problems that needed to be fixed.
Based on conversations with @spydr073 we have merged a smaller fix. Unsolved problems can be solved via additional patches.
So, I'm going to close this for now - I presume that additional fixes would be a new PR.

PR #235, opened 2022-04-20
Describe specifying the boost location to meson.

bredelings:


davidcerny (approved):
I had no trouble building RevBayes from this branch on macOS (12.3.1) and Scientific Linux (7.4, non-root) with meson (0.62.1) and ninja (1.10.2) following the updated instructions, so I'm approving this PR.

PR #244, opened 2022-06-04
Fixes1 from dev jsigao

bredelings:
Here are some fixes from the dev_jsigao branch that passed the integration testing.

Three of these are related to dirtiness flagging on trees, so probably @hoehna should look at them.

### DIFF
--- a/src/core/datatypes/trees/Tree.cpp
+++ b/src/core/datatypes/trees/Tree.cpp
@@ -431,6 +431,10 @@ void Tree::executeMethod(const std::string &n, const std::vector<const DagNode *
     if ( n == "parent" )
     {
         long index = static_cast<const TypedDagNode<long> *>( args[0] )->getValue()-1;
+        if (getNode( index ).isRoot() == true)
+        {
+            throw RbException("Root has no parent.");
+        }
         rv = long( getNode( index ).getParent().getIndex() )+1;
     }
     else if ( n == "child" )
--- a/src/core/moves/proposal/tree/NearestNeighborInterchange_nonClockProposal.cpp
+++ b/src/core/moves/proposal/tree/NearestNeighborInterchange_nonClockProposal.cpp
@@ -80,10 +80,23 @@ double NearestNeighborInterchange_nonClockProposal::getProposalTuningParameter( 
  */
 double NearestNeighborInterchange_nonClockProposal::doProposal( void )
 {
+
+    // reset flag
+    failed = false;
+    
     // Get random number generator
     RandomNumberGenerator* rng     = GLOBAL_RNG;
     
     Tree& tau = tree->getValue();
+    
+    // when the tree has fewer than 4 tips, nothing should be done
+    size_t num_tips = tau.getNumberOfTips();
+    size_t num_root_children = tau.getRoot().getNumberOfChildren();
+    if ( num_tips <= num_root_children)
+    {
+        failed = true;
+        return RbConstants::Double::neginf;
+    }
     
     // pick a random internal branch on which we are going to perform the NNI
     // the branch is represented by it's tipward node, so we can pick any internal node
@@ -244,6 +257,12 @@ void NearestNeighborInterchange_nonClockProposal::printParameterSummary(std::ost
  */
 void NearestNeighborInterchange_nonClockProposal::undoProposal( void )
 {
+    // we undo the proposal only if it didn't fail
+    if ( failed == true )
+    {
+        return;
+    }
+    
     // undo the proposal
     TopologyNode* parent = &stored_node_A->getParent();
     TopologyNode* node   = ( (picked_root_branch == false && picked_uncle == false) ? &parent->getParent() : &stored_node_B->getParent());
--- a/src/core/moves/proposal/tree/SubtreePruneRegraftProposal.cpp
+++ b/src/core/moves/proposal/tree/SubtreePruneRegraftProposal.cpp
@@ -94,11 +94,22 @@ bool SubtreePruneRegraftProposal::isDescendant(const TopologyNode &n, const Topo
  */
 double SubtreePruneRegraftProposal::doProposal( void )
 {
+    // reset flag
+    failed = false;
     
     // Get random number generator
     RandomNumberGenerator* rng     = GLOBAL_RNG;
     
     Tree& tau = tree->getValue();
+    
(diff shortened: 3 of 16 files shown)


src/core/datatypes/trees/Tree.cpp
```diff
@@ -431,6 +431,10 @@ void Tree::executeMethod(const std::string &n, const std::vector<const DagNode *
     if ( n == "parent" )
     {
         long index = static_cast<const TypedDagNode<long> *>( args[0] )->getValue()-1;
+        if (getNode( index ).isRoot() == true)
+        {
+            throw RbException("Root has no parent.");
```
hoehna, line 436: If we throw an error like this, a script might stop. If we return -1, then the script could handle this. Not sure what's better/safer.

src/core/moves/proposal/tree/NearestNeighborInterchange_nonClockProposal.cpp
```diff
@@ -80,11 +80,24 @@ double NearestNeighborInterchange_nonClockProposal::getProposalTuningParameter(
...
+    failed = false;
+    
     // Get random number generator
     RandomNumberGenerator* rng     = GLOBAL_RNG;
     
     Tree& tau = tree->getValue();
     
+    // when the tree has fewer than 4 tips, nothing should be done
+    size_t num_tips = tau.getNumberOfTips();
+    size_t num_root_children = tau.getRoot().getNumberOfChildren();
```
hoehna, line 94: This test is weird. Why not then test directly if num_tips <= 4?

src/core/moves/proposal/tree/SubtreePruneRegraftProposal.cpp
```diff
@@ -94,12 +94,23 @@ bool SubtreePruneRegraftProposal::isDescendant(const TopologyNode &n, const Topo
...
     
     // Get random number generator
     RandomNumberGenerator* rng     = GLOBAL_RNG;
     
     Tree& tau = tree->getValue();
     
+    // when the tree has fewer than 4 tips, nothing should be done
+    size_t num_tips = tau.getNumberOfTips();
+    size_t num_root_children = tau.getRoot().getNumberOfChildren();
+    if ( num_tips <= num_root_children)
```
hoehna, line 108: Why do you need to test the root children and not only test for num_tips <= 4?

src/core/moves/proposal/tree/SubtreeSwapProposal.cpp
```diff
@@ -97,12 +97,22 @@ bool SubtreeSwapProposal::isDescendant(const TopologyNode &n, const TopologyNode
...
+    failed = false;
     
     // Get random number generator
     RandomNumberGenerator* rng     = GLOBAL_RNG;
     
     Tree& tau = tree->getValue();
     
+    size_t num_tips = tau.getNumberOfTips();
+    size_t num_root_children = tau.getRoot().getNumberOfChildren();
+    if ( num_tips <= num_root_children)
```
hoehna, line 110: Why can't you test num_tips <= 4? Also, 4 should be fine. So maybe 3?

hoehna (changes requested):

src/core/datatypes/trees/Tree.cpp
bredelings: Hmm... if you return `-1`, is there a chance the script would continue and yield non-sensical results?

It looks like if you call `getNode( -1 )` revbayes might segfault:

```
const TopologyNode& Tree::getNode(size_t idx) const
{

    if ( idx >= nodes.size() )
    {
        throw RbException("Index out of bounds in getNode.");
    }

    return *nodes[idx];
}
```

I think we should throw an exception instead of returning `-1`.  And probably we should throw an exception also if the user does `getNode(-1)`.

bredelings:
@jsigao Any thoughts? :-)

src/core/datatypes/trees/Tree.cpp
bredelings: There aren't very many places in RevBayes that use `-1` to indication "no value".  So maybe its best for scripts to handle this situation first checking whether a node is the root node before computing its parent...

jsigao:
Thanks for making the PR, Ben, and thanks for reviewing it, Sebastian! Sorry for the delay in response. I'm working on some final paperworks now in order to formally graduate by next week, but I'll provide my input as soon as possible.

src/core/moves/proposal/tree/NearestNeighborInterchange_nonClockProposal.cpp
jsigao: Yeah, I agree this may appear a bit weird, especially due to my original comment about 'fewer than 4 tips'. By 'fewer than 4 tips' I meant 'fewer than or equal to 3 tips', _i.e._, `num_tips <= 3`.

To me, it seems that for the non-clock tree proposals to make sense, the tree that these proposals operate on needs to have at least one node that's not a tip nor root; _i.e._, the root needs to have at least one child that's not a tip.

I prefer to check for `num_tips <= num_root_children` instead of `num_tips <= 3` because the non-clock tree, in principle, could be fully bifurcating, in which case an NNI move still makes sense to operate on a 3-tip tree. [Personally, I perceive the non-clock tree proposals as proposals on _phylograms_ (trees whose nodes have no age) instead of on _unrooted trees_.] 

In another case---that's also highly unlikely or impossible with the current implementation or usage of revbayes but is biologically reasonable so might be relevant eventually---the tree could be multificating so that the root of a tree with more than 3 tips (_e.g._, a star phylogeny) might still have no grandchildren, which wouldn't be catched by checking `num_tips <= 3`.

To sum up, I wrote this check this way just to make it a bit more generic and potentially future proof. However, I do agree that this way comes with the cost of interpretability and clarity of the code. I'm happy to change it to `num_tips <= 3` if you would prefer that @hoehna.

src/core/moves/proposal/tree/SubtreePruneRegraftProposal.cpp
jsigao: Please see the response above to the previous comment on the same code.

src/core/moves/proposal/tree/SubtreeSwapProposal.cpp
jsigao: Please see the response above to the previous comment on the same code.

PR #247, opened 2022-06-16
Fix windows setwd

bredelings:
This fixes setwd(dir) on windows by using boost filesystem to set the current directory.  This means that we get to rely on boost filesystem to parse windows pathnames.  It also means that we no longer have to keep a string inside revbayes to track the current directory ourselves.

Incidentally, this means that `setwd("..")` now works.

We still handle `~`.

### DIFF
--- a/src/core/utils/RbFileManager.cpp
+++ b/src/core/utils/RbFileManager.cpp
@@ -11,7 +11,8 @@
 #include "RbFileManager.h"
 #include "RbSettings.h"
 #include "StringUtilities.h"
-#include "boost/filesystem/path.hpp"
+#include <boost/filesystem/path.hpp>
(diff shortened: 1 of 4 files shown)


src/core/utils/RbFileManager.cpp
```diff
@@ -11,7 +11,8 @@
 #include "RbFileManager.h"
 #include "RbSettings.h"
 #include "StringUtilities.h"
-#include "boost/filesystem/path.hpp"
+#include <boost/filesystem/path.hpp>
```
rachelwarnock, line 14: why are u using a diff format to reference the header files?

rachelwarnock (approved):
thanks Ben - this looks great!

PR #248, opened 2022-06-16
Fix MCMC getting stuck under the WAG and LG models.

bredelings:
This PR was in response to a request from @jembrown, where MCMC wasn't working with the WAG model, if I remember correctly.  

The issue is the matrix exponential can return bad values when branch lengths are very large.  If the rows sum to more than 1, then this can lead to large likelihoods even when the likelihood should be quite small.  Then the MCMC gets stuck on a large branch length.  

We already have a few places in RevBayes where we (partially) solved this issue by making sure that the rows of the transition probability matrix don't sum to more than 1.  But we need to do this also in `AbstractRateMatrix::exponentiateMatrixByScalingAndSquaring( )` too.

This patch
* modifies `AbstractRateMatrix::exponentiateMatrixByScalingAndSquaring( )` to avoid producing transition probability matrices where rows sum to more than 1.
* refactors the WAG matrix function to use the new generic_function infrastructure.
* refactors the LG matrix function to use the new generic_function infrastructure.
* allows estimating equilibrium frequencies with `fnLG(pi)`
* simplifies the generic_function infrastructure so that there is only one form

Arguably we should put the common code between the WAG and LG functions into a C++ `Func_Empirical(er, pi, name)` class.  But that could be done in a later patch, if we want more amino-acid matrices.


src/core/datatypes/phylogenetics/ratematrix/AbstractRateMatrix.cpp
```diff
@@ -582,13 +582,29 @@ void AbstractRateMatrix::exponentiateMatrixByScalingAndSquaring(double t,  Trans
...
+
     // now perform the repeated squaring
     TransitionProbabilityMatrix r(num_states);
     for (size_t i = 0; i < s; i++)
     {
         multiplyMatrices(p, p, r);
         p = r;
+    }
+
+    // sanitization part 2 -- ensure rows sum to 1
```
hoehna, line 599: To me this looks dangerous because it doesn't show that there is a problem. So the analysis could be happily continuing while there being serious issue. Shouldn't throwing an error be more appropriate?

hoehna (changes requested):

src/core/datatypes/phylogenetics/ratematrix/AbstractRateMatrix.cpp
bredelings: In summary, throwing an error isn't appropriate because round-off errors are not something we can avoid.  OK?

bredelings: Note that this isn't a new idea, I'm just implementing it in a new place.  This is the solution that we have used previously, it is just that RevBayes has many duplication versions of this code, and they don't all have the sanitization implemented.

bredelings: If we throw an error in the likelihood calculation, would that cancel an MCMC run, or would it in fact just force the proposal to be rejected?  Cancelling the MCMC run wouldn't solve the problem that @jembrown had, but forcing the proposal to be rejected might work.

One possibility would be to renormalize matrix rows if the total sum for each row is in `[0.99, 1.01]`, and force the proposal to be rejected otherwise.  We could do this by 
* setting all transition matrix entries to 0 -- maybe?
* throwing an exception and forcing proposals that throw an exception in the likelihood to be rejected
* other ideas?

Issue #246, opened 2022-06-15
setNumStatesPartition() overrides character exclusion


davidcerny:
When inferring phylogenies from discrete morphological data, it is often desirable to divide the latter into ordered and unordered characters, and then partition further by the number of character states. Intuitively, the following should work (example file provided below):

```
> morpho <- readDiscreteCharacterData("morphology.txt")
> ordered_vect <- [24, 35, 39, 60, 68, 71, 117, 145, 167, 169, 174, 180, 197, 199, 206, 214, 215, 222, 251, 269, 272, 286, 289, 303, 305, 307, 313, 322, 333, 338, 353, 360, 376, 378, 387, 393, 442, 446]
> ordered_temp <- morpho
> ordered_temp.excludeCharacter( 1:morpho.nchar() )    # exclude all characters before including just those that we want
> ordered_temp.includeCharacter(ordered_vect)
> ordered_temp.nchar()                                 # should be equal to ordered_vect.size(): here, 38
    38
> ordered_2state <- ordered_temp
> ordered_2state.setNumStatesPartition(2)                
> ordered_2state.nchar()                               # should be 0, since all ordered characters have three or more states
    373                                                                                
```

This is the total number of 2-state characters in the original matrix (`morpho`), indicating that the `setNumStatesPartition()` method overrides the distinction between included and excluded characters. This is undesirable, and also inconsistent with the behavior of the analogous `setCodonPartition()` method, which does respect the distinction. Note that switching the order of operations (partitioning by state number first, excluding later) does work.

[morphology.txt](https://github.com/revbayes/revbayes/files/8905794/morphology.txt)

davidcerny:
Fixed in PR #252.

PR #249, opened 2022-06-17
Dev cerny avgdistmat

davidcerny:
Restricts the scope of two methods of the `AbstractHomologousDiscreteCharacterData` class to included characters only, in order to address issue #246 .


davidcerny:
Bizarrely, these two trivial changes cause the MPI (but not the single-thread) version of RevBayes to segfault when attempting to start an analysis (either power posterior or regular MCMC) on a partitioned morphological dataset. Closing this for now until I can debug.

PR #251, opened 2022-07-02
Catch likelihood bug

bredelings:


davidcerny (approved):
Works as intended! I'm wondering if it might be worth providing a warning that is a bit higher-level than an assertion failure (maybe an `RbException`?), but hopefully the problem is obscure enough that it is more likely to be encountered by developers rather than general users.

hoehna (approved):

bredelings:
@davidcerny -- good point.  This isn't really a full solution to the problem, just a debugging aid.  One idea for a full solution would be to:
* allocate separate conditional likelihood arrays for each node
* allocate the arrays at the time of computation, instead of doing the allocation in advance.
If we did that, then I think the assert would be appropriate, because it shouldn't be possible for users to trigger this by writing valid revbayes code.

I went with asserts here because they can be compiled out with -DNDEBUG, so they don't slow down the program if its build in release mode.  Throwing an exception would require a check even in release mode -- I'm not sure how much CPU time that would take.

PR #252, opened 2022-07-04
Prevent NumStates methods from overriding character exclusion

davidcerny:
This branch includes minor changes to the `.setNumStatesPartition()` and `.getNumStatesVector()` methods of class `AbstractHomologousDiscreteCharacterData`, which are intended to solve issue #246. Note that a very similar solution was implemented in PR #249, which I closed because it crashed on one of the test scripts I was experimenting with. As it turns out, this was due to an unrelated problem in that script, which has now been addressed by PR #251.

### DIFF
--- a/src/revlanguage/datatypes/phylogenetics/characterdata/RlAbstractHomologousDiscreteCharacterData.cpp
+++ b/src/revlanguage/datatypes/phylogenetics/characterdata/RlAbstractHomologousDiscreteCharacterData.cpp
@@ -561,7 +561,7 @@ RevPtr<RevVariable> AbstractHomologousDiscreteCharacterData::executeMethod(std::
                 warn = true;
             }
 
-            if ( !v.isCharacterExcluded(i) && max + 1 == n)
+            if ( !v.isCharacterExcluded(i) && max + 1 == n) // only set partition for previously included characters
             {
                 v.includeCharacter(i);
             }
@@ -644,7 +644,7 @@ RevPtr<RevVariable> AbstractHomologousDiscreteCharacterData::executeMethod(std::
 
             for (size_t x = 0; x < v.getNumberOfStates(); x++)
             {
-                if ( max == x)
+                if ( !v.isCharacterExcluded(i) && max == x ) // only consider previously included characters
                 {
                     matVec[x]->includeCharacter(i);
                 }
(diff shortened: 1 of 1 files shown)


src/revlanguage/datatypes/phylogenetics/characterdata/RlAbstractHomologousDiscreteCharacterData.cpp
```diff
@@ -519,57 +519,60 @@ RevPtr<RevVariable> AbstractHomologousDiscreteCharacterData::executeMethod(std::
         size_t n = size_t( static_cast<const Natural&>( argument ).getValue() );
         for (size_t i = 0; i < nChars; i++)
         {
-            RevBayesCore::RbBitSet observed(v.getNumberOfStates());
-            size_t max = 0;
-            for (size_t j = 0; j < nTaxa; j++)
+            // only set state number partition for previously included characters
+            if ( !v.isCharacterExcluded(i) )
```
bredelings, line 523: Optionally you could avoid indenting this whole region with `if (v.characterExcluded(i)) continue` (if I'm reading this right).

src/revlanguage/datatypes/phylogenetics/characterdata/RlAbstractHomologousDiscreteCharacterData.cpp
```diff
@@ -600,57 +603,61 @@ RevPtr<RevVariable> AbstractHomologousDiscreteCharacterData::executeMethod(std::
 
         for (size_t i = 0; i < nChars; i++)
         {
-            RevBayesCore::RbBitSet observed(v.getNumberOfStates());
-            size_t max = 0;
-            for (size_t j = 0; j < nTaxa; j++)
+            // only set state number partition for previously included characters
+            if ( !v.isCharacterExcluded(i) )
```
bredelings, line 607: Same here.

bredelings (approved):

src/revlanguage/datatypes/phylogenetics/characterdata/RlAbstractHomologousDiscreteCharacterData.cpp
davidcerny: That's a great point, thank you! Just for future reference, do we have an established style guide for decisions like these?

PR #254, opened 2022-07-28
Clean up MC3 code + add some MC3 mixing statistics

bredelings:
This patch 
* cleans up the MC3 code
* adds mixing statistics that supplement the probability of swapping chains
* adds a comment and paper reference about the new mixing statistics.

There do not appear to be any MC3 tests in the testsuite, but I modified `test_JC_unrooted` (which is fast) to do the following:
```
mymcmc = mcmcmc(mymodel, monitors, moves, nchain=4, nruns=2, tuneHeat = TRUE)

mymcmc.burnin(generations=200,tuningInterval=20)
mymcmc.run(generations=2000)
```
According to my tests, there is no change to the output file `output{,4}/mcmcmc_JC_unrooted_uniform_run_1.out`.

### DIFF
--- a/src/core/analysis/mcmc/Mcmcmc.cpp
+++ b/src/core/analysis/mcmc/Mcmcmc.cpp
@@ -33,6 +33,35 @@
 #include <mpi.h>
 #endif
 
+/*
+ * Round trip statistics are inspired by the paper https://arxiv.org/pdf/cond-mat/0602085.pdf
+ *
+ * 1. The first idea is that simply having good acceptance probabilities for swapping adjacent temperatures is not good enough.
+ *    We need chains to make a round-trip from cold -> hot -> cold.
+ *    But it is possible to have a low rate of making round trips even if we have success swapping between adjacent temperatures.
+ *
+ * 2. The second idea is based on computing the fraction of the time that a chain at temperature j has most recently been at
+         the coldest chain versus the hottest chain.
+ *    The claim is that the fractions should be uniformly spaced.
+ *    For example, if we have 5 chains, then the fraction should be 100% for the cold chain, and then 75%, 50%, 25% and 0%.
+ *    In some cases, geometric heating produces very low round-trip rates, and this is because the fractions are NOT uniformly spaced.
+ *
+ * 3. The third idea is that one scenario where geometric heating works badly is when there is a "phase transition" at temperature T.
+ *    This means that the posterior distribution on one side of T looks very different than the posterior distribution on
+ *        the other side of T.
+ *    This can happen when there are a very large number of low-probability states.
+ *    As the temperature increases, the low-probability states increase in probability, and eventually overwhelm the high-probability
+ *        states because there are so many of the lower-probability states.
+ *    In such a case, we need closely-spaced temperatures near T, because the posterior changes rapidly as T changes.
+ *    However, we can tolerate wider spacing between temperatures further away from T.
+ *
+ * BDR: I worry that diffusion behavior limits the number of round-trips as the number of chains increases.
+ *      By diffusion behavior, I mean that a chain does not always move towards the coldest or hottest chain, but drifts randomly,
+          sometimes stepping away.
+ *      Does a chain take O(N^2) steps to move from cold -> hot, even if the temperatures are well placed?
+ * 
+ */
+
 using namespace RevBayesCore;
 
 Mcmcmc::Mcmcmc(const Model& m, const RbVector<Move> &mv, const RbVector<Monitor> &mn, std::string sT, size_t nc, size_t si, double dt, size_t ntries, bool th, double tht, std::string sm, std::string smo) : MonteCarloSampler( ),
--- a/src/core/analysis/mcmc/Mcmcmc.cpp
+++ b/src/core/analysis/mcmc/Mcmcmc.cpp
@@ -66,17 +66,17 @@ using namespace RevBayesCore;
 
 using std::string;
 
-double Mcmcmc::heat_for_chain(int i) const
+double Mcmcmc::heatForChain(int i) const
 {
     return chain_heats[i];
 }
 
-bool Mcmcmc::is_cold_chain(int i) const
+bool Mcmcmc::isColdChain(int i) const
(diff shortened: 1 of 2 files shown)


src/core/analysis/mcmc/Mcmcmc.cpp
```diff
@@ -530,6 +612,14 @@ void Mcmcmc::nextCycle(bool advanceCycle)
...
     }
     
+    for(int i=0;i<num_chains;i++)
+    {
+        int heat_rank = heat_ranks[i];
+        if (chain_prev_boundary[i] == boundary::coldest)
+            heat_visitors[heat_rank].first++;
+        if (chain_prev_boundary[i] == boundary::hottest)
+            heat_visitors[heat_rank].second++;
+    }
```
jsigao, line 622: Not entirely sure if heat_ranks needs to be synchronized (explicitly with a specific synchronization function) across processes, like the chain_heats, when MPI is used (seems unnecessary to me, but I never fully grasped the reason that the heats need to be explicitly synchronized...).

jsigao (commented):
Hi Ben @bredelings -- thanks for cleaning up the code and adding the statistics! I looked through the changes and didn't spot any issue. I just have a couple rather superficial comments left below.

src/core/analysis/mcmc/Mcmcmc.cpp
bredelings: Good point.  Relying on heat_ranks[] here might not be correct, since the heat_ranks[] might be out of date with respect to the chain_heats[].

It is problematic that the same information is stored in multiple places -- they can get out of sync.  I started doing additional refactoring to 
* compute chain_heat(i) from heat_temps[] and heat_ranks[]
* synchronize heat_temps[] and heat_ranks[]
However, I decided to avoid doing such invasive changes in the first round.  The code could really use a second round of refactoring.

bredelings: OK, this should be fixed now.

bredelings:
@jsigao Does this seem OK now?

src/core/analysis/mcmc/Mcmcmc.cpp
```diff
@@ -33,8 +33,84 @@
...
+          sometimes stepping away.
+ *      Does a chain take O(N^2) steps to move from cold -> hot, even if the temperatures are well placed?
+ * 
+ */
+
 using namespace RevBayesCore;
 
+using std::string;
+
+double Mcmcmc::heat_for_chain(int i) const
```
hoehna, line 69: Why use "_" to separate word? We usually use  camel case

src/core/analysis/mcmc/Mcmcmc.cpp
```diff
@@ -949,8 +1109,28 @@ void Mcmcmc::resetCounters( void )
 }
 
 
-void Mcmcmc::setHeatsInitial(const std::vector<double> &ht)
+void Mcmcmc::setHeatsInitial(const std::vector<double>& ht)
 {
+    // 1. Check that there are num_chains heats.
+    if (ht.size() != num_chains)
+        throw RbException()<<"Heat vector contains "<<ht.size()<<" heats, but there are "<<num_chains<<" chains.";
```
hoehna, line 1116: I'm really not sure that this style of coding is the most readable. But maybe I'm just getting out of touch...

src/core/analysis/mcmc/Mcmcmc.cpp
```diff
@@ -1666,47 +1757,31 @@ void Mcmcmc::swapRandomChains(void)
...
-        
-        std::vector<Mcmc::tuningInfo> tmp_mvs_ti = chain_moves_tuningInfo[j];
-        chain_moves_tuningInfo[j] = chain_moves_tuningInfo[k];
-        chain_moves_tuningInfo[k] = tmp_mvs_ti;
-        
-        
+
+        std::swap( chain_heats[j], chain_heats[k] );
+        std::swap( heat_ranks[j], heat_ranks[k] );
+        std::swap( chain_moves_tuningInfo[j], chain_moves_tuningInfo[k] );
```
hoehna, line 1763: Is this MPI safe? Not sure if I caught the details, but we often runs MC3 with MPI and use exactly as many cores as chains. So was this tested?

hoehna (changes requested):

src/core/analysis/mcmc/Mcmcmc.cpp
bredelings: I constructed an MC3 example and ran it under MPI with 1, 2, and 4 cores.  Given a fixed number of chains, the number of cores does not change the output.  I found and avoided a few mistakes that way.

I don't think any of the current tests use MC3, but we should probably add one that uses it.

Regarding the `std::swap( )` code, it does exact the same thing as the previous code.  It is just shorter.

src/core/analysis/mcmc/Mcmcmc.cpp
bredelings: I switched the new functions to use camel case.

src/core/analysis/mcmc/Mcmcmc.cpp
bredelings: Lots of types have `ostream& operator<<(ostream&, const T&)` defined.
All of them would work here.
This seems fine...

bredelings:
@hoehna Thanks for the review!
I changed the new functions to use camelCase, and I have definitely tested the changes with MPI.
More MPI testing would be appreciated though!

jsigao:
I think this all look good to me now! I'm ready to approve it as long as it's also okay with Sebastian @hoehna.

bredelings:
@jsigao Sebastian has requested changes, so it can't be merged until he approves it as well.  He is on vacation this week, so hopefully next week.

jsigao (approved):
@bredelings Got it. (Sorry I haven't done many PR reviews before, so still not super familiar with the workflow...) Just approved.

hoehna (approved):
Great, looks good now.

PR #261, opened 2022-08-22
Remove string plus number

bredelings:
This branch removes C++ overloads for operator+(string, number) and operator+(string, RbVector<T>).
Instead of using operator+, one can use `operator<<` (for streams) or `StringUtilities::toString( )` (for strings).
The Rev language still adds strings and numbers.

The number of places that relied on these overloaded operators was actually not very large.
Some places were already creating an `ostringstream` and using `operator<<` to construct an error message, so these places just got simpler.


bredelings:
We can either merge this into #253 , or merge #253 into development first, and then merge this one.

Issue #240, opened 2022-05-03
[BUG] Unable to read trees with sampled ancestors in development


brpetrucci:
**Describe the bug**
Since pulling from development on May 2nd, my scripts have errored when reading trees with sampled ancestors. The specific error is `Problem while working with tree: Node had bad index. Two nodes had same index of '0'.` This same error is not present when compiling and running the master branch. Furthermore, reading an ultrametric tree or one with extant and fossil tips (but no ancestors) does not lead to error. 

**To Reproduce**
From my tests any tree with sampled ancestors will error, but an example can be downloaded from [here](https://github.com/brpetrucci/ssefbd_evol22/blob/main/eeob565/simulation/replicates/1_base/trees/fbd/tree_fbd_1.nex), while the equivalent ultrametric tree can be found [here](https://github.com/brpetrucci/ssefbd_evol22/blob/main/eeob565/simulation/replicates/1_base/trees/ultrametric/tree_ultra_1.nex), and the tree with fossil and extant tips, but no SAs, [here](https://github.com/brpetrucci/ssefbd_evol22/blob/main/eeob565/simulation/replicates/1_base/trees/tree_1.nex). If `tree.nex` is the name of the file in question, to reproduce one need simply run `readTrees(file = "tree.nex")[1]`

**Computer info**
MacOS Big Sur 11.6, RevBayes development commit ebf2aa7 (Apr 29 2022)

wrightaprilm:
I am also having this issue on Mac Catalina 10.15.7 and Linux Ubuntu 22.04 LTS.

brpetrucci:
Tried a couple commits up to 8a31d1e (Mar 11) and problem persists.

brpetrucci:
Closed by mistake. Problem still persists in current (Aug 23) version of development

bredelings:
I'll try and take a look...   have you tried compiling a debug version and looking at it in `gdb` or the xcode debugger?

bredelings:
@brpetrucci Can you try to make  small tree that reproduces the problem?  I'm curious what the smallest tree that yield this type of error is.  This would serve as a "reduced testcase" that makes it easier to debug.

bredelings:
@brpetrucci I think I've fixed this on #263 - can you try that out?

brpetrucci:
Seem like it's working! Thank you Ben.

PR #263, opened 2022-08-25
Don't crash when reading some trees

bredelings:
Investigating #240, it seems that reading some trees with sampled ancestors does not work.

The problem seems to be that when calling `getRoot().makeBifurcating( as_fossils )`, new tip nodes are created.  All of these new nodes have index 0 because the index is never set.  Later, `orderNodesByIndex( )` crashes because multiple nodes have the same index.

The patch here does three things:
1. it distinguishes between an index of 0 and an index that has never been set.
2. it makes TopologyNode::getIndex( ) complain when the node index has never been set.
3. it changes make the call to `set_root( , )` always recompute node indices after the call to `getRoot().makeBifurcating( as fossils )`

Distinguishing between nodes with an unset index, and nodes with index 0 should help to prevent problems like this in the future and make it more clear what the underlying problem is if unset indices are referenced.

Should the `reindex` parameter to `Tree::makeInternalNodesBifurcating( )` should be removed?  Or are there other approaches to the reindexing that would be better?

### DIFF
--- a/src/core/datatypes/trees/TopologyNode.cpp
+++ b/src/core/datatypes/trees/TopologyNode.cpp
@@ -27,74 +27,33 @@
 
 using namespace RevBayesCore;
 
+using boost::optional;
 
 /** Default constructor (interior node, no name). Give the node an optional index ID */
-TopologyNode::TopologyNode(size_t indx) :
-    use_ages( true ),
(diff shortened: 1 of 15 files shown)


willpett (approved):
Thanks for figuring this out! I think your solution is a good one.

PR #264, opened 2022-09-01
Two FBD fixes

willpett:
- Fixes a small issue when initializing time intervals in FBD ranges models
- Registers the generic FBDP distribution (an alias to BDSTP)


bredelings (approved):
Looks good.

PR #266, opened 2022-09-14
Re-enable MPI test and fix an MPI compile error

bredelings:
Re-enable MPI test and fix an MPI compile error.


davidcerny (approved):
Both commits make sense to me and successfully address both the original problem itself (the MPI version failing to compile) and the reason why it wasn't picked up during the automated testing. Approving.

PR #267, opened 2022-09-15
Fix vectors in functions

bredelings:
Currently, you can define functions.  For example:

```
function sampleDirichlet(n, alpha)
{
  for(i in 1:n)
  {
     x[i] ~ dnGamma(alpha, 1.0)
  }
  return x/sum(x)
}
```

However, this function fails with the message:
```
> sampleDirichlet(10,2)
   Missing Variable:	Variable x[1] does not exist
```

The reason is each function has its own Environment that records names created in that function.  In this case, the function Environment include `n`, `alpha`, `x`, `x[1]`, `x[2]`, ... etc.

However, when we create vectors like `x`, RevBayes does not immediately record the Variable for each of the elements.
Instead, RevBayes sets the flag `needs_building` and defers construction of the vector until getRevObject( ) is called for `x`.

When building the vector, RevBayes needs to look up the names `x[1]`, `x[2]`, etc. and then call the function `v( )` on the resulting list of variables.  However, we need to look these names up in the function environment, not the global user workspace environment:

```
    // @TODO: We actually might need a different workspace here than the user workspace. (Sebastian)
    //        Environment &env = Workspace::userWorkspace();
```

This patch 
* modifies RevVariable to record the Environment to use when building a vector.
* modifies setToVectorVariable( ) and addIndex( ) to take an Environment* as an argument, so that we know what Environment to use.
* builds all vectors created during a function call after we finish executing the function, so that we don't have any dangling pointers to the function Environment.

With this patch, `sampleDirichlet(10,2)` works:

```
> function sampleDirichlet(n, alpha)
+ {
+   for(i in 1:n)
+   {
+      x[i] ~ dnGamma(alpha, 1.0)
+   }
+   return x/sum(x)
+ }
> sampleDirichlet(10,2)
   [ 0.148, 0.091, 0.090, 0.094, 0.055, 0.115, 0.188, 0.034, 0.042, 0.142 ]
```

### DIFF
--- a/src/revlanguage/functions/UserFunction.cpp
+++ b/src/revlanguage/functions/UserFunction.cpp
@@ -66,7 +66,7 @@ RevPtr<RevVariable> UserFunction::execute( void )
 RevPtr<RevVariable> UserFunction::executeCode( void )
 {
     // Create new evaluation frame with function base class execution environment as parent
-    Environment* function_frame = new Environment( getEnvironment(), "UserFunctionEnvironment" );
+    auto function_frame = std::make_shared<Environment>( getEnvironment(), "UserFunctionEnvironment" );
     
     // Add the arguments to our environment
     for ( std::vector<Argument>::iterator it = args.begin(); it != args.end(); ++it )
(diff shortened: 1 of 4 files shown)


bredelings:
I have an idea that might simplify this and avoid the need for extra pointers to the function_frame Environment in the first place...

PR #268, opened 2022-09-16
Fix vectors in functions2

bredelings:
Currently, you can define functions in Rev.  For example:

```
function sampleDirichlet(n, alpha)
{
  for(i in 1:n)
  {
     x[i] ~ dnGamma(alpha, 1.0)
  }
  return x/sum(x)
}
```

However, this function fails with the message:
```
> sampleDirichlet(10,2)
   Missing Variable:	Variable x[1] does not exist
```

The reason is that
*  each function has its own Environment that maps names to objects.  In this case, the function Environment includes the names `n`, `alpha`, `x`, `x[1]`, `x[2]`, ... etc.
* when we create vectors like `x`, RevBayes does not immediately lookup and record the RevVariable for each of the elements of `x`.
Instead, RevBayes sets the flag `needs_building` and defers construction of the vector until getRevObject( ) is called for `x`.
* when building the vector, RevBayes needs to look up the names `x[1]`, `x[2]`, etc. and then call the function `v( )` on the resulting list of variables.  However, we need to look these names up in the function environment, but RevBayes was using the global user workspace environment:
```
    // @TODO: We actually might need a different workspace here than the user workspace. (Sebastian)
    //        Environment &env = Workspace::userWorkspace();
```

This patch looks up element names like `x[1]` in the environment immediately and records their corresponding variable, instead of doing deferred lookup.  This ensures that we look up the element names in the right environment. 

In more detail, the patch:
* modifies RevVariable to remember the `RevPtr<RevVariable>` for each vector index, if the variable is a vector.
* modifies SyntaxIndexOperation to pass the `RevPtr<RevVariable>` for an index when we call `RevVariable::addIndex`.
* also refactors constructors for RevVariable, and removes the redundant field `is_reference_var`

With this patch, `sampleDirichlet(10,2)` works:
```
> function sampleDirichlet(n, alpha)
+ {
+   for(i in 1:n)
+   {
+      x[i] ~ dnGamma(alpha, 1.0)
+   }
+   return x/sum(x)
+ }
> sampleDirichlet(10,2)
   [ 0.148, 0.091, 0.090, 0.094, 0.055, 0.115, 0.188, 0.034, 0.042, 0.142 ]
```

### DIFF
--- a/src/revlanguage/functions/UserFunction.cpp
+++ b/src/revlanguage/functions/UserFunction.cpp
@@ -66,7 +66,7 @@ RevPtr<RevVariable> UserFunction::execute( void )
 RevPtr<RevVariable> UserFunction::executeCode( void )
 {
     // Create new evaluation frame with function base class execution environment as parent
-    auto function_frame = std::make_shared<Environment>( getEnvironment(), "UserFunctionEnvironment" );
+    auto function_frame = std::make_unique<Environment>( getEnvironment(), "UserFunctionEnvironment" );
     
     // Add the arguments to our environment
     for ( std::vector<Argument>::iterator it = args.begin(); it != args.end(); ++it )
--- a/src/revlanguage/functions/UserFunction.cpp
+++ b/src/revlanguage/functions/UserFunction.cpp
@@ -66,7 +66,7 @@ RevPtr<RevVariable> UserFunction::execute( void )
 RevPtr<RevVariable> UserFunction::executeCode( void )
 {
     // Create new evaluation frame with function base class execution environment as parent
-    Environment* function_frame = new Environment( getEnvironment(), "UserFunctionEnvironment" );
+    auto function_frame = std::make_shared<Environment>( getEnvironment(), "UserFunctionEnvironment" );
     
     // Add the arguments to our environment
     for ( std::vector<Argument>::iterator it = args.begin(); it != args.end(); ++it )
--- a/src/revlanguage/parser/SyntaxIndexOperation.cpp
+++ b/src/revlanguage/parser/SyntaxIndexOperation.cpp
@@ -145,14 +145,15 @@ RevPtr<RevVariable> SyntaxIndexOperation::evaluateLHSContent( Environment& env, 
         Container* c = dynamic_cast<Container*>( &theParentObj );
         if ( c != NULL )
         {
-            
             if ( theParentObj.getDagNode()->isConstant() && c->allowsModificationToCompositeContainer() )
             {
+                // We need to set this to a vector variable before we call addIndex on it.
+                theParentVar->setToVectorVariable();
+
                 for ( size_t i=1; i<=c->size(); ++i)
                 {
                     std::string elementIdentifier = theParentVar->getName() + "[" + std::to_string(i) + "]";
                 
-                    RevPtr<RevVariable> theElementVar = NULL;
                     if ( !env.existsVariable( elementIdentifier ) )
                     {
                         // create a new slot
@@ -162,12 +163,12 @@ RevPtr<RevVariable> SyntaxIndexOperation::evaluateLHSContent( Environment& env, 
                     }
                 
                     // get the slot and variable
-                    theElementVar  = env.getVariable( elementIdentifier );
+                    RevPtr<RevVariable> theElementVar  = env.getVariable( elementIdentifier );
                 
                     // set this variable as a hidden variable so that it doesn't show in ls()
                     theElementVar->setElementVariableState( true );
                     
-                    theParentVar->addIndex( int(i) );
+                    theParentVar->addIndex( int(i) , theElementVar );
                 }
             }
             else
@@ -187,15 +188,10 @@ RevPtr<RevVariable> SyntaxIndexOperation::evaluateLHSContent( Environment& env, 
             throw RbException("We cannot make a composite container from variable of type '" + theParentObj.getType() + "'.");
         }
     }
-    
+
     // compute the index and internal name for this variable
     int idx = (int)static_cast<Integer&>( indexVar->getRevObject() ).getValue();
     std::string identifier = theParentVar->getName() + "[" + std::to_string(idx) + "]";
-    
-    // mark the parent variable as a vector variable
-    theParentVar->setVectorVariableState( true );
-    theParentVar->addIndex( idx );
-
 
     if ( env.existsVariable( identifier ) == false )
     {
@@ -208,9 +204,12 @@ RevPtr<RevVariable> SyntaxIndexOperation::evaluateLHSContent( Environment& env, 
     // get the slot and variable
     RevPtr<RevVariable> the_var  = env.getVariable( identifier );
     
-    
     // set this variable as an element variable; which is by default a hidden variable so that it doesn't show in ls()
     the_var->setElementVariableState( true );
+
+    // mark the parent variable as a vector variable
+    theParentVar->setToVectorVariable();
+    theParentVar->addIndex( idx, the_var );
 
     return the_var;
 }
--- a/src/revlanguage/parser/SyntaxIndexOperation.cpp
+++ b/src/revlanguage/parser/SyntaxIndexOperation.cpp
@@ -162,18 +162,19 @@ RevPtr<RevVariable> SyntaxIndexOperation::evaluateLHSContent( Environment& env, 
                 {
                     std::string elementIdentifier = theParentVar->getName() + "[" + std::to_string(i) + "]";
                 
+                    // first ensure that the environment has a slot named `elementIdentifier`
                     if ( !env.existsVariable( elementIdentifier ) )
                     {
-                        // create a new slot
+                        // create a new variable for environment slot `elementIdentifier`
                         RevPtr<RevVariable> theElementVar = RevPtr<RevVariable>( new RevVariable( c->getElement(i-1) ) );
-                        env.addVariable(elementIdentifier,theElementVar);
+                        env.addVariable( elementIdentifier, theElementVar );
                         theElementVar->setName( elementIdentifier );
                     }
-                
-                    // get the slot and variable
+
+                    // then look up the element variable in the environment by its identifier
                     RevPtr<RevVariable> theElementVar  = env.getVariable( elementIdentifier );
                 
-                    // set this variable as a hidden variable so that it doesn't show in ls()
+                    // set the element variable as a hidden variable so that it doesn't show in ls()
                     theElementVar->setElementVariableState( true );
                     
                     theParentVar->addIndex( int(i) , theElementVar );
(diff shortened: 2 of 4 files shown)


src/revlanguage/parser/SyntaxIndexOperation.cpp
```diff
@@ -208,10 +204,13 @@ RevPtr<RevVariable> SyntaxIndexOperation::evaluateLHSContent( Environment& env,
     // get the slot and variable
     RevPtr<RevVariable> the_var  = env.getVariable( identifier );
```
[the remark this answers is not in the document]
mikeryanmay: Note also there's a slight stylistic mismatch with ! vs == false when we check for the object in the environment here, compared to how it was done the first time. (I prefer == false)

src/revlanguage/workspace/RevVariable.h
```diff
@@ -77,20 +78,19 @@ namespace RevLanguage {
...
-//        std::set<int>           element_indices;                                        //!< The indices of the elements if this is a vector variable.
-        bool                    is_element_var;                                         //!< Is this variable an element of a vector?
-        bool                    is_hidden_var;                                          //!< Is this a hidden variable?
-        bool                    is_reference_var;                                       //!< Is this a reference variable?
-        bool                    is_vector_var;                                          //!< Is this a vector variable?
-        bool                    is_workspace_var;                                       //!< Is this a workspace variable?
+
+        // variables related to building the components of a vector
+        boost::optional<std::vector<RevPtr<RevVariable>>> vector_var_elements;          //!< Elements if this is a vector variable.
+        bool                    needs_building = false;                                 //!< Do we need to construct the revobject?
```
mikeryanmay, line 85: I like the default members!

src/revlanguage/workspace/RevVariable.cpp
```diff
@@ -105,7 +74,7 @@ RevVariable::RevVariable(const RevVariable &v) :
 RevVariable::~RevVariable( void )
 {
     
-    if ( is_reference_var == false && rev_object != NULL )
+    if ( not isReferenceVariable() and rev_object != NULL )
```
mikeryanmay, line 77: I get that you can do this now, but it is stylistically a bit out of sync with other rev code. Not that I think it's a big deal---this is perfectly readable.

src/revlanguage/workspace/RevVariable.cpp
```diff
@@ -154,16 +121,19 @@ RevVariable& RevVariable::operator=(const RevVariable &v)
...
 /** Resize the vector to include this index. */
-void RevVariable::addIndex(size_t idx)
+void RevVariable::addIndex(int idx, const RevPtr<RevVariable>& element)
 {
-    
-    if ( idx > element_index_max )
-    {
-        element_index_max = idx;
-    }
+    assert( isVectorVariable() );
```
mikeryanmay, line 126: Any reason to prefer `assert` here over `RbException`? `RbException` might be a more pleasant user experience (less unpleasant anyway), given the context here is user-defined functions!

src/revlanguage/workspace/RevVariable.cpp
```diff
@@ -210,41 +181,33 @@ size_t RevVariable::getReferenceCount(void) const
...
+
             args.push_back( Argument( element_var ) );
+            i++;
         }
+
         // @TODO: We might need a to check if this should be dynamic or not. (Sebastian)
         bool dynamic = true;
-        Function* func = Workspace::userWorkspace().getFunction("v",args,!dynamic).clone();
-        func->processArguments(args,!dynamic);
+        Function* func = Workspace::userWorkspace().getFunction("v", args, not dynamic).clone();
```
mikeryanmay, line 209: Maybe use a smart pointer here?

src/revlanguage/workspace/RevVariable.cpp
```diff
@@ -210,41 +181,33 @@ size_t RevVariable::getReferenceCount(void) const
...
-        //        }
-        
-        // @TODO: We actually might need a different workspace here than the user workspace. (Sebastian)
-        Environment &env = Workspace::userWorkspace();
-        
-        size_t max_index = getMaxElementIndex();
         std::vector<Argument> args;
-        for (size_t i = 1; i <= max_index; ++i)
+        int i=1;
+        for (auto& element_var: *vector_var_elements)
```
mikeryanmay, line 193: `vector_var_elements` makes this code much more elegant than the previous.

src/revlanguage/workspace/RevVariable.cpp
```diff
@@ -210,41 +181,33 @@ size_t RevVariable::getReferenceCount(void) const
...
         {
-            std::string element_identifier = name + "[" + std::to_string(i) + "]";
-            RevPtr<RevVariable>& element_var = env.getVariable( element_identifier );
             // check that the element is not NULL
             if ( element_var == NULL || element_var->getRevObject() == RevNullObject::getInstance() )
```
mikeryanmay, line 196: Given how `vector_var_elements` now works, will there ever be cases where `element_var` is `NULL`?

mikeryanmay (commented):
Everything here makes sense to me. The most significant changes are in RevVariable, which now optionally keeps track of its elements in the case that it is a vector. This makes some code in RevVariable itself a little more straightforward, but more importantly allows vectors to be created inside user-defined functions.

I've compiled the code and plays around with the example function @bredelings provided and a few others, and everything seems to work! I've made mostly cosmetic comments.

src/revlanguage/workspace/RevVariable.cpp
bredelings: I personally prefer `if (condition_holds)` over `if ( condition_holds == true )` since
* you could also write `if ( (condition_holds == true) == true)`
* people don't usually write `if ( (i > 1) == true)`
* the former seems closer to the way people talk "If the condition holds, then..."
I'm not aware of any other project that compares boolean values to `true` to get another boolean.

Anyway, that's my 2 cents.

src/revlanguage/workspace/RevVariable.cpp
bredelings: I'm thinking of this as documenting a precondition, which is a common use case for asserts.  

Usually the standard is to use asserts for programmer errors and report run-time error messages for user errors.  Any code that calls `addIndex( )` without first checking that it is operating on a vector indicates a programmer error, not a user error.  Probably the existing code has already checked whether the variable is a vector.

RevBayes' lack of asserts indicates low engineering standards, according to the Stamatakis group's paper "The State of Software for Evolutionary Biology".  See Tables 2 and 3 where they score software according to the number of asserts per 1000 lines of code.  I think most programmers would agree with that paper.

Kids these day would call the lack of asserts a "code smell" :roll_eyes: 

Thoughts?

mikeryanmay: That's fair enough, but you could make the same argument that attempting to assign an idx < 1 is a programming error? So maybe that should be an assert, too?

src/revlanguage/functions/UserFunction.cpp
```diff
@@ -66,7 +66,7 @@ RevPtr<RevVariable> UserFunction::execute( void )
 RevPtr<RevVariable> UserFunction::executeCode( void )
 {
     // Create new evaluation frame with function base class execution environment as parent
-    Environment* function_frame = new Environment( getEnvironment(), "UserFunctionEnvironment" );
+    auto function_frame = std::make_shared<Environment>( getEnvironment(), "UserFunctionEnvironment" );
```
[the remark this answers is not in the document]
bredelings: Yeah, unique_ptr is probably more appropriate.  Let me see how that looks.

src/revlanguage/workspace/RevVariable.cpp
bredelings: Yeah, good idea.

src/revlanguage/functions/UserFunction.cpp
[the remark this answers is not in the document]
bredelings: A number of C++ experts have suggested that people change from writing
```
std::string s("text");
```
to 
```
auto s = std::string("text");
```
See https://subscription.packtpub.com/book/programming/9781786465184/1/ch01lvl1sec5/using-auto-whenever-possible

I have mixed feelings, but am open the idea of getting the type info from the right-hand-side.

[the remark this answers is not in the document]
mikeryanmay: I get that argument, but in this case the type is not `std::make_shared<Environment>`, it's `std::shared_ptr<Environment>`, so the type is not 100% explicit from the rhs.

src/revlanguage/workspace/RevVariable.cpp
bredelings: I see what you mean.

I'm not sure if `idx < 1` is actually a programmer error the way RevBayes is currently written :grimacing:  
In some cases, an `idx` of `0` can be passed down to `addIndex( )`, and in some other cases it is intercepted before it gets there.  I think its kind of a mess with no clear contract to enforce by assert.

Perhaps figuring out how indices are passed down to `addIndex( )` would be a good thing to unsnarl later, but I didn't want to make this patch bigger than it is.

src/revlanguage/workspace/RevVariable.cpp
bredelings: I just checked, and this happens if you define `x[4]` but not `x[3]`.

bredelings: OK, I added a comment.

src/revlanguage/functions/UserFunction.cpp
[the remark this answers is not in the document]
bredelings: I can change the `auto` to `shared_ptr<Environment>` if you prefer.

src/revlanguage/workspace/RevVariable.cpp
bredelings: I guess this raises the question about the extent to which
* we should enforce coding style when reviewing.
* we should treat common patterns in revbayes as a style guide.

I tend not to enforce my coding style when reviewing, for better or worse.
Joelle also said that if we want to have a style guide, we should write one and get agreement to it-- not say that common patterns in revbayes source constitute a style guide.

mikeryanmay: To be clear I don't think this is a big deal, I was just mentioning it because I had not seen it before. For me what matters is that the code is readable and maintainable, which this certainly is.

src/revlanguage/parser/SyntaxIndexOperation.cpp
```diff
@@ -162,12 +163,12 @@ RevPtr<RevVariable> SyntaxIndexOperation::evaluateLHSContent( Environment& env,
                     }
                 
                     // get the slot and variable
-                    theElementVar  = env.getVariable( elementIdentifier );
+                    RevPtr<RevVariable> theElementVar  = env.getVariable( elementIdentifier );
```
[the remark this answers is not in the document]
bredelings: I can see why Sebastian (I presume) wrote it this way, so I'll default to not changing it.

Let me know if you feel strongly, I can rewrite it if needed.

I pushed some comment changes to make it clearer -- feel free to add more tweaks.

src/revlanguage/parser/SyntaxIndexOperation.cpp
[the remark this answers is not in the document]
bredelings: I changed the comments for this case also.

src/revlanguage/workspace/RevVariable.cpp
bredelings: Short version: that's a good question, but I think that `(idx < 1)` is a user error and not a programmer error.  Thoughts?

mikeryanmay: Yes I see. I agree with this distinction now.

mikeryanmay (approved):
Looks all good to me.

PR #269, opened 2022-09-17
Fix compatibility with C++17

bredelings:
RevBayes won't compile out-of-the-box on current XCode,  This is because RevBayes uses some deprecated features in C++14 that are actually removed in C++17. XCode now defaults to C++17, and its libc++ implementation actually removes the deprecated features.  GCC also defaults to C++17. 

The removed features are mainly:
* use of the `register` keyword in the generated parser/lexer.
* use of `std::random_shuffle( )` in a few files.

With this PR, RevBayes continues to target C++14, but removes deprecated features so that it is possible to compile it in C++17 mode.
* the parser/lexer are regenerated using more modern flex/bison, and so don't use the `register` keyword
* a simple implementation of random_shuffle( ) is placed in namespace `deprecated`
* `std::random_shuffle( )` is replaced with `deprecated::random_shuffle( )`.  People should use `std::shuffle(start, end, rng)` instead.

To review this patch, I suggest just reading the 3 change sets.  Except that you can ignore the one that regenerates the parser & lexer code because that is computer-generated code that is not really readable.  I think if the tests pass, you don't really need to compile locally.

### DIFF
--- a/src/core/distributions/phylogenetics/tree/birthdeath/SampledSpeciationBirthDeathProcess.cpp
+++ b/src/core/distributions/phylogenetics/tree/birthdeath/SampledSpeciationBirthDeathProcess.cpp
@@ -629,7 +629,8 @@ void SampledSpeciationBirthDeathProcess::simulateTree( void )
             std::vector<size_t> taxon_idx;
             for (size_t i = 0; i < taxa.size(); i++)
                 taxon_idx.push_back(i);
-            std::random_shuffle(taxon_idx.begin(), taxon_idx.end());
+
+            deprecated::random_shuffle(taxon_idx.begin(), taxon_idx.end());
             
             // Set names for terminal taxa
             for (size_t i = 0; i < taxa.size(); i++)
--- a/src/core/functions/phylogenetics/cladogenetic/DECCladogeneticStateFunction.cpp
+++ b/src/core/functions/phylogenetics/cladogenetic/DECCladogeneticStateFunction.cpp
@@ -817,7 +817,7 @@ std::string DECCladogeneticStateFunction::simulateDataAugmentedCladogeneticState
         if (event_type == "s")
         {
             // choose the area inherited by the right node
-            std::random_shuffle( node_child_on.begin(), node_child_on.end() );
+            deprecated::random_shuffle( node_child_on.begin(), node_child_on.end());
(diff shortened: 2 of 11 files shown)


davidcerny (commented):
This all seems good to me; the changes are mostly straightforward (aside from those in the parser and lexer code, which admittedly elude me) and exactly as described.

My only question is about the recommendation provided in the PR description that "[p]eople should use `std::shuffle(start, end, rng)` instead", which seems to contradict the continued reliance on `deprecated::random_shuffle`. As far as I understand, `std::shuffle(start, end, rng)` is valid under both C++14 and C++17, so why not use it?

bredelings:
Yeah..... I agree.  I originally wrote code that switches from `boost::mt19937` to `std::mt19937`.  This allowed using `std::shuffle( )`.  In theory it would also allow using the rng to draw samples from different distributions, such as the geometric distribution (see https://en.cppreference.com/w/cpp/numeric/random/geometric_distribution). 

HOWEVER.  The current MCMC trace tests rely on generating the same trace on both GCC and CLANG given the same seed.  I'm not sure that using `std::shuffle(  ,rng)` is actually guaranteed to yield the same results using libstdc++ (GCC) and libc++ (CLANG).  I think that `std::mt19937` would produce the same random numbers as `boost::mt19937`, but the algorithms for sampling could legally be different.  

It seems plausible that the results for `std::shuffle( )` might actually be the same, even though they are not legally required to be the same.  But I think I read that the results for some other distributions are actually different :grimacing:.  So, I tried the current approach, placing random_shuffle inside `deprecated( )` to avoid any new uses being added.  This is just postponing the problem...

Thoughts?

spydr073 (approved):
The changes all look fine to me. 

Since the MCMC tests rely on psuedo-random number generators to produce exactly the same sequence given the same seed, it may be a good idea to "freeze" a given implementation of an RNG and `shuffle`, and include it directly in the RevBayes source code.  Then we don't have to worry about potential implementation differences between libstdc++/libc++/boost/etc..

davidcerny (commented):
Agreed – it sounds like while the random number generator issue is something that'll have to be addressed eventually, it would be too much trouble to go that route right now just to make the code C++17-compatible. Thanks for explaining this.

PR #271, opened 2022-09-20
Accept constant rho and user-supplied starting trees in dnFBDP()

davidcerny:
This PR includes two simple fixes that address the following issues:

(1) While `dnBDSTP()` and its alias, `dnFBDP()`, currently accept an `initialTree` argument, they ignore it and proceed to simulate a starting tree regardless of whether a user-specified initial tree has been provided.

(2) While `dnBDSTP()` and `dnFBDP()` are indeed supposed to be aliases, their behavior differs when the `rho` argument (extant sampling fraction) is specified: the former accepts it, while the latter throws a seemingly unrelated error:

```
No parameter with name "rhoTimeline" found to set
```

Here, the code was changed to ensure that the `rho` argument is accepted by both aliases.

### DIFF
--- a/src/revlanguage/distributions/phylogenetics/tree/Dist_FBDP.cpp
+++ b/src/revlanguage/distributions/phylogenetics/tree/Dist_FBDP.cpp
@@ -445,7 +445,7 @@ void Dist_FBDP::setConstParameter(const std::string& name, const RevPtr<const Re
     {
         Mu_timeline = var;
     }
-    else if ( name == "PhiTimeline" || name == "PsiTimeline" )
+    else if ( name == "PhiTimeline" || name == "rhoTimeline" )
     {
         Phi_timeline = var;
     }
--- a/src/core/distributions/phylogenetics/tree/birthdeath/BirthDeathSamplingTreatmentProcess.cpp
+++ b/src/core/distributions/phylogenetics/tree/birthdeath/BirthDeathSamplingTreatmentProcess.cpp
@@ -161,15 +161,20 @@ BirthDeathSamplingTreatmentProcess::BirthDeathSamplingTreatmentProcess(const Typ
     prepareTimeline();
     prepareProbComputation();
 
-
-    // We employ a coalescent simulator to guarantee that the starting tree matches all time constraints
-    RbVector<Clade> constr;
-    StartingTreeSimulator simulator;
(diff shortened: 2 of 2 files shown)


src/revlanguage/distributions/phylogenetics/tree/Dist_FBDP.cpp
```diff
@@ -445,7 +445,7 @@ void Dist_FBDP::setConstParameter(const std::string& name, const RevPtr<const Re
     {
         Mu_timeline = var;
     }
-    else if ( name == "PhiTimeline" || name == "PsiTimeline" )
```
bredelings, line 448: I think the idea here is that "Phi" and "Psi" are synonyms.  Is "rho" a synonym for phi/psi?

If so, should we keep the "Psi" synonym when adding the "rho" synonym?

bredelings:
Which issue has the code for testing this PR?

src/revlanguage/distributions/phylogenetics/tree/Dist_FBDP.cpp
davidcerny: According to [this post](https://revbayes.slack.com/archives/C08UR8U3X/p1662526895818889), upper-case "Phi" should be the synonym of "rho" (extant sampling fraction), while lower-case "phi" should be the synonym of "psi" (fossil recovery rate). So the synonymy between "Phi" and "Psi" in the original code was likely a mistake – I'm not sure upper-case "Psi" means anything at all. But @afmagee should feel free to correct me if I got that wrong!

bredelings: Ah, that makes sense.  I'll defer to you and @afmagee on this.  It would be interesting to hear what @afmagee says.

afmagee: Well, that's a bug.

We adopted a unified notation for continuous rates and their event-based analogs, lambda/Lambda, mu/Mu, phi/Phi. In the literature, continuous sampling is often psi, while discrete is often rho (then again, so are mass extinctions). So yes, big-P Psi should be rho

afmagee: Good catch!

afmagee (approved):
The checks are passing and this all looks fine to me

PR #273, opened 2022-09-30
Add a ./build.sh script for meson.

bredelings:
This PR adds `./build.sh` script for meson.  It is similar in structure to the `./build.sh` script for cmake, but contains some modifications:
* in both scripts, I rename the `-help true` option to `-help2yml true`, to reduce confusion with the `-h` and `--help` options.
* meson doesn't have anything corresponding to the `-boost_verbose true` and `-boost_debug true` options, so I removed those
* `GitVersion.cpp` is handled inside `meson/generate.sh` instead of `build.sh`
* `meson/build.sh` gives error messages and suggestions if it can't find meson or ninja
* `meson/build.sh` creates a python virtual environment and tries to install meson or ninja if it can't find them.

The easiest way to examine `meson/build.sh` is probably to checkout the `build-sh-for-meson` branch and run `kdiff3 projects/cmake/build.sh projects/meson/build.sh`.

That said, I suggest mostly trying out the `build.sh` to see how well it works.


davidcerny:
I ran it on macOS (12.6) and CentOS (8.4.2105) using different command-line options (`./build.sh` and `./build.sh -mpi=true -help2yml=true`) and got RevBayes to compile just fine. My only question would be about the `-cmd` option, which is listed in help, but isn't set to a default value in the beginning of the script (unlike the other command-line options on lines 9–17). The script will not run when this option is specified, regardless of which value it is set to (i.e., neither `./build.sh -cmd=true` nor `./build.sh -cmd=false` works).

bredelings:
@davidcerny Thanks for checking!  I added default values for `cmd` -- though I'm not sure they are needed.

I would expect an error with `./build.sh -cmd true` if you don't have gtk installed.

I think another issue is that if you first run with `-cmd true`, then running with `-cmd false` doesn't change the configuration unless you first delete the build directory.  Hmm...

bredelings:
OK, I pushed a possible fix.  
I still expect that `./build.sh -cmd true` will fail if GTK is not installed.
Running `./build.sh -Dcmd=false` should _not_ work.
Running `./build.sh -cmd=false` should _not_ work.
However, running `./build.sh -cmd false` afterwards should work.

davidcerny (approved):
Apologies – I managed to confuse myself on the proper syntax for setting flags. Although I'm still a bit confused (e.g., `./build.sh -mpi true` works while `./build.sh -mpi=true` doesn't, but for some reason, `./build.sh -mpi true -help2yml true` and `./build.sh -mpi=true -help2yml=true` both work), the script itself is very clear on how to set flags and even provides helpful examples, so this is really a non-issue. I'm approving this PR.

bredelings:
> .... `/build.sh -mpi true -help2yml true` and `./build.sh -mpi=true -help2yml=true` both work)

Huh. That is weird.  The second one "works", but the arguments don't have any effect.  It looks like it parses as a single flag with a key of `-mpi=true` and a value of `-help2yml=true` :roll_eyes:   If you just do `./build.sh -mpi=true` then if does not work.

I added an error message for arguments of the for `-key=value`, since they should be `-key value`.  With this  change, we get:
> $ ./build.sh -mpi=true
> ./build.sh: I don't understand '-mpi=true' - did you mean '-mpi true'?

Since I did not come up with the idea for parsing command-line options using shell-scripting, I think I will not try to make further changes.

Thanks for checking all this, and thanks for approving.

PR #274, opened 2022-09-30
Correctness fix for node dirty flagging

bredelings:
Apparently removing the cached value and recomputing the likelihood leads to a different likelihood!  That has to be a bug.

Only the BDSTP test is actually affected by this, so I updated on the BDSTP test results.


jsigao (approved):
This looks good to me! (though it's always harder to notice any errors made by myself so an approval from a second reviewer may be more ideal...)

PR #279, opened 2022-10-05
Fix constrained topo

bredelings:
This PR fixes `Segmentation fault` errors in NNI and SPR.  In both cases, a loop was created in getParent() edges, and this lead to an infinite loop in recursive-dirty-flagging which overflowed the stack.

The PR also adds an error message indicating that the NNI proposal does not (currently) handle a root node with only 2 children.

### DIFF
--- a/src/core/datatypes/trees/Tree.cpp
+++ b/src/core/datatypes/trees/Tree.cpp
@@ -1733,6 +1733,11 @@ void Tree::resetTaxonBitSetMap( void )
 
 TopologyNode& Tree::reverseParentChild(TopologyNode &n)
 {
+    // This routine makes n the new root by
+    // * first making n.getParent() the new root, and then
+    // * making n.getParent() into a child of n.
+    // It returns the original root.
+
     TopologyNode* ret = &n;
 
     if ( !n.isRoot() )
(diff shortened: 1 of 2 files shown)


davidcerny:
The code changes all make sense, the new error message is very helpful, and the PR successfully addresses the problem of non-clock NNI getting stuck in an infinite loop. That said, I would suggest postponing the merge until we solve the analogous issue with SPR, which currently also keeps getting stuck when a backbone is provided (even when the backbone tree and the initial tree are both unrooted). I bet the source of the problem is the same – a loop getting created somewhere inside `doProposal()` – and I'd be happy to look into it, but will not be able to do it until next week.

bredelings:
It looks like with SPR, we get a similar infinite loop in parent edges.  It comes from this line in `src/core/moves/proposal/tree/SubtreePruneRegraft/Proposal.cpp`:

``` C++
        TopologyNode& regraft = tau.reverseParentChild(*newParent);
```

bredelings:
OK, this fixes the SEGFAULT in SPR.

davidcerny (approved):
You are amazing, Ben – very happy to approve this! :-) I can confirm that our nonclock backbone test script which was previously causing RevBayes to either hang or segfault now runs without any issues (with NNI, SPR, or both), and produces a posterior distribution of correctly constrained trees. Thank you for your work on this!

PR #280, opened 2022-10-06
Don't crash in setwd(dir) if dir does not exist.

bredelings:
Without the patch we get (for example):
```
> setwd("asdf")
terminate called after throwing an instance of 'boost::filesystem::filesystem_error'
  what():  boost::filesystem::current_path: No such file or directory: "asdf"
Aborted
```
With the patch, we get
```
> setwd("asdf")
   Error:	setwd: The path "asdf" does not exist.
```

The PR:
* checks if the filename exists and is a directory first, before we try and set it as the current directory.
* modifies the read-eval-print loop not to crash on exceptions that are not RbException, such as boost::filesystem::filesystem::error.

### DIFF
--- a/src/revlanguage/functions/basic/Func_setwd.cpp
+++ b/src/revlanguage/functions/basic/Func_setwd.cpp
@@ -46,11 +46,15 @@ Func_setwd* Func_setwd::clone( void ) const
 /** Execute function */
 RevPtr<RevVariable> Func_setwd::execute( void )
 {
-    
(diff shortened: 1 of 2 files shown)


spydr073 (approved):
Both changes look good, and makes for a better user experience.

One thing that we may want to consider though is the use of boost libraries. In the past there have been pushes to minimize reliance on external libraries. Not sure if this will come up again in the future.

bredelings:
Long-term, would should replace boost::filesystem with std::filesystem.

But, in order to do this, we need to
* change the standard to c++17
* require gcc v9 or higher

Issue #239, opened 2022-05-01
[BUG] taxa(), ntips() and rootAge() don't work


siariazi:
Hello, 

I’m using RevBayes version 1.1.1 on Windows and now trying to reproduce the results of “Simple Diversification Rate Estimation” from here:  [https://revbayes.github.io/tutorials/divrate/simple.html](url)

When I run this line of code:
`taxa <- T.taxa()`

I get this message:
`Error: No function named 'taxa'`

I get the same message for `T.ntips()` and `T.rootAge()` as well.

In addition when I run this line:
`timetree ~ dnBDP(lambda=birth_rate, mu=0.0, rho=rho, rootAge=root_time, samplingStrategy="uniform", condition="survival", taxa=taxa)`

I get this message: 
```
Error:	Argument or label mismatch for function call.
   Provided call:
   dnBDP (RealPos<stochastic> 'lambda',
          Probability<constant> 'mu',
          RealPos<constant> 'rho',
          Natural<constant> 'rootAge',
          String<constant> 'samplingStrategy',
          String<constant> 'condition',
          Natural<constant> 'taxa',
          Natural<constant> 'Clade' )
   Correct usage is:
   dnBirthDeath (RealPos<any> lambda,
                 RealPos<any> mu,
                 Probability<any> rho,
                 RealPos<any> rootAge,
                 String<any> samplingStrategy {valid options: "uniform"|"diversified"},
                 String<any> condition {valid options: "time"|"survival"|"nTaxa"},
                 Taxon[]<any> taxa,
                 Clade[]<any> incompleteClades)
```
Where there is no clade provided in the sample data. 
Are those functions revised and there is a new tutorial covering them?

davidcerny:
Hi @siariazi, I unfortunately don't have access to a Windows machine, but I tried – and failed – to reproduce this error using RevBayes 1.1.1 on macOS.

In my experience, the error you describe is often due to the fact that RevBayes distinguishes between a vector of trees of length 1, and a single tree that forms an element of such a vector. In the following example, `T` is actually a vector of trees (as indicated by the square brackets, `[]`), and I get the same error as you:

```
> T <- readTrees("data/primates_tree.nex")
   Attempting to read the contents of file "primates_tree.nex"
   Successfully read file
> type(T)
    TimeTree[]
> taxa <- T.taxa()
   Error:	No function named 'taxa'
```

[...]

PR #281, opened 2022-10-06
Mention object type in error messages for missing method `object.method()`.

bredelings:
This is motivated by the common problem where people write `T <- readTrees("data/primates_tree.nex")`.  

Since the subscript `[1]` is missing, T is not a tree and so method calls like `T.taxa()` don't work.  This results in bug reports like #239.

This patch generates the "no method" error from `SyntaxFunctionCall::evaluateContent( )` instead of from `FunctionTable::findFunction( )`.  SyntaxFunctionCall has access to the type of the base variable, so we can add that to the error message.  We also suggest using `.methods()`.

```
> T = readTrees("primates.tree")
   Attempting to read the contents of file "primates.tree"
   Successfully read file
> T.ntips()
   Error:	Variable of type 'TimeTree[]' has no method called 'ntips'.  You can use '.methods()' to find available methods.
```

This will hopefully prevent bug reports like #239, as well as making `.methods()` more discoverable.

### DIFF
--- a/src/revlanguage/workspace/FunctionTable.cpp
+++ b/src/revlanguage/workspace/FunctionTable.cpp
@@ -287,7 +287,7 @@ std::vector<Function *> FunctionTable::findFunctions(const std::string& name) co
 
 
 /** Find function (also processes arguments) */
-const Function& FunctionTable::findFunction(const std::string& name, const std::vector<Argument>& args, bool once) const
+const Function* FunctionTable::findFunction(const std::string& name, const std::vector<Argument>& args, bool once) const
 {
     
     std::pair<std::multimap<std::string, Function *>::const_iterator,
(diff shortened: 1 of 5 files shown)


bredelings:
Fixes #239

src/revlanguage/workspace/FunctionTable.cpp
```diff
@@ -305,7 +305,7 @@ const Function& FunctionTable::findFunction(const std::string& name, const std::
         }
         else
         {
-            throw RbException("No function named '"+ name + "'");
+            return nullptr;
```
hoehna, line 308: Why does this now return a nullptr instead of the error message?

bredelings: I added the following comment:
```
            // Here we just return a nullptr to indicate failure.
            // The caller can produce better error messages, so it is the caller's job to throw an exception.
```

hoehna (approved):
Great, looks good.

PR #282, opened 2022-10-06
Remove degree-2 root nodes from non-clock trees by default in `readTrees` and `readTreeTrace`.

bredelings:
This PR modifies `readTrees( )` and `readTreeTrace( )` to remove degree-2 nodes at the root from non-clock trees by default.
The patch adds the argument `unroot_nonclock`, and sets it to `TRUE` by default.
When `TRUE`, trees that are read in are marked unrooted even if they were read in as rooted.
I have checked that the removal of degree-2 nodes works and correctly merges branch lengths for the 2 adjacent sub-branches.

The PR also fixes
* an infinite loop reading a newick tree with a missing right parenthesis.
* a segfault if the tree file does not exist and we read it as `treetype="non-clock"`


davidcerny (approved):
The newly added code is very clearly written and well-documented (I especially like `Tree::removeDegree2Node()`), and seems to work exactly as intended. There are just two things I found:

(1) This is very nitpicky, but the new exception added to `NewickConverter::createNode()` should proabably read "Not enough closing parentheses" (i.e., with the word "parentheses" in the plural).

(2) I was unsure what a typical use case would be for the `unroot_nonclock` option in the `readTreeTrace()` function, so the way I tested it out was to give RevBayes a trace of clock trees but (mis)specifying `treetype` as "non-clock". To see if the root removal was successful, I tried generating a summary tree using either `mapTree()` or `mccTree()`, expecting the resulting tree to have _n_ tips and 2_n_ − 2 (rather than 2_n_ − 1) nodes. However, passing the unrooted tree trace to these functions resulted in a segfault inside `TreeSummary::annotateTree()`. I don't think this is necessarily problematic, because my test was almost certainly not valid – I just wasn't sure how to get RevBayes to give me a trace of rooted non-clock trees that this option would ideally be applied to.

bredelings:
Hi David, thanks for looking at this!  Can you post the tree trace that crashes for you?

I just tried this out, and I didn't get a crash:
```
fn = "primates_BDP.trees"
t1 = readTreeTrace(fn, treetype="clock")
t2 = readTreeTrace(fn, treetype="non-clock",unroot_nonclock=FALSE)
t3 = readTreeTrace(fn, treetype="non-clock",unroot_nonclock=TRUE)

mccTree(t1).nnodes()
mccTree(t2).nnodes()
mccTree(t3).nnodes()

mapTree(t1).nnodes()
mapTree(t2).nnodes()
mapTree(t3).nnodes()
```

davidcerny:
[This](https://github.com/revbayes/revbayes/files/9833415/test_trace.txt) is the trace I'm having trouble with, but now I think I know why – it includes sampled ancestors, which we would not expect to get in non-clock trees. So it was just a case of me choosing an inappropriate test case.

bredelings:
I "fixed" the crash that you were seeing by not marking the trees as unrooted after removing the degree-2 root.  The code paths for unrooted trees are not very well tested, and so its probably safer to avoid marking trees unrooted for now.

I also got about a 4x speedup for the mccTree operation by replacing RbBitSet with boost::dynamic_bitset<>.  RbBitSet was based on vector<bool> which is really slow.  The interface for dynamic_bitset is _almost_ identical, so it was not very hard to rename the few member functions with different names, and then make RbBitSet into a typedef for boost::dynamic_bitset<>, which is _much_ more efficient.

PR #288, opened 2022-10-14
Fix false claim of conflicting constraints when simulating a constrained tree

bredelings:
Before simulating the constrained tree, we first sort clade constraints so that tip-ward clades are first.  However, this doesn't make sense for negated constraints, so we need to avoid checking for overlapping taxa between negated constraints and other constraints.

Another issue was that we said that there were conflicting constraints, but didn't say which ones.  So, print out the supposedly-conflicting constraints.  When regular constraints conflict, its because their taxon sets overlap, so also print out the overlapping taxa.

Another issue was that we printed out negated or "optional" constraints as if they were regular clade constraints.  So, print out `NOT A` for negative constraints and `(A OR B OR C)` for optional constraints. 

The underlying issue may be that the code for simulating initial trees is just really fragile, and needs a thorough rework.

A deeper issue is that we are pretending that a constraint of the form (A OR B OR B) is a `Clade`.  It really is a Constraint, but not a Clade.  We should separate `Clade` from constraints -- perhaps by creating classes `CladeConstraint`, `OrConstraint`, and `NegativeConstraint` or something like that.

### DIFF
--- a/src/core/datatypes/trees/Clade.cpp
+++ b/src/core/datatypes/trees/Clade.cpp
@@ -488,7 +488,7 @@ bool Clade::isNestedWithin(const Clade& c) const
  *
  * \return       The true/false value of whether the clade is a negative constraint.
  */
-bool Clade::isOptionalMatch(void) const
+bool Clade::isOptionalConstraint(void) const
 {
     return is_optional_match;
 }
--- a/src/core/datatypes/trees/Clade.cpp
+++ b/src/core/datatypes/trees/Clade.cpp
@@ -5,6 +5,7 @@
 #include <string>
 #include <utility>
 
+#include "StringUtilities.h"
 #include "RbVectorUtilities.h"
 #include "RbException.h"
 
@@ -672,22 +673,38 @@ size_t Clade::size(void) const
  */
 std::string Clade::toString( void ) const
 {
-    std::string s = "{";
-    
-    for (size_t i = 0; i < taxa.size(); ++i)
+    std::string s;
+
+    if (isOptionalConstraint())
     {
-        if ( i > 0 )
+        vector<string> cstrings;
+        for(auto& c: getOptionalConstraints())
         {
-            s += ",";
+            cstrings.push_back(c.toString());
         }
-        s += taxa[i].getName();
-        if ( std::find(mrca.begin(), mrca.end(), taxa[i]) != mrca.end() )
-        {
-            s += "*";
-        }
+
+
+        s = "(" + StringUtilities::join(cstrings, " OR ") + ")";
     }
-    s += "}";
-    
+    else
+    {
+        vector<string> tstrings;
+        for (size_t i = 0; i < taxa.size(); ++i)
+        {
+            string t = taxa[i].getName();
+            if ( std::find(mrca.begin(), mrca.end(), taxa[i]) != mrca.end() )
+            {
+                t += "*";
+            }
+            tstrings.push_back(t);
+        }
+
+        s  = "{" + StringUtilities::join(tstrings,", ") + "}";
+    }
+
+    if (isNegativeConstraint())
+        s = "NOT " + s;
+
     return s;
 }
 
--- a/src/core/datatypes/trees/Clade.cpp
+++ b/src/core/datatypes/trees/Clade.cpp
@@ -478,7 +478,7 @@ bool Clade::isOptionalConstraint(void) const
 /**
  * Does the provided clade overlaps with myself? An overlap is if we share at least one taxon.
  *
- * \param[in]    c    Theclade.
+ * \param[in]    c    The clade.
  *
  * \return       True/False, if there is an overlap.
  */
@@ -499,6 +499,36 @@ bool Clade::overlaps(const Clade& c) const
 
     return false;
 }
+
+/**
+ * Compute the set of overlapping taxa between clade c and myself.
+ *
+ * \param[in]    c    The clade.
+ *
+ * \return       The set of overlapping taxa.
+ *
+ * We use this method to give a more helpful error message
+ *  when we claim that clades conflict.
+ */
+set<Taxon> Clade::intersection(const Clade& c) const
+{
+    set<Taxon> both;
+    size_t N = taxa.size();
+    for ( size_t i=0; i<c.size(); ++i)
+    {
+        const Taxon& t = c.getTaxon(i);
+        for ( size_t j=0; j<N; ++j)
+        {
+            if ( taxa[j] == t )
+            {
+                both.insert(t);
+            }
+        }
+    }
+
+    return both;
+}
+
 
 
 /**
--- a/src/core/datatypes/trees/Clade.cpp
+++ b/src/core/datatypes/trees/Clade.cpp
@@ -750,7 +750,7 @@ void Clade::setAges(const std::vector<Taxon>& taxa)
 // We really should separate the idea of constraints (CladeConstraint,
 // NotCladeConstraint, OptionalCladeConstraint) from the idea of clades.
 
-bool clade_before(const Clade& c1, const Clade& c2)
+bool cladeBefore(const Clade& c1, const Clade& c2)
 {
     if (&c1 == &c2)
         return false;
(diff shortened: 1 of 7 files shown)


src/core/datatypes/trees/Clade.cpp
```diff
@@ -5,24 +5,22 @@
...
  * this default object.
  */
-Clade::Clade( void ) :
-    age( 0.0 ),
-    clade_name(""),
-    num_missing( 0 ),
-    taxa(),
-    is_negative_constraint(false),
-    is_optional_match(false)
+Clade::Clade( void )
```
hoehna, line 23: why is it better to not initialize the values with specific defaults? I think that's better style because it's more explicit.

src/core/datatypes/trees/Clade.cpp
```diff
@@ -32,13 +30,8 @@ Clade::Clade( void ) :
...
  */
 Clade::Clade( const Taxon &t, const RbBitSet &b ) :
-    age( 0.0 ),
     bitset( b ),
-    clade_name( "" ),
-    num_missing( b.size() > 1 ? b.size() - 1 : 0 ),
-    taxa(),
-    is_negative_constraint(false),
-    is_optional_match(false)
+    num_missing( b.size() > 1 ? b.size() - 1 : 0 )
```
hoehna, line 34: same as above

src/core/datatypes/trees/Clade.cpp
```diff
@@ -52,13 +45,9 @@ Clade::Clade( const Taxon &t, const RbBitSet &b ) :
...
  */
 Clade::Clade(const std::vector<Taxon> &n, const RbBitSet &b) :
-    age( 0.0 ),
     bitset( b ),
-    clade_name( "" ),
     num_missing( b.size() > n.size() ? int(b.size()) - int(n.size()) : 0 ),
-    taxa( n ),
-    is_negative_constraint(false),
-    is_optional_match(false)
+    taxa( n )
```
hoehna, line 50: Same as above

src/core/datatypes/trees/Clade.cpp
```diff
@@ -519,6 +501,26 @@ bool Clade::overlaps(const Clade& c) const
     return false;
 }
 
+set<Taxon> Clade::intersection(const Clade& c) const
```
hoehna, line 504: Could you write a one-line comment above what this method is doing and should be used for?

src/core/datatypes/trees/Clade.cpp
```diff
@@ -728,3 +731,117 @@ void Clade::setAges(const std::vector<Taxon>& taxa)
...
+
+// These should provide a "strict weak ordering" that returns
+// true only if c1 is ordered with respect to c2, and c1 < c2.
+// We do not treat conflicting constraints as unordered.
+//
+// Its not really clear how to handle negative constraints or optional constraints.
+// We really should separate the idea of constraints (CladeConstraint,
+// NotCladeConstraint, OptionalCladeConstraint) from the idea of clades.
+
+bool clade_before(const Clade& c1, const Clade& c2)
```
hoehna, line 743: Usually we use camel-case for our methods and class names. Maybe change this?

src/core/datatypes/trees/Clade.cpp
```diff
@@ -728,3 +731,117 @@ void Clade::setAges(const std::vector<Taxon>& taxa)
...
+    else if (not (c1.isNegativeConstraint() or c1.isOptionalConstraint()) and (c2.isNegativeConstraint() or c2.isOptionalConstraint()))
+        return true;
+    else if (c1.isNegativeConstraint() or c1.isOptionalConstraint() or c2.isNegativeConstraint() or c2.isOptionalConstraint())
+        return false;
+
+    else
+        return c1.getAge() < c2.getAge();
+}
+
+bool clade_within(const Clade& c1, const Clade& c2)
```
hoehna, line 758: camel case?

src/core/datatypes/trees/Clade.cpp
```diff
@@ -728,3 +731,117 @@ void Clade::setAges(const std::vector<Taxon>& taxa)
...
+        throw e;
+    }
+    else
+    {
+        // unordered!
+        return false;
+    }
+}
+
+bool clade_within_and_before(const Clade& c1, const Clade& c2)
```
hoehna, line 794: camel case

src/core/datatypes/trees/Clade.h
```diff
@@ -87,21 +88,24 @@ namespace RevBayesCore {
...
-        int                                         num_missing;
+        int                                         num_missing = 0;
         std::set<Taxon>                             mrca;
         std::vector<Taxon>                          taxa;
-        bool                                        is_negative_constraint;
-        bool                                        is_optional_match;
-        std::vector<Clade>                          optional_constraints;
-        
+        bool                                        is_negative_constraint = false;
+        boost::optional<std::vector<Clade>>         optional_constraints;
```
hoehna, line 98: what is this boost optional? Shouldn't a plain pointer work? Just thinking that not introducing more special language features makes the code harder to read. Just my old school way ...

src/core/utils/StringUtilities.h
```diff
@@ -17,6 +17,7 @@
 #include <stddef.h>
 #include <sstream> // IWYU pragma: keep
 #include <vector>
+#include <boost/optional.hpp>
```
hoehna, line 20: is boost optional actually used in this file?

hoehna (changes requested):
This is very useful, but I had a few stylistic suggestions to keep it more in line with the code base and some clarifications.

src/core/datatypes/trees/Clade.cpp
bredelings: They actually are initialized with specific defaults - but in the class declaration instead.
Then each constructor can mention which defaults are DIFFERENT from the global defaults in the class declaration.

src/core/datatypes/trees/Clade.cpp
bredelings: See above.

src/core/datatypes/trees/Clade.cpp
bredelings: see above

src/core/datatypes/trees/Clade.h
bredelings: Hmm... so the optional<T> is definitely an improvement over having a separate boolean flag, because (i) we don't have to worry about keeping the flag in sync and (ii) we don't allocate a clade if we aren't using it.  I used boost::optional here, because std::optional<T> isn't available until c++17. 

std::optional<T> is different than pointers because it allocates on the stack, instead of on the heap.  It replaces the common practice of i.e. returning something like `-1` to indicate "no value", which can lead to mistakes.  I think it makes code _easier_ to read (after you get used to it) by explicitly stating that we may-or-may-not have a value.  For example, you can write `if (optional_constraints.has_value())`...

I would argue that std::optional<T> is one of the few features that is definitely worth the time to learn (See https://en.cppreference.com/w/cpp/utility/optional).

I also found this: https://devblogs.microsoft.com/cppblog/stdoptional-how-when-and-why/

src/core/datatypes/trees/Clade.cpp
bredelings: Done.

src/core/datatypes/trees/Clade.cpp
bredelings: Fixed.

src/core/datatypes/trees/Clade.cpp
bredelings: Fixed.

src/core/datatypes/trees/Clade.cpp
bredelings: Fixed.

src/core/utils/StringUtilities.h
bredelings: Oops, it was temporarily, but not anymore.  Thanks for catching this. Fixed.

hoehna (approved):

PR #295, opened 2022-10-27
Improve speed and memory usage for DNA data by about 5x

bredelings:
While looking for a problem in a someone's data set, I noticed that RevBayes was spending a lot of time copying and creating DNA characters in `readDiscreteCharacterData( )`.

One big culprit is that a `vector<double> weights` field was added to DiscreteCharacterState.  This is slow to copy because it requires a memory allocation for each character.  It also expands the memory footprint.  This field seems to have originally been part of PoMoState and PoMoState4, so I moved the field back to those classes.  Probably the idea needs a better design.

I did some testing, and this reduces the time to read a large DNA matrix from 6.2s to 1.4s -- about 4.5x faster.

It reduces the memory usages for that matrix from 3100M to 610M, about 5x smaller.

### DIFF
--- a/src/core/datatypes/phylogenetics/character/DiscreteCharacterState.cpp
+++ b/src/core/datatypes/phylogenetics/character/DiscreteCharacterState.cpp
@@ -10,9 +10,7 @@ using namespace RevBayesCore;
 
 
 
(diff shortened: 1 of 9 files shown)


hoehna (approved):
This looks good and was probably long overdue.

Issue #250, opened 2022-06-23
[BUG] maximum number of monophyletic constraints is 15


anoukwillemsen:
There appears to me a maximum to the number of monophyletic constraints you can give. This maximum number appears to be 15, when any additional constraints are included I get the following error: "Segmentation fault (core dumped)"

This error occurred when I tried to construct a timetree with constraints on the calibration nodes. If I limit the number to 15 constraints, does not matter which constraints, the error does not occur and I am able to construct my timetree

Can this error please be fixed so I am able to include more than 15 monophyletic constraints?

I am working on Linux Ubuntu 20.04.4 LTS with RevBayes v.1.1.1

bredelings:
Hi Anouk,

Can you please provide an example script and data set that produces this problem?  If you can do that, I will take at look.

bredelings:
Hi @anoukwillemsen, #288 might fix this.  Let me know if you want more info.

bredelings:
Looks like #299 actually does fix this.

bredelings:
You can test this out with the preview release https://github.com/revbayes/revbayes/releases/tag/v1.2.1-preview8

The executables will should available in about an hour.

willgearty:
I'm currently using the Windows version of RevBayes (v1.2.1, build from 04666d), and this does not appear to be fixed. My analysis works just fine when I have 15 constraints, but when I add a single additional constraint it no longer progresses.

Issue #162, opened 2021-03-25
[BUG] Error loading topological constraints above a certain number of species


NickCrouch:
There appears to be an issue when trying to define topological constraints that contain a large number of species. Specifically, when defining a constraint (either using `source()` or by directly copying and pasting) an error is produced:
`Syntax error while reading character '"' at position 8178 in command:`
The precise number can change slightly, reflecting I think differences in the lengths of taxa within the constraint, but always falls within a certain range (e.g. 8179, 8187). 

To demonstrate I will use the taxonomic data provided by Jetz et al. (2012) from birdtree.org. First, I use this function to quickly create a line of code reflecting a particular constraint. 

```{r}

make.constraint.line <- function(x, name){

seq.nums <- seq(1, length(x), by=10)

res <- lapply(1:length(seq.nums), function(ee){
  start <- seq.nums[ee]
  end <- start + 9
  tmp <- x[start:end]
  tmp[!is.na(tmp)]
})

res.lines <- lapply(res, make.single.line)

num.res.lines <- length(res.lines)

if(num.res.lines>1){
 target.rows <- seq(1, by=1, length.out = num.res.lines-1)
 for(i in 1:length(target.rows)){
  res.lines[[i]] <- paste0(res.lines[[i]],",")
 }
}

res.lines[[1]] <- paste0("clade(", res.lines[[1]])
res.lines[[num.res.lines]] <- paste0(res.lines[[num.res.lines]], ")")

constraint.name <- paste0("constraint_",name)
res.lines[[1]] <- paste0(constraint.name, " = ", res.lines[[1]])
return(res.lines)
}

make.single.line <- function(y){
  y <- paste(y, collapse=" ")
  y <- gsub(' ','","',y)
  y <- paste('"',y,'"', collapse="")
  y <- gsub(" ","",y)
  return(y)
}
```
Now we can quickly generate dummy constraint files with different numbers of species

```{r}
# Taxonomic data
dat <- read.csv("BLIOCPhyloMasterTax.csv")

# sample species
species.sample <- sample(dat$TipLabel, 300, replace=FALSE)

# make the control line
control.line <- make.constraint.line(species.sample, "sample")

# export
fileConn <- file("exampleConstraint.Rev")
writeLines(as.character(control.line), fileConn)
close(fileConn)
```
[...]

PR #299, opened 2022-11-03
Fix constraints for real this time.

bredelings:
This is fixes problems reported in #162 and #250.  It is a followup to a previous PR #288 that partially fixed this problem, but not completely.

PR #288 used a partial order to sort the clades, which std::sort cannot handle.

Instead just sort the clades by the number of taxa they contain.  This should ensure that each clade comes before any other clade that includes it.  We still need to sort negative and optional constraints after real clades for this to work.

Since we aren't using cladeWithin anymore, we need to check for conflicting constraints separately.  Factor this out into a new function `checkConstraints(clades)`.  Also check that nested clades aren't older than clades that contain them.  Make sure to print out clades that supposedly conflict, so the user can tell why there are errors.

The order `cladeBefore` is still used in a few places.  Just comparing numbers with NaNs doesn't provide an order that you can sort on.  We need to place all clades without ages either before or after clades with ages, so place them before.  So fix that.

Finally, remove cladeWithinAndBefore -- this is no longer used, and in retrospect wasn't really a good idea anyway.  I don't think we should have two identical clades with different ages, so we don't need ages in order to sort the clades.  We can still have conflicts in clade ages, but sorting by age didn't do a good job finding them.  We now check for those conflicts in the new `checkConstraints( )` function.

LONG TERM, we should probably solve the constraints in a different way -- see note in `TopologyConstrainedTreeDistribution.cpp`.  We can construct a star tree and impose clades on it without sorting the clades.  Also, we don't have a method to solve general constraints, so an alternative approach would be to let the MCMC search solve the constraints.

### DIFF
--- a/src/core/datatypes/trees/Clade.cpp
+++ b/src/core/datatypes/trees/Clade.cpp
@@ -742,11 +742,17 @@ void Clade::setAges(const std::vector<Taxon>& taxa)
     }
 }
 
-// These should provide a "strict weak ordering" that returns
-// true only if c1 is ordered with respect to c2, and c1 < c2.
-// We do not treat conflicting constraints as unordered.
+// std::sort needs a "strict weak ordering" that returns true
+//    only if c1 is ordered with respect to c2, and c1 < c2.
+//
+// cladeWithin only provides a partial ordering, so we can't use std::sort with that.
+// We would need to use a topological sort, which is more complex.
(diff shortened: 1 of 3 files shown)


wrightaprilm (approved):
I’ve now tested this on the dataset provided in #250, as well as a dataset I had with a ton of constraints. Works as expected and produces no errors. Great work, @bredelings

PR #301, opened 2022-11-07
switching submodules to use ssh instead of https

hoehna:
[Github doesn't allow https anymore but requires ssh for authentication. This should also be the default for the submodules.](https://github.com/revbayes/revbayes/pull/300)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "tests"]
 	path = tests
-	url = https://github.com/revbayes/tests.git
+	url = git@github.com:revbayes/tests.git


bredelings:
Hi Sebastian.  It looks like the tests failed because they don't have permission to check out the tests repo via ssh.  I think github supports both http and ssh, but (i) you need ssh to push and (ii) if you use ssh you must have a github account.

An alternative approach is to leave the submodule as https (in the revbayes repo) and then do
```
cd tests
git remote rm origin
git remote add origin git@github.com:revbayes/revbayes.git
```
Then the tests submodule will be https by default, but ssh for you.

bredelings:
This doesn't work :-(

PR #306, opened 2022-11-09
Fix deprecated `set-output::value` command in github actions.

bredelings:
We are now supposed to write `echo "name=value" >> $GITHUB_OUTPUT`.

Since this is for the github actions scripts, and they seem to be passing, I would suggest that review should just be a quick glance over the code changes.


brpetrucci (approved):
Seems good to me!

PR #312, opened 2022-12-12
Fix GHA tests - remove obsolete gcc-7 and gcc-8 checks

bredelings:
Tests are failing because the ubuntu-latest image doesn't have gcc-7 and gcc-8.

This branch removes those checks (for gcc from 2017/2018) and adds a check for gcc-11 (for 2021).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,30 +12,17 @@ jobs:
   build:
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


bredelings:
Per our discussion on the call today
* this only affects clusters
* it doesn't affect people with older hardware, as newer linux software run just fine on old hardware.

PR #315, opened 2022-12-15
Select jupyter with --jupyter command-line flag.

bredelings:
This allows using jupyter notebooks with the standard rb executable.


wrightaprilm (approved):
This works and we have updated the revbayes_kernel to reflect these changes.

PR #316, opened 2022-12-24
fix: add memory lib

starsareintherose:


bredelings:
Hi Guoyi,  thanks for your contribution!

Your should probably be working on the development branch, not the master branch.  I rebased your branch on development so that the tests will run correctly.  You might need to pull with `git pull --rebase`.

Can you say a few words about why we need to `#include <memory>` here?

starsareintherose:
Okay, because the compile issue, I use the meson to compile revbayes and then I received report like this
```
../src/revlanguage/functions/UserFunction.cpp: In member function ‘virtual RevLanguage::RevPtr<RevLanguage::RevVariable> RevLanguage::UserFunction::executeCode()’:
../src/revlanguage/functions/UserFunction.cpp:69:32: error: ‘make_unique’ is not a member of ‘std’
   69 |     auto function_frame = std::make_unique<Environment>( getEnvironment(), "UserFunctionEnvironment" );
      |                                ^~~~~~~~~~~
../src/revlanguage/functions/UserFunction.cpp:13:1: note: ‘std::make_unique’ is defined in header ‘<memory>’; did you forget to ‘#include <memory>’?
   12 | #include "Workspace.h"
  +++ |+#include <memory>
   13 | #include "Argument.h"
../src/revlanguage/functions/UserFunction.cpp:69:55: error: expected primary-expression before ‘>’ token
   69 |     auto function_frame = std::make_unique<Environment>( getEnvironment(), "UserFunctionEnvironment" );
      |                                                       ^
```
This is because smart pointers need it. My c++ version is 
```
c++ (GCC) 12.2.0
```

bredelings:
Your patch is clearly correct.  Thanks for finding this problem!

I am not sure why our tests did not catch this problem, because we have gcc-12 tests.  On my laptop, compilation with gcc-12 works.  So perhaps some other include file is including <memory> accidentally.

All the test pass with your patch,  but apparently sending the test results to slack is failing.  I will approve this PR I figure out how to fix that.

Thanks again!

bredelings:
Hi, can do this patch as a branch within the revbayes repository?
The checks currently use a "secret" that is only known inside the revbayes repository, so they won't pass on branches of other repos.
If this is too complicated, I'll just re-submit this as a new PR.

starsareintherose:
I will try to resubmit

bredelings (approved):

bredelings:
Apparently, its possible to merge the PR even in the slack step fails.  It is not a required step.  So, I should be able to merge this when the build tests finish.

PR #323, opened 2023-02-14
Changed simulateDivergenceTime to avoid overflow.

brpetrucci:


bredelings:
This all looks good.  Thanks for checking my math!  You were right that the first and second cases agree, so the only problem was the divide by zero when dividing by exp(-age*x).  I also like that you've refactored the formula to look like (... + age).

bredelings (changes requested):
What do you think about adding two assert statements:
- `assert(age >= 0)` at the beginning of the function?
- `assert(0 <= t and t <= age)` at the end of the function?

brpetrucci:
> What do you think about adding two assert statements:
> 
> * `assert(age >= 0)` at the beginning of the function?
> * `assert(0 <= t and t <= age)` at the end of the function?

That sounds good! Do I just add those two lines, or are there headers or something of the sort I need to change as well?

bredelings:
Hmm, perhaps you should add `#include <cassert>` as well.  Often some other header includes <cassert>, so it might compile without it, but it can't hurt to add it.

brpetrucci:
Made those changes! Let me know what you think.

bredelings (approved):

PR #326, opened 2023-02-23
Fix multi summary

bredelings:
RevBayes was crashing whenever we annotated a tree trace multiple times.  This was because it marked the individual runs "clean" after the first annotation.  Then, during the second annotation, it skipped computing the summary statistics (i.e. split frequencies) because all the individual runs were clean.

Solved by separately recording whether the summary statistics are computed yet.

In addition, this patch
1. Improves readability for TreeSummary.cpp
2. Makes mccTree substantially faster.
3. Renames splitFrequency to splitCount since it is an integer, and creates a new function splitFrequency that is a fraction.  Similar for topologyCount/topologyFrequency.

You could review this patch as one big change, but I recommend reviewing each of the 4 parts separately.  The readability patch is the largest, and should make no functional changes.

--

BTW, for the "readability" part of the patch, I'm using 2 c++  features introduced in 2011, and one introduced in 2017.
(i) range-based for loops: `for(int i: v) { foo(i) }` to iterate over values inside v, where v is (say) a vector<int>. https://en.cppreference.com/w/cpp/language/range-for

(ii) the `auto` type: `auto iter = x.begin()` instead of `vector<TreeTrace*>::iterator iter = x.begin()`, where the type of `iter` can be deduced. https://en.cppreference.com/w/cpp/language/auto. 

(iii) decomposing pairs and tuples: You can now write `auto [x,y] = p` instead of `x=p.first; y=p.second`.  In for loops, you can write `for(auto& [tree,count]: tree_counts)` to iterate over a map. https://en.cppreference.com/w/cpp/language/auto

### DIFF
--- a/src/core/analysis/mcmc/output/TreeSummary.cpp
+++ b/src/core/analysis/mcmc/output/TreeSummary.cpp
@@ -11,6 +11,7 @@
 #include <string>
 #include <utility>
 #include <vector>
+#include <optional>
 
 #include "NewickConverter.h"
 #include "ProgressBar.h"
@@ -734,7 +735,7 @@ void TreeSummary::annotateTree( Tree &tree, AnnotationReport report, bool verbos
             // annotate CCPs
             if ( !n->isTip() && report.conditional_clade_probs )
             {
-                double parentCladeFreq = splitFrequency( parent_split );
+                double parentCladeFreq = splitCount( parent_split );
                 double ccp = condCladeAges[split].size() / parentCladeFreq;
                 n->addNodeParameter("ccp",ccp);
             }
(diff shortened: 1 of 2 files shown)


brpetrucci (approved):

wrightaprilm (approved):
- [x] Compiles
- [x] Computes multiple tree summaries sequentially
- [x] Computes different options of mcc (`positiveBranchLengths = true` vs not)
- [x] Computes the same MCC, maj rule, and MAP as prior versions of RevBayes
- [x] New C++ vectorization very readable.

davidcerny:
I'm not qualified to comment on every detail of the code changes, either, but renaming "frequencies" to counts where appropriate is very helpful and makes it clearer what the code actually does. Additionally, I can confirm that the PR:

1. Removes the problem I described on Slack (computing multiple consecutive summaries of the same trace);
2. Produces the same MCC trees as v1.2.1
3. Provides extremely impressive speed-ups: computing the MCC tree for a ~4,000-element trace of 200-tip time trees takes about 29 seconds instead of >54,000 seconds for v1.2.1.

davidcerny (approved):
I'm not qualified to comment on every detail of the code changes, either, but renaming "frequencies" to counts where appropriate is very helpful and makes it clearer what the code actually does. Additionally, I can confirm that the PR:

1. Removes the problem I described on Slack (computing multiple consecutive summaries of the same trace);
2. Produces the same MCC trees as v1.2.1
3. Provides extremely impressive speed-ups: computing the MCC tree for a ~4,000-element trace of 200-tip time trees takes about 29 seconds instead of >54,000 seconds for v1.2.1.

bredelings:
Thanks @wrightaprilm and @davidcerny for the checks!  I didn't have time to do all of those myself.

And, WOW.  54,000 seconds is incredibly slow.  For something that doesn't need to be slow.

Issue #327, opened 2023-02-23
[BUG] .probability() member procedure not available for some distributions


jembrown:
**Describe the bug**

The `.probability()` member procedure fails with the error message: ” Error: Couldn’t find member procedure called ‘probability’ ” using the current release (v1.2.1) for some distributions (e.g., `dnBernoulli`). A little exploration suggests that this error carries over across discrete distributions (Bernoulli, Binomial, Poisson, Multinomial), but works fine with continuous distributions (Gamma,Normal). Interestingly, `.lnProbability()` seems to work fine for all distributions.

**To Reproduce**

```
> x ~ dnBernoulli(0.5)
> x.methods()
MixtureLikelihoods = MixtureLikelihoods ()
clamp = clamp (Natural<any> x)
integrateOut = integrateOut (Bool<any> x)
lnMixtureLikelihoods = lnMixtureLikelihoods ()
lnProbability = lnProbability ()
methods = methods ()
probability = probability ()
redraw = redraw ()
setValue = setValue (Natural<any> x)
unclamp = unclamp ()
> x.probability()
   Error:	Couldn't find member procedure called 'probability'
> x.lnProbability()
   -0.6931472
```

**Expected behavior**

`.probability()` should return a probability for all stochastic nodes, regardless of distribution.

bredelings:
OK, so looking at the evaluation of `x.probability()` and `x.lnProbability()`, it seems that the `MethodTable` object for `x` contains methods for both functions:
```
      ["MixtureLikelihoods"] = 0x55555841e2c0,
      ["clamp"] = 0x55555848caa0,
      ["integrateOut"] = 0x55555848cc00,
      ["lnMixtureLikelihoods"] = 0x5555584807a0,
      ["lnProbability"] = 0x555558480c00,
      ["methods"] = 0x555558477eb0,
      ["probability"] = 0x555558480d20,
      ["redraw"] = 0x555558480e10,
      ["setValue"] = 0x55555847d7c0,
      ["unclamp"] = 0x55555847ec30
```

However, the function pointer for `probability` is an object of type `const RevLanguage::MemberProcedure *`, whereas the entry for `lnProbability` points to an object of type `const RevLanguage::DagMemberFunction<RevLanguage::Real> *`.

bredelings:
So, for `x.probability()`, we end up trying to execute the methods of a Natural, which just knows about the procedure name "methods".

jembrown:
Thanks, Ben!

PR #328, opened 2023-02-23
Make the .probability() member function work again.

bredelings:
This makes the .probability() member function work again.

I've tacked on a change that fixes some caching in the github actions tests.  I don't think you need to look at that.


bredelings:
Fixes #327

jembrown (approved):
Thanks!

PR #331, opened 2023-03-01
Fixing issue FBDSP

hoehna:
The FBDSP (and maybe other distributions) didn't work because the function 'prepareProbComputation' wasn't overloaded correctly. The virtual function was constant but the overwriting function wasn't. Thus this new function was never called. This should be fixed. Not sure if that fixes all dnFBDSP issues.

### DIFF
--- a/src/core/distributions/phylogenetics/AbstractFossilizedBirthDeathRangeProcess.cpp
+++ b/src/core/distributions/phylogenetics/AbstractFossilizedBirthDeathRangeProcess.cpp
@@ -289,7 +289,6 @@ double AbstractFossilizedBirthDeathRangeProcess::computeLnProbabilityRanges( boo
 
                     std::vector<double> psi(ages.size(), 0.0);
                     
-                    bool test = true;
                     for (size_t j = 0; j < num_intervals; j++)
                     {
                         double t_0 = ( j < num_intervals-1 ? times[j+1] : RbConstants::Double::inf );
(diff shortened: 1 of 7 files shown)


bredelings:
Hmm... you might want to add the `override` keyword, which is supposed to catch exactly this sort of issue.

https://en.cppreference.com/w/cpp/language/override

bredelings:
The code to modify `q_tilde[i]` seems sketchy.

```
            q_tilde_i[i] = - anagenetic[i] - symmetric[i] * (birth[i] + death[i] + fossil[i]) * dt + (1.0 - symmetric[i]) * q_tilde_i[i];
```

If this code is run multiple times it will yield different values.  It almost looks like the right-hand-side should have `q_tilde[i-1]`, but I'm presuming that it is instead modifying a previous value of `q_tilde[i]` that was computed elsewhere.

hoehna:
Thanks for the tips.

- I added the override statement to these functions, although we probably should do that in a lot of places and haven't done so.
- I now only marked the member variables mutable if they were actually changed by 'prepareProbComputation'
- there is now also a comment about this
- I (hope I) didn't touch the code that computes 'q_tilde_i[i]'. I have no idea what's the correct version.

In the end, this was just a half-way quick fix from me before I forget about this and move to other things. The FossilizedBirthDeathSpeciationProcess and abstract base class require some more clean up, but I hope the authors would do so. I can also look into that a later time, but thought this change rather helps and should be merge better now than later.

bredelings (approved):

PR #332, opened 2023-03-01
fix missing check if optional constraints exist

hoehna:
The following code crashed before:
> taxa <- simTree(8).taxa()
> simStartingTree(taxa, [ clade(taxa[1],taxa[2]) ])

The issue was that in the starting tree simulator we didn't check if optional constraints exist and used them either way. This should be fixed now.

### DIFF
--- a/src/core/datatypes/trees/Clade.cpp
+++ b/src/core/datatypes/trees/Clade.cpp
@@ -355,7 +355,7 @@ size_t Clade::getNumberOfTaxa( void ) const
  * \return       The optional clade constraints
  */
 
-std::vector<Clade> Clade::getOptionalConstraints(void) const
+const std::vector<Clade>& Clade::getOptionalConstraints(void) const
 {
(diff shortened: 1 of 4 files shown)


bredelings:
Nice.

From what I remember, optional constraints are really a different type of constraint -- that is, an optional constraint with options {C1,C2,C3} is really the constraint (C1 or C2 or C3).  So, optional constraints really are not clades, and we should really separate clades and constraints in the future.

Given that, perhaps isOptionalConstraint( ) is more accurate than hasOptionalConstraints( )?  I don't have strong feelings about this.  I guess there are two questions that have the same answer.
- is this an optional constraint?
- does it have any options?

I slightly prefer isOptionalConstraint( ) but I'll approve, and either way should be OK.

bredelings (approved):

bredelings:
If its not too hard, it would be nice to put your two-line crashing code into the tests somewhere.

PR #333, opened 2023-03-01
Fixing the problem with .RevBayes.ini 

hoehna:
Recent versions of RevBayes crashed because of corrupt .RevBayes.ini

This fixes it by writing the directory name as a string and not path.

### DIFF
--- a/src/core/utils/RbSettings.cpp
+++ b/src/core/utils/RbSettings.cpp
@@ -163,7 +163,7 @@ void RbSettings::initializeUserSettings(void)
 
 void RbSettings::listOptions() const
 {
-    std::cout << "moduledir = " << moduleDir.string() << std::endl;
+    std::cout << "moduledir = " << moduleDir << std::endl;
     std::cout << "outputPrecision = " << outputPrecision << std::endl;
     std::cout << "printNodeIndex = " << (printNodeIndex ? "true" : "false") << std::endl;
     std::cout << "tolerance = " << tolerance << std::endl;
@@ -234,7 +234,9 @@ void RbSettings::setOption(const std::string &key, const std::string &v, bool wr
 
     if ( key == "moduledir" )
     {
-        moduleDir = value;
+        // Read from stream to handle quotes.
+        std::istringstream input(value);
+        input >> moduleDir;
     }
     else if ( key == "outputPrecision" )
     {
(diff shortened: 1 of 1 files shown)


bredelings:
Hmm... I've seen this in older versions as well. Why is this supposed to fix it?

hoehna:
if the ini file had modulesdir="modules", specifically with the quotes, then internally the directory was \"modules\", but when writing that back into the file created "&"modules&"". The .string() function avoids this.

bredelings:
Ah, so the quotes are causing the problem.  Hmm.

bredelings (approved):

bredelings:
Maybe add a comment about the affect line like:
```
//    NOTE: Don't write quotes around the module dir!
```

src/core/utils/RbSettings.cpp
```diff
@@ -321,8 +321,8 @@ void RbSettings::writeUserSettings( void )
     path settings_file_name = user_dir / ".RevBayes.ini";
 
     std::ofstream writeStream( settings_file_name.string() );
-    assert( moduleDir == "modules" or is_directory(moduleDir) );
-    writeStream << "moduledir=" << moduleDir << std::endl;
+    assert( moduleDir.string() == "modules" or is_directory(moduleDir) );
+    writeStream << "moduledir=" << moduleDir.string() << std::endl;
```
bredelings, line 325: I think we want the quotes - we just don't want to add more quotes every time.  Quotes help to handle spaces or weird characters in the filename.

Instead of removing the quotes, I think we should just read the module dir from a stream, to match writing it to a stream.

Can you instead modify `RbSettings::setOption( )` to do the following:
```
        // Read from stream to handle quotes.
        std::istringstream input(value);
        input >> moduleDir;
```

I check that this works.

bredelings (changes requested):

src/core/utils/RbSettings.cpp
hoehna: OK, I tried your suggestion and that works too. I don't have a strong preference either way but can see that reading without quotes could be complicated. However, I think we separate the key-value pair by "=" and if we remove the trailing spaces, then a path with spaces but without quotes should still work (not tested).

bredelings (approved):

src/core/utils/RbSettings.cpp
bredelings: Yeah, I was thinking about a filename with trailing spaces.... probably pretty rare.

PR #335, opened 2023-03-02
Reintegrate tests

bredelings:
I talked to @bjoelle about reintegrating the tests into the main repo, and it seems workable.  So this PR does that.

This PR doesn't change any of the contents of the tests directory yet.

One question is whether we should leave the website in `tests/revbayes.github.io` as a submodule, or simply clone it when we run the tests.  The main different would be that if we have a submodule, then the revbayes repo must specify which SHA it is running the tests against, whereas if we just clone the website then the tests would always run against the current website.  So, if the website is changed in a way that doesn't work, then the tests will fail.

Currently this PR leaves the website as a submodule.


hoehna (approved):
As we discussed over slack, this looks like a good step forward.

PR #340, opened 2023-03-24
Work on Multispecies coalescent process

hoehna:
This was some older work on:
- multispecies coalescent, specifically, the conjugate prior models
- the empirical sampling distribution


src/revlanguage/datatypes/phylogenetics/trees/RlTreeVector.cpp
```diff
@@ -0,0 +1,114 @@
+#include "RlMultiValueEvent.h"
```
bredelings, line 1: So RlTreeVector.cpp only includes code for MultiValueEvent?

It looks like this file was copied from RlMultiValueEvent.cpp so that it could be modifed, but was never modified.

hoehna: Good point. Don't remember what the purpose was but looks like we can delete it.

hoehna: That one has been removed.

bredelings (approved):
Based on a quick read of the changes, this looks good.

PR #277, opened 2022-09-30
Fixes2 from dev jsigao

bredelings:
Some speed fixes from the dev_jsigao branch.

PR #342, opened 2023-04-18
Cache p matrix

bjoelle:


bredelings:
Wait... It looks like this maybe duplicates and/or reverts #277?

bredelings:
Actually, I think maybe this branch didn't revert stuff and just added a few fixes that weren't included in #277.

PR #343, opened 2023-04-18
Cleanup RbFileManager

bjoelle:


bjoelle:
Ok I messed something up when doing the rebase but I'm not sure what - @bredelings could you take a look ? or close this and remove the branch if these changes are no longer needed ?

bredelings:
This represents an intermediate stage of removing RbFileManager, and it looks like it actually reverts some of the changes.  I think we can delete it.

bjoelle:
works for me, thanks !

Issue #336, opened 2023-03-08
[BUG] Columns without any data not handled correctly in any +I model


hoehna:
**Describe the bug**
If we have a column in the sequence alignment were all characters for this site are missing or gaps (treated as the same) and we assume any substitution model with the +I model option, then RevBayes has a segmentation issue. The column is internally recognized as being invariant, and in the +I model we multiply the probability of the site being invariant with the base frequency of the observed invariant site. Since for a gap-only column we don't know the base frequency index, the index is set to -1 which retrieves arbitrary memory for the frequency. This leads to unreproducible behaviour and invalid MCMC as the likelihood can change without the parameters changing.

**To Reproduce**
See the example scripts. Since the seed is set, the MCMC should be identical if you rerun it. It doesn't. You can also see the issue when you enable the DEBUG_MCMC preprocessor macro.

[invar_sites_issue.zip](https://github.com/revbayes/revbayes/files/10919724/invar_sites_issue.zip)

**Expected behavior**
Not sure how to handle gap-only columns. Probably should be removed from the alignment?

bjoelle:
@hoehna I would expect the behaviour of the +I models to match with the other models - do they also remove columns only composed of gaps ?

bjoelle:
yes, problem appears to be fixed now (at least the test script gives consistent results on repeated runs) so I will close this

PR #345, opened 2023-04-19
adding a new move to slide the root age

hoehna:
For coalescent models, we don't have a root age variable. So we need some "decent" moves to change the root age because our existing moves only change the internal nodes but not the root. This PR adds a simple root time slide move.


src/core/moves/proposal/tree/RootTimeSlideProposal.cpp
```diff
@@ -0,0 +1,213 @@
...
+    static std::string name = "RootTimeSlide";
+    
+    return name;
+}
+
+
+double RootTimeSlideProposal::getProposalTuningParameter( void ) const
+{
+    // this proposal has no tuning parameter
+    return RbConstants::Double::nan;
```
bjoelle, line 72: isn't delta the tuning parameter ? that's what it says in the parameter description

src/core/moves/proposal/tree/RootTimeSlideProposal.h
```diff
@@ -0,0 +1,63 @@
...
+#include "StochasticNode.h"
+#include "Tree.h"
+
+namespace RevBayesCore {
+    
+    /**
+     * The node-age slide proposal operator using a Uniform distribution.
+     *
+     * This node-age proposal is a Uniform-sliding proposal on rooted subtrees without changing the topology.
+     * That is, we pick a random node which is not the root.
```
bjoelle, line 16: the doc still refers to non-root nodes

src/core/moves/proposal/tree/RootTimeSlideProposal.cpp
hoehna: Good point. Probably lost this while copying ....

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
```diff
@@ -680,6 +680,7 @@ void RevBayesCore::AbstractPhyloCTMCSiteHomogeneous<charType>::compress( void )
         std::map<std::string,size_t> patterns;
         for (size_t site = 0; site < num_sites; ++site)
         {
+            bool all_missing = true;
```
hoehna, line 683: here is also some new code that removes site columns that only contain missing entries. Per se, this should not be a problem. However, if the +I model is used, then we get illegal memory access because we multiply the probability with the base frequency of the observed invariant site. For an all gap column, we don't know this index, that we access the base frequency vector at a random position. See https://github.com/revbayes/revbayes/issues/336

bjoelle (approved):

PR #350, opened 2023-04-26
add dirty flag for tree variable

mlandis:
Update behavior for dnGLHBDSP to appropriately refresh tree variable upon restore.


sswiston (approved):
Reviewed changes with Dr. Landis and observed that merging development into dev-tensorphylo passed tests.

PR #351, opened 2023-05-03
Constructors and destructors don't take template arguments.

bredelings:
This also makes revbayes compile under C++20.


mikeryanmay (approved):
Approving.

PR #353, opened 2023-05-18
Leave ancestors alone

bredelings:
Stop modifying Newick trees when reading them or writing them.  Currently, when we write out trees, we modify sampled ancestors that are tip nodes on 0-length branches to be "knuckles" -- nodes with 1 child -- by default.  Then, when we read the trees back in, we convert these knuckles back to bifurcating nodes in `makeNodesBifurcating(reindex, as_fossils)`.

What the problem with this?
- if the tree has sampled ancestors, we lose the index for the sampled ancestors tips.  
- if the tree does not have sampled ancestors, we leave the topology alone, but its hard to avoid reindexing all the nodes.

So, the current approach messes up the node indices, which breaks ancestral state reconstruction, restoring from checkpoints, etc.

*Proposed solution:* do not modify trees EITHER when we write them out OR when we read them back in (i.e. "leave ancestors alone").  This means that the trees are written out with sampled ancestors as tips on a 0-length branch -- not as knuckles -- because we write out trees directly as used in the MCMC.  In addition to previous code that flagged knuckles as sampled ancestors, we add code to flag tips on 0-length branches as sampled ancestors when converting to TimeTree.

THIS FIXES MULTIPLE NODE-INDEXING BUGS.

So how do things work in the new paradigm? 
- we still can't do MCMC on trees with knuckles -- phyloCTMC throws an exception if you try.
- but it is now possible to read trees with knuckles -- they aren't converted to bifurcating nodes automatically.
- you can still log trees with knuckles -- but you do this by calling a function to modify the trees: `fnCollapseSA(psi)`.
- you can also convert any trees that you read in to bifurcating trees by calling `psi.makeBifurcating()` -- this is not new.
- the user option `collapseSampledAncestors` has no effect -- sampled ancestors are never collapse by writing newick.

Other changes:
- we no longer convert ultrametric trees to time trees automatically when we read them.  They must be explicitly converted to time trees.
   * the justification is that (i) this is a bad criterion -- it was already failing for any trees with serially sampled tips (ii) we can convert them to time trees later -- this needs to work for trees both with and without serially sampled tips.
- we extract node attributes from indices _after_ we call `setRoot(root, true)` so that the index isn't over-written by reindexing.
- it is no longer OK to call makeNodesBifurcating( ) with reindex=false.  This doesn't make sense, because it will either (i) do nothing or (ii) crash.
- we automatically label tips on 0-length branches as sampled ancestors when converting to a Time Tree.


bredelings:
Based on feedback from @jsigao, I think we need to
1. Save/restore sampled_ancestor flags, similarly to how to save/restore indices.
2. Avoid reindexing after loading indices from the Newick.
3. Make trees with bifurcating sampled ancestors answer `true` to `isUltrametric( )`.

src/core/datatypes/trees/Tree.cpp
```diff
@@ -1172,21 +1172,8 @@ void Tree::initFromString(const std::string &s)
 {
     NewickConverter converter;
     Tree* bl_tree = converter.convertFromNewick( s );
-    if ( bl_tree->isUltrametric() == true )
```
hoehna, line 1175: I'm wondering if this breaks some stuff. If I remember correctly, trees are always read in first using branch lengths. Then, only if they are assumed to be ultrametric, then we convert the branch lengths into ages. Maybe this happens somewhere  else now?

src/core/datatypes/trees/Tree.cpp
```diff
@@ -1296,6 +1283,11 @@ bool Tree::isUltrametric( void ) const
 
 void Tree::makeInternalNodesBifurcating(bool reindex, bool as_fossils)
 {
+    // If reindex is false, then makeInternalNodesBifurcating will either
+    // * do nothing (if no sampled ancestors)
+    // * crash ( if there are sampled ancestors)
+    assert(reindex);
```
hoehna, line 1289: shouldn't we then throw an error?

src/core/io/NclReader.cpp
```diff
@@ -1919,9 +1919,6 @@ Tree* NclReader::translateNclSimpleTreeToBranchLengthTree(NxsSimpleTree& nTree,
         tau->getNode(nodes[i]->getIndex()).setBranchLength( brlens[i] );
     }
     
-    bool fossils_only = true;
```
hoehna, line 1922: why is this removed? I would rather make all internal nodes bifurcating and work with 0-length branches. Our code, such as PhyloCTMC, always assumes bifurcating nodes.

src/core/io/NewickConverter.cpp
```diff
@@ -52,45 +52,37 @@ Tree* NewickConverter::convertFromNewick(std::string const &n, bool reindex)
...
-    t->makeInternalNodesBifurcating( reindex, true );
-
     // trees with 2-degree root nodes should not be rerooted
     t->setRooted( root->getNumberOfChildren() == 2 );
     
     // make this tree first a branch length tree
     // hence, tell the root to use branch lengths and not ages (with recursive call)
     root->setUseAges(false, true);
     
-    // if this tree is ultrametric, then we should use ages and transform so!
```
hoehna, line 83: be careful if you remove this. The above code uses "root->setUseAges(false, true);" where the first argument is false, but within the if clause it is set to true. Simply removing this seems dangerous to me.

hoehna (changes requested):
My main concern is that I would prefer internally to use 0-length branches, as PhyloCTMC and PhyloBM assume bifurcating trees. Either we recode all this functions to allow for any number of nodes, or we stick internally to allow only bifurcating nodes (i.e., 0-length branches for SA).

src/core/datatypes/trees/TopologyNode.cpp
```diff
@@ -262,9 +336,14 @@ void TopologyNode::addNodeParameter(const std::string &n, const std::string &p)
         std::cerr << "Illegal node parameter with name '" << n << "' and value "<< p <<".\n";
```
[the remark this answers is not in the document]
bredelings: It probably would -- I didn't write this code though.  I usually try not to do too much modification of code that I didn't write.  It takes too much work to fix up all code that I come near.

I think either RBOUT or an exception would be fine here though -- or deleting the error message.  Do you have a preference?

src/core/datatypes/trees/TopologyNode.cpp
```diff
@@ -216,6 +219,77 @@ void TopologyNode::addBranchParameters(std::string const &n, const std::vector<s
...
+        {
+            return node_comment.substr(name.size()+1);
+        }
+    }
+
+    // Not found
+    return {};
+}
+
+bool TopologyNode::setNodeParameter(const std::string& name, const std::string& value)
```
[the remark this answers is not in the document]
bredelings: It always sets the value, but in the case where there is already a `name=value` comment, it replaces the existing comment instead of appending a second value for the name.

I had a comment in the header file, I'll add one here as well.

src/core/io/NclReader.cpp
bredelings: Not modifying trees while we are reading them was one of the main goals of this PR.

Modifying the trees while we are reading them has caused several bugs so far.  First, if reindex == false, then either
* the tree contains no "knuckles" and we didn't do anything.
* the tree contains "knuckles" and we crashed because they don't have a node index.

So, I changed reindex to be always `true`, and then this changed the `index` attribute in some cases, which caused problems with ancestral state reconstruction, and with loading trees from a checking.

Overall, I think we should not be modifying trees while reading them.

Its fine for phyloctmc to throw an exception if it sees a tree that it doesn't like.

src/core/io/NewickConverter.cpp
```diff
@@ -52,45 +52,37 @@ Tree* NewickConverter::convertFromNewick(std::string const &n, bool reindex)
...
 
     // set the branch lengths
     for (size_t i = 0; i < nodes.size(); ++i)
     {
         t->getNode( nodes[i]->getIndex() ).setBranchLength( brlens[i] );
     }
 
-    // make all internal nodes bifurcating
-    // this is important for fossil trees which have sampled ancestors
-    t->makeInternalNodesBifurcating( reindex, true );
```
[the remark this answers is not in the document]
bredelings: This was indeed important when we converted sampled ancestors to "knuckles" when writing Newick, and then converted them back to tips with 0-length branches while reading Newick.

But now, we just write the trees without knuckles -- no modification on writing.  So, we don't need a modification when we read them.

bredelings:
Thanks for the feedback.  Note that my responses are in the order of the comments they are responding too, which may not be the logical order for reading them.  So maybe take a look at all the comments before responding.

src/core/datatypes/trees/Tree.cpp
jsigao: I think when `initFromString` is called (presumably either when reinitializing from checkpoint or doing posterior predictive simulation), different from the `readTrees` or `readTreeTrace` functions, no `treetype` like argument would be provided (nor accessible from `initFromString`), so it wouldn't know whether the tree should be a timetree or not. However, as Sebastian pointed out, calling the `TreeUtilities::convertTree` function would be necessary if the tree needs to be a timetree, given that the return of `convertFromNewick` is a tree whose node ages have not been set.

In that case, perhaps the desired tree type could be determined by checking whether `*this` is a branch-length tree or a timetree (i.e., by checking the age of any given node in the tree, say the root node, is `nan` or not)? This should work if it is safe to assume that `initFromString` is called after the model DAG has been initialized (and thus `*this` should have been initialized to the desired tree type by the corresponding tree simulation function [if it's a stochastic node] or by assignment [if it's a constant node]).

bredelings:
@hoehna I rewrote the PR description to catch you up the discussion that we were having on slack.  The discussion was primarily focused on fixing indexing problems caused the the `reindex` argument to `makeInternalNodesBifurcating( )`.

src/core/io/NclReader.cpp
hoehna: OK, still seems weird to me that we may be inconsistent internally in sometimes allowing for degree-1 nodes and sometimes allowing 0-length branches. This really should be made consistent; for writing, reading and internal use.

src/core/io/NewickConverter.cpp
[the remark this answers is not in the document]
hoehna: OK, sounds fine.

src/core/datatypes/trees/Tree.cpp
hoehna: Yes, that would be better.

src/core/datatypes/trees/Tree.cpp
jsigao: Thanks for the new commits @bredelings!

Related to another comment by Sebastian, as we now removed `root->setUseAges(true, true);` in the `convertFromNewick` function, perhaps we might want to add that to somewhere in the TreeUtilities::convertTree function? Currently the `bl_tree` will be created first by `convertFromNewick` and then converted to `time_tree` by `TreeUtilities::convertTree` if necessary, so the `use_ages` attribute of each TopologyNode will still be false in the finally converted timetree. Looks like this is probably innocuous given that we are not checking `use_ages` otherwhere, but perhaps it might be safer to make it consistent.

In addition, I just noticed that there is [one place](https://github.com/revbayes/revbayes/blob/development/src/core/analysis/mcmc/output/TreeSummary.cpp#L1318) we are calling `initFromString` where the tree has not been initialized so the tree type is not known. It looks like there we should know what the desired tree type is as the goal there is to check whether the tree converted from the string is contained in a certain credible interval of trees which is already read in as a `TraceTree` object. Perhaps we could check the type of first tree in the `TraceTree` to determine whether we would like the string to be converted to a timetree or not? Would be great to hear your thought on this @hoehna .

bredelings:
OK, initialization from checkpoints now works with both test-cases from #356.  Three bugs fixed:
- correctly convert trees to time trees
- touch all dag nodes after loading from checkpoint
- don't mark parents of sampled-ancestors-tips as sampled ancestors.

PR #357, opened 2023-05-25
updating FBD test output

mikeryanmay:
This is just a minor update to mvTipTimeSlideUniform. Previously, the new age for a sample was drawn from a uniform distribution between 0 and the age of the parent, leading to low acceptance rates, especially for older samples. Now, the new age is further bounded by the minimum and maximum age of the sample.

I also updated the outputs for integration tests that use this proposal.


hoehna:
This may be a duplicate from the coalescent branch. Please check.

mikeryanmay:
Yes, it appears to be _almost_ a duplicate. I see that you implemented two new moves FossilTipTimeUniformProposal and FossilTipTimeSlideUniformProposal on dev-coalescent, which implement essentially the same logic (using taxon age data to specify min/max bounds). 

My changes are to the original move TipTimeSlideUniformProposal, which I think will now behave more-or-less the same as your FossilTipTimeUniformProposal. It seems to be like we don't need both? Perhaps the ideal solution would be to make mvTipTimeSlideUniform (and its alias mvFossilTimeSlideUniform) (which are using TipTimeSlideUniformProposal) aliases for mvFossilTipTimeSlideUniform (which uses your TipTimeSlideUniformProposal), and get rid of FossilTipTimeUniformProposal?

bredelings (approved):

bredelings:
The fix here seems equivalent (or better) than the one on the branch.  

I'd prefer to merge this now.

People working on the coalescent branch can decide how to respond -- they could either drop the fossil-specific move/proposal, or add an alias, whichever they would like.

PR #358, opened 2023-06-06
Use more up-to-date-version of openlibm package on mingw.

bredelings:
Update the version of openlibm to make the windows test pass.  The old version is no longer available.


hoehna (approved):
Looks good.

PR #367, opened 2023-07-24
Bdstp segfault fix

afmagee:
Fixes a segfault that can otherwise be caused by using the `initialTree` option of the BDSTP class.


hoehna (commented):
Nice, with a test case!

hoehna (approved):

PR #368, opened 2023-07-26
Fixing compile errors because of missing cmath include

hoehna:
This is a very minor fix which adds <cmath> to make RevBayes compile properly again.

### DIFF
--- a/src/core/datatypes/trees/Clade.cpp
+++ b/src/core/datatypes/trees/Clade.cpp
@@ -1,6 +1,7 @@
 #include "Clade.h"
 
 #include <algorithm>
+#include <cmath>
 #include <iostream>
 #include <string>
 #include <utility>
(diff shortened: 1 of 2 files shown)


mlandis:
Accepting this fix. I was looking into issues concerning the use of `math.h` (C header) vs `cmath` (C++ header) and noticed we don't consistently use one library across RevBayes classes. In general, we should prefer C++ headers to avoid various types of compile errors (some being very sensitive to C++XX dialect choice) so I'll add a note to the Issue Tracker suggesting we port functions/includes over for other classes.

mlandis (approved):
Accepting this fix. I was looking into issues concerning the use of `math.h` (C header) vs `cmath` (C++ header) and noticed we don't consistently use one library across RevBayes classes. In general, we should prefer C++ headers to avoid various types of compile errors (some being very sensitive to C++XX dialect choice) so I'll add a note to the Issue Tracker suggesting we port functions/includes over for other classes.

PR #369, opened 2023-07-26
Fixing representation of sampled ancestors in tensorphylo functions

brpetrucci:
The recent pull request #353 made it so sampled ancestors needed to be represented as 0-length branches, but internal tensorphylo functions require sampled ancestors to be degree-2 nodes since they consider bifurcations as speciation events. These changes (by @mikeryanmay and I) change the representation of the SAs to degree-2 nodes when communicating with TP functions, and back to bifurcating when needed (i.e. in the `setValue` method). All changes were to `GeneralizedLineageHeterogeneousBirthDeathSamplingProcess` functions.


hoehna:
Do you think you could add a small test that makes sure this keeps working in the future?

brpetrucci:
> Do you think you could add a small test that makes sure this keeps working in the future?

That'd be a great idea, but I'm not sure if it's currently possible to write tests that use functions that depend on tensorphylo, as is the case with this one. @mikeryanmay @mlandis is that the case?

brpetrucci:
@hoehna @bredelings I confirmed with Mike May that we currently don't have the possibility of implementing tests involving tensorphylo in the github actions. 

While we can work on that for the future, it'll be pretty involved and, given that the changes in this PR are very small (just regarding the representation of SAs between nodes and branches when communicating with TP functions), maybe it isn't necessary for this PR in particular.

bredelings (approved):

PR #377, opened 2023-09-08
Allow mnFile and mnModel to write JSON output if given `format="json"`

bredelings:
This PR allows JSON output for mnFile and mnModel if they are given the argument `format="json"`.  If seems to work fine for vectors of vectors.

This PR makes use of nlohmann::json, which provides a C++ type `json` that is really easy to work with.

In order to make it possible to serialize objects as JSON, we extend the Printable class with a `toJSON() const` method.

The DagNode class already has a method `getValueAsString()`.  This PR adds a method `getValueAsJSON()` to complement that.

We introduce a datatype to represent output formats:
``` C++
struct SeparatorFormat
{
    std::string separator;
    SeparatorFormat(const std::string& s):separator(s) {}
};

struct JSONFormat
{
};

typedef std::variant<SeparatorFormat,JSONFormat> SampleFormat;
``` 
This allows querying to format to determine if it has a separator, and only getting the separator if there is one.

Most monitor class constructors continue to take a `std::string separator`, but `VariableMonitor` and `ModelMonitor` now take a `SampleFormat`.  Interestingly, if you pass a std::string separator to the `SampleFormat` argument, it automatically constructs a `SeparatorFormat` from that string.

There are a lot of monitor classes that derive from `VariableMonitor` but ignore the vector of `nodes` that VariableMonitor is supposedly monitoring.  They override the virtual functions `printFileHeader() and `monitorVariables(unsigned long gen)`  to print their own stuff in CSV/TSV format.  Such classes could be extended to support JSON output in the future, but this PR does not add that ability.


src/core/monitors/VariableMonitor.h
```diff
@@ -4,19 +4,32 @@
 #include <cstddef>
 #include <vector>
 #include <iosfwd>
+#include "variant.h"
 
 #include "AbstractFileMonitor.h"
 #include "MonteCarloAnalysisOptions.h"
 
+struct SeparatorFormat
```
mlandis, line 12: For future consideration, is it conceivable that file writers/readers outside of VariableMonitor could make use SeparatorFormat? For example, the `write` method assumes csv/tsv format and accepts a separator flag, though it could probably be adapted to support json. If so, it may be worthwhile separating SeparatorFormat and JSONFormat out into their own file(s).

src/core/utils/variant.h
```diff
@@ -0,0 +1,24 @@
+#ifndef UTIL_VARIANT_H
+#define UTIL_VARIANT_H
+
+#include <variant>
+
+template <typename T, typename U>
```
mlandis, line 6: Nice solution for the specific use case of handling alternative formatting types (JSON vs. CSV/TSV). Seems like templated variant classes could make it difficult to extend for more types (e.g. if XML support was also added).

tests/test_Partition/scripts/mcmc_Partition_codon.Rev
mlandis, line 1: Ran and tested locally. It is outside the scope of this pull request, but it'd be nice if json format is capable of monitoring higher dimensional variables (e.g. Q matrices across partitions). I didn't see Q in the output json, though I see RateGenerator::toJSON() was implemented, so this will probably be handled soon.

src/core/utils/variant.h
bredelings: What are you thinking?

I was thinking that we could extend this by create (for example) a struct XMLFormat, and adding it to the variant.  Then we'd have:
```
typedef std::variant<SeparatorFormat,JSONFormat,XMLFormat> SampleFormat;
```
Code that does `to<SeparateFormat>(format)` should be unaffected.

src/core/monitors/VariableMonitor.h
bredelings: Yeah, that might be a good idea.  In preparation for that, I move the file format types out into their own file `core/util/FileFormat.h`

tests/test_Partition/scripts/mcmc_Partition_codon.Rev
bredelings: So, I looked at `tests/test_Partition/scripts/mcmc_Partition_gene.Rev`.  It is true that `mnModel( )` doesn't print the variable `Q`, where Q[i] is a rate matrix.

However, if you give `Q` to `mnFile(  )`, then it actually does print the Q matrix.

So I think this might be more an issue of which variables `mnModel( )` chooses to ignore?

Issue #354, opened 2023-05-21
[BUG] Specifying an initial tree in dnBDSTP overwrites fossil ages


davidcerny:
**Describe the bug**
To accommodate tip age uncertainty in the FBD-specimen process, we can place a uniform prior on the age range associated with each specimen, and use MCMC to sample exact ages in proportion to their posterior density (see "Sampling Fossil Occurrence Ages" in the [current FBD-specimen tutorial](https://revbayes.github.io/tutorials/fbd/fbd_specimen.html)). To do this, we have to access the age range stored in the `TimeInterval` member of class `Taxon`:

```
fossils = fbd_tree.getFossils()
for(i in 1:fossils.size())
{
    t[i] := tmrca(fbd_tree, clade(fossils[i]))
    a_i = fossils[i].getMinAge()
    b_i = fossils[i].getMaxAge()

    F[i] ~ dnUniform(t[i] - b_i, t[i] - a_i)
    F[i].clamp( 0 )
}
```

The `TimeInterval` member is initialized using `TaxonReader.cpp`, but specifying the `initialTree` argument in `dnBDSTP` apparently overwrites it, setting both the minimum and maximum age of each fossil in the tree to 0. As a result, initializing an FBD-specimen analysis using a user-supplied starting tree is only viable if the user is willing to treat specimen ages as fixed values.

**To Reproduce**
See [example script 1](https://github.com/revbayes/revbayes/files/11523634/test.txt), [data 1](https://github.com/revbayes/revbayes/files/11523629/data.zip), and [example script 2](https://github.com/revbayes/revbayes/files/11523633/test_2.txt), [data 2](https://github.com/revbayes/revbayes/files/11523628/data_2.zip). One absolute path has to be changed. A draw from `dnBDSTP` that does not specify an `initialTree` results in the expected maximum and minimum ages being printed to the screen; commenting it out and drawing after an `initialTree` is specified results in these being equal to 0.

**Expected behavior**
Specifying an `initialTree` should not alter the `TimeInterval` member of the objects stored in the `taxa` vector.

**Screenshots**
N/A

**Computer info**
macOS v12.6.5, RevBayes 1.2.1, latest `development` version (rapture-2179-g991117).

afmagee:
This is definitely not something one would expect _a priori_. The following is not yet a solution to the problem, just a sketch of what's going on so as to help arrive at a solution.

[...]

PR #386, opened 2023-10-07
Fix a starting tree issue w/ dnBDSTP()

davidcerny:
This branch fixes issue #354 by introducing a new function called `startingTreeInitializer()` following @afmagee's suggestion. The function takes (1) a user-supplied starting tree (`treeToChange`) and (2) a vector of `Taxon` objects (`taxaToCopy`) as its input, checks that the number and names of the taxa in `taxaToCopy` match the number and names of the tips in `treeToChange`, and returns a tree in which every tip has been replaced by a `taxaToCopy` element of the same name. The function is then called by dnBDSTP(). This ensures that the tips of the starting tree "know" their age ranges. 

I'm open to feedback if this doesn't seem like the best solution. For example, checking the initial tree is something other distributions could benefit from as well –e.g., dnConstrainedTopology() – whereas replacing taxa to account for age uncertainty is only really applicable to FBD models: dnBDSTP() for now, and in the future possibly dnFBDSP() as well. So going forward, the function could be split in two. And maybe there are other checks one could come up with to ensure the starting tree is appropriate, too. But for now, this fixes an issue that has affected multiple projects of mine.

[Here](https://github.com/revbayes/revbayes/files/12838213/BDSTP_init_tree_check.zip) is a simple test based on an empirical dataset: a starting tree is passed to dnBDSTP() and no topology moves are specified, so the analysis just samples the ages of tips and internal nodes on a fixed topology. An R script is provided to calculate tip heights, showing that these differ between the 10 sampled trees. The analysis is therefore successfully sampling tip ages from their ranges – something that was not possible before.


hoehna (approved):
Looks good. Maybe next time also add a new test script to make sure this stays working in the future.

PR #392, opened 2023-10-17
Add and change AbstractHomologousDiscreteCharacterData methods

davidcerny:
This branch makes the following changes to the class `AbstractHomologousDiscreteCharacterData`, originally suggested by Alessio Capobianco:

(1) Add a `.getInvariantSiteIndices()` method, which returns the indices of all characters that are invariant, either with or without accounting for ambiguities. The code is based on the pre-existing `.getNumInvariantSites()` method, which has been refactored to remove redundancies.

(2) Add a `.removeExcludedCharacters()` method, which entirely removes excluded characters from the matrix. The idea behind this method is to do for characters what we already do for taxa, where we have both an `.excludeTaxa()` method (which marks one or more taxa as excluded but keeps them in the matrix) and a `.removeTaxa()` method (which removes them entirely).

(3) Add a test to check these methods work as intended.

(4) Use the words "exclude" and "remove" in a consistent manner by renaming the existing method `.removeMissingSites()` to `.excludeMissingSites()` (since it marks sites consisting entirely of missing data as excluded but keeps them in the matrix) and `.removeRandomSites()` to `.replaceRandomSitesByMissingData()`. This latter method neither excludes nor removes any sites; instead, it irreversibly changes their coding.

### DIFF
--- a/src/core/datatypes/phylogenetics/characterdata/AbstractHomologousDiscreteCharacterData.cpp
+++ b/src/core/datatypes/phylogenetics/characterdata/AbstractHomologousDiscreteCharacterData.cpp
@@ -77,7 +77,7 @@ void AbstractHomologousDiscreteCharacterData::fillMissingSitesMask(std::vector<s
 }
 
 
-void AbstractHomologousDiscreteCharacterData::removeMissingSites( void )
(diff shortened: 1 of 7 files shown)


hoehna (approved):
Very nice with a good test!

Issue #374, opened 2023-08-15
[BUG] Ancestral state reconstruction missing data assigned state 0


king-ben:
I noticed in the RevGadgets plots missing data at the tips as state 0 in all cases. This problem is already in the annotated trees produced by ancestralStateTree, so it seems to be a Rev problem rather than RevGadgets. In addition, all nodes are annotated with a posterior of 1.0, which is not correct. I noticed both problems seem to be there in the tutorial as well, but less noticeable than with my example.

**To Reproduce**
https://github.com/king-ben/revbayes_asr_bug/tree/master

**Expected behavior**
taxa with missing data should have state estimated, in general should be high posterior probability for the same state as phylogenetically related taxa.

**Screenshots**
<img width="293" alt="Screenshot 2023-08-15 at 15 45 24" src="https://github.com/revbayes/revbayes/assets/57047255/d1277cb7-fe20-4c18-b4fe-3b5ebdb97ecb">
<img width="336" alt="Screenshot 2023-08-15 at 15 45 10" src="https://github.com/revbayes/revbayes/assets/57047255/78c5881f-404e-463a-b88b-12e1d4dc37d5">

**Computer info**
Revbayes 1.2.1. Mac OS Ventura 13.5

bredelings:
This should be fixed in `development`.

PR #396, opened 2023-10-21
Fix problem reconstructing ancestral state with missing data.

bredelings:
This problem was caused by a change I made in the RbBitSet implementation a while back.

RbBitSet is now an alias to `boost::dynamic_bitset<>`.  So, RbBitSet(n, true) doesn't construct an all-1s bitset anymore.  Instead, the second argument is a `long int` that represents a bitmask.  So, passing in `true` initializes the bitset to `0000..001` 😬.

While fixing this problem, I noticed a few other cases where RbBitSet(n, true) occurred and fixed them too.


bredelings:
Fixes #374

hoehna (approved):

PR #399, opened 2023-10-24
Clean version of coalescent classes

hoehna:
This is a merge from the branch that Ronja and I had been working on for a while. I don't remember all the new features, but the main new and important ones include:

- restructuring of the coalescent models, including heterochonous samples, see https://revbayes.github.io/tutorials/coalescent/
- new models for autocorrelated random variables (e.g., many variants of CPPs)
- iid proposals
- integrating over tree samples


src/core/distributions/phylogenetics/tree/coalescent/ConstantPopulationCoalescent.cpp
```diff
@@ -1,23 +1,23 @@
-#include <cstddef>
```
bredelings, line 1: This looks like a mistake...

bredelings:
One thing that is missing is updated help information in `help/md`.  There are several new functions and moves which need documentation, and the updated coalescent distributions need  documentation.

src/core/distributions/phylogenetics/tree/coalescent/AbstractCoalescent.h
```diff
@@ -1,7 +1,7 @@
 #ifndef AbstractCoalescent_H
 #define AbstractCoalescent_H
 
-#include <cstddef>
```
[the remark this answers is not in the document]
hoehna: Yes, that must have happened during merging ...

src/core/distributions/phylogenetics/tree/coalescent/ConstantPopulationCoalescent.cpp
hoehna: Yes, that must have happened during merging ...

src/core/distributions/math/GeometricDistribution.cpp
```diff
@@ -45,15 +45,15 @@ GeometricDistribution* GeometricDistribution::clone( void ) const
 double GeometricDistribution::computeLnProbability( void ) 
 {
     
-    double v = *value;
+    long v = *value;
```
bredelings, line 49: What is the rationale for switching to `long` here?  Should we use `long long` instead?  On windows, `long` is 32 bits and `long long` is 64 bits :grimacing:

src/core/distributions/math/GeometricDistribution.cpp
```diff
@@ -45,15 +45,15 @@ GeometricDistribution* GeometricDistribution::clone( void ) const
...
     
     // check that the value is inside the boundaries
     if ( v < 0.0 )
     {
         return RbConstants::Double::neginf;
     }
     
-    return RbStatistics::Geometric::lnPdf(*value, p->getValue());
+    return RbStatistics::Geometric::lnPdf(int(v), p->getValue());
 }
```
bredelings, line 57: Should we use `long` or `long long` here also?

src/core/distributions/phylogenetics/tree/coalescent/ConstantPopulationCoalescent.cpp
```diff
@@ -130,42 +196,84 @@ std::vector<double> ConstantPopulationCoalescent::simulateCoalescentAges( size_t
...
+    }
+    
+    // now simulate the ages
+    
+    // allocate the vector for the ages
+    std::vector<double> coalescent_ages = std::vector<double>(n,0.0);
+    
+    // j is the number of active lineages at the current time
+    size_t current_num_lineages = num_taxa_at_present;
+    double theta = Ne->getValue();
```
bredelings, line 228: Again, theta = 2*Ne*mu, so why are we writing `theta = Ne`?

tests/test_Coal_skyline/scripts/mcmc_Coal_skyline_fixed.Rev
```diff
@@ -41,9 +41,21 @@ for (i in 1:NUM_INTERVALS) {
 
 }
 
+num_events_pi <- rep(1, NUM_INTERVALS)
```
bredelings, line 44: Can you add a short comment about what this new code is doing?

src/revlanguage/workspace/RbRegister_Move.cpp
```diff
@@ -25,7 +25,7 @@
 
 #include <sstream>
 #include <vector>
-#include <cstdio>
```
bredelings, line 28: This should change back to <cstdio> - probably a merge issue.

src/revlanguage/functions/math_helper/Func_SmoothenTimeline.h
```diff
@@ -0,0 +1,48 @@
+#ifndef Func_SmoothenTimeline_H
+#define Func_SmoothenTimeline_H
```
bredelings, line 2: This should probably be SmoothTimeLine.  The verb for making something smooth is "to smooth".  Apparently some dictionaries list "smoothen" as well, but it sounds a bit odd.

src/revlanguage/moves/Move_IndependentPriorSampler.h
```diff
@@ -0,0 +1,211 @@
...
+const RevLanguage::MemberRules& RevLanguage::Move_IndependentPriorSampler<rlValueType>::getParameterRules(void) const
+{
+    
+    static MemberRules move_member_rules;
+    static bool rules_set = false;
+    
+    if ( !rules_set )
+    {
+        move_member_rules.push_back( new ArgumentRule( "x", rlValueType::getClassTypeSpec(), "The variable on which this move operates.", ArgumentRule::BY_REFERENCE, ArgumentRule::STOCHASTIC ) );
+        move_member_rules.push_back( new ArgumentRule( "MetropolisHastings"   , RlBoolean::getClassTypeSpec(), "If TRUE, use the regular Metropolis-Hastings acceptance ratio. If FALSE, always accept this move and sample every variable uniformly.", ArgumentRule::BY_VALUE, ArgumentRule::ANY, new RlBoolean( true ) ) );
```
bredelings, line 150: Can you explain why someone would want this to be false?

My impression is that this move proposes a new value from the prior.  If that is the case, then always accepting is equivalent to ignoring likelihoods that condition on this node.

src/core/distributions/math/GeometricDistribution.cpp
hoehna: The reason is that the GeometricDistribution derives from TypedDistribution<long>, which has `long* value` defined. So this is simply the corrected version and doesn't hide what the internal value actually is.

src/core/distributions/math/GeometricDistribution.cpp
hoehna: Well, since `value` is of type `long*` as defined by `TypedDistribution<long>`, I would stick with long for consistency.

src/core/distributions/phylogenetics/tree/coalescent/ConstantPopulationCoalescent.cpp
```diff
@@ -69,53 +70,118 @@ double ConstantPopulationCoalescent::computeLnProbabilityTimes( void ) const
 {
     
     // variable declarations and initialization
-    double ln_prob_times = 0;
-    
+    double ln_prob = 0;
+    double theta = Ne->getValue();
```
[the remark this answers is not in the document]
hoehna: Well, in some scenarios theta is defined as 4*Ne*mu, but often for Bayesian coalescent models which define a coalescent prior on a genealogy, theta is redefined. I can change the names of the variables if that avoids confusions.

src/core/distributions/phylogenetics/tree/coalescent/ConstantPopulationCoalescent.cpp
hoehna: Because for some arbitrary reasons, the equations of the coalescent use theta as the variable name and define it that way. It's yet another theta ;)

src/revlanguage/moves/Move_IndependentPriorSampler.h
hoehna: If I remember correctly, that was used to show that the incorrect version can give biased estimates. Don't remember when or why we even tested that (@jsigao). We can remove this option for the released version if you prefer.

tests/test_Coal_skyline/scripts/mcmc_Coal_skyline_fixed.Rev
hoehna: Done.

src/revlanguage/workspace/RbRegister_Move.cpp
hoehna: Done.

src/revlanguage/functions/math_helper/Func_SmoothenTimeline.h
hoehna: Done.

src/core/distributions/phylogenetics/tree/coalescent/ConstantPopulationCoalescent.cpp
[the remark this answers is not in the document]
bredelings: True about the 4 * Ne *mu (for diploids). 

I've seen a variety of ways to rescale the coalescent process... for example if you want the coalescent rate to be 1, then the mutation rate is theta/2, and if you want the mutation rate to be 1, then the coalescent rate is 2/theta.  I guess I'm just confused about how theta would ever be equal to Ne.

If a variable name change would clear this up, then maybe that would be good.

src/revlanguage/moves/Move_IndependentPriorSampler.h
bredelings: Ah, I see.  Let's remove it for the released version though.

src/core/distributions/phylogenetics/tree/coalescent/ConstantPopulationCoalescent.cpp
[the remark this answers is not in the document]
bredelings: OK, I see you changed theta -> ne.  That is much less confusing to me.

src/core/distributions/math/GeometricDistribution.cpp
bredelings: OK, maybe change `int` to `long` then.

bredelings: Or replace `int(v)` with just `v`.

bredelings:
I guess this needs help files for:
- dnCoalescent( )
- dnCoalescentDemography( )
- dnCoalescentSkyline( )
- fnSmoothTimeLine( )
- mvRandomNaturalWalk( )
- mvVectorElementSwap( )
- mvEmpiricalTree( )
- mvFossilTipTimeSlideUniform( )
- mvFossilTipTimeUniform( )
- mvIidPrior( )

bredelings:
Would it also be possible to change the function arguments to dnCoalescent( ), dnCoalescentSkyline( ), and dnCoalescentDemography( ) to stop calling the population size "theta"?

bredelings:
I recall that Mike May already submitted a path to mvTipTimeSlideUniform( ) to be more efficient in the case of fossil tips by proposing a new time that is within the taxon age bounds, if there are any.  I checked TipTimeSlideUniformProposal( ), and it is indeed taking the taxon age bounds into account. 

Given that, perhaps we don't need fossil-specific tip moves?  It seems nicer to have a single move for fossil and non-fossil tips.

bredelings:
One small issue is that it looks like TipTimeSlideUniformProposal( ) is not really doing a "slide" -- its just sampling from a uniform.  It doesn't even take a proposal size parameter.  So perhaps it should be renamed to TipTimeUniformProposal( )...

src/core/distributions/math/GeometricDistribution.cpp
hoehna: OK, I changed that throughout for the geometric distribution.

hoehna:
Thanks @bredelings for the feedback. I now did the following and some answers:

- I changed the geometric distribution to consistently use long. If that should change, then we would need to change the Rev language types.
- I internally renamed the variable for theta, but for users through Rev I'm keeping the argument because this is how it is described in the publication (line 872; https://academic.oup.com/mbe/advance-article/doi/10.1093/molbev/msae073/7648822)
- I removed the option to do the invalid non-MH proposal for the iid sampler
- I removed the duplicate fossil tip move, but kept the correct names as aliases to hopefully not break anything.
- help documentation will follow soon

bredelings:
Thanks for the changes, and congratulations on the publication!  That looks very nice :-)

Hmm.... in that case, can you please:
* add a synonym for "theta", such as "N" or "Ne"
* mention that "theta" is not related to mutation in the documentation?
I guess "theta" is just meaning "parameter" here, instead of being related to the 2Nu/4Nu thing.

I see you deleted the existing mvTipTimeSlideUniform, which I guess makes sense since your version has more options.  I'll at that more later.

rbillenstein (approved):

Issue #401, opened 2023-10-26
[BUG] FBD model with sampled ancestors


hoehna:
**Describe the bug**
Once a tip is set as a sampled ancestor, then node age moves change the tip age. This should neither happen and leads to strange model behavior.

**To Reproduce**
See attached script. Note that the oldest fossil gets a new age and is stuck at 49, the max origin time.

**Expected behavior**
Fossil tips should not change their age, only if there is a specific fossil 
move.

[test_FBD_dating.zip](https://github.com/revbayes/revbayes/files/13172969/test_FBD_dating.zip)

bredelings:
We talked about this on the call today.

Sample ancestors can either have a fixed age, or can have an age within an interval.  If we try and scale the tree and move a sampled ancestor node outside of its interval, then the tree should have probability 0, effectively blocking all moves that rescale the whole tree.

General consensus was that MCMC moves need to be able to identify sampled ancestor nodes.  Tree scaling moves would need to avoid scaling them, whereas moves that slide the internal node age for a single node could just do nothing.

One solution would be to make setAge( ) throw an exception if a move tries to change the age for a sampled ancestor.

bredelings:
OK, so it looks like mvRootTimeSlideUniform is indeed moving fossils if the root is a sampled ancestor.  This does not appear to be a recent change in behavior though.

bredelings:
Fixed by #447

PR #402, opened 2023-10-26
Fix fbd move

hoehna:
This should fix issue https://github.com/revbayes/revbayes/issues/401


src/core/datatypes/trees/TopologyNode.cpp
```diff
@@ -1858,14 +1861,7 @@ void TopologyNode::setAge(double a, bool propagate)
     // we also need to recompute the branch lengths of my children
     for (auto& child: children)
     {
-        if ( child->isTipSampledAncestor() )
```
bredelings, line 1861: This doesn't seem like the right fix.  The internal-node-sampled-ancestor is supposed to have the same time as the tip-sampled-ancestor.  So, the real problem is that that something is trying to change the time of the internal-node-sampled-ancestor.  Why is this happening?

bredelings (changes requested):

src/core/datatypes/trees/TopologyNode.cpp
hoehna: Because we have our standard move, the mvNodeTimeSlideUniform that changes internal node ages. We could change the move, but I think that would not be safe. It would require knowledge for all moves that they are only allowed to change internal nodes if those nodes are not "fake" because of sampled ancestors. That would mean new developers of node time change moves would need to know about this behavior even though the developer might never work with fossils.

bredelings: So, what happens when mvNodeTimeSlideUniform changes the internal node age for a sampled ancestor to be different than the corresponding tip node for the sampled ancestor?

bredelings:
So apparently we have always updated the age of sampled ancestor tip nodes when we update the age of the corresponding internal node.

PR #403, opened 2023-10-30
Handle the index for a node not being an integer (e.g. "index=NA")

### LINKED ISSUE #400: Error in readTreeTrace
**Describe the bug**
readTreeTrace is generating and error message in rb 1.2.2 compiled from source, it does however run perfectly under the rb executable (1.2.2).

**To Reproduce**
> tree_init_fn = "Master_Resolved_Const.trees"
> treetrace = readTreeTrace(file=tree_init_fn,  burnin=0.0, thinning=1, separator="")

**Screenshots**
Captura de pantalla 2023-10-25 a la(s) 9 53 22 a m

**Expected behavior**
Under the rb 1.2.2 executable the command runs smoothly and reads the treetrace.
Captura de pantalla 2023-10-25 a la(s) 10 02 07 a m

**Computer info**
MacBookPro Intel running on Sonoma 14.0.
rb 1.2.2 compiled from source.

**Additional context**
The abnormal behavior appears using the above (and similar) treetrace files. These files have been processed in R and re-written in the same format as produced by monitors. Another (related?) problem appears when using the original treetrace file under the compiled version, which disappears under the rb executable.
I'm guessing there is a problem with indices in the trees.

bredelings:
Fixes #400

### DIFF
--- a/src/core/datatypes/trees/Tree.cpp
+++ b/src/core/datatypes/trees/Tree.cpp
@@ -1405,7 +1405,15 @@ bool Tree::tryReadIndicesFromParameters(bool remove)
             // 4. Set the index if there is one.
             if (value)
             {
-                int index = stoi(*value) - 1;
+		int index = -1;
+		try
+		{
+		    index = stoi(*value) - 1;
+		}
+		catch (...)
+		{
+		    throw RbException()<<"tryReadIndicesFromParameters: index '"<<*value<<"' is not an integer";
+		}
 
                 if (index < 0)
                     throw RbException()<<"tryReadIndicesFromParameters: Negative node index "<<index;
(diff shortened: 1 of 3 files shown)


davidcerny:
This all looks good to me. I assume lines 1433–1436 in `Tree.cpp` are for exceptions that we can't handle using `RbException` but that can still be handled using the standard library?

Also, that's interesting that `dnMultiSpeciesCoalescentInverseGamma` is parameterized using shape and scale rather than shape and rate (which is how `dnInverseGamma` does it). To prevent confusion, it might be worth calling these arguments `kappa` and `theta` rather than `alpha` and `beta`, maybe?

bredelings:
> This all looks good to me. I assume lines 1433–1436 in Tree.cpp are for exceptions that we can't handle using RbException but that can still be handled using the standard library?

Right.

> Also, that's interesting that dnMultiSpeciesCoalescentInverseGamma is parameterized using shape and scale rather than shape and rate (which is how dnInverseGamma does it). To prevent confusion, it might be worth calling these arguments kappa and theta rather than alpha and beta, maybe?

So, that change just updates RbHelpDatabase to match the `*.md` files.  The change in the parameters has already been made (by someone else) -- this just updates the documentation.

I do agree that there is some confusion about the parameter names.  Usually I use alpha and beta for the shape and scale, but I see that wikipedia suggests k and theta.  I have more opinions, but I suppose the main thing is that we should be consistent. Buuuuut... other people made these changes and I don't have time to make all the parameterizations consistent right now.

davidcerny (approved):

bredelings:
Thanks!

PR #404, opened 2023-10-31
Fix slice error

bredelings:
First, complain when the user changes the number of states in the Q matrix after using it to create a phyloCTMC object.


hoehna (approved):
Looks good.

Issue #405, opened 2023-11-02
[BUG] ASR - basic_string::_M_construct null not valid


mariosandovalmx:
**Describe the bug**
I was trying to run the script "Inferring ancestral states and rates of morphological evolution and testing for irreversibility". Everything runs fine, but when I try to read the tree trace and construct the ancestral states tree there is an error :  `basic_string::_M_construct null not valid`

The full message is: 

```
Compiling MAP ancestral states from 10002 samples in the ancestral state trace, using a burnin of 2500 samples.
   
   Calculating ancestral state posteriors...
   
   basic_string::_M_construct null not valid
```

**To Reproduce**
`anc_tree = ancestralStateTree(tree=phylogeny, ancestral_state_trace_vector=anc_states, include_start_states=false, file="C:/Users/sando/Desktop/Revbayes/Irreversibility/output/"+CHARACTER+"_ase_irrev.tree", burnin=0.25, summary_statistic="MAP", site=1)`

**Computer info**
I'm using windows 11 and RevBayes v1.2.2

**Additional context**
My phylogeny contains 428 species and my traits matrix has three states:  0= absence, 1 = presence, and  ? = Unknown.

bredelings:
Thanks for the report.  Can you please attach a complete script and data so that we can reproduce the problem?

PR #406, opened 2023-11-02
Request a script and data files in the issue template.

bredelings:
People keep on providing just one line of a failing script, see #405.  So then we begin most PRs by responding "please send us a complete script with data files so that I can run it".


davidcerny (approved):
Makes a lot of sense, and should hopefully make it somewhat easier for us to reproduce submitted issues.

Issue #293, opened 2022-10-22
[BUG] mvSlice crashing in a very complex model


davidcerny:
I recently ran a pair of FBD analyses that employed the same branching-process and substitution models, and only differed from each other with respect to the clock model: in one (represented by the `mvSlice_test_independent_partitions.Rev` script in [this](https://github.com/revbayes/revbayes/files/9845149/mvSlice_test.zip) ZIP archive), every partition got a separate set of branch rates, while in the other, rates were apportioned out to branches using rjMCMC (`mvSlice_test_rjMCMC_over_partitions.Rev`). The overall models were both very complex (independent clocks: 689 different moves per iteration; rjMCMC: 823 different moves per iteration – I tried to quantify complexity using the number of DAG nodes, too, but that resulted in a segfault in `DagNode::getPrintableChildren()`).

However, while the analysis with independent partition-specific clocks worked fine with my initial choice of priors, for the rjMCMC analysis, I had to replace an `mvSlice` move on the alpha parameter of the discrete gamma rate heterogeneity model (line 200 in `mvSlice_test_rjMCMC_over_partitions.Rev`) with `mvScale` moves (lines 206–208). The former caused RevBayes to crash, with the following gdb output:

```
Thread 2 received signal SIGABRT, Aborted.
0x00007ff810fbe00e in ?? ()
(gdb) backtrace
#0  0x00007ff810fbe00e in ?? ()
#1  0x00007ff810ff41ff in ?? ()
#2  0x0000000000000003 in ?? ()
#3  0x00007ff7bfefb588 in ?? ()
#4  0x00000001020bc600 in ?? ()
#5  0x00007ff7bfefb528 in ?? ()
#6  0x00007ff7bfefb550 in ?? ()
#7  0x00007ff810f3fd24 in ?? ()
#8  0x000000010138f738 in vtable for RevBayesCore::RbOrderedSet<RevBayesCore::DagNode*> ()
#9  0x00006000fffff9df in ?? ()
#10 0xffffffffbc10fbc0 in ?? ()
#11 0x00000000000000c8 in ?? ()
#12 0x0000000101441008 in RevBayesCore::RandomNumberFactory::randomNumberFactoryInstance()::singleRandomNumberFactory ()
#13 0x0000000101440fe8 in guard variable for boost::system::detail::stdcat_mx_holder<void>::mx_ ()
#14 0x00007ff7bfefb680 in ?? ()
[...]

PR #407, opened 2023-11-02
Fix UPP distribution

hoehna:
This (hopefully) fixes the issue https://github.com/revbayes/revbayes/issues/293

We should be propagating the call to touch if the parameter values change. Thus, we need to implement:
-  getAffected
-  keepSpecialization
-  restoreSpecialization
-  touchSpecialization


bredelings:
One issue is that the mvUPPAllocation does seem to initiate any touching.  It just does `dist.redrawValue()`:

https://github.com/revbayes/revbayes/blob/a4a40ae7d4f9807c1bac58c4f9f057c2dd765d21/src/core/moves/proposal/mixture/UPPAllocationProposal.h#L149

And that just does `this->value = simulate()`:

https://github.com/revbayes/revbayes/blob/a4a40ae7d4f9807c1bac58c4f9f057c2dd765d21/src/core/distributions/UniformPartitioningDistribution.h#L378

But perhaps I am missing something.

davidcerny:
Unfortunately, this still fails on the case from #293 with the same error:

```
   Running MCMC simulation
   This simulation runs 1 independent replicate.
   The simulator uses 823 different moves in a random move schedule with 2139 moves per iteration
   

Iter        |      Posterior   |     Likelihood   |          Prior   |   num_samp_anc   |    origin_time   |    elapsed   |        ETA   |
------------------------------------------------------------------------------------------------------------------------------------------
0           |       -8414.05   |       -7135.28   |       -1278.77   |              0   |       250.4522   |   00:00:00   |   --:--:--   |
   Error:	Issue in 'UPP-Allocation' on 'rel_branch_rates[74]' after move because posterior of -5646.09 and -5646.35/-5646.35. The move was rejected.
   Error:	Problem processing line 500 in file "/Users/David/Downloads/mvSlice_test/mvSlice_test_rjMCMC_over_partitions.Rev"
```

Issue #391, opened 2023-10-17
[BUG] Constructor function uses setParameter instead of class constructors


bjoelle:
I encountered this while tracking down why `a = Probability(-1)` works when the checks in the class constructors should block that
basically `ConstructorFunction::execute` in `ConstructorFunction.cpp` builds the class from a template and then calls setParameter or setConstParameter, and `Probability::setConstParameter` has no checks on the value passed to it

we could solve the specific issue by adding checks to `Probability::setConstParameter` but this seems like a problem that should have a more generic solution

davidcerny:
This is the source of issue #348.

bredelings:
Perhaps a more generic solution would modify how a ConstructorFunction works?  Or, even more generally, we could express some types (like probability) in terms of a base type and a constraint.  So,
-  RealPos = (Real, x >=0)
-  Probability = (RealPos, x>=0 and x <= 1)
Right now these constrained types don't really seem to _enforce_ constraints, they mostly serve as a note that the constraint exists for the purpose of finding argument-type mismatches.

Looking at the Probability class in `src/revlanguage/datatypes/basic/Probability.cpp`, it looks like the value can be set in 3 ways:
- Probability::Probability( double x )
- Probability::Probability( RevBayesCore::TypedDagNode<double> *x )
- void Probability::setConstParameter(const std::string& name, const RevPtr<const RevVariable> &var)

Only the first way checks the value.  We could certainly add a check to `setConstParamter( )`.  I wonder if we could also add a check to `Probability::Probability( RevBayesCore::TypedDagNode<double> *x )`.

davidcerny:
Any further thoughts on this? My take would be that for now, an inelegant fix is probably better than keeping the error in until we get around to restructuring the whole type system (which unfortunately really is a mess; see also issue #308). Happy to open a quick PR for this if that's what we agree to do.

bredelings:
Yeah, in practice would should probably fix it, even before we have a wider solution.

@bjoelle Do you have any ideas about a "more generic solution"?  Perhaps one that is less invasive than the one I mentioned?

bjoelle:
so my ideas on a more generic solution would be
- as you mentioned, switch the ConstructorFunction to use a constructor ? pro is that it would apply to every class without added effort, con is I'm not sure how well that would work because I'm not sure why things were setup in that way in the first place
[...]

Issue #348, opened 2023-04-23
[BUG] The Probability() function can create probabilities < 0 and > 1


davidcerny:
**Describe the bug**
The `Probability()` function can cast any real number to type Probability, not just those that are >= 0 and <= 1.

**To Reproduce**
```
a <- -13/12
type(a)
b := Probability(a)
type(b)
```

**Expected behavior**
RevBayes should throw an error message stating that the value is outside of the valid range for probabilities. It is not clear to me why this doesn't happen, since both the type constructor and the `Real::convertTo()` function check for this condition.

**Screenshots**
N/A

**Computer info**
RevBayes v1.2.1 on Red Hat Enterprise Linux 8.4 and macOS 12.6.5.

davidcerny:
Solved by PR #410.

PR #409, opened 2023-11-06
fix probability constructor function

hoehna:
This should make a simple and RevBayes consistent fix to the discussion https://github.com/revbayes/revbayes/issues/391

It also fixes this issue: https://github.com/revbayes/revbayes/issues/348

### DIFF
--- a/src/revlanguage/datatypes/basic/Probability.cpp
+++ b/src/revlanguage/datatypes/basic/Probability.cpp
@@ -127,7 +127,7 @@ const MemberRules& Probability::getParameterRules(void) const
     if ( !rules_set )
     {
 
-        argumentRules.push_back( new ArgumentRule( "x", Real::getClassTypeSpec(), "The value.", ArgumentRule::BY_CONSTANT_REFERENCE, ArgumentRule::ANY ) );
+        argumentRules.push_back( new ArgumentRule( "x", Probability::getClassTypeSpec(), "The value.", ArgumentRule::BY_CONSTANT_REFERENCE, ArgumentRule::ANY ) );
 
         rules_set = true;
     }


bredelings (changes requested):
Thanks @hoehna.

1. Can we make the corresponding changes to RealPos.cpp at the same time, and synchronize them? 
2. RealPos doesn't have any argument rules, and doesn't have a setConstParameter( ) member function.  Why is that?
3. RealPos checks the parameter values when constructing from  a TypedDagNode<double> -- we should do this in Probability.cpp also I think.
4. Probability.cpp should check that the value is in [0,1] when we do setConstParameter( ) as well.

davidcerny:
This solves the problem with a single line change, which is great. My only complaint would be that the error message we get when trying to create an invalid probability:

```
a <- Probability(-0.0000000001)
   Error:	Argument or label mismatch for function call.
   Provided call:
   Probability (Real<constant> )
   
   Correct usage is:
   Probability (Probability<any> x)
```

is decidedly less informative than the exception we throw in one of the constructors (`Creation of Probability with value x=<> outside standard probabilities (0,1)`).

@bredelings:

> Probability.cpp should check that the value is in [0,1] when we do setConstParameter( ) as well.

So I've been trying to implement a check like that, and found it surprisingly difficult. The best I could come up with was:

```
double test = std::stod( var->getRevObject().getDagNode()->getValueAsString() );

if ( name == "x" && test >= 0 && test <= 1 )
{
    real = var;
}
else if ( (name == "x" && test < 0) || (name == "x" && test > 1) )
{
    throw RbException() << "Creation of " << getClassType() << " with value x=" << test << " outside standard probabilities (0,1)";
}
```

But this is not very reliable; it doesn't prevent the creation of invalid probabilities whose string representation is rounded / truncated:

```
> Probability(-3)
   Error:	Creation of Probability with value x=-3 outside standard probabilities (0,1)
> Probability(1.00001)
   Error:	Creation of Probability with value x=1.00001 outside standard probabilities (0,1)
> Probability(1.000001)
   1.000001
```

I was mostly surprised that `DagNode` has no simple `.getValue()` method, only `.getValueAsString()` and `.getValueAsJSON`.

bredelings:
@hoehna It seems like requiring the input to Probability to be a Probability defeats the purpose of Probability in some cases.

Looking for `Probability( )` in the tutorials, I see stuff like:
``` R
mix_p <- Probability(1.0 - expected_number_of_mass_extinctions/NUM_BREAKS)
```
and
``` R
rho       = Probability(extant_taxa / total_taxa)
```

It seems like the point here is that Probability can take a `Real` or `RealPos` and return a `Probability`.  It so, then the meaning of `Probability( x )` is something like
> I assume that `x` is in [0,1] even though it has type Real.

However, if `Probability( 1 - p )` now fails because `1-p` isn't **known* to be a Probability, then this defeats the point.  We want to be able to do:
``` R
> x ~ dnBinomial(size=10, p=0.5)
> y ~ dnBinomial(size=10, p = x/10)
   Error:	Argument or label mismatch for function call.
   Provided call:
   dnBinomial (Natural<constant> 'size',
               RealPos<deterministic> 'p' )
   
   Correct usage is:
   dnBinomial (Probability<any> p,
               Natural<any> size)
   
> y ~ dnBinomial(size=10, p = Probability(x/10))
```

Am I overlooking something?

bredelings:
Another possible route we can take is illustrated by Pr #410

PR #410, opened 2023-11-06
Alternate probability fix

bredelings:
An alternative approach to fixing `Probability(-1)` and `Probability(2)`.

Not very generalized.

### DIFF
--- a/src/revlanguage/datatypes/basic/Probability.cpp
+++ b/src/revlanguage/datatypes/basic/Probability.cpp
@@ -26,7 +26,7 @@ Probability::Probability( double x ) : RealPos( x )
     
     if ( x < 0.0 || x > 1.0)
     {
-        throw RbException() << "Creation of " << getClassType() << " with value x=" << x << " outside standard probabilities (0,1)";
+        throw RbException() << "Creation of " << getClassType() << " with value x=" << x << " outside standard probabilities [0,1]";
     }
     
 }
@@ -35,7 +35,8 @@ Probability::Probability( double x ) : RealPos( x )
 /** Construct from double */
 Probability::Probability( RevBayesCore::TypedDagNode<double> *x ) : RealPos( x )
 {
-    
+    if (x->getValue() < 0 or x->getValue() > 1)
+	throw RbException() << "Creation of " << getClassType() << " with value x=" << x << " outside standard probabilities [0,1]";
 }
 
 
@@ -169,6 +170,9 @@ void Probability::setConstParameter(const std::string& name, const RevPtr<const 
     if ( name == "x" )
     {
         real = var;
+	RevBayesCore::TypedDagNode<double>* x = static_cast<const Real&>( real->getRevObject() ).getDagNode();
+	if (x->getValue() < 0 or x->getValue() > 1)
+	    throw RbException() << "Creation of " << getClassType() << " with value x=" << x << " outside standard probabilities [0,1]";
     }
     else
     {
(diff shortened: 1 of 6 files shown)


davidcerny:
This is great, and exactly what I had in mind too. Maybe we could follow @bjoelle's suggestion and avoid code duplication by doing something like this?

```
void Probability::validate( double x ) const
{
    if ( x < 0.0 || x > 1.0)
    {
        throw RbException() << "Creation of " << getClassType() << " with value x=" << x << " outside standard probabilities (0,1)";
    }
}


/** Construct from double */
Probability::Probability( double x ) : RealPos( x )
{
    validate(x);   
}


/** Construct from a DAG node containing a double */
Probability::Probability( RevBayesCore::TypedDagNode<double> *x ) : RealPos( x )
{
    validate( x->getValue() );
}


/** Set a member variable */
void Probability::setConstParameter(const std::string& name, const RevPtr<const RevVariable> &var)
{
    RevBayesCore::TypedDagNode<double>* test = static_cast<const Real&>( var->getRevObject() ).getDagNode();

    if ( name == "x")
    {
        validate( test->getValue() );
        real = var;
    }
    else
    {
        RevObject::setConstParameter(name, var);
    }
}
```

bredelings:
This is a good idea... maybe for the moment we just add a validate( ) method to Probability( ) instead of making it a generic framework.  But could go either way.

src/revlanguage/datatypes/basic/Probability.cpp
```diff
@@ -23,19 +23,25 @@ Probability::Probability( void ) : RealPos( 1.0 )
...
 Probability::Probability( RevBayesCore::TypedDagNode<double> *x ) : RealPos( x )
 {
-    
+    validate();
+}
+
+void Probability::validate(double x) const
+{
+    if (x < 0 or x > 1)
+	throw RbException() << "Creation of " << getClassType() << " with value x=" << x << " outside standard probabilities [0,1]";
```
davidcerny, line 39: I'm wondering if we can use `std::setprecision` to print more decimals for `x` than C++ does by default, because this can currently lead to somewhat confusing error messages:

```
> a <- Probability(1.000001)
   Error:	Creation of Probability with value x=1 outside standard probabilities [0,1]
```

src/revlanguage/datatypes/basic/RealPos.cpp
```diff
@@ -25,7 +25,7 @@ RealPos::RealPos( RevBayesCore::TypedDagNode<double> *x ) : Real( x )
     
     if ( x->getValue() < 0.0 )
```
davidcerny, line 26: I'm wondering under what conditions an error like this one would be thrown, because the constructors for `RealPos` (unlike those for `Probability`) are not directly accessible to the user (i.e., through a Rev function). Can we maybe expand the test suite to include `ReaPos` as well?

davidcerny (changes requested):

src/revlanguage/datatypes/basic/RealPos.cpp
bredelings: I'm not sure if command-line input can trigger this or not, but its probably still good to have a check here.

src/revlanguage/datatypes/basic/Probability.cpp
bredelings: OK, I modified RbException so that std::setprecision now works with it.  Then I fixed the problem with the error message that you point out above.

davidcerny (approved):
Works perfectly! Merging.

Issue #412, opened 2023-11-24
[BUG] readTreeTrace() can create 0-length traces


davidcerny:
**Describe the bug**
I just realized I'd never opened an issue for the problem I was trying to fix with PR #227, so I belatedly do so here. By supplying a particular value – equal to either 1, or the length of the trace – to the `burnin` argument of the `readTreeTrace()` function, it is possible to create an empty `TraceTree` object, which understandably causes its class methods and summary functions to fail. They don't fail particularly gracefully, either: for example, `mccTree()` will just keep printing asterisks ad infinitum. (I once filled up my cluster storage allocation with a 100GB+ Slurm out file produced by such a failed function call!)

**To Reproduce**
[Here](https://github.com/revbayes/revbayes/files/13454291/test_trace_960.txt) is an example trace with 960 trees in it. All of the following causes the resulting `TraceTree` object to be empty, i.e., to contain 0 trees:

```
trace = readTreeTrace("test_trace_960.txt", treetype="non-clock", burnin=1.0)
mcc_tree = mccTree(trace)
```

```
# Writing "1" rather than "1.0" in the hope that RB will interpret it as an integer and not a probability
trace = readTreeTrace("test_trace_960.txt", treetype="non-clock", burnin=1)
mcc_tree = mccTree(trace)
```

```
trace = readTreeTrace("test_trace_960.txt", treetype="non-clock", burnin=960)
mcc_tree = mccTree(trace)
```

**Expected behavior**
I can't imagine a situation in which a 0-length trace would be desirable (what would you do with it?), so an error should be thrown when the user inadvertently tries to create one.

**Computer info**
macOS 12.7.1, RevBayes 1.2.2, latest version of the `development` branch (rapture-2367-ga4a40a).

davidcerny:
Solved by PR #413.

PR #413, opened 2023-11-24
Make it impossible to create 0-length traces

davidcerny:
This is intended to fix issue #412 and supplant PR #227 with a (hopefully) somewhat more elegant solution. The basic idea is still to prevent the user from creating 0-length traces in the first place, rather than to make sure that the class methods of `TraceTree` or summary functions like `mapTree()` can handle them. This is mostly done for the sake of brevity, and because I can't think of any use for a 0-length trace anyway.

@bredelings's idea of switching the order of argument types, so that `1` (or `1.0`) will get preferentially interpreted as an integer rather than a probability, was almost enough to solve the problem. The only other way to create a 0-length trace is to specify a burnin length that exactly matches the total number of trees in the trace, so I added an extra check that throws an error if burnin >= length(trace). The current RB version already fails gracefully when the length of the burnin _exceeds_ the length of the trace, but with a somewhat enigmatic error message that we can probably improve on: `map::at:  key not found`.

### DIFF
--- a/src/revlanguage/functions/io/Func_readTreeTrace.cpp
+++ b/src/revlanguage/functions/io/Func_readTreeTrace.cpp
@@ -204,6 +204,11 @@ RevPtr<RevVariable> Func_readTreeTrace::execute( void )
 
         for(size_t i = 0; i < rv->getValue().size(); i++)
         {
+            if ( burnin >= rv->getValue()[i].getValue().size() )
+            {
+                throw RbException("Specified burnin equals or exceeds the total number of trees in the trace.");
+            }
+            
             (*rv)[i].getValue().setBurnin(burnin);
         }
     }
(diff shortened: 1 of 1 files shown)


davidcerny:
(Note that the other issue mentioned in the discussion under PR #227 – `mapTree()` and `mccTree()` failing if there is just one tree in the trace – no longer occurs.)

bredelings:
Do you have any thoughts on moving the check for burnin down further, either into the Trace::setBurnin( ) function, or into routines that try to access entries of the trace when there are no entries after the burnin?

I could just approve this now, and leave such ideas a later work.

davidcerny:
I'm happy to add something like `if ( b >= values.size() ) throw RbException(...)` to `Trace.h` instead, if you think that's a better solution – I definitely don't have any strong preference here. I have to admit I don't like the second idea (i.e., modifying multiple methods so that they can deal with 0-length traces, if I understand your suggestion correctly) quite as much; it sounds like it would require lots of extra code just to handle something that should never really happen anyway. Again, maybe there _is_ some use for 0-length traces that I'm overlooking, but I can't think of any.

bredelings (approved):

PR #417, opened 2023-12-05
Fixing the AvgDistMat_ExpError validation script

davidcerny:
Fixing the validation script for supertree analyses relying on the `fnAverageDistanceMatrix()` function and the `dnExponentialError()` distribution, in accordance with Sebastian's advice: `rBDP()` is used to generate random "true" trees without having them appear as stochastic nodes in the resulting model. The test now passes.


validation/scripts/Validation_AvgDistMat_ExpError.Rev
```diff
@@ -23,101 +25,76 @@ moves     = VectorMoves()
...
+species <- simTree(4).taxa()
 
-# Get the names of the taxa in the average distance matrix. We will use this to add
-# tips to the tree
+# Construct a variable for the tree drawn from the birth-death process specified above
 
-for (i in 1:n_species) {
-    species[i] <- taxon(taxonName=species_names[i])
-}
+true_tree <- rBDP(n=1, lambda=speciation, mu=extinction, rootAge=tree_height, condition="nTaxa", taxa=species)
```
hoehna, line 42: I find this slightly misleading as the simulated tree will actually not be used for the validation analysis. As you see later, only ADM will be used as a placeholder. During the validation analysis, we actually simulate new datasets from "dnExponentialError(distMat=mat, lambda=0.005)" and use this as the values for ADM in each replicate.

So my suggestion is to not simulate a tree here, rather use a fixed one, to avoid confusion. Or even better, clamp `exp_error_dist` to itself:
p_error_dist.clamp( p_error_dist )

hoehna (changes requested):

validation/scripts/Validation_AvgDistMat_ExpError.Rev
davidcerny: Hmm, I've made the requested change and the test still passes (i.e., the coverage frequencies are what they should be), so I guess that's good, but at this point I just don't understand how the validation tests are supposed to work. My understanding was that they try to implement the protocol of Talts et al. (2020), which should look like this:

$$
\tilde{\theta} \sim p(\theta) \hspace{3cm} (1)
$$

$$
\tilde{y} \sim f(y | \tilde{\theta}) \hspace{3cm} (2)
$$

$$
\theta \sim p(\theta | \tilde{y}) \hspace{3cm} (3)
$$

i.e., simulate a "true"value from the prior, simulate data conditional on that true value according to some sampling distribution $f$, then re-infer parameter values conditional on the simulated data, and exploit the fact that $\tilde{\theta} \sim p(\theta | \tilde{y})$. But what would $\tilde{\theta}$ correspond to in this modified script? We don't seem to be making a distinction between $\tilde{\theta}$ (simulated true value) and $\theta$ (re-inferred estimates), so how does the test just not fall apart conceptually?

hoehna (approved):
I'm happy about these modifications.

PR #426, opened 2024-01-30
Ou refactor

kopperud:
proposal to change the `PhyloOrnsteinUhlenbeckREML` class such that it better corresponds with the FitzJohn (2012; Methods Ecol Evol) supplement

notable changes
* Change the name/suffix from `REML` to `Pruning`. Reason being, I think it's not the REML likelihood, and so calling it REML is misleading at best. The likelihoods computed using this pruning algorithm are identical to the likelihoods computed using the generalized-least-squares (or vcv) method, as long as the root treatment is the same (i.e. root state = theta). If it really were a REML likelihood, then the likelihoods would not be the same
* Remove storing of the normalizing factors, since they were set to 1.0 and were not used anyway
* Compute the log-normalizing factor for the node, instead of the normalizing factor, saving some calls to `exp(...)`
* Rename "contrasts" (which they were not) to "means"
* Rename "contrast_uncertantity" to "variances"
* Add some comments to explain where the equations come from

question: what to do about the related classes, for example the state-dependent OU process (which is on a different branch)?


src/revlanguage/distributions/phylogenetics/branchrate/Dist_PhyloOrnsteinUhlenbeckPruning.cpp
```diff
@@ -132,17 +132,17 @@ const TypeSpec& Dist_PhyloOrnsteinUhlenbeckREML::getClassTypeSpec(void)
  *
  * \return Rev name of constructor function.
  */
-std::string Dist_PhyloOrnsteinUhlenbeckREML::getDistributionFunctionName( void ) const
+std::string Dist_PhyloOrnsteinUhlenbeckPruning::getDistributionFunctionName( void ) const
 {
     // create a distribution name variable that is the same for all instance of this class
-    std::string d_name = "PhyloOrnsteinUhlenbeckREML";
+    std::string d_name = "PhyloOrnsteinUhlenbeckPruning";
```
hoehna, line 138: This is a bit dangerous to do as it breaks backwards compatibility. Would it be OK to keep `PhyloOrnsteinUhlenbeckREML` as an alias? This is imply from the `Rev` side of things.

hoehna (changes requested):
I suggest to keep `PhyloOrnsteinUhlenbeckREML` as an alias for backwards compatibility.

src/revlanguage/distributions/phylogenetics/branchrate/Dist_PhyloOrnsteinUhlenbeckPruning.cpp
kopperud: That sounds good to me!

hoehna (approved):
Now it looks good.

PR #434, opened 2024-03-19
More BLeSS efficiencies

davidcerny:
Some additional improvements to the performance of ln(pdf), again following very helpful suggestions by Claudio Kozický ([@qak](https://github.com/qak)). The main changes are as follows:

(1) The "alignment" of the average distance matrix and the distance matrix of a proposed tree (i.e., the step of ensuring that we are subtracting distances corresponding to the same pair of taxa) is done using `std::sort` rather than `std::find`, which is up to two orders of magnitude faster for 10,000-by-10,000 matrices;

(2) Indexing is done using `uint32_t` rather than `size_t`, which speeds up memory access (and consequently, the matrix subtraction step).

In practice, the performance improvement gained by these changes can be difficult to observe, because the total amount of time spent on evaluating the (log) probability density of a proposed tree is now dominated by the time it takes to convert the tree to a distance matrix. However, some experimentation with `std::chrono::steady_clock` will show that the ln(pdf) calculation implemented in this branch is about 12% faster for 10,000-by-10,000 matrices (~0.63 vs. ~0.55 s).

### DIFF
--- a/src/core/math/distributions/DistributionExponentialError.cpp
+++ b/src/core/math/distributions/DistributionExponentialError.cpp
@@ -10,7 +10,6 @@
 
 #include <cmath>
 #include <numeric>
-#include <iostream>
 #include <algorithm>
 
 #include "DistributionDirichlet.h"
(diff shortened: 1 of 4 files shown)


src/core/math/distributions/DistributionExponentialError.cpp
```diff
@@ -21,6 +23,28 @@
 
 using namespace RevBayesCore;
 
+/*!
```
hoehna, line 26: Wouldn't it make sense to add this function to StringUtilities?

hoehna (changes requested):
Looks good, the only suggestion I have is to move the string index sort to StringUtilities so that it can be used more broadly.

src/core/math/distributions/DistributionExponentialError.cpp
davidcerny: Great point – moved it there!

bredelings (approved):

Issue #436, opened 2024-03-23
[BUG] Mixture model weights not estimated correctly


jembrown:
**Describe the bug**

I am attempting to estimate two sets of state frequencies associated with sites in an alignment using a profile mixture model. To do this, I'm passing a vector of two Q matrices to `dnPhyloCTMC` as the `Q` argument and a vector of two weights (with a `dnDirichlet` prior) as the `siteMatrices` argument. While the two sets of state frequencies are accurately estimated in the Q matrices, the weights just sample from the prior (i.e., the MCMC samples uniformly between 0 and 1).

**To Reproduce**

I am attaching a simulated dataset (`simData.fa`) of 113 taxa where ~90% of sites have a very strong bias towards A and C, and ~10% of sites have a very strong bias towards G and T. I have also attached the tree used to simulate the data. Both the nucleotide frequencies and mixture weights are easily and correctly estimated by other software (e.g., Bali-Phy). The script used to set up the profile mixture model is also attached (`Profile_mixture.rev`).

**Expected behavior**

The mixture weights (`mix[1]` and `mix[2]`) should be correctly estimated as ~10% and ~90%.

**Computer info**

I've observed this behavior while running RevBayes on Mac OS 14.3, using both the currently released version (1.2.1), as well as a version compiled from the `development` branch (v 1.2.2) on 3/21/24.

**Additional context**

(1) This problem may be related to some other issue with the structure of the DAG. Lines 28-29 in the script (`pi_1 := pi[1]` and `pi_2 := pi[2]`) don't seem like they should be necessary, but if they're removed the `pi` values are not written to the file specified by the model monitor.

(2) @bredelings looked into this a bit and has suggested that the problem has to do with a touch not propagating as it should through the DAG. Forcing RevBayes to recalculate conditional likelihoods for all nodes seems to fix the estimation problem.

Files available here: [mixtureBug.zip](https://github.com/revbayes/revbayes/files/14733416/mixtureBug.zip)

jsigao:
[...]

PR #437, opened 2024-03-28
Correctly updating phylogenetic likelihood when site mixture probability changes

jsigao:
This PR is supposed to fix an issue identified in this [bug report #436](https://github.com/revbayes/revbayes/issues/436) by @jembrown. The phylogenetic likelihood wasn't updated when only the site mixture probability vector, v<sub>mix</sub>---i.e., either the site rate probs or the site matrix probs---changed at an MCMC step. This PR Introduced an additional flag to allow revbayes to execute the `sumRootLikelihoods` function (instead of doing nothing, which was the cause of this bug) without having to recomputing any of the partials when only v<sub>mix</sub> is updated.

The bug fix is validated by comparing the output of running the example dataset provided in [#436](https://github.com/revbayes/revbayes/issues/436). The two computation options (as pointed out by @bredelings) when only v<sub>mix</sub> changes---1) recomputing the phylogenetic likelihood entirely and 2) only executing `sumRootLikelihoods` without updating any partials---produce identical outputs and both achieve the expected behavior stated in [#436]: 

> The mixture weights (mix[1] and mix[2]) should be correctly estimated as ~10% and ~90%.

Option 2 is chosen to be the implementation of this bug fix because of its computational efficiency. Option 2 is about 25% faster than option 1 for the example dataset, consistent with the specified 25% of proposal weight on updating v<sub>mix</sub>.


bredelings:
Thanks @jsigao for looking into this carefully and checking the times and accuracy!

Hmm... so one other option would be to remove the `mixture_probs_dirty` flag, and to do `this->lnProb = sumRootLikelihood();` unconditionally.  It seems slight simpler, and maybe more robust.  What do you think?

jsigao:
Thanks for the suggestion @bredelings! I'd be fine with always doing `this->lnProb = sumRootLikelihood(); `unconditionally. I guess though I don't know, as you previously mentioned, if there is any case that the `computeLnProbability` function in phyloctmc is called but the likelihood does not need to be updated at all. Even that may happen, considering that the computation cost of `sumRootLikelihood` should be fairly small, perhaps it'd still be more robust and virtually as efficient as it could be to always executing `sumRootLikelihood` unconditionally?

bredelings:
As you say, if there is some case where this is a slowdown, probably it is not much.  And I think we should value robustness over efficiency here.  So, if you are OK with it, lets do `this->lnProb = sumRootLikelihood();` unconditionally!

jembrown:
This plan sounds good to me. Do we need an additional commit to do `this->lnProb = sumRootLikelihood();` unconditionally?

bredelings (approved):

Issue #423, opened 2024-01-15
[BUG] Problem with mvCollapseExpandFossilBranch


davidcerny:
**Describe the bug**
This is a problem I first mentioned on Slack in November 2023, but I wanted to create a GitHub issue for it as well to make sure it doesn't slip through the cracks. I also followed @bredelings's advice to run the example dataset below with multiple seeds to find the one that crashes first.

Recently, I tried to run a tip-dating analysis on a small morphological matrix under the FBD model and a white-noise relaxed clock, and noticed problematic behavior from the `mvCollapseExpandFossilBranch()` move. When the `DEBUG_MCMC` preprocessor macro is enabled, the analysis crashes with the following error after ~380 iterations (~96,000 moves):

```
   Error:       Issue in 'CollapseExpandFossilBranch' on 'fbd_tree' after move because posterior of -1889.28 and -inf/-inf. The move was rejected.
```

Without the macro, the analysis will seemingly run just fine, but the move will never be accepted, and no ancestors will ever be sampled:

```
            Name             | Param         |  Weight  |  Tried  | Accepted | Acc. Ratio | Parameters
======================================================================================================
CollapseExpandFossilBranch     fbd_tree         6.0000    2399416   0          0.0000
```

As far as I can tell, this problem is separate from issue #401, another recently described bug having to do with sampled ancestors under the FBD model.

**To reproduce**
[This zipped directory](https://github.com/revbayes/revbayes/files/13943142/samp_anc_bug_test.zip) contains (1) a character matrix, (2) a taxon data file with tip age ranges, (3) a starting tree, and (4) a script to run the analysis. One absolute path has to be changed in the script.

**Expected behavior**
The `mvCollapseExpandFossilBranch()` move should be accepted at least some of the time.

**Computer info**
Red Hat Enterprise Linux 8.4 (Ootpa), RevBayes 1.2.2, `development` branch, commit rapture-2364-ga6a40c (two PRs behind the tip of the branch).

willgearty:
I noticed this in an FBD analysis I recently ran, not a single move was ever accepted.

bjoelle:
So after investigating, I think everything is behaving correctly, *but* the default removal probability in BDSTP is set to 1 so unless you set it explicitly `r=0` it will not allow SAs
if I set r to 0 in your script I do get a non-zero amount of SAs
[...]

PR #440, opened 2024-04-04
BDSTP configuration updates

bjoelle:
This PR is to address issue #423 with two main changes
 * Refactoring dnBDSTP, dnFBDP and dnPhylodynamicBDP so they now share common code instead of being duplicates, and deviations from the default behaviour are more clearly documented
 * Adding a check to mvCollapseExpandFossilBranch so it errors out if the corresponding tree is set with r=1. This is a bit wonky because it involves two separate components so comments welcome if there is a better way.

FBD tutorial has also been edited so it no longer recommends r=1 

Bonus fix : test output is now correctly uploaded by the automated build when tests fail


src/revlanguage/distributions/phylogenetics/tree/Dist_PhylodynamicBDP.cpp
```diff
@@ -201,26 +33,12 @@ RevBayesCore::AbstractBirthDeathProcess* Dist_PhylodynamicBDP::createDistributio
 const std::string& Dist_PhylodynamicBDP::getClassType( void )
 {
 
-    static std::string rev_type = "Dist_PhylodynamicBDP";
+    static std::string rev_type = "Dist_PhylodynamicBDSTP";
```
davidcerny, line 36: Wondering if it's worth making the old name an alias to preserve backward compatibility...

davidcerny (approved):
This is amazing, thank you so much, Joëlle! The code changes all look good to me, and I've verified the following:

- In `dnBDSTP`, the `r` argument no longer has a default value, so the distribution can't be called without explicitly specifying removal probability;
- If `r` is set to 1 and `mvCollapseExpandFossilBranch` appears in the script, a warning is thrown;
- If the `dnPhylodynamicBDSTP` alias is used in the same setting, the same warning is thrown also;
- When using the same seed, the likelihoods are exactly identical to those produced by the current `development` branch both when sampled ancestors are allowed (`dnBDSTP` with `r=0` or `dnFBDP`) and when they are disallowed (`dnBDSTP` with `r=1`).

The current `dnPhylodynamicBDP` is unable to find an initial state for my dataset, which indicates that it does something different from merely setting `r=1` in `dnBDSTP` (since that combo does work). In contrast, `dnPhylodynamicBDSTP` from this branch does behave simply as `dnBDSTP` with `r=1`. Unfortunately, I can't tell if the divergence of `dnPhylodynamicBDP` from `dnBDSTP` with `r=1` was just a bug, or if there was a deeper reason for it.

The only change I'd maybe suggest would be to keep `dnPhylodynamicBDP` as an alias of `dnPhylodynamicBDSTP` for backward compatibility, but I don't insist on it.

src/revlanguage/distributions/phylogenetics/tree/Dist_PhylodynamicBDP.cpp
bjoelle: added

Issue #366, opened 2023-07-21
[BUG] Alloc Issues with Power Posteriors from Tutorial


mmyen:
**Describe the bug**
A few days ago, when I was going through the ["General Introduction to Model selection" tutorial](https://revbayes.github.io/tutorials/model_selection_bayes_factors/bf_intro.html), I got a `std::bad_alloc` error while trying to set up the stepping stone sampling and path sampling code. I did not notice any provided data files and scripts on the page, so I decided to copy and paste in an old model from an earlier tutorial page (["Introduction to MCMC"](https://revbayes.github.io/tutorials/mcmc/poisson.html)). 

My model by itself ran without any issues, so I'm a little confused as to why adding the power posteriors and stepping stone or path sampling methods (I tried both; neither worked) would make my code break.

**To Reproduce**
First, clear the RevBayes workspace (`clear()`) and quit out of RevBayes (`quit()`). Then, run this script (compressed for uploadability):
[bayes_model_sel_2.Rev.zip](https://github.com/revbayes/revbayes/files/12134711/bayes_model_sel_2.Rev.zip)

**Expected behavior**
I expected a marginal likelihood to be printed/shown on the terminal. 

**Screenshots**
I got these screenshots (potentially sensitive information redacted):
<img width="607" alt="Screen Shot 2023-07-21 at 2 06 38 PM" src="https://github.com/revbayes/revbayes/assets/91394943/d1b1b77c-f395-4afa-9801-96d76d0fe174">
<img width="598" alt="Screen Shot 2023-07-21 at 2 03 35 PM" src="https://github.com/revbayes/revbayes/assets/91394943/37ece8f3-3f88-47d7-bef3-f180cad27b8e">

**Computer info**
I am using RevBayes v1.2.1 on a Mac.

**Additional context**
I also noticed that there were some [files](https://github.com/revbayes/revbayes_tutorial/tree/390a3d78ce63dc7c08cb37e172eefff67b61c42d/RB_BayesFactor_Tutorial/scripts) in the tutorial repository that found the power posterior distributions and the stepping-stone and path sampling marginal likelihoods, which is what was described in the ["General Introduction to Model selection" tutorial](https://revbayes.github.io/tutorials/model_selection_bayes_factors/bf_intro.html). Can someone please let me know the reason they're not linked on the page itself?

bredelings:
Hmm... so what is going on here is that both the monitor and the poster posterior are trying to write to the same file at the same time:
```
[...]

PR #443, opened 2024-04-09
Better error handling for MarginalLikelihoodEstimator

bjoelle:
This PR attempts to prevent issues similar to #366 by 1) failing gracefully instead of segfaulting when the power or likelihood column cannot be found 2) actually giving an error on a non-numeric input instead of silently failing


davidcerny:
Not a reviewer on this one, but this looks like a very nice bug fix to me – I'd be happy to approve & merge it!

src/core/analysis/mcmc/MarginalLikelihoodEstimator.cpp
```diff
@@ -97,10 +97,19 @@ MarginalLikelihoodEstimator::MarginalLikelihoodEstimator(const path &fn, const s
...
+            // check for broken lines
+            if( columns.size() <= powerColumnIndex || columns.size() <= likelihoodColumnIndex ) 
+                throw RbException() << "Please check format of file " << fn << ", missing power and likelihood columns in some lines";
+
+            double p, l;
             // get the power entry
             std::string tmp = columns[powerColumnIndex];
-            double p = atof( tmp.c_str() );
+            try {
+                p = std::stod( tmp.c_str() );
```
hoehna, line 109: Doesn't std::stod takes a std::string as input instead of a c-array of characters?

src/core/analysis/mcmc/MarginalLikelihoodEstimator.cpp
```diff
@@ -111,7 +120,11 @@ MarginalLikelihoodEstimator::MarginalLikelihoodEstimator(const path &fn, const s
         
             // get the likelihood entry
             tmp = columns[likelihoodColumnIndex];
-            double l = atof( tmp.c_str() );
+            try {
+                l = std::stod(tmp.c_str());
```
hoehna, line 124: Doesn't std::stod takes a std::string as input instead of a c-array of characters?

hoehna (changes requested):
In general that looks like nice error catching. Just for code cleanness, perhaps use the string directly as input for stod.

src/core/analysis/mcmc/MarginalLikelihoodEstimator.cpp
bjoelle: you're right, I just copied the previous code - fixed now

hoehna (approved):
Looks good.

Issue #415, opened 2023-12-04
[BUG] vectorFlatten


igoryanovich:
**Describe the bug**
1) vectorFlatten() works hardly as intended, only returning a vector with one element. 
2) in the [documentation,](https://revbayes.github.io/documentation/vectorFlatten.html) the return type is RealPos while the input is Real. This can hardly be correct.  

**To Reproduce**
1) vectorFlatten(v([1,2], [3,4]))  results in   [ 2.000 ]
2) vectorFlatten(v([1,-2.0], [3,4])) returns  [ -2.000 ], but upon calling type(vectorFlatten(v([1,-2.0], [3,4]))), it's said to be RealPos[] despite -2.0.

bredelings:
The second issue seems to come from here: https://github.com/revbayes/revbayes/blob/a4a40ae7d4f9807c1bac58c4f9f057c2dd765d21/src/revlanguage/functions/math/Func_vectorFlatten.h

So, we could:
* change RealPos to Real
* change RealPos to `T` and make this into a template class.  I think we can replace `RevBayesCore::RbVector< double>` with `RevBayesCore::RbVector< T::valueType >`

However, I don't have time to do this myself right now.

PR #445, opened 2024-04-10
changed vectorFlatten to use Real everywhere

bjoelle:
This is a small change to address issue #415 (potentially as a temporary solution before we make this a templated class as suggested by @bredelings in the linked issue)


bredelings:
Yeah... let's do that for now.  And maybe make a note on #415 that the issue is partially addressed.

bredelings (approved):

Issue #339, opened 2023-03-23
[BUG] Deterministic variables when restarting from checkpoint


davidcerny:
**Describe the bug**
I performed a partitioned, topologically constrained FBD analysis using the MPI version of RevBayes, running 10 chains in parallel and saving checkpoint files every 100 iterations. To get around issue #184, I then used the single-thread version of RevBayes to restart each of the 10 runs individually. The analysis made one attempt to draw an initial state and then threw the following error, originating from the `setValueFromString()` member function of class `DeterministicNode`: 

```
Error:	Cannot set a deterministic node from a string.
```

I tried to investigate which variable was responsible for this, since the overall model contained a number of deterministic nodes. However, when I recompiled the exact same version of RevBayes with `-debug true` and ran it within GDB, I was unable to reproduce the issue, as the analysis instead threw a different error:

```
Assertion failed: (isnan(sum) || (0 <= sum and sum <= 1.00000000001)), function computeInternalNodeLikelihood, file PhyloCTMCSiteHomogeneous.h, line 264.
```

**To Reproduce**
A zip file with the script, data files, and the checkpoint state file is available [here](https://github.com/revbayes/revbayes/files/11056197/detnode_restart_error_example.zip). One absolute path has to be changed in the script.

**Expected behavior**
The analysis should be successfully re-initialized. I would expect RevBayes to re-initialize deterministic nodes conditional on the current values of their parent nodes, instead of attempting to set their values from a string.

**Screenshots**
N/A

**Computer info**
I encountered this problem with both the stable (rapture-1925-g7b872a) and development (rapture-2131-g72b6d3) versions of RevBayes 1.2.1 on Red Hat Enterprise Linux 8.4 and macOS 12.6.3.

jsigao:
Hi David, after glancing over the checkpoint file (`zhang_et_wang_19_rjMCMC_partially_constrained_run_1.state`), I think this may be related to a similar checkpointing issue that I encountered and tried to fix a short while ago. Specifically, the issue is caused by the mismatch between the ways that a stochastic-variable vector is written to and read from the checkpoint file. 

[...]

Issue #184, opened 2021-09-11
[BUG] initializeFromCheckpoint crash - MPI related


BenjaminGuinet:
Hello everyone, I'm writing this mail in the hope that someone has already encountered the same kind of problem when trying to use mpirun on the cluster.

I use mpirun to run a Revbayes script:

`mpirun -np 6 --oversubscribe /beegfs/data/niwradel/TOOLS/revbayes/projects/cmake/rb-mpi /beegfs/data/niwradel/these/Rev_datation_analysis/scripts/Revbayes_main_6.Rev`

**Here is the Revbayes_main_6.Rev content :** 

```
D <- readDiscreteCharacterData(file="/beegfs/data/niwradel/these/Rev_datation_analysis/data/Concatenated_aligned_busco_seqs3_constrain_complet_name_SNPs_sampled_6.fasta")
n_sites <- D.nchar()
mi =1
source("/beegfs/data/niwradel/these/Rev_datation_analysis/scripts/BDP2.Rev")
source("/beegfs/data/niwradel/these/Rev_datation_analysis/scripts/GMC2.Rev")
mymodel = model(Q)
monitors.append(mnModel(filename="/beegfs/data/niwradel/these/Rev_datation_analysis/output/TimeTree_mcmc_6_2.log",printgen=400))
monitors.append(mnFile(filename="/beegfs/data/niwradel/these/Rev_datation_analysis/output/TimeTree_mcmc_6_2.trees",printgen=400, timetree))
monitors.append(mnScreen(printgen=200, extant_mrca))
## Setting-Up & Executing the MCMC
mymcmc = mcmc(mymodel, monitors, moves,nruns=2,combine="mixed")
mymcmc.burnin(generations=1000,tuningInterval=300)

mymcmc.initializeFromCheckpoint("/beegfs/data/niwradel/these/Rev_datation_analysis/scripts/checkpoint6_2.txt")# #comment this out for the first run
mymcmc.run(generation=200000,checkpointFile="/beegfs/data/niwradel/these/Rev_datation_analysis/scripts/checkpoint6_2.txt", checkpointInterval=100)

# Building the MAP tree
treetrace = readTreeTrace("/beegfs/data/niwradel/these/Rev_datation_analysis/output/TimeTree_mcmc_6_2.trees", treetype="clock", burnin=0.25)
map_tree = mapTree(treetrace, "/beegfs/data/niwradel/these/Rev_datation_analysis/output/TimeTree_mcmc_6_2.tree")

[...]

PR #448, opened 2024-04-14
Don't flatten arrays of stochastic variables in checkpoint files

davidcerny:
This is another trivial, single-line bug fix that we've never gotten around to implementing, this time for issue #339. @jsigao suggested this as a solution and implemented it on [his branch](https://github.com/revbayes/revbayes/commit/ec20e7875f3f572c7ad55b0767b40c7492538b4c); I verified that it's sufficient to get rid of the issue. While the plan was to make it just one part of a more extensive checkpointing fix, it's been more than a year, and it seems worth fixing the issue even without a more general solution to MPI and MC^3 checkpointing problems. However, precisely because the problem of getting checkpointing to work under MPI is still [unresolved](https://github.com/revbayes/revbayes/issues/184), I'm holding off from adding a full-blown checkpointing test.

### DIFF
--- a/src/core/analysis/mcmc/Mcmc.cpp
+++ b/src/core/analysis/mcmc/Mcmc.cpp
@@ -202,7 +202,7 @@ void Mcmc::checkpoint( void ) const
 {
     // initialize variables
     std::string separator = "\t";
-    bool flatten = true;
+    bool flatten = false;
     
     createDirectoryForFile( checkpoint_file_name );
     


bredelings:
Does this also affect logging multidimensional arrays to CSV files?  I think @mlandis was looking at this, but it might be a separate issue...

davidcerny:
If you happen to have a test case, I'd be happy to try it out!

bjoelle (approved):

PR #449, opened 2024-04-15
Disable FBDR models temporarily

bjoelle:
This PR is to remove the FBD range models from use
At the moment they show problematic behaviour and it's not clear whether they're calculated correctly
Further tests are in progress, in the meantime users can use an old version of these models on the `rapture_FBD` branch


hoehna (approved):
Looks good.

Issue #398, opened 2023-10-22
[BUG] fnPruneTree() can produce malformed Newick strings


davidcerny:
This bug was originally identified by Alessio Capobianco during the Munich RevBayes hackathon.

**Describe the bug**
The `fnPruneTree()` function can cause a tip to detach from the rest of the tree. Printing such a pruned tree results in a malformed Newick string where the corresponding label is still present, but there is no comma separating it from the outermost parenthesis, so it gets interpreted as a root label rather than a tip. Here is an example (2nd tree in the `pruned.trees` file linked to below, with node indices and branch lengths removed for clarity):

```
(((((Crocodylus_siamensis,Crocodylus_johnstoni),((Crocodylus_porosus,Crocodylus_moreletii),Crocodylus_rhombifer)),((Crocodylus_niloticus,(Crocodylus_novaeguineae,Caiman_latirostris)),(Crocodylus_palustris,Alligator_sinensis))),(((((Alligator_mississippiensis,Gavialis_gangeticus),Mecistops_cataphractus),((Melanosuchus_niger,Tomistoma_schlegelii),Crocodylus_acutus)),((Crocodylus_intermedius,Paleosuchus_palpebrosus),(Paleosuchus_trigonatus,Crocodylus_mindorensis))),(Caiman_yacare,Osteolaemus_tetraspis))))Caiman_crocodilus;
```

Internal RevBayes methods like `.ntips()` or `.taxa().size()` also return one fewer tips than they should, even before the tree is written to a file – so this is definitely a problem with the pruning algorithm rather than the Newick writer. I _think_ I've managed to track it down to the `recursivelyPruneTaxa` method of the `Tree` class. In this initial loop:

```
if( n->isTip()  )
{
    bool prune = prune_map[n->getIndex()];

    if( prune == false )
    {
        nodes.insert(nodes.begin(), n);
    }

    return prune;
}
```

all the tips to be retained get correctly added to the `nodes` vector, so the issue must occur later, when we are trying to add internal nodes too. Unfortunately, I haven't been able to get any further with this.

**To Reproduce**
[This zipped directory](https://github.com/revbayes/revbayes/files/13063426/pruning.zip) contains (1) a trace of two trees to be pruned, (2) a script to prune them and check the number of retained tips, and (3) the output trace of pruned trees. One absolute path has to be changed in the script.

**Expected behavior**
[...]

PR #450, opened 2024-04-15
Stop tree pruning from using tips as root

bjoelle:
This is an attempt to fix issue #398 by stopping the algorithm from selecting a tip as root of the tree when pruning a non-clock tree


davidcerny (approved):
Awesome work, thanks a lot Joëlle! I can confirm it fixes the issue on my end too.

I'm happy to merge this as-is, but I have two suggestions. The second one especially should be treated as optional, because it goes beyond simply fixing the issue and I don't know how much work it would involve:

1. In my test script, the check that (# of pruned tips + # tips in the pruned tree) = # tips in the full tree is currently only printed to the screen. It might be worth printing it to a plain-text file that is accessible to the test runner.

2. Would it be difficult to allow for the possibility that by pruning an unrooted tree, we might obtain a rooted tree? I noticed that in full tree 2, there is a node that unites the _Osteolaemus_ lineage and the _Crocodylus niloticus_ lineage to the exclusion of the _Caiman crocodilus_ lineage. But in the pruned tree, this node disappears and _Caiman crocodilus_ is in a polytomy with the remaining two lineages. In some cases at least, getting a rooted tree by pruning the outgroups from an unrooted tree might be exactly what the user wants to do.

bjoelle:
thanks for the review @davidcerny !
for 1) I will add this to the test
for 2) I am not super familiar with that part of the code but I think you could get a result like that if you load the tree as "clock" instead of "non-clock" ? then it will be treated as rooted. But I don't think it's a good idea to allow from conversions from unrooted to rooted from within the pruning function (although it could be a separate rooting function, maybe that even already exists ? either way that's a separate PR I think...)

Issue #272, opened 2022-09-21
[BUG] Tuning the scaling move on a mixture variable leads to +Inf values


acapomorphic:
**Describe the bug**
NaN or Inf produced during MCMC

**To Reproduce**
[Test_nan.Rev.txt](https://github.com/revbayes/revbayes/files/9616480/Test_nan.Rev.txt)

**Expected behavior**
Normal MCMC ;)

bredelings:
I tried the test script, and eventually the prior because -inf and stayed that way.  Since Inf - Inf = NaN, the Inf's could be the root of the problem.

Its not necessarily wrong for a proposed state to have prior=-Inf.  However, such proposals should not be accepted.  So that may be the major bug.

Its also possible that the prior should not be -Inf, that would be the minor bug.

In case someone has time to look at this bug, the following fragment should make it easier to make gdb stop revbayes at the place where we are generating an NaN or Inf.  But I don't have time to look into this further right now.

```
#include <fenv.h>
...
feenableexcept(FE_DIVBYZERO|FE_OVERFLOW|FE_INVALID);
```

We should probably also add some `assert( )` statements that an MH move never switches to a new state where the posterior probability is NaN, Inf, or -Inf.

bredelings:
One issue here is that RevBayes is using NaNs to indicate "we don't actually have a value for this number".  So, trapping on the first NaN doesn't work yet.  We should probably replace these NaNs with std::optional<double> to indicate that we may or may not have a double.  This makes more sense, and would also allow trapping on NaNs.

bjoelle:
ok so this problem comes from the interaction of tuning, the `mvScale`, and the mixture variable `branch_deltas`
what happens is the following : the mixture variable spends a lot of time being constant at 0 -> as a result the move is accepted often (because it doesn't do anything) -> as a result the tuning increases the size of the proposal -> at some point the size of the proposal is so large that it gives a scaling factor of +inf so a proposed value of +inf -> the HR of this proposal is log(factor) = +inf so it's accepted -> the value becomes +inf and everything breaks

note that the `mvSlideBactrian` which the tutorial recommend to use with mixture variables has an upper and lower bound on the scaling factor so does not suffer from this issue => we could do the same here
**but** it doesn't resolve the fact that acceptance rates are wrongly calculated for moves on mixture variables which seems like the underlying issue

bjoelle:
I will PR an upper and lower bound on the tuning for the scaling move because that seems like a good idea regardless of anything else

PR #452, opened 2024-04-16
added bounds on tuning parameter for mvScale

bjoelle:
This is a partial fix for the issue #272 where a bad interaction between tuning and mixture variable leads to overflow on the scaling factor used in mvScale
similar bounds are implemented in e.g. mvSlideBactrian


hoehna (approved):
Looks good.

Issue #159, opened 2021-03-09
segmentation fault or hang dealing with many topological constraints


zhangchicool:
The morphological dataset has 54 hard topological constraints in a FBD dating analysis (attached). It fail to launch in both Mac (Big Sur 11.2.2) and Ubuntu (20.04.2 LTS). Mac gave "segmentation fault" and Ubuntu hung after saying "Successfully read one character matrix from file 'pterosaur_discrete.nex'".  Removing some constraints, say just leaving 5 to 10 constraints enabled, the program seems to run fine.

The same full set of constraints was tested in MrBayes and had no problem to run (also attached).

RevBayes was compiled from source from the develop branch (Build from development (rapture-641-g7d4aab) on Mon Mar  8). 

Thanks for looking at this issue.

[topo_constraints.zip](https://github.com/revbayes/revbayes/files/6106641/topo_constraints.zip)

zhangchicool:
Another question related, does RevBayes have an option to set multiple backbone constraints? 

For example, let A B C form a clade, D E F form another clade, etc. and the rest of the taxa, say, X Y Z, can move freely in the tree (in or out of the defined constraints).

bredelings:
There seems to be a problem trying to draw an initial value that fits the constraints.

After I added a hack to my local code, the example ran fine:
```
diff --git a/src/core/distributions/phylogenetics/tree/birthdeath/AbstractBirthDeathProcess.cpp b/src/core/distributions/phylogenetics/tree/birthdeath/AbstractBirthDeathProcess.cpp
index 52061613..64e2e0ba 100644
--- a/src/core/distributions/phylogenetics/tree/birthdeath/AbstractBirthDeathProcess.cpp
+++ b/src/core/distributions/phylogenetics/tree/birthdeath/AbstractBirthDeathProcess.cpp
@@ -10,6 +10,8 @@
 #include "RbMathCombinatorialFunctions.h"
 #include "TopologyNode.h"
 #include "AbstractRootedTreeDistribution.h"
+#include "RandomNumberGenerator.h"
+#include "RandomNumberFactory.h"
 #include "Tree.h"
 
 namespace RevBayesCore { class Taxon; }
@@ -127,9 +129,15 @@ std::vector<double> AbstractBirthDeathProcess::simulateDivergenceTimes(size_t n,
     {
         double t = 0.0;
 
+        auto rng = GLOBAL_RNG;
+        double u = rng->uniform01();
+        t = min + u*(origin-min);
+
[...]

Issue #384, opened 2023-10-03
[BUG] Segfault w/ backbone constraints for nonclock trees unless initial tree is specified


davidcerny:
**Describe the bug**
When using the `dnConstrainedTopology` distribution with nonclock trees (i.e., when the `treeDistribution` argument if of type `Distribution__BranchLengthTree`), a backbone constraint can only be employed when an initial tree is specified as well. If the `initialTree` argument is omitted, passing a constraint tree to the `backbone` argument will cause a segfault. For clock trees (i.e., when `treeDistribution` is of type `Distribution__TimeTree`), this doesn't happen – the analysis might fail to find a suitable initial state, but there will be no segfault.

I tried to investigate further using the backtrace functionality in GDB, but unfortunately, I'm not able to parse the output beyond seeing that the problem has to do with `TopologyNode`:

```
Thread 2 received signal SIGSEGV, Segmentation fault.
0x000000010436c9bc in std::__1::__tree<RevBayesCore::TopologyNode const*, std::__1::less<RevBayesCore::TopologyNode const*>, std::__1::allocator<RevBayesCore::TopologyNode const*> >::size (this=0x70) at /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.1.sdk/usr/include/c++/v1/__tree:1071
1071	    const size_type& size() const _NOEXCEPT {return __pair3_.first();}
(gdb) backtrace
#0  0x000000010436c9bc in std::__1::__tree<RevBayesCore::TopologyNode const*, std::__1::less<RevBayesCore::TopologyNode const*>, std::__1::allocator<RevBayesCore::TopologyNode const*> >::size (this=0x70) at /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.1.sdk/usr/include/c++/v1/__tree:1071
#1  0x00007ff7bfef8940 in ?? ()
[...]

PR #453, opened 2024-04-17
Stop BD simulation from hanging in situations with many constraints

bjoelle:
This is a fix for issue #159 where the combination of many topological constraints made it hard to find good ages for the simulated tree and led to the BD simulator hanging in an infinite loop
Now if a good time cannot be found after several tries the age will simply be set based on the minimum

To be honest I don't think the trials would be needed at all, but removing them would require changing a lot of test output


src/core/distributions/phylogenetics/tree/birthdeath/AbstractBirthDeathProcess.cpp
```diff
@@ -126,10 +127,18 @@ std::vector<double> AbstractBirthDeathProcess::simulateDivergenceTimes(size_t n,
...
-        do {
+        int ntries = 0;
+        
+        do { 
             t = simulateDivergenceTime(origin, present);
-        } while ( t < min );
+            ntries ++;
+        } while (t < min && ntries < 1000);
+
+        if(t < min) {
```
hoehna, line 137: Do you think this is safe or would it be better to throw an error?

bjoelle: my thinking was that this is primarily used to initialize the run so it doesn't matter too much what the ages are as long as they work with the constraints...
if we just throw an error then we're back with the initial issue which is that some analyses cannot start because it can't find a valid tree

bredelings:
I think the real solution here is to separate initialization from sampling.
So we need a function that randomly initializes a variable, but maybe not
from the correct distribution.

One solution that might not be too hard here would be to create a virtual
function for sampling an initial point, where the default implementation is
to sample from the true distribution. Then for the birth death process,
provide an alternate implementation that prioritizes termination over
correctness.

Then I guess you would need to call the initialization sampling function in
the right place...

bjoelle:
I would agree that this may a good idea to separate simulation and initialization, since a lot of BD models don't have good simulation functions under certain constraints (fossils, or like here topological constraints), but I don't know if it's possible to distinguish the two uses from the Rev side of things at the moment
since this is a case where we cannot simulate properly anyway (because of the constraints on topology) I think we could do what I suggested as a temporary fix while someone (me if I find the time) goes looking further ?
especially since that issue exists since 2021

bjoelle:
as an alternative I just thought of, I could add a warning to the simulation if we get into the fallback behaviour
that way it doesn't block the analysis but the user can decide if it's ok or not that the simulation is not fully BD

davidcerny:
This is just my 5 cents, but I'm personally not too concerned about the fact that the initial state might not be a valid draw from the prior. After all, we already allow users to supply their own initial trees, which (presumably) don't constitute valid draws, either.

I also wanted to point out that this discussion is relevant to issue #384 as well, i.e., the fact that it's currently impossible to find an initial state for non-clock trees with backbone constraints. This one, too, should be easy to fix if we're willing to decouple initialization from simulation (e.g., we could take all the tips that are not in the backbone, and randomly drop them onto the backbone).

bredelings:
@bjoelle Hmm... perhaps the warning might be the way to go for now.  The rest of RevBayes is not set up to handle the distinction between simulation and initialization yet, but perhaps logically separating the two different things in the api is a step in the right direction?

@davidcerny I think allowing users to set the value is different.  Presumably if a user sets the value, they know that it doesn't come from the distribution.  But people also use RevBayes to simulate stuff, so it would be problematic if it silently simulates from the wrong distribution.

Long term, here are three things that might be possible:
- allow distributions/variables to have an empty value, and simulate the value when it is first needed/requested.  Currently we do weird stuff because we try to simulate after every parameter change, even before all the parameters are initialized.
- allow distributions/variables to have an "can't simulate" value.  When someone writes `x ~ dnVeryComplex( )` then writes `x` at the prompt, we could say "x does not have a value; cannot be directly simulated" or something like that.  The idea being that some distributions can only be simulated from via MCMC.
- MCMC routines would start by looking for variables with a "can't simulate" value and then initialize them using the initialization routine

davidcerny:
That would be a very elegant solution, yes!

hoehna:
Perhaps it might be worth looking into SimulationCondition (https://github.com/revbayes/revbayes/blob/7c2b560ca900b988881d16a4dbc1902bf902ce36/src/core/distributions/Distribution.h#L61)

From my perspective it would be super nice if we could adopt their usage. Currently there are two options: (1) MCMC which only requires a valid state (non-zero probability), and (2) VALIDATION, which should make the validation scripts work, for example, in BD models not condition the simulation in n taxa.

bjoelle:
ok as suggested I added an argument to all tree simulations to decide whether we're ok with returning a state that is not drawn from the true distribution - note that I haven't checked the behaviour of all simulation functions, so I just assumed that they did the correct thing unless explicitly specified
Note nb2 is that a lot of the constructor functions use draws from the simulation as well - in that case I assumed that we didn't care about the true distribution (so `alwaysReturn = true`) but I can change it if that's not correct

hoehna (approved):

PR #454, opened 2024-04-18
Compute average distance matrices more efficiently

davidcerny:
Apologies for implementing all of this in a piecemeal manner, but I just noticed the same trick I'd used in PR #432 to make `ExponentialError::lnPdf` more performant can also be exploited in `TreeUtilities::getAverageDistanceMatrix`. (In contrast, the even faster `std::sort`-based solution eventually implemented in PR #434 wouldn't work here, because we are working with string vectors of different lengths.) It results in an approximately 16-fold speed-up when computing 10,000-by-10,000 average distance matrices.


hoehna (approved):
looks good

Issue #421, opened 2024-01-04
[Help Wanted] Problems with Mass Extinction Estimation Tutorial in RevBayes Development Version on Intel Mac


minghao-du:
Dear RevBayes Developers,

I have attempted to run the Mass Extinction Estimation tutorial from the RevBayes Tutorials but encountered an error message:
`Error: The age of tip 'Acynodon_adriaticus' in the initial tree is outside of specified bounds.`

When I replace the command in the tutorial
`taxa <- readTaxonData("data/crocs_taxa.txt", delim=TAB)`
with
`taxa <- T.taxa()`
the program runs correctly. For details, you can refer to my previous inquiry on the forum at https://groups.google.com/g/revbayes-users/c/shztH9ovxcs.

I compared the Age of Acynodon_adriaticus using both methods and found it to be 75.4715 in each case, so I am unsure why running `taxa <- readTaxonData("data/crocs_taxa.txt", delim=TAB)` results in an error.

I am using the development version of RevBayes, running it on an Intel version of Mac.

Thank you for your assistance.

Best,
Minghao

davidcerny:
Hi Minghao,

It was me who [added this warning](https://github.com/revbayes/revbayes/pull/386) to the development branch last October. It was supposed to have a rather narrow use case: if (1) you want to estimate a tree that includes fossils, (2) the ages of those fossils are read in from a taxon data file, and (3) you want to initialize your analysis with a user-specified tree, then the tip ages of that initial tree indeed have to be within the age ranges listed in the taxon file.

In the [Mass Extinction Estimation tutorial](https://revbayes.github.io/tutorials/divrate/efbdp_me.html), this is not really the case: the tree (`crocs_T1.tre`) is fixed (treated as data) rather than estimated, and the taxon data file (`crocs_taxa.txt`) lists point ages rather than ranges. Moreover, as you point out, these point ages are identical to the tip ages in the tree – and therefore redundant with respect to the latter. This part of the tutorial was intended to teach the users general best practices when it comes to handling fossils in RevBayes, but ironically, these are perhaps not the best in this particular situation. For now, you can feel free to go ahead and change

```
taxa <- readTaxonData("data/crocs_taxa.txt", delim=TAB)
```

to

```
taxa <- T.taxa()
```

as you were advised; it doesn't really change what the analysis is doing in any way. A more permanent solution would be to edit the tutorial itself, so that instead of stating:

[...]

PR #455, opened 2024-04-22
Optional precision in initial tree age checking

davidcerny:
This PR is meant to provide a more lasting solution to issue #421. I have added a new argument called `ageCheckPrecision` to `dnBDSTP`, which determines how many decimal places should be used when comparing the tip ages of a user-specified initial tree to the ages (given either as ranges or fixed values) provided in a taxon data file. By default, the argument is set to 4. I have verified that the problem described in #421 does not resurface unless it is set to a value of 8 or greater.

We already have a test for using `dnBDSTP` with an initial tree, so I haven't added a new one. This would be a nice place to implement the idea that @bjoelle brought up on Slack a while ago, i.e., to have a test that's supposed to fail – that would help verify that it does in fact matter what value `ageCheckPrecision` is set to. However, as far as I know, we haven't figured out how to do that.


bjoelle (approved):
looks good to me
maybe as a test we can add the script from #421 ?

src/core/distributions/phylogenetics/tree/birthdeath/BirthDeathSamplingTreatmentProcess.h
```diff
@@ -53,7 +53,8 @@ namespace RevBayesCore {
                                                     const std::string &cdt,
                                                     const std::vector<Taxon> &tn,
                                                     bool uo,
-                                                    Tree *t);  //!< Constructor
+                                                    Tree *t,
+                                                    const TypedDagNode<long> *age_check_precision);  //!< Constructor
```
hoehna, line 57: Why is this a DAG node? DAG nodes should only be used if the intention is that the variable might change during an analysis.

src/core/utils/TreeUtilities.h
```diff
@@ -62,7 +62,7 @@ namespace RevBayesCore {
         void                    rescaleTree(TopologyNode& n, double factor);                                                            //!< rescale tree ages below a node by a factor
         void                    setAges(TopologyNode& n, const std::vector<double>& ages);                                              //!< set ages of a node and children from a vector
         void                    setAgesRecursively(TopologyNode& n, double age);                                                        //!< set age of a node and rescale its children
-        Tree*                   startingTreeInitializer(Tree& treeToChange, std::vector<Taxon>& taxaToCopy);                            //!< make sure starting tree satisfies age constraints, and assign min/max ages to its tips
+        Tree*                   startingTreeInitializer(Tree& treeToChange, std::vector<Taxon>& taxaToCopy, const long& agePrecision);  //!< make sure starting tree satisfies age constraints, and assign min/max ages to its tips
```
hoehna, line 65: doesn't need to be a `const long&` and instead can be `long` because a constant reference takes up as much space as a long.

src/revlanguage/distributions/phylogenetics/tree/Dist_BDSTP.cpp
```diff
@@ -92,6 +92,13 @@ RevBayesCore::AbstractBirthDeathProcess* Dist_BDSTP::createDistribution( void )
     {
         init = static_cast<const TimeTree &>( initial_tree->getRevObject() ).getDagNode()->getValue().clone();
     }
+    
+    // number of decimal places to use when checking the initial tree against taxon ages
+    RevBayesCore::TypedDagNode<long>* pr = NULL;
+    if ( age_check_precision != nullptr )
+    {
+        pr = static_cast<const Natural &>( age_check_precision->getRevObject() ).getDagNode();
```
hoehna, line 100: Again, why as DAG node and not directly a value?

src/revlanguage/distributions/phylogenetics/tree/Dist_BDSTP.cpp
```diff
@@ -365,6 +373,7 @@ void Dist_BDSTP::addCommonRules(MemberRules& dist_member_rules) const
         dist_member_rules.push_back( new OptionRule( "condition", new RlString("time"), optionsCondition, "The condition of the process." ) );
         dist_member_rules.push_back( new ArgumentRule( "taxa"  , ModelVector<Taxon>::getClassTypeSpec(), "The taxa used for initialization.", ArgumentRule::BY_CONSTANT_REFERENCE, ArgumentRule::ANY ) );
         dist_member_rules.push_back( new ArgumentRule( "initialTree" , TimeTree::getClassTypeSpec() , "Instead of drawing a tree from the distribution, initialize distribution with this tree.", ArgumentRule::BY_VALUE, ArgumentRule::ANY, NULL ) );
+        dist_member_rules.push_back( new ArgumentRule( "ageCheckPrecision", Natural::getClassTypeSpec(), "If an initial tree is provided, how many decimal places should be used when checking its tip ages against a taxon file?", ArgumentRule::BY_CONSTANT_REFERENCE, ArgumentRule::ANY, new Natural(4) ) );
```
hoehna, line 376: I suggest to use `ArgumentRule::BY_VALUE`

hoehna (changes requested):

davidcerny:
Made the changes requested by @hoehna, added a test as suggested by @bjoelle, and as a bonus, made it possible to use `Phi` as an alias of `rho` in `dnFBDP`.

hoehna (approved):
Yes, that looks much safer and cleaner.

PR #460, opened 2024-04-26
Stairwayplot method

hoehna:
Here I added two main new features:
1. The stairway plot approach (https://www.nature.com/articles/ng.3254)
2. Leave one out cross validation (https://academic.oup.com/sysbio/article/72/3/616/7050004)


bredelings:
It looks like the tests here almost pass.  The only exception is:
```
>>>> Test failed: steppingstone
	mismatch: ml_Partition_uniform.errout
	mismatch: ml_Partition_uniform_one_stone_at_a_time.errout
```

src/core/analysis/mcmc/MarginalLikelihoodEstimator.cpp
```diff
@@ -44,7 +45,7 @@ MarginalLikelihoodEstimator::MarginalLikelihoodEstimator(const path &fn, const s
         // Initialize
         std::string commandLine;
     
-        //    RBOUT("Processing file \"" + fn + "\"");
+        RBOUT("Processing file \"" + fn.string() + "\"");
```
davidcerny, line 48: This line is what causes the stepping stone test not to pass – the actual values of marginal likelihoods stay the same. We can probably just regenerate the expected output.

davidcerny:
I'm still hoping to look into this in greater depth and submit a more detailed review. For now, the PR good to me as it provides some very cool new functionality, but I'd suggest adding integration tests to go with it. The stairway plot test could be based on the [corresponding tutorial](https://revbayes.github.io/tutorials/stairwayplot/), but it'd be great to also have a test for LOO-CV – what do you think, @hoehna?

bredelings (changes requested):
In addition to some tests, it looks like this needs help files for the next distributions, functions, moves, analyses, etc.

bredelings:
I've modified the tutorial so that it now says this will hopefully be included in 1.2.6

src/core/simulators/CoalescentSFSSimulator.cpp
```diff
@@ -0,0 +1,454 @@
...
+    std::ofstream out_stream_trees;
+    if ( write_trees )
+    {
+        createDirectoryForFile( file_name_trees );
+        out_stream_trees.open( file_name_trees.string(), std::fstream::out );
+    }
+    
+    RandomNumberGenerator* rng = GLOBAL_RNG;
+    
+#ifdef RB_MPI
```
bredelings, line 66: This looks like a hack ... what is the story here?

davidcerny: @bredelings I've replaced this by the solution which @bjoelle implemented in #662, and which was meant to solve the same issue – make sure that if we have multiple processes, they receive different random seeds.

src/core/analysis/mcmc/MarginalLikelihoodEstimator.cpp
davidcerny: When resolving the merge conflicts, I just decided to just leave the line commented out.

davidcerny:
We should consider making the following changes to the names of the new Rev elements added by this PR:

1. `BSPInterval` should be called `fnBSPInterval` for consistency with other functions 
2. `ConditionalPosteriorOrdinate` would fit the pattern of `mcmc`, `pathSampler`, `powerPosterior`, etc. better it it started with a lowercase "c" (but see `ValidationAnalysis`). But I'm not sure the name itself is quite correct. [Lewis et al. (2014)](https://academic.oup.com/sysbio/article-abstract/63/3/309/1648360) wrote of conditional _predictive_ ordinates, and [Lartillot (2023)](https://academic.oup.com/sysbio/article/72/3/616/7050004) used the term _cross-predictive_ ordinates.
3. `mvAdaptiveRJSwitch` could use a longer and more descriptive alias: `mvAdaptiveReversibleJumpSwitch` 

(1) and (3) are relatively non-invasive, so I went ahead and made the corresponding changes. For (2), I did not, since `ConditionalPosteriorOrdinate` is already in use in the [tutorial](https://revbayes.github.io/tutorials/stairwayplot/). We might want to simply add an alias or just leave it the way it is.

davidcerny:
Hmm, it looks like `fnBSPInterval` can deal with situations where `size(x)` > `size(n)` by simply ignoring the extraneous elements:

```
> a <- [0.1, 0.2, 0.3, 0.4]
> b <- [2, 1, 4]
> fnBSPInterval( a, b )
   [ 0.100, 0.100, 0.200, 0.300, 0.300, 0.300, 0.300 ]
```

but not with situations where `size(n)` > `size(x)`:

```
> a <- [0.1, 0.2, 0.3]
> b <- [2, 1, 4, 6]
> fnBSPInterval( a, b )
   Error:	Vector index out of range: 3 of 3
```

It might be worth checking that both vectors are of the same length.

src/revlanguage/moves/mixture/Move_AdaptiveReversibleJumpSwitch.cpp
```diff
@@ -0,0 +1,176 @@
...
+    {
+        move_member_rules.push_back( new ArgumentRule( "x", Real::getClassTypeSpec(), "The variable on which this move operates.", ArgumentRule::BY_REFERENCE, ArgumentRule::STOCHASTIC ) );
+        move_member_rules.push_back( new ArgumentRule( "waitBeforeLearning", Natural::getClassTypeSpec(), "Number of tries before learning.", ArgumentRule::BY_VALUE, ArgumentRule::ANY, new Natural(1000) ) );
+        move_member_rules.push_back( new ArgumentRule( "waitBeforeUsing", Natural::getClassTypeSpec(), "Number of tries before using.", ArgumentRule::BY_VALUE, ArgumentRule::ANY, new Natural(10000) ) );
+        move_member_rules.push_back( new ArgumentRule( "updateEvery", Natural::getClassTypeSpec(), "How frequent to update.", ArgumentRule::BY_VALUE, ArgumentRule::ANY, new Natural(1) ) );
+
+        /* Inherit weight from Move, put it after variable */
+        const MemberRules& inheritedRules = Move::getParameterRules();
+        move_member_rules.insert( move_member_rules.end(), inheritedRules.begin(), inheritedRules.end() );
+
```
davidcerny, line 123: Here, we are inheriting not just the `weight` argument, which we want, but also the `tuneTarget` argument, which I don't think is meaningful for `mvAdaptiveRJSwitch`. Can we get rid of it? At the very least, we should point out in the help file that it has no effect, but not including it in the first place would be much more elegant.

help/md/mvRandomCategoryWalk.md
```diff
@@ -0,0 +1,20 @@
+## name
+mvRandomCategoryWalk
+## title
+Random walk on vector of naturals (e.g., from multinomial distribution) keeping the sum.
```
bredelings: maybe "keeping the sum constant."?

src/core/distributions/phylogenetics/tree/coalescent/demography/LinearDemographicFunction.cpp
```diff
@@ -165,8 +165,20 @@ double LinearDemographicFunction::getWaitingTime(double time, double lambda) con
...
     {
         double alpha = ( N1-N0 ) / (t1 - t0);
-        return (N0 + (time-t0) * alpha) * lambda;
+        // we compute the value for which the waiting time is longer than this demographic function (outside its range)
+        return ( exp(alpha*lambda + log(N0 + alpha * t0)) - N0 )  /  alpha;
+//        double max_lambda = t1 / (N0 + (time-t0) * alpha);
+//        if ( max_lambda < lambda )
+//        {
+//            // simply return a waiting time that is just outside this demographic function range
+//            return t1-time+1.0;
```
bredelings: This commented out value matches more closely what ExponentialDemographicFunction is doing.

Should these comments be removed?

help/md/dnStairwayPlot.md
```diff
@@ -0,0 +1,46 @@
...
+	# obtain the number of individuals and the number of sites
+	N_IND   = abs(obs_sfs.size()-1)
+	N_SITES = round(sum(obs_sfs))
+
+	# now specify a different theta per interval
+	for (i in 1:(N_IND-1)) {
+	  theta[i] ~ dnLognormal( ln_theta_mean, theta_sd )
+	}
+
+	sfs ~ dnStairwayPlot( theta, numSites=N_SITES, numIndividuals=N_IND, folded=USE_FOLDED, monomorphicProbability="rest", coding="all" )
```
bredelings: Here you seem to supply an unfolded spectrum... but possibly dnStairwayPlot allows `folded=true` even for unfolded spectra?  In that case do you simply fold the spectrum and compute the likelihood of the folded version?

Also, can you simulate from this?  Or is it only used for computing likelihoods?

src/core/moves/proposal/mixture/AdaptiveReversibleJumpProposal.cpp
```diff
@@ -0,0 +1,343 @@
...
+    variable( n ),
+    stored_value( 0 ),
+    stored_index( 0 ),
+    wait_before_learning( n0 ),
+    wait_before_using ( c0 ),
+    updates_every( ue ),
+    num_tried ( 0 ),
+    proposal_distribution( NORMAL )
+{
+    if (wait_before_using < wait_before_learning)
```
bredelings, line 30: What is the argument here that this adaptation doesn't affect the equilibrium distribution?

src/core/moves/proposal/mixture/AdaptiveReversibleJumpProposal.h
```diff
@@ -0,0 +1,67 @@
...
+#include <set>
+#include <string>
+
+#include "Proposal.h"
+#include "StochasticNode.h"
+
+namespace RevBayesCore {
+    
+    /**
+     * The adaptive reversible jump proposal to switch between two elements of an RJ-Mixture.
```
bredelings: Since this is an adaptive move, what's the argument that the adaptation doesn't modify the equilibrium?

help/md/mvRandomCategoryWalk.md
hoehna: Sure.

src/core/distributions/phylogenetics/tree/coalescent/demography/LinearDemographicFunction.cpp
hoehna: No, I don't think so. In the constant model we also simply extrapolate. The exception is the exponential function. Which can easily lead to degenerate distributions. Importantly, we just need to pass a value that is outside the range. Then the calling function must know and handle this.

help/md/dnStairwayPlot.md
hoehna: You should be able to simulate. I added a direct simulator.

Well, it's true that you can mismatch the size of the value to what is expected.

src/core/moves/proposal/mixture/AdaptiveReversibleJumpProposal.cpp
hoehna: In the RJ move we usually propose values from the prior. This could be bad (performance wise). So instead here we want to "learn" the distribution for proposing. However, you might only actually want to start learning after some initial MCMC steps are done to remove this "burnin".

src/core/moves/proposal/mixture/AdaptiveReversibleJumpProposal.h
hoehna: The two arguments that are updated are `sampled_mean` and `sampled_variance`. We are going to propose based on these values. These values "should" stabilize with more samples. This is the same as in the mvAVMVN.

help/md/dnStairwayPlot.md
bredelings: Every time I try simulating from dnStairwayPlot( ), I get something like this:
```
> sfs ~ dnStairwayPlot( theta, numSites=N_SITES, numIndividuals=N_IND, folded=USE_FOLDED, monomorphicProbability="treelength" )
> sfs
   [ 1.000, 1.000, 1.000, 1.000, 1.000, 1.000, 1.000, 1.000, 1.000, 1.000, 1.000, 1.000, 1.000, 1.000, 1.000, 1.000, 1.000, 1.000, 1.000, 1.000, 1.000 ]
```
Am I missing something?

src/core/moves/proposal/mixture/AdaptiveReversibleJumpProposal.h
bredelings: Ah, OK.  Can you insert a brief note that says something like this?

help/md/dnStairwayPlot.md
davidcerny: So @hoehna actually added a proper `simulator()` method to the distribution to fix this issue, but there was a type mismatch related to drawing from the multinomial distribution that prevented it from compiling. Since it was necessary for people to be able to use this branch (#980), I removed the commit in question and force-pushed. Sebastian is planning to look into the issue asap.

davidcerny: Now fixed: Sebastian and I have refactored `redrawValue()`, and made sure to redraw after initialization. So if you examine a draw from `dnStairwayPlot` before clamping it, you will no longer see a vector of 1's.

src/core/moves/proposal/mixture/AdaptiveReversibleJumpProposal.h
davidcerny: Done.

bredelings:
BTW, I see the dnStairwayPlot example actually runs now, nice!  I wonder if it should have `coding="no-monomorphic"`, since we are setting `obs_sfs[obs_sfs.size()] <- 0`

OK, the last main concern that I had here was that if you set numIndividuals=40 and folded=TRUE, then
* simulation produces a vector of length 21 (as expected)
* but you can clamp a vector of length 41

Clearly the probability of simulating a vector of length 41 is zero.  It seems like it would be more consistent to have a `foldSFS( )` function and then do something like `sfs.clamp( foldSFS(unfolded_sfs) )`.  And if you try to clamp an SFS of the wrong size, there should be an error message.

If we do `sfs.clamp( [0,1,2,3,4,,5,6,7] )` then we get `Error:	Vector index out of range: 8 of 8`.  And if we clamp a folded sfs of size 21.  But it also works if we clamp a vector of length 22, 23, etc.  Length 41 works, but also 42, 43, 44, ...

davidcerny:
> I wonder if it should have `coding="no-monomorphic"`, since we are setting `obs_sfs[obs_sfs.size()] <- 0`

I think it's more likely that the example accidentally omitted a few lines that should have been there (based on what we see in the [test](https://github.com/revbayes/revbayes/blob/dev_stairwayplot/tests/test_StairwayPlot/scripts/mcmc_StairwayPlot_BSP.Rev) [scripts](https://github.com/revbayes/revbayes/blob/dev_stairwayplot/tests/test_StairwayPlot/scripts/mcmc_StairwayPlot_RJ.Rev)). My understanding is that in the original `obs_sfs` vector, the first element represents the number of monomorphic sites at which all the sampled alleles carry the ancestral state, and the last element represents the number of monomorphic sites at which all the alleles carry the derived state. These four lines taken together

```r
TOTAL_NUM_SITES <- sum(obs_sfs)
obs_sfs[1] <- 0
obs_sfs[obs_sfs.size()] <- 0
obs_sfs[1] <- TOTAL_NUM_SITES - sum(obs_sfs)
```

merge those two monomorphic categories into a single entry, since they are indistinguishable to the model. But they don't get rid of the monomorphic category as such.

> OK, the last main concern that I had here was that if you set numIndividuals=40 and folded=TRUE, then
> - simulation produces a vector of length 21 (as expected)
> - but you can clamp a vector of length 41
>
> Clearly the probability of simulating a vector of length 41 is zero. 

That's a great point, thanks! It's actually a mistake I made in the example. The original example provided by Sebastian used an undefined variable (`folded=USE_FOLDED`). I tried to make it self-contained, so I replaced it by `folded=TRUE`. It should have actually been `folded=FALSE`. In order to use `folded=TRUE`, we would have had to introduce an extra bit of data-handling code, which you can actually see in the test scripts:

```r
if (USE_FOLDED == TRUE) {
    NUM_IND = abs(obs_sfs.size() - 1)
    tmp_sfs[1] <- obs_sfs[1] + obs_sfs[NUM_IND+1]
    for (i in 1:ceil(NUM_IND/2)) {
        if (i != NUM_IND/2) {
            tmp_sfs[i+1] <- obs_sfs[i+1] + obs_sfs[NUM_IND-i+1]
	    } else {
            tmp_sfs[i+1] <- obs_sfs[i+1]
	    }
    }
    obs_sfs <- tmp_sfs
}
```

However, both of your points still stand: there should be a length check (which would expose my mistake), and we could provide a `foldSFS()` function as a more user-friendly alternative to handling the folding at the Rev level.

Issue #468, opened 2024-05-12
[BUG] Crash when applying the [] operator to MatrixBoolean


davidcerny:
**Describe the bug**
As pointed out by @bredelings, the `[]` operator as defined in `MatrixBoolean.cpp` returns `boost::dynamic_bitset<>`, whereas the corresponding member function expects `RbVector<Boolean>`. As a result, an attempt to subset a `MatrixBoolean` object using this operator causes RevBayes to crash:

```
libc++abi: terminating due to uncaught exception of type std::bad_cast: std::bad_cast
zsh: abort      ./rb
```

**To reproduce**
Consider the following script, in which an `AverageDistanceMatrix` is created by averaging the path-length distance matrices of two trees with partially overlapping leaf sets. Since neither of the two trees contains both `A` and `E`, the distance between these two taxa is not meaningful, and should not be used in any further comparisons or calculations. (The same is true of the trivial distances along the diagonal.) This is indicated by the fact that the corresponding entries of the Boolean mask (a `MatrixBoolean` object) are set to `FALSE`. We can access this Boolean mask using `ADM.mask()`, but for the reasons described above, we cannot subset it:

```
speciation <- 0.1
extinction <- 0.09
tree_height <- 1.0

taxa1 <- [taxon("A"), taxon("B"), taxon("C"), taxon("D")]
taxa2 <- [taxon("B"), taxon("C"), taxon("D"), taxon("E")]

tree1 ~ dnBDP(lambda=speciation, mu=extinction, rootAge=tree_height, condition="nTaxa", taxa=taxa1)
tree2 ~ dnBDP(lambda=speciation, mu=extinction, rootAge=tree_height, condition="nTaxa", taxa=taxa2)

distMatrices[1] := fnTreePairwiseDistances(tree1)
distMatrices[2] := fnTreePairwiseDistances(tree2)

ADM := fnAverageDistanceMatrix(distMatrices)
ADM.mask()[1]
```

**Expected behavior**
The code above should return the following:

```
[ FALSE, FALSE, TRUE, TRUE, TRUE ]
```

**Computer info**
macOS 14.4.1 (Sonoma), ARM64 architecture, RevBayes 1.2.3, latest version of the `development` branch (rapture-2661-g8bcc50).

PR #469, opened 2024-05-12
Improvements to MatrixBoolean and AverageDistanceMatrix

davidcerny:
This PR is primarily intended to fix issue #468. Specifically, it makes the following three changes:

1. The type of the `rv` argument of one of the `executeMethod(...)` utility functions of class `MatrixBoolean` has been changed from `boost::dynamic_bitset<>` to `RbVector<Boolean>`.
2. The `.getElement()` method of class `AverageDistanceMatrix` returns a vector that no longer contains two doubles but rather a single string. This makes for nicer printing: instead of `[ 1.905, 1.000 ]`, we now get `[ 1.90471, TRUE ]`, which better indicates what's going on (i.e., each element of `AverageDistanceMatrix` is a numeric value associated with a Boolean that indicates whether or not this value should be used in further calculations).
3. A test (based on the script from #468) has been added to showcase these changes.

Interestingly, change (1) results in a very slight but consistent drop in the performance of `ExponentialError::lnPdf` (by about ~50 ms for 10,000-by-10,000 matrices), which I previously did my best to optimize (see #432, #434). It's not clear to me why: that function does subset a `MatrixBoolean` object at one point (see [here](https://github.com/revbayes/revbayes/blob/8bcc503a10936e8ef1c3ee095c9c31b0ee35a8bf/src/core/math/distributions/DistributionExponentialError.cpp#L240)), but it doesn't do so at the Rev level, i.e., it doesn't employ the `executeMethod` utility function to do it. Instead, it uses a lower-level operator which still returns `boost::dynamic_bitset<>`. However, the time difference is small enough that in any actual analysis of average distance matrices, it seems to be absorbed by the stochastic variation in the time it takes to convert a tree to a distance matrix – overall, there is no clear performance hit.

### DIFF
--- a/src/core/datatypes/math/MatrixBoolean.cpp
+++ b/src/core/datatypes/math/MatrixBoolean.cpp
@@ -150,7 +150,7 @@ void MatrixBoolean::executeMethod(const std::string &n, const std::vector<const 
         size_t index = (size_t)static_cast<const TypedDagNode<long> *>( args[0] )->getValue() - 1;
         boost::dynamic_bitset<> tmp = elements[index];
         
-        RbVector<Boolean> rv;
+        RbVector<Boolean> rv( tmp.size() );
         for (size_t i = 0; i < tmp.size(); i++)
         {
             rv[i] = (Boolean)tmp[i];
@@ -161,7 +161,7 @@ void MatrixBoolean::executeMethod(const std::string &n, const std::vector<const 
     {
         boost::dynamic_bitset<> tmp = this->getUpperTriangle();
         
-        RbVector<Boolean> rv;
+        RbVector<Boolean> rv( tmp.size() );
         for (size_t i = 0; i < tmp.size(); i++)
         {
             rv[i] = (Boolean)tmp[i];
--- a/src/core/datatypes/math/MatrixBoolean.cpp
+++ b/src/core/datatypes/math/MatrixBoolean.cpp
@@ -13,6 +13,7 @@
 #include <vector>
 
 #include "MatrixBoolean.h"
+#include "RbBoolean.h"
 #include "RbException.h"
 #include "RbVector.h"
 #include "RbConstants.h"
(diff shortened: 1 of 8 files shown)


davidcerny:
Previously, the tests were not passing because no particular order was specified for the rows and columns of the average distance matrix (ADM), and as it turns out, the "default" order was different on different CPU architectures. I changed the code so that the taxon labels associated with every row and column are alphabetically sorted.

One advantage of this solution is that it allows a further speed-up in `ExponentialError::lnPdf`. Instead of always sorting both the ADM and the distance matrix of a proposed tree in every single MCMC iteration, we can exploit the fact that in any given analysis, the ADM is constant, and thus only needs to be sorted once. Since we now do it upon its creation, the `lnPdf` function only needs to calculate one vector of indices instead of two.

bredelings (approved):
Hi David, this looks good as is.

A thought for the future: if you are concerned about speed you could add a member function like `M.element(i,j)` that just checks one bit.  Unlike M[i][j], this doesn't need to create an M[i] object each time, which should make it faster.

Unlike R, there is on special syntax for this.  However, we could in theory change the grammar to allow operator `[]` to have multiple arguments, as in `M[i,j]`.  However, we might want `M[i]` and `M[i,j]` to call different functions, and I'm not quite sure how to do that.

In any case, this PR is fine as is.

davidcerny:
Hi Ben,

Just to make sure I'm not misunderstanding, would this apply to the Rev level, the C++ level, or both? I'm not too concerned about how long it takes to access a matrix element in a Rev script, but I would definitely like these operations to be optimized for performance in internal functions.

For what it's worth, I've tried defining

```
bool MatrixBoolean::element(size_t i, size_t j) const
{
    return elements[i][j];
}
```

and

```
const double& DistanceMatrix::element( size_t i, size_t j ) const
{
    return matrix[i][j];
}
```

and then I changed `ExponentialError::lnPdf` so that it uses

```
if (z.getMask().element(i, j))
{
    double difference = z.getDistanceMatrix().element(i, j) - distMat.element(k, l);
    dist += difference * difference;
}
```

instead of the current

```
if (z.getMask()[k][l])
{
    double difference = z.getDistanceMatrix()[i][j] - distMat[k][l];
    dist += difference * difference;
}
```

As far as I can tell, this results in no measurable changes in performance even when subtracting 10,000-by-10,000 matrices. But again, perhaps I misunderstood?

bredelings:
Hi David,

I meant only on the Rev level.  On the C++ level `[i][j]` isn't creating a new `RbVector<Boolean>` so it should be fast.

-BenRI

davidcerny:
That makes sense – thank you for the clarification! :)

PR #479, opened 2024-06-13
Ensure that we can check `args.size()` even when no arguments are given.

bredelings:
This is one way to set `args` equal to an empty `String[]` when no arguments are given.


davidcerny (approved):
I liked the idea of creating something like `VectorArgs()` that would be analogous to `VectorMoves()` and `VectorMonitors()`, but this is simpler and works too. Approving.

PR #480, opened 2024-06-23
Fix issue #354 once again

davidcerny:
This PR restores the starting tree code in `BirthDeathSamplingTreatmentProcess.cpp`, which was introduced there in PR #386 to fix issue #354, but which was then (accidentally, it seems) removed in PR #399. As a result, it's once again possible to sample fossil ages from their respective priors even when a user-supplied starting tree is employed, and the starting tree is checked for compatibility with fossil ages.


hoehna (approved):

dwbapst:
Is it possible there are any other accidental code reversions resulting from PR #399  ? Has this been checked?

Issue #477, opened 2024-06-04
[BUG] Unable to compute probability for fbd_tree in v. 1.2.2


ZGriebenow:
I am attempting to run an FBD analysis in v. 1.2.2 (docker) on Ubuntu Linux 20.04, and am encountering an error in which a starting state with a computable probability for a given node cannot be found after 1000 tries.

This occurs for the fbd-tree variable when I use a backbone topology. Conversely, when a backbone topology is not implemented, the probability of the starting state is not computable with respect to the uniform random variable that accounts for fossil stratigraphic range. Since v. 1.2.3 is not yet available on Docker with an Ubuntu-compatible architecture, I have not yet assessed whether this bug occurs in that version.

I have successfully run mcmc_CEFBDP_Specimens.Rev under the same conditions, but consultation with another user and repeated mention of this same error on the RevBayes Google group suggests that this is indeed a difficult-to-replicate bug, rather than the result of misspecifications within my data or script. Hopefully, I am correct in that supposition.

Regards, Zach

[bug-report_122_docker.zip](https://github.com/user-attachments/files/15539482/bug-report_122_docker.zip)

davidcerny:
Based on my own experience, my guess would indeed be that this is due to one of the FBD-related issues that got fixed in v1.2.3 – in some ways, the v1.2.2 release really was uniquely bad when it comes to FBD. (See in particular issues #159, #354, and their respective fixes in PRs #453, #386, both of which occurred between v1.2.2 and v1.2.3.) Often, these MCMC initialization problems are most easily overcome by supplying your own starting tree, but in v1.2.2, it was impossible to have a starting tree and stratigraphic ranges at the same time.

Is there any chance you might be able to run v1.2.3 on Ubuntu outside of Docker?

davidcerny:
@ZGriebenow Unfortunately, I'm unable to reproduce this, because the zipped folder does not contain the `clade_constraints` directory that is referenced in both of your scripts. Could you please provide the constraints as well, just to check if this issue still occurs in the latest release (v1.2.4)?

ZGriebenow:
@davidcerny My apologies; I am attaching the folder containing the necessary text files (and many others).
[clade_constraints.zip](https://github.com/user-attachments/files/15877461/clade_constraints.zip)

davidcerny:
Hi @ZGriebenow, apologies for the late reply. I finally got a chance to play with your dataset. There's a lot going on.

First, some problems that are unrelated to potential bugs:

[...]

PR #481, opened 2024-06-24
Sampling fossil ages with dnConstrainedTopology + initial tree

davidcerny:
This PR is intended to fix issue #477, which turned out to be essentially the same as issue #354 except that it affected `dnConstrainedTopology` rather than `dnBDSTP`: supplying a user-specified initial tree overwrote the ages of fossil taxa (min = max = 0), making it impossible to sample them from the [min, max] intervals specified in a taxon info file. I have added extra code to the `TopologyConstrainedTreeDistribution` constructor that checks whether a starting tree is present and whether it contains fossils. If both of these conditions are satisfied, it runs `TreeUtilities::startingTreeInitializer()` on the initial tree to replace each of its tips with the previously specified taxon of the same name. As with `dnBDSTP`, a new `ageCheckPrecision` argument has been added to specify how strict we should be when assessing whether the ages of the fossils in the initial tree are compatible with the age ranges specified in the taxon info file.

I have verified that the code works as intended, and that it doesn't affect other uses of `dnConstrainedTopology` (e.g., when it is used for non-clock trees).

### DIFF
--- a/src/core/distributions/phylogenetics/tree/TopologyConstrainedTreeDistribution.cpp
+++ b/src/core/distributions/phylogenetics/tree/TopologyConstrainedTreeDistribution.cpp
@@ -27,6 +27,7 @@
 #include "Tree.h"
 #include "TreeChangeEventHandler.h"
 #include "TreeChangeEventMessage.h"
+#include "TreeUtilities.h"
 #include "TypedDagNode.h"
 #include "TypedDistribution.h"
 
@@ -44,7 +45,10 @@ using namespace RevBayesCore;
  *
  * \param[in]    c         Clade constraints.
  */
-TopologyConstrainedTreeDistribution::TopologyConstrainedTreeDistribution(TypedDistribution<Tree>* base_dist, const std::vector<Clade> &c, Tree *t) : TypedDistribution<Tree>( NULL ),
+TopologyConstrainedTreeDistribution::TopologyConstrainedTreeDistribution(TypedDistribution<Tree>* base_dist,
+                                                                          const std::vector<Clade> &c,
+                                                                          Tree *t,
+                                                                          long age_check_precision = 4) : TypedDistribution<Tree>( NULL ),
 //    active_backbone_clades( base_dist->getValue().getNumberOfInteriorNodes(), RbBitSet() ),
     active_clades( base_dist->getValue().getNumberOfInteriorNodes(), RbBitSet() ),
     backbone_topology(NULL),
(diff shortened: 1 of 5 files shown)


bjoelle:
@davidcerny I'm not sure about putting special handling code into the topology class, especially things that are directly duplicated from another distribution - if we start having more distributions which need to do things their own way it will quickly become messy
isn't there a way to delegate the handling of the initial tree directly to `base_distribution` ?

davidcerny:
@bjoelle Hmm, I guess to me this solution didn't seem that bad, because the code that actually does the work is packaged into a separate function (`TreeUtilities::startingTreeInitializer()`), and what's duplicated between distributions is just a function call plus some basic exception handling.

For the particular case from #477, i.e., `dnConstrainedTopology` + `dnBDSTP`, the base distribution _does_ handle the initial tree already, but it doesn't help, because `dnConstrainedTopology` goes on to ignore whatever initial tree the base distribution used (or modified). So maybe a better way to go about things would be to allow `dnConstrainedTopology` to pull the initial tree from its base distribution and set it as its own initial tree. But that would also require some amount of special handling, because not all base distributions (AFAIK) support user-specified initial trees.

bjoelle:
I guess I just feel like the `dnConstrainedTopology` should not be the one knowing/caring about fossil special handling ?
is there a way to create a `setInitialTree` function in tree distributions which gets called as `base_distribution -> setInitialTree` and then the BDSTP can do the special handling there ?
maybe this gets too complicated and I'm worrying about nothing...

bredelings:
I kind of like the structure that Joelle suggested, because it seems like tree distributions could apply arbitrary properties to their trees, and only the distribution would know what those properties are.  In this case, the property is taxon ages, but other tree distributions could in theory add other things.

From another perspective, maybe the problem is that taxon names and taxon properties are confused by having a Taxon object?  That is, it isn't possible to specify just the name, since a Taxon object contains both.  A tree should just specify the taxon _name_, and taxon properties should really be a different thing. 

At an even higher level, the fact that there is such high demand for initial trees really indicates that we should get around to fixing the problem where an initial tree is needed to start the MCMC... but that can't be fixed here.

bredelings:
Maybe we should merge this as is, but add a TODO or FIXME note to reimplement it better later?  That way the best does not become the enemy of the good.

davidcerny:
Sorry for the delay in replying; it's been a busy few days.

I was actually going to ask how we should proceed. I don't insist on having this merged if it doesn't look like a good idea, but since the issue is real and fixable, we should fix it in some other way instead. I agree 100% it would be more satisfying to solve it at a higher level of abstraction, possibly by circumventing the need for a user-specified initial tree altogether. But the progress on making this happen has been very slow (I distinctly remember us having the exact same discussion at the SSB 2023 meeting a year and a half ago!), and people should be able to run their analyses even before these more fundamental solutions have finally been implemented.

If I understand @bjoelle's idea correctly, a member function called `setInitialTree` would be added to those distributions that currently accept an `initialTree` argument (which, aside from `dnConstrainedTopology` itself, would seem to be just `dnBDSTP` and its derived classes, `dnOBDP`, and `dnEBDP`, though `dnUniformSerialSampledTimeTree` could arguably also benefit from it). When the `initialTree` argument isn't `NULL`, this function would be called by the distribution's constructor, but it could also be called from outside the distribution by `dnConstrainedTopology`. There would have to be some way for `dnConstrainedTopology` to check whether its `base_distribution` does in fact contain the `setInitialTree` member function. (Alternatively, we could add an `initialTree` argument and a `setInitialTree`member function to _all_ tree distributions, but that would be a pretty far-reaching change, especially if we eventually want to eliminate the need for initial trees altogether.)

Does that sound right, and is that what I should aim for instead?

davidcerny:
> Alternatively, we could add an initialTree argument and a setInitialTreemember function to all tree distributions, but that would be a pretty far-reaching change

Hmm... it sounds like @mlandis's suggestion of including an `initValue` argument as deep as `TypedDistribution<T>` and letting inheritance take care of things would make this a lot easier, if we want to go down that route.

bredelings (approved):

Issue #483, opened 2024-06-26
[BUG] MCMC Stops After 200 Iterations When Using Checkpoint with FBD Model and Mass Extinction in version 1.2.4


minghao-du:
**Describe the bug**
When performing phylogenetic inference using an FBD model with mass extinction, after using a checkpoint, the MCMC only continues for 200 iterations and then stops.

**To Reproduce**
The data and code I used are in this file. Simply run the _scripts/FBD_Model.Rev_ file.
[FBD_Checkpoint.zip](https://github.com/user-attachments/files/15983631/FBD_Checkpoint.zip)

I am attempting to conduct phylogenetic inference using both molecular and morphological data (total evidence dating). While using the FBD model as the tree prior, I also included mass extinction. I fixed the mass extinction time at 30 but inferred the probability of mass extinction.

**Expected behavior**
After using the checkpoint to continue the analysis, it should run normally instead of stopping after 200 iterations.

**Screenshots**
<img width="468" alt="image" src="https://github.com/revbayes/revbayes/assets/51898298/933d0060-cd6a-414d-9606-2197f0e50dbc">

When continuing the analysis using the checkpoint, it stops at this screen and does not proceed with further iterations.

**Computer info**
I am using a Mac, and the version of RevBayes I am using is development (rapture-2776-g62de34).

**Additional context**
1. This issue seems to occur only in version 1.2.4. When I used version 1.2.2 (development, rapture-2545-g7c2b56), this issue did not occur.
2. On one occasion, after continuing the run, the Posterior and Likelihood values for iterations 1100 and 1200 were -inf, but I could not reproduce this result. Is it possible that the program is still running, but abnormal Posterior and Likelihood values are causing each iteration to take a very long time?

bjoelle:
so the problem appears to be that the move `FossilTipTimesSlide` gets stuck because the fossil age ranges are not restored properly when initializing from checkpoint
I don't understand why this didn't happen in 1.2.2 but it's possible that it was masked by another bug

bjoelle:
should be resolved by PR #484

Issue #482, opened 2024-06-25
[BUG] Significant Fossil Age Estimation Errors When Using Checkpoint in FBD Model Analysis


minghao-du:
**Describe the bug**
When using the FBD model for phylogenetic inference, continuing the analysis with Checkpoint leads to significant changes in the inferred fossil ages, even erroneous changes.

**To Reproduce**
In the tutorial "Combined-Evidence Analysis and the Fossilized Birth-Death Process for Analysis of Extant Taxa and Fossil Specimens," simply add Checkpoint to the script.

All scripts and output are in this file.
[revbayes_fbd_fbd_specimen_checkpoint.zip](https://github.com/user-attachments/files/15970289/revbayes_fbd_fbd_specimen_checkpoint.zip)

**Expected behavior**
Continuing the previous analysis using Checkpoint should not produce anomalous results.

**Screenshots**
<img width="1177" alt="image" src="https://github.com/revbayes/revbayes/assets/51898298/ada528ac-28fd-4233-aa73-08ec97298acd">
This is a computational result: I ran for 5000 iterations, then used Checkpoint to run another 5000 iterations, and then used Checkpoint again at 10000 iterations to run another 5000 iterations. It seems that every time Checkpoint is used, the fossil ages change significantly, with the age estimates even being 0 between 5000-10000 iterations. All fossil ages have this problem.

**Computer info**
I am using a Mac, and the version of RevBayes I am using is development (rapture-2776-g62de34).

**Additional context**
Perhaps checking the tree log files before and after using Checkpoint might help identify where the problem occurs.

bjoelle:
should be fixed by PR #484

PR #484, opened 2024-06-27
Restarting FBD trees properly from checkpoint

bjoelle:
This PR is designed to fix issue #483 by restoring fossil age ranges correctly when restarting an FBD analysis from checkpoint, using a similar method as what David did for initial trees

Bonus fix: fixed inconsistent orderings of the taxon map so the MRCA references still work properly


bjoelle:
Update: I've also added sorting to the vector returned by `Tree::getFossilTaxa` which hopefully should solve #482 as well

davidcerny (commented):
Hmm... based on @minghao-du's test cases, this unfortunately seems to fix neither #482 nor #483 – at least not fully.

For issue #482, the tip ages no longer have staircase-like traces, but they are not successfully re-initialized either. Once the analysis has been restarted, their traces just remain completely flat, with no moves ever getting accepted:

Screenshot 2024-07-09 at 11 08 18

(Incidentally, I seem to remember us running into this exact same problem before, but I can't find the corresponding issue in the repo. Nor do I remember if it's ever been successfully resolved.)

For issue #483, instead of the re-initialized analysis getting stuck at iteration 1200, I now can't re-initialize it all. Instead, I'm running into the following error:

```
> restart = TRUE
> source("/Users/David/Downloads/issue483/scripts/FBD_Model.Rev")
   Processing file "/Users/David/Downloads/issue483/scripts/FBD_Model.Rev"
   Successfully read one character matrix from file 'data/mol_1.nex'
   Successfully read one character matrix from file 'data/morph_1.nex'
   Processing file "scripts/model_ME.Rev"
   Processing of file "scripts/model_ME.Rev" completed
   Processing file "scripts/model_FBD.Rev"
   Processing of file "scripts/model_FBD.Rev" completed
   Processing file "scripts/model_JC.Rev"
   Processing of file "scripts/model_JC.Rev" completed
   Processing file "scripts/model_Morph.Rev"
   WARNING: There are 1 characters incompatible with the specified coding bias. These characters will be excluded.
   Processing of file "scripts/model_Morph.Rev" completed
   Processing file "scripts/model_MCMC.Rev"
   Error:	Tip names of the initial tree do not match the taxon names.
   Error:	Problem processing line 16 in file "scripts/model_MCMC.Rev"
   Error:	Problem processing line 48 in file "/Users/David/Downloads/issue483/scripts/FBD_Model.Rev"
```

bjoelle:
@davidcerny thanks for testing this - I will look into this further !

davidcerny:
Ah, you're 100% right – that's actually a problem with my `TreeUtilities::startingTreeInitializer()` function. Still worth fixing for sure, and it's weird to me that it only resurfaces when restarting the analysis rather than right from the outset, but it's not something that should prevent us from going ahead with this PR.

Would you like me to approve now, or do you want to give a try to fixing the issue with stuck tip ages first?

bjoelle:
@davidcerny I don't have the time to work on this before next week, so I propose that I revert the last 2 commits related to fossil ordering, and then we can merge this as a fix for 483 ? thanks !

bjoelle:
actually, I get likelihood=-inf when restarting (which is why the ages don't move anymore - actually nothing moves) but that also happens on `development` so it seems unrelated to my changes

davidcerny (approved):

PR #487, opened 2024-07-09
Update dnPhyloCTMC.md

### LINKED ISSUE #486: `dnPhyloCTMC` documentation
I'm new to RevBayes, and have been trying to get my head around how I can interact with `dnPhyloCTMC()`.  I've gleaned some aspects of its behaviour from the tutorial pages, but the entry in the [reference pages](https://revbayes.github.io/documentation/dnPhyloCTMC.html) is somewhat minimal.

- One thing I didn't see listed there was a `.clamp()` method. Perhaps `.clamp()` is conceptually an operator (I've seen it listed in that category elsewhere), but its method-like syntax meant I expected to see it on the reference page, and there might be value in stating its supported arguments there.  I couldn't find a separate reference page for `.clamp()`, so it's hard to infer its behaviour.
- Another thing that would be helpful to have documented is the 'return value'.  I suspect that I'm thinking of this in the wrong way, and the "domain type" gives the answer to the question I'm trying to ask, but I don't get from the function page or the [domain page](https://revbayes.github.io/documentation/AbstractHomologousDiscreteCharacterData.html) a clear picture of what exactly is returned, or what the pathway is to get from the distribution to the likelihood of the clamped data.

Would it be possible for the documentation to be expanded with some pointers on those topics?

ms609:
This PR adds some basic documentation to dnPhyloCTMC, incorporating the very helpful comments from @davidcerny at #486.

One outstanding question is whether the dummy single-iteration MCMC analysis is helpful to include in the analysis, given the existence of a `.lnProbability()` method and the potential for confusion where `tree` features in multiple `dnPhyloCTMC` objects.

This will close #486.

### DIFF
--- a/help/md/dnPhyloCTMC.md
+++ b/help/md/dnPhyloCTMC.md
@@ -1,9 +1,218 @@
 ## name
-dnPhyloCTMC
+`dnPhyloCTMC`: Distribution of a phylogenetic continuous-time Markov chain
+
 ## title
+The parameters of a phylogenetic model – a tree topology with branch lengths, a substitution model that describes how observations evolve over the tree, etc. – collectively form a distribution called the _phylogenetic continuous-time Markov chain_.
+
 ## description
+dnPhyloCTMC gives the probability distribution of tip labels on a phylogenetic tree given an phylogenetic continuous-time Markov chain model.
+
 ## details
+
+The likelihood of observed tip labels (specified via a clamped `AbstractHomologousDiscreteCharacterData` object) is computed using Felsenstein's pruning algorithm, with partial likelihoods stored for each branch of the tree. It is automatically outputted in the `Likelihood` column of the `mnFile()` and `mnScreen()` monitors (which can be suppressed with `likelihood = FALSE`).
+
 ## authors
 ## see_also
+- Tutorial on [graphical models](https://revbayes.github.io/tutorials/intro/graph_models)
+
+- Tutorial on [specifying a phylogenetic continuous-time Markov chain](https://revbayes.github.io/tutorials/ctmc/) model
+
 ## example
+
+```rb
+# Create stochastic node with the tip distribution given by the topology `tree` with the 
+# rate matrix parameter q_matrix
+x ~ dnPhyloCTMC(tree = tree, Q = q_matrix)
+
+# Clamp observed characters to the node
+chars <- readDiscreteCharacterData("myData.nex")
+x.clamp(chars)
+```
+
+Simpler distributions contain methods to compute the probability of a particular value
+```rb
+x ~ dnNormal(0.0, 1.0)
+x.clamp(0.42)
+x.lnProbability()
+```
+
+The likelihood of observed characters given fixed parameter values (tree with branch lengths, site rates, etc.)
+can be computed by running a dummy single-iteration MCMC analysis:
+
+```rb
+mymodel = model(tree) # will incoroporate all phyloCTMC objects with `tree` as a parameter
+monitors = VectorMonitors()
+monitors.append( mnScreen(printgen=1) )
+mymcmc = mcmc(mymodel, monitors, moves)
+mymcmc.run(generations=1)
+```
+
+The value is given in the Likelihood column of the first (i.e. 0th-iteration) row.
+
+
+## Methods
+
+MixtureLikelihoods = MixtureLikelihoods ()
+
+[] = [] (Natural<any> index)
+
+addMissingTaxa = addMissingTaxa (String|Taxon|String[]|Taxon[]<any>...
+
+applyMissingSitesMask = applyMissingSitesMask (AbstractHomologousDi...
+
+chartype = chartype ()
+
+clamp = clamp (AbstractHomologousDiscreteCharacterData)[https://revbayes.github.io/documentation/AbstractHomologousDiscreteCharacterData.html])<any> x): Clamp the draw from this distribution to a `AbstractHomologousDiscreteCharacterData` object.
+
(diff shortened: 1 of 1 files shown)


help/md/dnPhyloCTMC.md
```diff
@@ -1,9 +1,218 @@
...
+monitors = VectorMonitors()
+monitors.append( mnScreen(printgen=1) )
+mymcmc = mcmc(mymodel, monitors, moves)
+mymcmc.run(generations=1)
+```
+
+The value is given in the Likelihood column of the first (i.e. 0th-iteration) row.
+
+
+## Methods
```
davidcerny, line 53: This list mixes up actual methods of `dnPhyloCTMC()`, like `clamp`, `probability`, and `redraw`, with methods of `AbstractHomologousDiscreteCharacterData` – i.e., methods that cannot be applied to the distribution itself, but only to draws from it. This is not accurate.

help/md/dnPhyloCTMC.md
```diff
@@ -1,9 +1,218 @@
...
+
+varGcContentByCodonPosition = varGcContentByCodonPosition (Natural<...
+
+
 ## references
+- citation: Felsenstein J., 1973. Maximum Likelihood and Minimum-Steps Methods for Estimating Evolutionary Trees from Data on Discrete Characters. Systematic Biology 22:3, 240--249
+  doi = 10.1093/sysbio/22.3.240
+- citation: Felsenstein, J. (1981). Evolutionary trees from DNA sequences: A maximum likelihood approach. Journal of Molecular Evolution. 17 (6): 368–376.
+  doi : 10.1007/BF01734359
+- citation: Hohna, S., Landis, M.J. and Heath, T.A. 2017. Phylogenetic inference using `RevBayes`. Curr. Protoc. Bioinform.
```
davidcerny, line 213: Höhna not Hohna

davidcerny (changes requested):

help/md/dnPhyloCTMC.md
davidcerny, line 53: This list mixes up actual methods of `dnPhyloCTMC()`, like `clamp`, `probability`, and `redraw`, with methods of `AbstractHomologousDiscreteCharacterData` – i.e., methods that cannot be applied to the distribution itself, but only to draws from it. This is not accurate.

davidcerny, line 53: This list mixes up actual methods of `dnPhyloCTMC()`, like `clamp`, `probability`, and `redraw`, with methods of `AbstractHomologousDiscreteCharacterData` – i.e., methods that cannot be applied to the distribution itself, but only to draws from it. This is not accurate.

davidcerny, line 53: This list mixes up actual methods of `dnPhyloCTMC()`, like `clamp`, `probability`, and `redraw`, with methods of `AbstractHomologousDiscreteCharacterData` – i.e., methods that cannot be applied to the distribution itself, but only to draws from it. This is not accurate.

help/md/dnPhyloCTMC.md
```diff
@@ -1,9 +1,218 @@
...
+
+computeSiteFrequencySpectrum = computeSiteFrequencySpectrum (Bool<a...
+
+computeStateFrequencies = computeStateFrequencies ()
+
+excludeAll = excludeAll ()
+
+excludeCharacter = excludeCharacter (Natural<any> pos)
+
+excludeCharacter = excludeCharacter (Natural[]<any> )
```
davidcerny, line 77: This list mixes up actual methods of `dnPhyloCTMC()`, like `clamp`, `probability`, and `redraw`, with methods of `AbstractHomologousDiscreteCharacterData` – i.e., methods that cannot be applied to the distribution itself, but only to draws from it. This is not accurate.

ms609: Ah, that makes sense –  I hadn't clocked that distinction, but it's obvious now you point it out.

help/md/dnPhyloCTMC.md
```diff
@@ -1,9 +1,45 @@
...
+# Create stochastic node with the tip distribution given by the topology `tree` with the 
+# rate matrix parameter q_matrix
+x ~ dnPhyloCTMC(tree = tree, Q = q_matrix)
+
+# Clamp observed characters to the node
+chars <- readDiscreteCharacterData("myData.nex")
+x.clamp(chars)
+
+# Calculate the probability of the observed characters under the given distribution
+x.lnProbability()
```
davidcerny, line 32: While it's indeed possible for `tree` to be just a topology (e.g., a draw from `dnUniformTopology()` or a similar distribution), the output of `x.lnProbability()` will then be `nan`, because we can't evaluate the likelihood without branch lengths. Perhaps it might be better to provide a slightly expanded, self-contained example:

```
chars <- readDiscreteCharacterData("myData.nex")
taxa = chars.taxa()
# Draw a tree with branch lengths
tree ~ dnUniformTopologyBranchLength(taxa, branchLengthDistribution=dnExp(10.0))
q_matrix <- fnJC(4)
```

and then go on to show how to draw, clamp, and evaluate, as in your current code snippet.

Also, in this context, I would refer to `q_matrix` as an argument rather than a parameter.

davidcerny (changes requested):
This looks good – almost there, I think! Thanks for addressing the previous feedback.

help/md/dnPhyloCTMC.md
ms609: Good suggestions, thanks; implemented.

davidcerny (approved):

PR #488, opened 2024-07-09
Stopping rules

ms609:
Propagates the changes of #485 to related documentation pages, with examples of stopping rules in practice and brief descriptions of their motivation and use.


help/md/srGeweke.md
```diff
@@ -1,9 +1,58 @@
...
+stopping_rules[3] = srMinESS(50, file = paramFile, freq = convergenceFreq)
+stopping_rules[4] = srGelmanRubin(1.01, file = paramFile, freq = convergenceFreq)
+stopping_rules[5] = srGeweke(prob = 0.001, file = paramFile, freq = convergenceFreq)
+stopping_rules[6] = srStationarity(prob = 0.01, file= paramFile ,freq = convergenceFreq)
+
+mymcmc = mcmc(mymodel, monitors, moves)
+# Begin the MCMC run.
+# The checkpoint file will allow the run to be resumed from where it left off
+# if you later opt for more stringent stopping rules
+mymcmc.run(rules = stopping_rules, checkpointFile = "checkpoint.ckp", checkpointInterval = 1000)
```
davidcerny, line 52: I think these examples should ideally be (1) self-contained, (2) simpler, and (3) specific to a given rule. What's given here is actually the most complex case possible, since all available stopping rules are used together. I'd suggest something like this:

```
# Binomial example: 7 successes out of 20 trials
r ~ dnExp(10)
p := Probability(ifelse(r < 1, r, 1))
n <- 20
k ~ dnBinomial(n, p)
k.clamp(7)

mymodel = model(k)

moves = VectorMoves()
moves.append( mvSlide(r, delta=0.1, weight=1) )

paramFile = "parameters.log"

monitors = VectorMonitors()
monitors.append( mnScreen(printgen=100, p) )
monitors.append( mnModel(filename=paramFile, printgen=100, p) )

# Stop when the test statistic becomes significant at alpha = 0.001
stopping_rules[1] = srGeweke(prob=0.001, file=paramFile, freq=10000)

# Create the MCMC object
mymcmc = mcmc(mymodel, monitors, moves)

# Begin the MCMC run
mymcmc.run(stopping_rules)
```

(This particular example apparently does well on `srGeweke` and `srMinESS` but terribly on `srGelmanRubin` and `srStationarity`. Even so, it's still useful for documentation purposes.)

davidcerny (changes requested):

help/md/srGelmanRubin.md
```diff
@@ -1,9 +1,54 @@
 ## name
 srGelmanRubin
 ## title
+Gelman–Rubin stopping rule
+
 ## description
+Allow an MCMC run to terminate once the specified criterion has been met.
+The Gelman–Rubin rule compares the variance between chains with the variance within chains; its value tends to unity (1) as runs converge.
```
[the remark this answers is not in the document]
ms609: Yes! I hadn't realised that these were the same thing.  Done.

help/md/srGelmanRubin.md
```diff
@@ -1,9 +1,54 @@
...
+
 ## details
+When passed as an element of the `StoppingRules[]` argument to `mcmc.run()` or `mcmcmc.run()`, the run will terminate once *all* specified convergence rules ([`srGelmanRubin()`], [`srGeweke()`], [`srMinESS()`], [`srStationarity()`]) have been fulfilled; or once *any* threshold rules ([`srMaxTime()`], [`srMaxIteration()`]) are met.
```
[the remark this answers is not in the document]
ms609: OK - removed from the individual SR pages - on reflection I suppose that this is describing the behaviour of mcmc rather than the rule itself, and this is clear from the example.

help/md/srGeweke.md
ms609: Yes, on reflection, I see that this is a better approach.  

In adapting this code for srMaxIteration, I wanted to add an explanation of why one might use srMaxIteration rather than just passing an `nGen` argument to `mcmc`.  I realise I'm not sure how the behaviour differs.  What is the motivation for a separate srMaxIteration function – do stopping rules cause `nGen` to be ignored?

davidcerny: Hmm, good question. It looks like the `generations` (not `nGen`) argument behaves as just another stopping rule. If you combine it with `srMaxIteration` and give them different values, the total number of iterations will be determined by whichever is reached first. I guess the only point of `srMaxIteration` is that you can combine it with other stopping rules in the same vector and pass it to `mcmc.run()` as a single argument, instead of having to specify two arguments.

help/md/srGelmanRubin.md
```diff
@@ -15,32 +15,30 @@ The Gelman–Rubin rule compares the variance between chains with the variance w
...
-# Add convergence rules: Stop as soon as ALL have been met
-stopping_rules[3] = srMinESS(50, file = paramFile, freq = convergenceFreq)
-stopping_rules[4] = srGelmanRubin(1.01, file = paramFile, freq = convergenceFreq)
-stopping_rules[5] = srGeweke(prob = 0.001, file = paramFile, freq = convergenceFreq)
-stopping_rules[6] = srStationarity(prob = 0.01, file= paramFile ,freq = convergenceFreq)
+# Stop when the potential scale reduction factor falls below 1.01
+stopping_rules[1] = srGelmanRubin(1.01, file = paramFile, freq = 1000)
 
+# Create the MCMC object
 mymcmc = mcmc(mymodel, monitors, moves)
```
davidcerny, line 38: We should add `nruns=2` here, because the Gelman-Rubin statistic is only defined for 2 or more chains.

help/md/srStationarity.md
```diff
@@ -12,32 +12,30 @@ Allow an MCMC run to terminate once the specified criterion has been met.
...
-# Add convergence rules: Stop as soon as ALL have been met
-stopping_rules[3] = srMinESS(50, file = paramFile, freq = convergenceFreq)
-stopping_rules[4] = srGelmanRubin(1.01, file = paramFile, freq = convergenceFreq)
-stopping_rules[5] = srGeweke(prob = 0.001, file = paramFile, freq = convergenceFreq)
-stopping_rules[6] = srStationarity(prob = 0.01, file= paramFile ,freq = convergenceFreq)
+# Stop when stationarity has been accomplished at significance level = 0.05
+stopping_rules[1] = srStationarity(prob = 0.05, file = paramFile, freq = 1000)
 
+# Create the MCMC object
 mymcmc = mcmc(mymodel, monitors, moves)
```
davidcerny, line 37: Same as above; this stopping rule also only works with more than 1 chain. That might be useful to mention in the description, too.

davidcerny (changes requested):

davidcerny (approved):

PR #490, opened 2024-07-10
Add `build/` to .gitignore

ms609:
https://revbayes.github.io/developer/setup/ tells the user to add `build/` to `.gitignore`; but as the repo does not contain any tracked files in the `build/` directory, it's not clear why `build/` shouldn't just be included in the default `.gitignore` file.

This would also simplify the process of switching between branches using Git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ src/generated_include_dirs.cmake
 src/*/CMakeLists.txt
 
 # ----------------------------------------------
+build/
 projects/cmake/*
 !projects/cmake/README.txt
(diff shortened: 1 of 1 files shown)


ms609:
It might also be worth adding VSCode-generated files to the `.gitignore`?

Candidates would be

```
*.tlog
*.filters
*.vcxproj
*.stamp
*.depend
*.obj
*.sln
*.recipe
```

bredelings:
Sounds like a good idea.

bredelings (approved):

PR #491, opened 2024-07-10
Missing `S` in `Boost_INCLUDE_DIRS`

ms609:
Reported variable should match message.

I believe `S` is intended to match the output of `find_package()`.

https://stackoverflow.com/questions/63660054/cmake-finds-boost-but-boost-include-dir-is-not-set


bredelings (approved):
Thanks!

PR #493, opened 2024-07-11
multiple rules generate ../rb.exe

ms609:
I encountered a few issues when trying to install RevBayes from source.

One was:

```
[cmake] CMake Error:
[cmake]   Running
[cmake] 
[cmake]    'C:/Strawberry/c/bin/ninja.exe' '-C' 'C:/Users/<username>/revbayes/build' '-t' 'recompact'
[cmake] 
[cmake]   failed with:
[cmake] 
[cmake]    ninja: error: build.ninja:148: multiple rules generate ../rb.exe
```

When I removed the line `set_target_properties(${RB_EXEC_NAME}   PROPERTIES PREFIX "../")`, the build succeeded.

I've proposed removing this line in the present pull request. Perhaps, however, I am not aware of conditions where this line is necessary – e.g. if RevBayes is also being built?  If so, it should probably be retained, but enclosed in an appropriate conditional.


bredelings:
I'm inclined to merge this if the tests pass.  I'm not a cmake expert, but it looks like the PREFIX property is supposed to be something like "lib", not "../".

That said, there's now a merge conflict.

ms609:
conflict resolved.

bredelings (approved):

PR #494, opened 2024-07-11
Policies relevant to Boost

ms609:
Using cmake 3.30, I encountered a variety of warning messages whilst attempting to configure Boost. RevBayes compiles successfully under the new policy settings, and setting these policies may help users to better diagnose issues with Boost installation (or at least spare them some blind alleys).


bredelings (approved):
Very good, thanks!

bredelings:
There is also `-DBoost_DEBUG=ON` (from the command-line) or `set(Boost_DEBUG ON)` (from CMakeLists.txt).  Perhaps these policies replace that?

bredelings:
One of these three changes makes the build fail on Mac-ARM, because it can't find the Boost include files.  I'm not sure why this is so...

ms609:
I've converted this PR to draft whilst I commence a process of elimination...

ms609:
Sorted – some builds were not happy with the "new" policy for CMP0167, which arrived in cmake 3.30; setting this to "OLD" allows all builds to pass.

My reading of the situation is that this suggests that the make process will need revision in order not to rely on the FindBoost module, which will be removed in a future version of cmake.

bredelings:
Might the 3rd policy pass on Mac-Arm now that both INCLUDE_DIRS bugs are fixed?

ms609:
Apparently so!  I see that the Windows build is now failing, but I don't think that's caused by these changes.

bredelings:
A think the windows test variable is some other problem.  I have been able to make it go away by rerunning the failing tests.  PR #500 has a change to help diagnose why the failures are occurring.

PR #495, opened 2024-07-11
Big object support

ms609:
Big object support was necessary for me to build RevBayes.

The MSVC option is potentially unnecessary – I didn't end up getting RevBayes to build under msvc.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -7,11 +7,13 @@ project(RevBayes)
 
 set(CMAKE_CXX_STANDARD 17)
 
-# Congifure compiler to support big objects
-if(MSVC)
-  add_compile_options(/W4 /EHsc /bigobj)
-else()
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wa,-mbig-obj")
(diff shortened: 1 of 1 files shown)


bredelings (approved):

src/CMakeLists.txt
```diff
@@ -7,6 +7,13 @@ project(RevBayes)
 
 set(CMAKE_CXX_STANDARD 17)
 
+# Congifure compiler to support big objects
+if(MSVC)
+  add_compile_options(/W4 /EHsc /bigobj)
+else()
```
bredelings, line 13: I think you need to gate this on WIN32.  Otherwise it breaks non-windows builds.

So something like `elseif(WIN32)` perhaps.

ms609: I gated the whole block – I figured it was more conservative, even if few non-windows users are likely to be using MSVC

bredelings:
Also note that currently our automated tests build the windows executables _on Linux_ using meson instead of cmake.

We could check if there's another way to run the automated tests on windows.  Alternatively, we could emulate a windows compilation by running windows software on Linux under wine.

PR #500, opened 2024-07-13
Distributions for transformed random variables

bredelings:
This PR implements distributions for random variables that have been transformed by applying a function `f`.

1. The following scalar transformations have been implemented:
- tnLog()  
- tnExp()
- tnLogit()
- tnInvlogit()
- tnShift(Real delta)
- tnShift(RealPos Delta) -> this version preserves the positivity of Distribution__RealPos
- tnScale(Real lambda)
- tnScale(RealPos lambda) -> this version preserves the positive of Distribution__RealPos
- tnScale(Probability lambda) -> this version preserves the probability property of a Distribution__Probability

For example `dnNormal(mu,sigma)` is equivalent to `tnShift(tnScale(dnNormal(0,1),sigma),mu)`.

2. The PR also implements vector transformations.  You can either specify a vector transfomation (including the log-jacobian) or a scalar transformation that is applied to every dimension. The following transformations have been implemented on vector distributions:
- tnExp() 
- tnLog()
- tnLogit()
- tnInvlogit()

3. The PR also adds the functions logit(x) and invlogit(x).  The function invlogit(x) is just a new alias to the existing function logistic(x) -- invlogit(x) is the name from R.


davidcerny (approved):
This all looks great to me – it's a feature-rich, well-documented PR that substantially extends RB's math functionality. I only have two suggestions, but I'm happy to leave it up to you whether they are actually worth addressing @bredelings.

(1) I know this is basic math, but I'm wondering if it might make sense for the documentation to remind the users of what the support of a distribution must be for a given transformation to apply (e.g., nonnegative reals for `tnLog`, the unit interval for `tnLogit`, etc.).

(2) To me, it looks like the validation test _showcases_ the new functions but does not really _validate_ them. Would it make sense to include a few trivial examples where the expected result is first computed by other means (e.g., using the `logit`, `log`, and `exp` functions, or using the `offset` argument of `dnExp`) and then compared to the output of a given transform? I'm thinking of something along the lines of our [topologyProb test](https://github.com/revbayes/revbayes/blob/master/tests/test_topologyProb/scripts/Test_topologyProb.Rev).

bredelings:
Oops, I merged before I read your comment.  I'll address those via a followup PR.

PR #502, opened 2024-07-16
Fixing broken checkpointing in FossilTipTimeSlideUniformProposal

bjoelle:
This should fix the checkpointing issue identified by @davidcerny in PR #484 
The tuning parameter wasn't stored properly for FossilTipTimeSlideUniformProposal, which stopped the operator from working when restarting from checkpoint


src/core/moves/proposal/tree/FossilTipTimeSlideUniformProposal.cpp
```diff
@@ -90,7 +90,7 @@ const std::string& FossilTipTimeSlideUniformProposal::getProposalName( void ) co
 double FossilTipTimeSlideUniformProposal::getProposalTuningParameter( void ) const
 {
     // this proposal has no tuning parameter
-    return RbConstants::Double::nan;
+    return lambda;
 }
```
bredelings, line 94: Is the comment false now?

bjoelle: oops, you're right - fixed now !

davidcerny:
This still doesn't quite work for me @bjoelle – now the restarted analysis is just hanging indefinitely:

Screenshot 2024-07-17 at 9 53 51

Looking at the `state_moves.state` file, this might be because the tuning operator / lambda is still equal to `nan` in the following moves: `FNPR`, `CollapseExpandFossilBranch`, `NodeTimeSlideUniform`, `RootTimeSlideUniform`. I think these require analogous changes.

bjoelle:
@davidcerny I don't think this is from the operators, the bug was causing fixed parameters, not hanging - for sure I checked `NodeTimeSlideUniform` and `RootTimeSlideUniform` and they were fine 
it was working here, so to debug further I would need the seed & checkpoint files you are using ?

bjoelle:
thanks, will look into this when I have time

bredelings (approved):

bredelings:
I don't know if you want to wait to merge this until @davidcerny's example is fixed, or merge the partial fix first.

bjoelle:
I don't know when I can work on this further, and the fix appeared to work on at least some runs, so I think I will merge this now

PR #503, opened 2024-07-16
Eliminating the WithOffset distributions

davidcerny:
This PR adds an offset parameter (set to 0 by default) to the exponential and lognormal distributions, eliminating the need for separate "exponential with offset" and "lognormal with offset" distributions. On the user side of things, everything should stay exactly the same:

- `x ~ dnExp( rate )`: `x` is successfully drawn and is of type `RealPos` (i.e, no need to specify `offset`)
- `x ~ dnExp( rate, offset )` and `offset` is greater than or equal to 0: `x` is of type `RealPos`
- `x ~ dnExp( rate, offset )` and `offset` is less than 0: `x` is of type `Real`


davidcerny:
@bredelings I'd be happy to also add a note to the documentation saying that the `offset` argument is deprecated in favor of `tnShift()`.

bredelings (approved):

PR #504, opened 2024-07-18
Modify `type( )` so that you can see the full type-spec

bredelings:
This adds an extra argument `verbose` that defaults to `FALSE`.  When `verbose=True` is specified, it shows the full typespec and not just the last element:

``` R
> type(0.5)
   Probability
> type(0.5,TRUE)
   Probability, RealPos, Real, RevObject
```


davidcerny:
Not sure this still requires review given that it has already been merged. Still, a very nice addition!

PR #505, opened 2024-07-19
Document `mcmc.initializeFromCheckpoint()`

ms609:
I wasn't clear how this method could be used, so have expanded the documentation to record my observations of its behaviour.

I'm undecided as to the best way to avoid duplication between mcmc and mcmcmc; perhaps there's a clever way to use templates, but for now I've just pointed to mcmc from mcmcmc.

This also fixes a couple of typos in the existing examples.


help/md/mcmc.md
```diff
@@ -11,28 +11,41 @@ An MCMC analysis is initiated using the `mcmc.run()` method.
...
+
+    # Resume analysis from the checkpoint file
+    mymcmcObject.initializeFromCheckpoint( "output/out.ckp" )
+
+    # Conduct an additional 400 generations
+    mymcmcObject.run( generations = 400 )
+
+    # Stopping rules are defined on the total number of generations
+    # This command will have no effect, as 400 generations have already been performed.
+    mymcmcObject.run( generations = 400, rules = [ srMaxIteration(400) ] )
```
davidcerny, line 48: Ah, so there _is_ a difference between `generations` and `srMaxIteration` after all! Congratulations on uncovering this :) However, maybe the final line is a bit too complicated, and

```
mymcmcObject.run( rules = [ srMaxIteration(400) ] )
```

would suffice to illustrate the point?

help/md/mcmcmc.md
```diff
@@ -9,7 +9,7 @@ The Mcmcmc analysis object produced by a call to this function keeps copies of t
 
 An MCMCMC analysis is initiated using the `mcmcmc.run()` method.  
 The `StoppingRule[]` argument provides a mechanism to automatically terminate a run once a set of rules are met: perhaps once the run has attained convergence, or after a certain amount of time has passed.  The run will be terminated once *all* convergence rules ([`srGelmanRubin()`], [`srGeweke()`], [`srMinESS()`], [`srStationarity()`]) have been fulfilled; or once *any* threshold rules ([`srMaxTime()`], [`srMaxIteration()`]) are met.
-The parameters `checkpointFile` and `checkpointInterval` generate snapshots of the current state of the MCMCMC run from which the run can be continued if interrupted using the `mcmc.initializeFromCheckpoint()` method.
+The parameters `checkpointFile` and `checkpointInterval` generate snapshots of the current state of the MCMCMC run from which the run can be continued if interrupted using the `mcmc.initializeFromCheckpoint()` method. More details are given on the documentation page for [`mcmc()`].
```
davidcerny, line 12: I would maybe say "An example is" instead of "More details are". But I wouldn't object to just including a checkpointing example on this page too.

davidcerny (changes requested):

help/md/mcmc.md
ms609: I've simplified the example, and rewritten the text to clarify the interaction between the parameters.

davidcerny (approved):

PR #506, opened 2024-07-19
How the `weight` parameter is used in move proposals

ms609:
I've determined by experiment that the number of moves proposed per generation equals the sum of all move proposal weights.

I think it is logical to document this in each move proposal function; perhaps there are also other places this should be mentioned?

The former wording of the documentation for the `weight` parameter gave me the impression that a fixed number of moves were proposed per iteration, with the ratio of moves adjusted according to the weights – rather than the number of moves being a function of the weight.

I am left unclear what an 'iteration' means in this context.  Is this simply a measure of the progress of a run?  Is likelihood only computed after `sum(weights)` moves have been conducted, i.e. each iteration corresponds to a proposal?  As this has quite large consequences for MCMC design, it is important for this to be obvious from the documentation.


bredelings:
> I am left unclear what an 'iteration' means in this context. Is this simply a measure of the progress of a run? Is likelihood only computed after sum(weights) moves have been conducted, i.e. each iteration corresponds to a proposal? As this has quite large consequences for MCMC design, it is important for this to be obvious from the documentation.

Some MCMC software does one move per iteration.  Others do a move on _every parameter_ per iteration.  RevBayes is in the latter group.  Additionally, some parameters take more repetitions in order to result in an accepted move, which leads to weights higher than one.

I personally prefer that each iteration should have one _accepted_ proposal for each parameter per iteration.  I think of iterations more as a timescale for logging.  If each parameter is updated once each iteration, then logging per iteration makes sense.  If the autocorrelation-time for some parameters is high, you might still want to subsample by some factor (not for the sake of correctness, but to avoid wasting disk space by logging nearly-identical samples).  However, now the logging interval is a multiple of something more sensible ("iterations" where each parameter is updated once.)

bredelings:
Yeah, the previous text was rather misleading.  How about "The number of times (on average) this move will be applied during each iteration." ?  Possibly mentioning "the number" and "during" would make this clearer.  

Alternatively, the longer version is fine too.  Perhaps "The weight specifies the number of times (on average) this move will run during each iteration. (The number of moves proposed in each iteration is the sum of the weights of all moves)."

bredelings:
It makes sense to give a longer explanation somewhere -- but the parameter description should not be too long perhaps?

Thanks for all this great feedback!  I suspect many users just remain confused and don't report anything, much less suggest an alternative.

ms609:
Yes, brevity is good! How about:

> The weight defines the number of times (on average) this move will be attempted during each iteration.

It's a fair point about "proposed"; there's also (perhaps less) scope for confusion with "run" (which is used elsewhere to mean "independent replicate"). I think "be attempted" makes it clear that we count the number of suggested rather than accepted moves.

I agree that a more complete description of how an MCMC iteration is implemented in RevBayes would be helpful, somewhere.  I'm not sure where I would head for this, as a user – the "Documentation" page seems natural, but is restricted to functions; "Tutorials" isn't the first place I'd look to understand these aspects of the software, even though some tutorials come close to touching on these topics.

bredelings:
> Yes, brevity is good! How about:

> >   The weight defines the number of times (on average) this move will be attempted during each iteration.

I like "attempted".  Other possibilities could be "executed" or "performed".  Whichever sounds best to you.

> I agree that a more complete description of how an MCMC iteration is implemented in RevBayes would be helpful, somewhere. I'm not sure where I would head for this, as a user

One possibility would be to add a "see also" to the help file (*.md), and then add a help file about moves and weights that isn't attached to a specific command.  The downside of that is that I'm not sure how to do that for each move, where as putting this as documentation on the "weight" parameter automatically shows up in the help for every move.  :man_shrugging:

ms609:
Would it be possible to include the link in the parameter text?  I'm not sure how this string is processed in different contexts.

e.g. `The weight defines the number of times (on average) this move will be attempted during each iteration (see [new_page])`, or perhaps better
`The weight defines the number of times (on average) this move will be attempted during each [MCMC iteration](new_page.md).`

ms609:
Another way of looking at this is that this information belongs in `mcmc.md`.  The move proposals are part of the MCMC process (and its accompanying functions), and this is likely to be one of the first places someone looks when attempting to diagnose a misbehaving mcmc run.

davidcerny:
Yes, I also think that information is better placed in `mcmc.md`, because the meaning of the `weight` parameter is ultimately determined by the `moveschedule` argument of the `mcmc` function. The current explanation (number of times a move will be attempted per iteration) works for `moveschedule="random"` or `moveschedule="random"`. But note that with `moveschedule="single"`, RevBayes will imitate the behavior of other software like BEAST or MrBayes, and count each attempted move as a separate iteration.

bredelings:
That's tricky then, if the weight can mean different things depending on the move schedule.  We could write something like "The weight defines how often this move will be attempted during each iteration.  The exact meaning depends on the "moveschedule" parameter of mcmc".

I'm not sure if markdown links work in the text help in the rb command line environment, but might work in the HTML help that we generate for the website.

wrightaprilm:
I like that wording, Ben. It lines up well with how we phrased this distinction in a recent[ paper](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC10933576/):

> Others move a whole set of parameters at each step (e.g., RevBayes), with the weights representing how many times a move is attempted for a particular parameter during each step.

There is an MCMC Troubleshooting guide that will be released at the same time as the final typeset of that paper, so it's good to have consistent language across related sources.

davidcerny:
Hmm, I really like the current explanation by @ms609 (from commit 5f2ccd4); it's both clear and concise (or at least as concise as it can be given this additional complication). I don't think we should write "The weight defines how often this move will be attempted during each iteration", because that's simply not true when `moveschedule="single"`.

davidcerny:
Note also this detailed explanation from [Höhna et al. (2017)](https://currentprotocols.onlinelibrary.wiley.com/doi/10.1002/cpbi.22), a paper that is already referenced in the documentation for  `dnPhyloCTMC`:

> There are several important details about this algorithm. First, the next move is either selected _randomly_ or _sequentially_. If the moves are selected randomly, then a move is selected randomly in proportion to its weight. Random selection of moves has, in general, the best performance, although sequential moves are useful for consecutive updates of dependent or correlated parameters. Second, we apply _k_ moves per iteration, where _k_ is equal to the sum of weights of all moves. This has the advantage that models with many parameters still require the same number of iterations but perform more moves per iteration (similar to the MCMC procedure in PhyloBayes; Lartillot et al., 2009). Nevertheless, we provide the option to run `RevBayes` with a single, randomly chosen move per iteration with the option `moveschedule=single`.

wrightaprilm:
That's quite thorough and helpful text. I'm fine with what's in this PR for brevity, but perhaps incorporating the above would be better.

davidcerny:
In `RlMove.cpp`, we could just say "The weight determines the relative frequency with which this move will be attempted." That's somewhat vague, but always accurate.

bredelings:
> In RlMove.cpp, we could just say "The weight determines the relative frequency with which this move will be attempted." That's somewhat vague, but always accurate.

That would work if we _also_ mention "moveschedule" in `mcmc`.

davidcerny:
Thanks to @ms609, we do that as of 5f2ccd4!

bredelings:
No, I mean that would work if we reference `moveschedule` and `mcmc` from the help on `weight`.

bredelings:
Although I'm fine merging as-is if other people are.

davidcerny:
> No, I mean that would work if we reference `moveschedule` and `mcmc` from the help on `weight`.

Ah, gotcha. I agree; that would make a lot of sense.

davidcerny (approved):

Issue #508, opened 2024-07-19
[BUG] Intermittent NaN in dnPhyloCTMC.lnProbability()


ms609:
**Describe the bug**
dnPhyloCTMC.lnProbability() sometimes returns NaN instead of the lnProbability.

**To Reproduce**

Save this script to file:
```rb
clear()
seed(42)

# Read the primates data from the validataion data directory
chars <- readDiscreteCharacterData("../validation/data/primates_cytb.nex")

taxa <- chars.names()
outgroup <- clade(taxa[1])
nTaxa <- chars.size()
nEdge <- 2 * nTaxa - 3

topology ~ dnUniformTopology(taxa, outgroup = outgroup)

tree_length ~ dnGamma(2, 2)
rel_br_lengths ~ dnDirichlet( rep(1.0, nEdge) )
br_lengths := tree_length * rel_br_lengths

phylogeny := treeAssembly(topology, br_lengths)

Q4 = fnJC(4)
phy ~ dnPhyloCTMC(tree = phylogeny, Q = Q4, type = "DNA")
phy.clamp(chars)
phy.lnProbability()
```

Source the script a few times. Sometimes you will see the likelihood of `-28378.14`.  Other times you will see `nan`.  There seems to be no pattern.

It's perhaps notable that adding 
```rb
invPhy ~ dnInv(dnPhyloCTMC(tree = phylogeny, Q = Q4, type = "DNA"))
invPhy.clamp(chars)
invPhy.lnProbability()
```
to the end of the script (on branch `dnInverse`) sees `phy.lnProbability() == - invPhy.lnProbability()`; i.e. both are or are not `nan` on the same run

**Expected behavior**
The likelihood should always be computed correctly

**Screenshots**
![image](https://github.com/user-attachments/assets/6005291d-b8c2-487a-8dbe-9602d889309a)

**Computer info**
Win10; RevBayes 1.2.5

PR #509, opened 2024-07-19
Fix branch length NaN

bredelings:
Correctly initialize a boolean variable that determines whether we set the initial branch lengths in TreeAssemblyFunction.


bredelings:
Fixes #508

ms609 (approved):
That was quick work! Thank you!  The changes look sensible insofar as my rather limited knowledge can judge.  I've merged the changes into my local code and I no longer see the NaN outcome, so this seems to have fixed the issue for me.

Issue #507, opened 2024-07-19
[BUG] Modifying source file during run causes failure


ms609:
**Describe the bug**
If the content of a file being sourced changes during analysis, this can cause RevBayes's reading frame to shift and a run to fail.

**To Reproduce**

- Create a file `test.Rev` in the `rb.exe` directory, containing the following:

```rb
    # Create a simple model (unclamped)
    a ~ dnExponential(1)
    mymodel = model(a)
    
    # Create a move vector and a monitor vector
    moves[1] = mvScale( a, lambda = 1.0, weight = 40000.0 )
    monitors[1] = mnFile( a, filename = "output/out.log" )
    
    # Create an mcmc object
    mymcmcObject = mcmc( mymodel, monitors, moves )
    
    # Run a short analysis
    mymcmcObject.burnin( generations = 500, tuningInterval = 100 )
    mymcmcObject.run( generations = 400, checkpointFile = "output/out.ckp", checkpointInterval = 100 )
    
    # print the summary of the operators (now tuned)
    mymcmcObject.operatorSummary()

    # Resume analysis from the checkpoint file
    mymcmcObject.initializeFromCheckpoint( "output/out.ckp" )

    # Conduct an additional 400 generations
    mymcmcObject.run( generations = 400 )

    # Stopping rules are defined on the total number of generations
    # This command will have no effect, as 400 generations have already been performed.
    mymcmcObject.run( generations = 400, rules = [ srMaxIteration(400) ] )
```

- Launch two instances of `rb.exe`
- In the first instance, execute `source("text.rev")`
- Whilst the burnin phase is running (be quick!), modify the content of `test.Rev`.  I deleted line 5, which contains the comment `# Create a move vector and a monitor vector`. Save the file.
- In the second `rb` instance, execute `source("text.rev")`
- The first instance eventually throws an error:

```
   Syntax error while reading characters "erations" at position 75 to 82 in command:
       mymcmcObject.run( generations = 400, rules = [ srMaxIteration(400) ] )erations = 400, rules = [ srMaxIteration(400) ] )
   Error:       Problem processing line 27 in file "test.rev"
```

**Expected behavior**
Once a file has been `source`d, modifying its content should not affect the behaviour of RevBayes.  Or (as a less preferred option), RevBayes should write-protect the file until it has finished with the file.

**Computer info**
Win 10; RevBayes version (1.2.5-preview)

**Additional context**
I've not tried all permutations, but I'm not quite clear whether the bug occurs only when
[...]

PR #510, opened 2024-07-19
Avoid crashing if a file changes while we are source( )-ing it.

bredelings:
We can simply read the whole file into a string and execute the string.  That way the execution sees the way the file looked when we ran the source command.


bredelings:
Fixes #507

ms609 (approved):
This seems like the natural way to handle this case, and is consistent with how I had expected the `source()` command would behave.

davidcerny (approved):
Looks great!

Issue #463, opened 2024-04-30
[Help Wanted] Running MCMC under the prior also removes the node calibrations


bjoelle:
Running the `mcmc.run` with `underPrior = TRUE` seems to remove all clamped nodes, including node calibrations (at least if node calibrations are set following this tutorial [here](https://revbayes.github.io/tutorials/relative_time_constraints/))
I don't think this is proper behaviour as node calibrations are generally considered part of the prior - in particular this makes difficult to check the effective priors on the calibrations, which is generally good practice

bjoelle:
thanks, that is helpful to know !

bredelings:
Yeah, "remove the dnPhyloCTMC and run the MCMC as usual" seems like it corresponds more closely to what BEAST people mean by "run under the prior".  What they really mean is "run without the sequence data", which is exactly what happens when you remove `dnPhyloCTMC( )`.

davidcerny:
@bjoelle Do you think this issue can be closed now, or would you suggest adding a brief note about this behavior to the documentation for `mcmc.run`?

bjoelle:
@davidcerny I don't think the core issue is resolved ? I don't think this is correct behaviour for this parameter, and I don't think documenting the bug is enough to make sure people will not run into it again...

bredelings:
Hmm... I think the current behavior is correct.  The prior does not include calibration likelihoods.  BEAST people have a habit of calling the prior + calibration likelihoods "the prior", but it isn't a prior.  This is extra confusing because people like Sandra Álvarez-Carretero, Mario dos Reis, Bruce Rannala, etc. use a different framework where they modify the prior instead of adding calibration likelihoods.

It seems that the problem with this parameter is not whether it is technically correct (I think it is correct) but a matter of perception:  what do people expect it to do?  Right now it is confusing and misleading, given how (some) people use the phrase "under the prior".

bjoelle:
I don't think this is related to BEAST culture to be honest - in my understanding, node calibrations are considered part of the prior because they're set by the user based on information which is external to the inference, which is pretty much the definition of a prior
certainly if you want to know the influence of model choices on the results of the inference you need to include them

bredelings:
> certainly if you want to know the influence of model choices on the results of the inference you need to include them

True enough.  I think its more helpful to say "prior to what".  Let's consider an analysis with a birth-death prior, fossil information, and DNA data.  In this case using you can talk a distribution that is
1. prior to fossil information and DNA data
[...]

PR #515, opened 2024-07-20
Make it possible to run node-dating analysis without character data

davidcerny:
This PR is intended as a solution to issue #463. It adds an argument to the `.run()` and `.burnin()` methods called `suppressCharacterData`, whose usage is analogous to `underPrior`. It automatically detects all stochastic nodes associated with character data, and selectively unclamps them. It makes no assumptions about the distribution these nodes are drawn from (i.e., distributions other than `dnPhyloCTMC` will still be detected) or about the type of character data (in a hypothetical node-dating analysis based on continuous characters, these will be detected also). If both `underPrior` and `suppressCharacterData` are set to `TRUE`, the former overrides the latter, and a warning is printed to the screen alerting the user to this. Finally, I added a few sentences to the documentation of `mcmc` and `mcmcmc` explaining the difference between these two options.

I've run analyses on the dataset from our [node-dating tutorial](https://revbayes.github.io/tutorials/dating/nodedate) with a fixed seed to confirm that the sampled ages and likelihoods are exactly the same under this new option as those obtained by simply not clamping `dnPhyloCTMC`.


bjoelle (approved):
this looks great, thanks !
it may be good to add a test with the new option, if you have the time

bredelings:
So, on the plus side, I notice you handle continuous character data on trees, which is good.

On the negative side, this approach feels wrong to me statistically.  Or perhaps, insufficiently general.  My intuition says that if we merge then, then in 6 years you will write a paper on on why it is insufficient and/or incomplete.

But I won't block merging.

davidcerny:
@bredelings I guess there could be two problems, one implementation-related and one conceptual:

1. Selectively calling `.setPriorOnly( true )` on only those DAG nodes that are of type "character data" is not the right way to eliminate the influence of character data on the analysis;
2. Eliminating the influence of character data, even if done correctly, is not the right way to get at the "induced prior" or "effective prior" (= tree prior + calibrations) that people want to explore in node-dating.

I don't think problem (2) is much of a problem – it was [you](https://github.com/revbayes/revbayes/issues/463#issuecomment-2086083616) who pointed out that when BEAST users say "run under the prior", what they really mean is "run without the sequence data". But feel free to correct me.

To safeguard against problem (1), I compared my solution (`suppressCharacterData=TRUE`) to those previously suggested by @hoehna: use `.setValue()` instead of `.clamp()` for calibrations, then run the analysis with `underPrior=TRUE`; or remove `dnPhyloCTMC`, as well as any nodes that end up being disconnected from the model after doing so, and run the analysis as usual. Specifically, I applied all three of these strategies to our node-dating tutorial, ran the analysis for 5 million iterations (sampling every 500), summarized the results using an MCC tree with mean node ages, and compared those ages, as well as the induced distribution of ages for the calibrated node.

`suppressCharacterData=TRUE` yields exactly the same results as `.setValue()` + `underPrior=TRUE`, and both are close to indistinguishable from removing `dnPhyloCTMC`, just as one would hope if the three approaches are in fact sampling from the same distribution:

Screenshot 2024-07-22 at 9 59 35
Screenshot 2024-07-22 at 10 13 05
<br><br>

Now, success in one case obviously doesn't address your criticism that the solution is not sufficiently general. But in that case, it would be very helpful if you could describe a case where the solution might fail.

davidcerny:
More changes:

1. I've made it possible to restart MCMC(MC) runs performed with `underPrior=TRUE` or `suppressCharacterData=TRUE` from a checkpoint file;
2. I have added a test for the new option as recommended by @bjoelle;
3. The test includes checkpointing, as requested in #519.

bredelings:
> > I guess I'd like to see a statistical argument that for any type of data that we conconceivably have, leaving out "character data" will always be the thing that makes sense.
> Yes, that would be a much more ambitious claim than I'm willing to make – which is why I'm not making it! I'm only trying to address one particular use case that often arises in node-dating analyses, because a series of papers in the early 2010s drew attention to the divergence of the effective prior from the user-specified calibration "priors", and recommended that users run their node-dating analyses without character data first to figure out what their effective prior actually is. BEAST makes that easy; RevBayes – currently – doesn't.

Right now I think the documentation reads like it IS making that claim.  Can you update the documentation to make it more clear that we are not claiming that leaving out character data is the right thing to do?

davidcerny:
@bredelings I rephrased the final paragraph of the documentation in light of our conversation this morning as best I could. Please check whether it reads okay to you now.

src/core/analysis/mcmc/Mcmc.cpp
```diff
@@ -554,11 +571,42 @@ void Mcmc::initializeSampler( bool prior_only )
...
+        
+        if (suppress_char_data)
+        {
+            std::string node_value = the_node->getValueAsString();
+                
+            /* Whenever character data are printed to the screen, the string "character matrix"
+             * is always included: this is true of both discrete data (e.g., DNA, discrete
+             * morphology) and of continuous data.
+             */
+            bool chardata_found = node_value.find("character matrix") != std::string::npos;
```
bredelings, line 583: It looks like this string "character matrix" comes from `core/datatypes/phylogenetics/characterdata/AbstractCharacterData.cpp`.  Instead of print the node value, it would be cleaner to check if the value is derived from AbstractCharacterData using dynamic_cast.

src/core/analysis/mcmc/Mcmc.cpp
```diff
@@ -783,8 +846,40 @@ void Mcmc::initializeSamplerFromCheckpoint( void )
...
     {
+        if (suppress_char_data)
+        {
+            std::string node_value = node->getValueAsString();
+                
+            /* Whenever character data are printed to the screen, the string "character matrix"
+             * is always included: this is true of both discrete data (e.g., DNA, discrete
+             * morphology) and of continuous data.
+             */
+            bool chardata_found = node_value.find("character matrix") != std::string::npos;
```
bredelings, line 857: Also here.

src/core/analysis/mcmc/Mcmc.cpp
davidcerny: Hmm... I'm having trouble figuring out how to access the node value for casting, since `DagNode` doesn't have a `getValue()` function like `StochasticNode` does. If you're willing to make the change yourself by any chance, please feel free to commit to the branch!

help/md/mcmcmc.md
```diff
@@ -7,10 +7,12 @@ The Mcmcmc analysis object keeps a model and the associated moves and monitors.
 ## details
 The Mcmcmc analysis object produced by a call to this function keeps copies of the model and the associated moves and monitors. The Mcmcmc analysis object is used to run Markov chain Monte Carlo (Mcmcmc) simulation on the model, using the provided moves, to obtain a sample of the posterior probability distribution. During the analysis, the monitors are responsible for sampling model parameters of interest.
 
-An MCMCMC analysis is initiated using the `mcmcmc.run()` method.  
-The `StoppingRule[]` argument provides a mechanism to automatically terminate a run once a set of rules are met: perhaps once the run has attained convergence, or after a certain amount of time has passed.  The run will be terminated once *all* convergence rules ([`srGelmanRubin()`], [`srGeweke()`], [`srMinESS()`], [`srStationarity()`]) have been fulfilled; or once *any* threshold rules ([`srMaxTime()`], [`srMaxIteration()`]) are met.
```
bredelings, line 11: What removing the comments about StoppingRule intentional?

davidcerny: Hmm? Nothing of substance was removed, just rephrased in accordance with changes made to `mcmc.md` by PR #505.

src/core/analysis/mcmc/Mcmc.cpp
bredelings: I thought of a few approaches but none of them worked.  Essentially you are looking for the type of a node, but the DagNode class does not expose the type.  Extracting the type by printing the value is going through the back door, and would only work for some types.

We could maybe try casting from DagNode to StochasticNode<AbstractCharacterData>, but I think that StochasticNode<T> does not inherit from StochasticNode<AbstractCharacterData> even if T inherits from AbstractCharacterData.

The fact that DagNode doesn't allow looking up the type of a node makes me wonder if (a) we should expand the DagNode interface or (b) we should not be doing that.

bredelings:
OK, so here's my list of concerns:

One issue with this PR is that we check if a node counts as "character data" by printing the _value_ to a string, and then checking if the string contains the sub-string "character matrix".  I don't know of a better way to do this off-hand.  Possibly some more work could be done to make the type accessible from the DagNode.  But multiple different data types would probably need to be characterized as "character data", so we would need a list.

The other issue is that you could have fossil data, character data, and other data.  In that case, suppressing only the characters might not be  correct.  In that case, if you want to see the effect of fossil data on the analysis you probably want to look both (i) fossils-only and (ii) fossils plus non-character data.  Since there are more than just the two alternatives (with fossils, without fossils) to look at, using a boolean option (which only looks at two alternatives) doesn't seem correct.

More generally, when you have lots of types of data (for example, DNA data, amino acid data, continuous trait data, discrete trait data, fossil data) you need to look at how your estimates changes when you leave out different categories of data.  For example, if continuous-trait data and DNA data lead to very different divergence-time estimates (which seems to have occurred in issue #565) it is important to be aware of this.  The right solution there is to comment out lines of the model to remove data of different classes, and this seems to be something we should be putting in the tutorials.  The need to consider the effect of fossil data is an example of this more general problem.  Thinking about the problem in this way seems like it could be important for when RevBayes is used for teaching.

OK, that's my spiel.  What do other people think?

wrightaprilm (approved):

davidcerny:
My point of view:

(1) The solution proposed here is simple, transparent about what it does, and works as intended. Given all this, and given that there's no obviously better alternative, I'm not too concerned about its inelegance. I would be concerned if detecting character data in this way was prone to yield false positives or false negatives, but that has not been demonstrated.

(2) I, too, agree that suppressing character data is not always the right and/or only thing to do, and the documentation explicitly points that out. The availability of this option doesn't take away users' freedom to conduct alternative or additional sensitivity analyses. It just makes it easy to run one that's very common and recommended by multiple studies, but much more difficult to set up in RevBayes than in competing software such as BEAST or MCMCTree.

(3) This PR solves an open issue that multiple people have run into. It can be enlightening to think about how to futureproof RevBayes and guard against hypothetical scenarios, but I don't believe this should come at the cost of making actual, real-world analyses that users want to run in the here and now unreasonably complicated.

bredelings:
Just to double-check, what makes something count as character data again?  Is this a way of referring to data that is mapped to nodes in a tree?

mikeryanmay:
I like the idea of having a more user-friendly way of running analyses without character data. It's really frustrating teaching people how to do this, because commenting out "likelihood functions" (as has been pointed out) often leads to disconnected graphs and you end up having to edit a whole bunch of stuff to make your script work. The solution presented here is to selectively suppress dag nodes that include the phrase "character matrix" when their value is printed as a string (intended for screen printing). While this appears to get the job done, it feels a bit fragile and narrow to me. As has been noted, it also doesn't let people exclude particular types of data, as it is an all-or-nothing strategy.

Personally, I would prefer a more generic and robust solution. For example, we could add a method to `RevLanguage::StochasticNode` that causes the underlying `RevBayesCore::StochasticNode` to return 0 when it is asked for its probability. This would not rely on string matching for "character matrix", would allow users to generically exclude particular nodes, wouldn't depend on potentially fraught philosophical arguments about what counts as data, and would be relatively easy to implement (I think). Of course, using this method would be a little bit more work than flipping a boolean, but I feel like that's what you get with a scripting language (though people are probably going to be following a tutorial in any case).

bredelings:
Copying this from slack to avoid losing it. (more discussion on slack)

We (David, Rachel, Laura, April, and myself) had a very interesting and eye-opening discussion.  Some issues that came up were:
- if you comment out both `y.clamp(...)` and `y ~ distribution` then the model can become disconnected.  RevBayes will then complain that some of your moves operate on variables that are not in the model.
- if you comment out just the `y.clamp(...)` part ---- we aren't quite sure what it does?  The observations are removed, but it seems like y is now part of the MCMC state, which could make it mix more slowly.
- @amwright suggests writing a tutorial and paper about how drop out different data types (dna, discrete morphological, continuous morphological, fossil ages, etc.) when looking for conflicting signals between them.
- @mikeryanmay suggests allowing the user to specify a list of nodes to suppress, which should still be easy to use, but is more general and avoids some of the possible issues with deciding what to suppress.
- @davidcerny would also like to flag the question raised by : what happens to fossil ages when you run an FBD analysis with `underPrior=TRUE`? If the ages are clamped in the script, they will become unclamped, but what does that mean in practice? Will they become fixed? Laura says no. Will they still be sampled but become completely unconstrained?
- @rachelwarnock is also comparing BEAST X/BEAST 2 and RevBayes implementations of the FBD and hasn't yet got them to match.  It would be interesting to add tests that the implementations yield the same result given the same parameters and data. @bredelings is interested in whether we could put tests for the FBD tree probability into the current testiphy test suite.

bredelings:
In theory if you have a node `y ~ distribution` with no downstream observations, then you can both (i) remove it from the MCMC state (ii) draw it from the posterior by simple simulation each iteration right before you log it.

When creating a `model` (which copies the DAG) we could therefore flag any nodes with no downstream clamps as simulation-only.  We could decide to keep nodes in the model if they have moves that affect them.

bredelings:
Hmm.. so if we have 
```
x ~ dist
y ~ dist(x)
z ~ dist(y)
y.clamp(...)
```
then setting the probability of y 1 seems not correct.

bredelings:
There are also a few different interfaces we could use.  One would be like:
```
x ~ dist
x.clamp(data)
x.suppress()
```

Another would be more like
```
mymodel = model(x, drop=c(y,z))
```

@mikeryanmay Any others?

mikeryanmay:
My original idea was more like x.suppress(), because I was thinking of a suppressed node as a special type of dag node, which you would specify as you made the object. I also felt like this solution would be more straightforward to implement. However, as you say it would allow people to suppress nodes with dependents, which is probably undesirable. I suppose we could add the restriction that x.suppress() only works on clamped nodes (with an error message), which are unlikely to have dependents, but there will be corner cases where clamped nodes do have dependents (eg [state-dependent branch-rate models using data augmentation](https://revbayes.github.io/tutorials/cont_traits/state_dependent_bm.html)). It seems like it might be complicated to enforce the behavior that suppressed nodes can't have dependents in this setup---we would probably have to end up deferring this check to addParameter for the dependent node.

I am coming around to the idea of providing a list of suppressed nodes to model. For one, it allows us to check/enforce special behavior for nodes with dependents (whether or not they are clamped) or anything else we deem is necessary. For another, it puts all of the suppressed nodes in one place in a user's script, which will probably be less error prone.

Aside from these two, no other alternatives really jump out at me. I think if we are to provide a list, the model object is the right place to do it, rather than the analysis object.

I feel like we should choose one of these two solutions and implement it. I have a slight preference for the list solution. I think implementing both is doable, but I actually feel that having multiple ways of doing the same thing (with potentially slightly different behavior) can sometimes lead to confusion.

davidcerny:
While the basic idea behind this PR (i.e., calling`.setPriorOnly()` on some nodes but not others) seems useful, I've come around to the view that limiting it to nodes associated with (automatically detected) character data is not flexible enough, and that the right solution is to provide some sort of Rev interface for the `.setPriorOnly()` method. So my original thinking was also along the lines of `x.suppress()`, but I think providing a list of nodes to be suppressed to the `model` function is a better idea – it's great we now all seem to be in agreement on that. It forces the user to spell out exactly what they want to do, whereas having to add `.suppress()` calls to previously defined nodes one by one seems error-prone.

I'm closing this PR. Any subsequent discussion about the implementation details of this alternative solution can take place on our new Discussions page: https://github.com/revbayes/revbayes/discussions/624.

Issue #516, opened 2024-07-23
[BUG] FBD analyses w/ FossilTipTimeSlideUniformProposal hanging after restart


davidcerny:
**Describe the bug**
When the `mvFossilTimeSlideUniform` move is applied to all fossils simultaneously, the initial analysis will run fine, but will hang indefinitely if an attempt is made to restart it from a checkpoint file.

**To Reproduce**
This chunk of Rev code from the `model_FBDP.Rev` script in the [combined-evidence FBD tutorial](https://revbayes.github.io/tutorials/fbd/fbd_specimen.html) is the culprit:

```
for(i in 1:fossils.size())
{
    t[i] := tmrca(fbd_tree, clade(fossils[i]))

    a_i = fossils[i].getMinAge()
    b_i = fossils[i].getMaxAge()

    F[i] ~ dnUniform(t[i] - b_i, t[i] - a_i)
    F[i].clamp( 0 )
}

# Add a move to sample the fossil times #
moves.append( mvFossilTimeSlideUniform(fbd_tree, origin_time, weight=5.0) )
```

Interestingly, if we place the move inside the loop and only apply it to one fossil at a time, like this:

```
for(i in 1:fossils.size())
{
    t[i] := tmrca(fbd_tree, clade(fossils[i]))

    a_i = fossils[i].getMinAge()
    b_i = fossils[i].getMaxAge()

    F[i] ~ dnUniform(t[i] - b_i, t[i] - a_i)
    F[i].clamp( 0 )

    # Add a move to sample the fossil times #
    moves.append( mvFossilTimeSlideUniform(fbd_tree, origin_time, fossils[i], weight=5.0) )
}
```

the restarted analysis will run just fine. Similarly, if we replace `mvFossilTimeSlideUniform` by the closely related `mvFossilTipTimeUniform`, there are no checkpointing issues, regardless of whether the move is applied to a single fossil (inside the loop) or all fossils simultaneously (outside the loop).

**Expected behavior**
We should be able to restart the analysis from a checkpoint file regardless of whether `mvFossilTimeSlideUniform` is applied to a single fossil or all fossils simultaneously.

**Computer info**
macOS 14.5 (Sonoma), ARM64 architecture, RevBayes 1.2.5-preview, latest version of the development branch (ac87f90).

bredelings:
Can you figure out what it is doing while it is hanging?  If you compile a debug version, run it inside gdb, and do CTRL-C, you should be able to see what loop it is getting stuck in.

[...]

PR #517, opened 2024-07-23
Another checkpointing fix for FossilTipTimeSlideUniformProposal 

davidcerny:
This PR fixes issue #516.


bredelings:
Hmm... that is confusing.  I'm not seeing why changing from a pointer to a reference would have any effect.  What am I missing?

davidcerny:
Yep, honestly I have no idea why it works, but it does. ¯\_ (ツ)_/¯ Only figured it out by comparing the `doProposal` methods of `FossilTipTimeSlideUniformProposal` and `FossilTipTimeUniformProposal` line by line...

bjoelle:
doesn't work for me - with seed(123) it still hangs at the same place

bjoelle:
issue seems two-fold - 1) rounding error on the tip age which means that a tip age can fall below its specified age range + 2) improper handling of the SAs in the operator where it uses the parent age as the max age instead of the grandparent age

bjoelle:
problem (2) is also present in `FossilTipTimeUniformProposal` actually, but that one never checks whether max_age > min_age so it happily keeps going regardless

so much duplicated code between these two proposals too....

bjoelle:
ok this should fix it
 * Note that the test output changed due to my changes to the proposals, so it would be great if someone could check my changes carefully
 * I'm not sure if the pointer changes are necessary but left them in for coherence with the other proposal (these two should probably be merged at some point)
 * there doesn't appear to be a good way to control whether the tree ages should be forcibly adjusted, at least not without further extensive changes - `setValue` has an argument `f = false` (force ?) which appears like it could serve for this but as far as I can tell it's never used at the moment => further comments and feedback on this are welcome

davidcerny:
Ah, that's what I get for not running the analyses with a fixed seed! So I was wrong both about the source of the problem and the key difference between `FossilTipTimeSlideUniformProposal` and `FossilTipTimeUniformProposal` (it was the age check rather than the handling of the `node` variable), _and_ when I thought my solution worked – because the problem is stochastic and I just got lucky the few times I ran the analysis after making the changes. Good to know!

I will look further into these changes either later today or tomorrow. I also agree it would make a lot of sense to consolidate these two moves to prevent unnecessary code duplication, and can give it a shot once we're satisfied with the current implementation. (I'm also not happy with what these moves are called on the Rev side of things; they have different sets of acceptable aliases.)

davidcerny:
Finally got a chance to go over the changes in depth. It all looks good to me. Just a few questions to make sure I understand what's going on:

1. Previously, we never allowed `taxon_max_age` or `provided_max_age` to be greater than `parent_age`. Now, we allow it if and only if the node in question is a sampled ancestor. Did I get that right?

2. @bjoelle In one of your comments above, you mention that `FossilTipTimeUniformProposal` doesn't check whether `max_age` > `min_age`, unlike `FossilTipTimeSlideUniformProposal` which does. But it's not clear to me where the latter does this?

3. Also, do I understand correctly that even after your changes, this check is still not performed within `FossilTipTimeUniformProposal`, but rather offloaded onto `AbstractRootedTreeDistribution::setValue`? I.e., if `max_age` < `min_age`, then `my_new_age = min_age + (max_age - min_age) * rng->uniform01()` will be less than `min_age`, and will fail the taxon age check test that was added to `setValue`.

bjoelle:
@davidcerny thanks for checking ! to answer your questions
1. yes correct, in case of a SA the limit is the age of the _grandparent_ (because parent_age = tip_age by definition in that situation)
2. so my comment isn't very clear - there is no explicit check for `max_age < min_age` but the loop `while ( new_age < min_age || new_age > max_age )` will go on forever if it happens (so `FossilTipTimeSlideUniformProposal` hangs, unlike `FossilTipTimeUniformProposal` which just fails silently). I guess I could add an explicit check but this situation should never happen if the taxon ages are set correctly.
3. yes correct I didn't add an explicit check to either proposal, but it's not a bad idea to do so any way so I will add it

davidcerny:
Perfect, thanks so much @bjoelle! I think this is ready for merging. Or we can wait for someone else's review, if you prefer.

bjoelle (approved):

Issue #521, opened 2024-07-23
[Help Wanted] Moves seemingly never tried


ms609:
**Describe the desired behavior.**
I'm running an MCMC analysis and noticing that almost no moves are being accepted.  

It's not impossible that this reflects an underlying issue with my model.  What strikes me as odd, however, is that certain move combinations are never being tried, despite high weights.

**What is going wrong?**

Here's an extract from my script:
```rb
moves.append( mvScale(tree_length, tuneTarget = 0.26) )
moves.append( mvScale( u, lambda = 1, weight = 20.0, tuneTarget = 0.26) )
# [...]
mymc3 = mcmcmc(mymodel, monitors, moves, nruns = 2, nchains = 6, tuneHeat = FALSE, 
  swapMethod = "both", swapMode = "multiple", combine = "mixed")
mymc3.run(generations = 40000, tuningInterval = 2000)
mymc3.operatorSummary()
```

The run is going hundreds of generations without accepting any proposals.  Here's the operator summary for chain 1; the picture's basically the same for the other chains too.

```
                  Name                  | Param              |  Weight  |  Tried   | Accepted | Acc. Ratio| Parameters
===============================================================================================================================
NNI                                      topology                14.5000      28791         16      0.0006
SubtreePruneRegraft                      topology                 2.9000       5800          4      0.0007
Scaling                                  tree_length              1.0000          0          0      0.0000 lambda = 0.0010
BetaSimplex                              rel_br_lengths          29.0000          0          0      0.0000 alpha = 100.0000
DirichletSimplex                         rel_br_lengths           2.9000          0          0      0.0000 alpha = 100.0000
Scaling                                  u                       20.0000          0          0      0.0000 lambda = 0.0010
```

I'd like to know what to make of the zeroes in the "Tried" column.  It looks like only tree rearrangements are being attempted, and other moves are ignored, despite being in the move schedule.  The "Weight" corresponds to the originally specified weights, which I presume means that the weights have not been subjected to tuning, though the alpha/lambda values have been.  I expected the values of "Tried" to have the same(ish) ratio as the values of "Weight".

**Computer info**
Win 10, RevBayes 1.2.5

bjoelle:
could you upload the whole script so we can try it ourselves ? thanks !

ms609:
I'll send a copy by e-mail.

bjoelle:
[...]

PR #522, opened 2024-07-24
Correct operatorSummary for MC^3 when moves are tuned

davidcerny:
This PR should fix issue #521 by making sure MC^3 counts attempted and accepted moves correctly. If the problem is more fundamental than just bad counting, then this should at least make it easier to diagnose it.

### DIFF
--- a/src/core/analysis/mcmc/Mcmcmc.cpp
+++ b/src/core/analysis/mcmc/Mcmcmc.cpp
@@ -631,7 +631,7 @@ void Mcmcmc::nextCycle(bool advanceCycle)
  *
  * \param[in]     o     The stream to which we print the summary.
  */
-void Mcmcmc::printMoveSummary(std::ostream &o, size_t chainId, size_t moveId, Move &mv) const
+void Mcmcmc::printMoveSummary(std::ostream &o, size_t chainId, size_t moveId, Move &mv, bool current_period) const
 {
     
     std::streamsize previousPrecision = o.precision();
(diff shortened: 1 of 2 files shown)


bjoelle (approved):
looks good, thanks ! @davidcerny do you want to merge this now or wait for a test to be added ?

ms609 (commented):
Yes, this fixes the initial issue for me too.

I've noticed two possibly related issues when conducting a run in two stages, i.e.
```rb
# Stage 1
mymc3.run(generations = 5000)
mymc3.operatorSummary()

# Stage 2
mymc3.run(generations = 35000)
mymc3.operatorSummary()
```

- [x] In the mcmcmc.run() call in stage 2, the "Iter" count starts again from zero; in `mcmc`, the iteration count does not reset, but continues counting from 5000 - the desired behaviour, I think

- [ ] In my test file, in calls to `mcmcmc.operatorSummary()`, The "MCMCMC chains swapping between" section of the operator summary lists a small number of accepted swaps in  stage 1, but zero accepted swaps in stage 2.  It's possible that this is because the chains really aren't mixing – but I thought it worth flagging in case there's a related counting issue at play.  The number of "Tried" seems to be correct.

davidcerny:
Thank you both!

1. @bjoelle I think I'd prefer to merge this now if that's okay. It looks like there is currently no way to test specifically for the operator summary without messing with the test runner, which I don't feel entirely comfortable doing. If / when it becomes possible, I'd be happy to add an MC^3 validation test.

2. @ms609 The resetting of the iteration counter should now be fixed as well.

3. @ms609 Unfortunately, I think the poor acceptance ratios during your second run are just a feature of your model / set-up. I can't reproduce the problem with the following toy example:

```
seed(12345)

p ~ dnUniform(0, 1)
n <- 20
k ~ dnBinomial(n, p)
k.clamp(7)
mymodel = model(k)

moves = VectorMoves()
moves.append( mvSlide(p, delta=0.1, tune=TRUE, weight=1) )

monitors = VectorMonitors()
monitors.append( mnScreen(printgen=10) )

myMcmcmc = mcmcmc(mymodel, monitors, moves, nchains=4, deltaHeat=5)

# Stage 1
myMcmcmc.run(generations=500, tuningInterval=50)
myMcmcmc.operatorSummary()

# Stage 2
myMcmcmc.run(generations=500, tuningInterval=50)
myMcmcmc.operatorSummary()
```

src/core/analysis/mcmc/Mcmcmc.cpp
```diff
@@ -620,6 +620,16 @@ void Mcmcmc::nextCycle(bool advanceCycle)
         if (chain_prev_boundary[i] == boundary::hottest)
             heat_visitors[heat_rank].second++;
     }
+    
+    for (size_t i = 0; i < num_chains; i++)
+    {
+        
+        if ( chains[i] != NULL )
+        {
+            setCurrentGeneration(current_generation);
```
bredelings, line 629: Running this num_chains times seems redundant?  Did you mean something like `chains[i]->setCurrentGeneration(current_generation)`?

Is this to avoid having the current generation reset to 0 for each chain?

davidcerny: So for some reason, `chains[i]->setCurrentGeneration(current_generation)` did not work for me – only this did. Lemme check if I can pull it out of the loop though...

davidcerny: Yes indeed, I can. Thanks @bredelings!

davidcerny:
Hmm... so while this fixes the iteration count problem for `mnScreen`, `mnModel` still behaves weirdly, skipping a lot of iterations. But that's not new – if anything, it's a very slight improvement over the current state of affairs, where stage 2 simply overwrites stage 1. So I'd still be in favor of merging this, and opening a separate issue for the interaction between `mcmcmc` and `mnModel`.

Issue #524, opened 2024-07-25
[Help Wanted] Accessing current values of parameters during an MCMC run


ms609:
**Context**
I'm trying to understand the behaviour of a misbehaving model.

Here's some relevant code illustrating one aspect of the model:
```rb
u ~ dnGamma( shape = 1, rate = 1 )
moves.append( mvScale( u ) )
u # Initial value is stochastic

condRates := [ [ 0, u],
               [ 4, 0] ]
Q4cond := fnFreeK(condRates)


m_morph[2] ~ dnInverse( dnPhyloCTMC(tree = phylogeny, Q = Q4cond, type = "Standard")

condChars <- readDiscreteCharacterData("data.nex")
m_morph[2].clamp(condChars)
m_morph[2].lnProbability()
```

I'm keen to sanity-check the relative likelihood values of each component of the model.  Hence I'm doing:

```rb
mymc3 = mcmc(mymodel, monitors, moves, nruns = 2, moveschedule = "single")
for (i in 1:40) {
  mymc3.run(generations = 10) # A small increment
  print( "u = ", u )
  print( "P1 =", m_morph[1].lnProbability(), "P2 =", m_morph[2].lnProbability() )
}
```

My MCMC monitor shows that new values of `u` are being proposed, but the `print()` command only shows the original value of `u`, and the original value of `m_morph[2].lnProbability()`; these are not updating to reflect the current state of the chain.

**Desired outcome**
I'd like to access the current value of `u`, and the current status of `m_morph[2]`, from within RevBayes, either via the console, or by writing calculated values to a monitor 
(e.g. by `mnScreen( u, m_morph[2].lnProbability() )`)

davidcerny:
So for `u`, you should just be able to add it to your monitors, no? Like this:

```
monitors.append( mnModel(filename="mymodel.log", u, printgen=1) )  # print every single iteration
monitors.append( mnScreen(printgen=1, u) )                                             # likewise
```

Since `u` is a stochastic node, it should even be included in the `mnModel` monitor by default, I believe. Perhaps I'm misunderstanding your problem?

When it comes to checking the individual components of your likelihood, your best shot is probably to define a deterministic variable and explicitly tell RevBayes to monitor it:

```
m2_logL := m_morph[2].lnProbability()
monitors.append( mnScreen(printgen=1, u, m2_logL) )
```

Unfortunately, this [doesn't always work](https://github.com/revbayes/revbayes/issues/425).

ms609:
[...]

PR #534, opened 2024-08-15
Document `model`

ms609:
Outline the basic use of `model()`, and highlight the `.graph()` method.

The function is not so complicated, but I thought it worth noting that it works with copies of workspace objects (per #524) – perhaps this doesn't necessarily earn a place in the example, but I figured, why not angle for maximum clarity and include this in the example code as well as §Details.

### DIFF
--- a/help/md/model.md
+++ b/help/md/model.md
@@ -1,9 +1,43 @@
 ## name
 model
 ## title
+Create a model object
+
 ## description
+Creates a model object that can be graphed or subjected to Bayesian inference.
+
 ## details
+`model(x)` creates a model object by creating a copy of all elements and 
+parameters that influence or are influenced by the likelihood of `x`.
+
+Because `model` works with copies of objects, conducting an mcmc(mc) analysis
+on a model object will not change the values of the objects in the RevBayes
+workspace.
+
 ## authors
 ## see_also
 ## example
+    # Create a simple model (unclamped)
+    a ~ dnExponential(1)
+    b ~ dnExponential(a)
+    mymodel = model(b) # equivalent to model(a) or model(a, b)
+    
+    # Save a DOT visualization of the model to file
+    mymodel.graph("mymodel.dot")
+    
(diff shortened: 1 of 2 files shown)


davidcerny (changes requested):
I think you need to run `./help/md2help.pl` to add the changed content of `./help/md/model.md` to `./src/core/help/RbHelpDatabase.cpp`. This makes it accessible from the console.

ms609:
When I run  `$ perl md2help.pl` (on Windows 10)

I see
```
/**
 * This file was generated automatically.
 * It is not intended to be human readable.
 * See help/README.md for details.
 */

#include "RbHelpDatabase.h"

using namespace std;

RevBayesCore::RbHelpDatabase::RbHelpDatabase()
{
}
```

No files (at least none tracked by git) are modified.

davidcerny:
You're right; I don't think the script works as described / intended. I had to run it like this: `perl md2help.pl md/model.md`, and even then it just printed text to the console that I then had to insert into `RbHelpDatabase.cpp` manually. Could you please do that? Happy to merge afterwards.

davidcerny:
Good point! Let's bring that conversation to Slack, perhaps?

davidcerny (approved):

davidcerny:
Hmm... it looks like the replacement of `help_arrays` by `help_strings` in the regenerated `RbHelpDatabase.cpp` file breaks the "References" and "See also" sections in some cases – they won't even show up in the console. (E.g., compare the output of running `?dnBivariatePoisson` on this branch and on development.) Not merging this quite yet.

bredelings:
I regenerated the help database and squashed some of the changes.

I wasn't able to push to this branch on @ms609 's repo -- I pushed to a new branch and made a new PR #538.  

I'm not quite sure why I was able to generate the missing sections.  I ran `projects/cmake/build.sh` and also `projects/meson/generate.sh`.

Issue #492, opened 2024-07-11
[Feature Request] Inference under a conditional


ms609:
# The problem

I am developing a model in which I observe data x under the model M(θ) with parameter θ, and know that the data x satisfy condition C.

I wish to perform inference under the conditional, which requires the calculation of 

π(x | θ, C) = Pr(x | θ) / Pr({X satisfies the conditional} | θ)

Pr(x | θ) can be computed with 
```rb
m_morph ~ dnPhyloCTMC(tree = phylogeny, ...)
m_morph.clamp(x1)
m_morph.lnProbability()
```

Pr({X satisfies C} | θ) is given by
```rb
conditional ~ dnPhyloCTMC(tree = phylogeny, ...)
conditional.clamp(x2)
conditional.lnProbability()
```

Hence π(x | θ, C) is given by
```rb
likelihood = m_morph.lnProbability() - conditional.lnProbability()
```

So far, so good. Now I wish to run an MCMC(MC) using the conditioned likelihood.

```rb
mymodel = model(phylogeny)
mymcmc = mcmc(mymodel, ...)
mymcmc.run() # will assume likelihood = m_morph.lnProb() + conditional.lnProb()
```

# The solution

The key function is `getModelLnProbability` in `src/core/analysis/mcmc/Mcmc.cpp` (and its counterpart in `Mcmcmc.cpp`). 

```c++
double Mcmc::getModelLnProbability(bool likelihood_only)
{
    double pp = 0.0;
    
    const std::vector<DagNode*> &n = model->getDagNodes();
    for (std::vector<DagNode*>::const_iterator it = n.begin(); it != n.end(); ++it)
    {
        
        DagNode *the_node = *it;
        if (likelihood_only == false)
        {
            pp += the_node->getLnProbability();
        }
        else if (the_node->isClamped() == true )
        {
            pp += the_node->getLnProbability();
        }

    }
    
    return pp;
}
```
 
The key question here is how to identify a distribution as a conditional. We need to either an additional `if` statement to replace `pp +=` with `pp -=` when a node is identified as representing a conditional distribution, or to design a conditional node such that its `getLnProbability()` method returns a negative value.

Possibilities include:

1. Specify something in the `model()` object that identifies nodes as conditional.
  This seems like a non-starter – it is too late in the workflow, and would require a dramatic change to how the user interacts with that function; specifying an arbitrary node from the model graph would no longer suffice.
[...]

Issue #444, opened 2024-04-09
Debug build is too large to run on Windows


bjoelle:
This isn't a critical issue for users but I've just spent two days troubleshooting this so I want to save it for future reference
Apparently Windows has a size limit on executables that it will run (somewhere around 3.5G) and the debug build is currently above so it won't start, with a very cryptic error message
Release build works fine on the same machine & setup (Windows 10 64bits Home edition, MinGW compiler)

I _think_ this could be solved by building some of it (e.g. the libs folder) as a shared library instead but I don't know enough cmake to do this properly

bredelings:
Can you add a link to something that discusses this size limit?  I haven't found anything yet.

I can try making the libraries into DLLs / shared objects in meson.

bjoelle:
yes, those are the links I found when troubleshooting
it also fit with my issue - the build worked fine in both release and debug, then a new PR added a few files and the debug build no longer ran (with the error in your second link) whereas the release was fine, and both were fine on Linux
based on some other googling I tried looking at the debug strings and the largest ones seem to come from boost templates, so if there is a way to configure the debug build to only make debug info for RB files and not the dependent libraries I guess that would also work, but this is such a weird problem I found very little info online about it

bredelings:
In meson (and probably CMake) converting static to shared libraries is as simple as replacing `static_library( )` with `shared_library( )`.  Then it seems that we can no longer have two libraries A and B that each depend on the other.  So, we make rb-core, depend on rb-libs, and rb-revlanguage depend on rb-core.

Then the interesting part begins:
- libs/tensorphylo/Loader.cpp uses RbException.  We can fix this by replacing all ~4 uses with std::runtime_exception
- src/core depends on src/revlanguage in a few places, which causes linker errors.  I found a few possible places by running `git grep 'include.*Rl'` in `src/core`

ms609:
As a partial workaround, I've got a debug build running on a Windows machine (albeit on a Linux environment) in VSCode using [WSL](https://code.visualstudio.com/docs/remote/wsl).  No help for debugging Windows-specific issues, of course, but it does mean I'll be able to start debugging using my Windows PC.

PR #535, opened 2024-08-15
dnInverse

ms609:
This pull request addresses #492 by adding a new distribution function `dnInverse(x)` that inverts the probabilities associated with the distribution `x`.

Queries:
- Does this belong as a `dn` function?  It has some parallels with the new `tn` transformation functions, though it cannot be straightforwardly accommodated in the TransformedDistribution class
- How can the distribution access the underlying properties (e.g. `taxa`) of a distribution of type `AbstractCharacterData` in order to correctly print its output?

I believe that the implementation is mathematically sound and functions, so think this could be ready to merge – though it would be nice to also address these outstanding issues, for which input from experienced developers would be welcome!


ms609:
I've discovered an issue with the implementation, which I don't see an obvious way to address.  The probability is calculated correctly on the first call to `.lnProbability()`, but not on subsequent calls.  Here's a revised test file:

```rb
seed(42)

exp ~ dnExp(1)
exp.clamp(42)

inv_exp ~ dnInv(dnExp(1))
inv_exp.clamp(42)
print("Expect: 42", -exp.lnProbability(), inv_exp.lnProbability())

chars <- readDiscreteCharacterData("../validation/data/bears.nex")

taxa <- chars.names()
n_taxa <- chars.size()
n_edge <- 2 * n_taxa - 3

br_lengths ~ dnDirichlet( rep(1.0, n_edge) )

k ~ dnUniform(1, 10)
kRates := [ [ 0, k],
            [ 1, 0] ]
qK := fnFreeK(kRates)
phy ~ dnPhyloCTMC(tree = phylogeny, Q = qK, type = "Standard")
phy.clamp(chars)

invPhy ~ dnInv(dnPhyloCTMC(tree = phylogeny, Q = qK, type = "Standard"))
invPhy.clamp(chars)
invPhy

k.clamp(1)
print(k, "Expect equal:", -phy.lnProbability(), invPhy.lnProbability())
# succeeds
k.clamp(5)
print(k, "Expect equal:", -phy.lnProbability(), invPhy.lnProbability())
# does not succeed: -phy decreases where invPhy increases.
k.clamp(6)
print(k, "Expect equal:", -phy.lnProbability(), invPhy.lnProbability())
# same again
```

Can anyone shed light on what might be happening here?

Thanks!

bjoelle:
when you're doing `setValue` and `redrawValue` it looks as if you're only updating the value for the base disribution, not the value stored by `InverseDistribution` so that may be part of the issue
does it also happen with the `dnExp` variable or just with the more complex model ? can you check that `qK`, ìnvPhy` and `phy` are updated as they should be ?

ms609:
Thanks for taking a look.  

With `dnExp` I see:
```rb
inv_exp ~ dnInv(dnExp(1))
inv_exp.clamp(42)
print("Expect: 42", -exp.lnProbability(), inv_exp.lnProbability())
# Success

exp.clamp(21)
inv_exp.clamp(21)
print("Expect: 21", -exp.lnProbability(), inv_exp.lnProbability())
# Success
```

but this feels slightly different as I'm clamping the value drawn from the distribution rather than one of its parameters – I'm not sure what an equivalent test would look like.

I'll look into your other suggestions anon

ms609:
Thanks for taking a look.  I've worked through your suggestions as best I can:

> `setValue` and `redrawValue`

This is an interesting observation. 
- `setValue` currently updates the base distribution via `dist->setValue()`; I have added `this->value`.
- `redrawValue` currently updates the value stored by `InverseDistribution` (`this->value`).  I have added a call to `dist->redrawValue()` here, to also update the base distribution.  Neither of these additions have fixed the issue.

As `computeLnProbability` only makes use of the base distribution, the value of the  `InverseDistribution` should not be relevant to this calculation. Perhaps I could even remove the code that modifies `this->value` – are there other side-effects of setting `this->value` that might influence the calculation or updating of the model?

> does it also happen with the dnExp variable or just with the more complex model ?

With `dnExp` I see:
```rb
inv_exp ~ dnInv(dnExp(1))
inv_exp.clamp(42)
print("Expect: 42", -exp.lnProbability(), inv_exp.lnProbability())
# Success

exp.clamp(21)
inv_exp.clamp(21)
print("Expect: 21", -exp.lnProbability(), inv_exp.lnProbability())
# Success
```

but this feels slightly different as I'm clamping the value drawn from the distribution rather than one of its parameters – I'm not sure what an equivalent test would look like.

> can you check that qK, invPhy and phy are updated as they should be ?

Printing `qK` confirms that it is updated as expected; can you suggest how I could test that updates to `invPhy` and `phy` are correct?

bjoelle:
sorry I was thinking of `phylogeny` actually, `phy` is clamped so it shouldn't move
but probably the best way is to run all of this in a debugger so you can directly check what values are used to do the likelihood calculation and whether something is not being updated correctly

ms609:
Using a debugger sounds like the way to go.  I'd been shying away from this because I feared the learning curve – fears which have proven well founded!  I just about got things set up when I ran up against the "file too big" issue you reported in #444. Did you eventually find a workaround for that issue?

bjoelle:
Unfortunately my current workaround is to do all my RevBayes-related work on my Linux laptop lol

Also I didn't mention it before but if you manage to make it work and would like some more introduction to debuggers/how things work in RevBayes I'm happy to do so, just tell me and we can have a Zoom call or something

ms609:
That's very generous, thank you!  I've dug out my decade-old Ubuntu machine, which is going to take a while to upgrade to current packages. An orientation session once (if?) I manage to get RevBayes built could be very helpful.

ms609:
Thanks to @bjoelle's patient coaching I've been able to fix the aforementioned issues: `print(invPhy)` now reports the values set in the object itself (copied from the base dist), and overriding the `touch` function to also touch the base distribution means that the likelihood is now calculated correctly.

ms609:
I've got two branches in a tangle here so I'm closing this PR and have created a new one from `revbayes/dnInverse` at #562.

PR #537, opened 2024-08-21
Fix #532

### LINKED ISSUE #532: [BUG] No valid initial state for FBD when specifying `rootAge`
**Describe the bug**
RevBayes is unable to simulate a viable starting tree when the FBD process is conditioned on the root age rather than the origin age. (Thanks to @bjoelle for figuring out this was the source of my initialization troubles!)

**To reproduce**
[This ZIP archive](https://github.com/user-attachments/files/16621166/fbd_init_test.zip) contains two scripts that both draw from the same process-age prior, but pass the resulting value to `dnBDSTP` either via the `originAge` or the `rootAge` argument. The former analysis starts just fine; the latter exhausts the 1000 attempts without finding a valid initial state. One absolute path has to be changed in the scripts.

**Expected behavior**
RevBayes should find a valid initial state regardless of whether we specify `rootAge` or `originAge`.

**Computer info**
macOS 14.5 (Sonoma), ARM64 architecture, RevBayes 1.2.5-preview, latest version of the `development` branch (f2a0610).

davidcerny:
This PR fixes issue #532 by adjusting the root age of the simulated tree. Hacky but arguably acceptable, since we are just trying to find a viable initial state.

### DIFF
--- a/src/core/distributions/phylogenetics/tree/birthdeath/BirthDeathSamplingTreatmentProcess.cpp
+++ b/src/core/distributions/phylogenetics/tree/birthdeath/BirthDeathSamplingTreatmentProcess.cpp
@@ -1491,12 +1491,20 @@ void BirthDeathSamplingTreatmentProcess::redrawValue( SimulationCondition condit
         if ( starting_tree == NULL )
         {
             // SH 20221212: The simulateTree functions hangs in certain situations. It's more robust to use the coalescent simulator.
-//            simulateTree();
+            // simulateTree();
             
             RbVector<Clade> constr;
             // We employ a coalescent simulator to guarantee that the starting tree matches all time constraints
             StartingTreeSimulator simulator;
             RevBayesCore::Tree *my_tree = simulator.simulateTree( taxa, constr );
+            
+            // When conditioning on the root age, increase the root age of the simulated tree to make it equal to the process age
+            if ( use_origin == false )
+            {
+                double new_root_age = getOriginAge();
+                my_tree->getRoot().setAge( new_root_age );
+            }
+            
             // store the new value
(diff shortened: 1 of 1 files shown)


bredelings:
The `redrawValue( )` function takes a simulation condition argument that says whether we are trying to simulate from the correct distribution or just trying to get an initial tree to start MCMC.  Should we be checking that?  (I note that we are already using the coalescent to simulate here...  but maybe that should depend on the simulation condition?)

The other question I had is whether the root and the origin are the same thing.  I kind of thought that at the origin there is a single particle, and the root age is actually the time where this particle splits into two particles.  If so, is there a way to set the origin time instead of the root time?  (If the root time works, that's probably more important, but if can set the origin time separately, that seems nicer.)

Thanks!

davidcerny:
> The other question I had is whether the root and the origin are the same thing. I kind of thought that at the origin there is a single particle, and the root age is actually the time where this particle splits into two particles. If so, is there a way to set the origin time instead of the root time? (If the root time works, that's probably more important, but if can set the origin time separately, that seems nicer.)

That's correct; they are not the same thing, and we can condition BDSTP on either one depending on whether we specify the `originAge` argument or the `rootAge` argument. (Internally, within `AbstractRootedTreeDistribution`, this is handled a bit differently: there is a single `process_age` and a Boolean called `use_origin` that records whether it pertains to the origin or to the root.) Currently, the simulator won't yield a valid initial state if we condition on the root age; this PR fixes that.

Your first point goes back to the discussion we had under PR #453, but it's beyond the purview of this PR, I would say.

bredelings:
Oh, I see.

bredelings (approved):

PR #538, opened 2024-08-22
Doc model

bredelings:
Outline the basic use of model(), and highlight the .graph() method.

The function is not so complicated, but I thought it worth noting that it works with copies of workspace objects (per https://github.com/revbayes/revbayes/issues/524) – perhaps this doesn't necessarily earn a place in the example, but I figured, why not angle for maximum clarity and include this in the example code as well as §Details.

### DIFF
--- a/help/md/model.md
+++ b/help/md/model.md
@@ -1,9 +1,43 @@
 ## name
 model
 ## title
+Create a model object
+
 ## description
+Creates a model object that can be graphed or subjected to Bayesian inference.
+
 ## details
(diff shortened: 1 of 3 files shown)


davidcerny (approved):

davidcerny:
Thanks a lot, Ben; the documentation on this branch is both up-to-date and correctly formatted. It's a minor mystery to me how you managed to generate it, though. For me, simply building RevBayes using cmake did not automatically incorporate all the `.md` files `RbHelpDatabase.cpp` (not even when I did it with the `-help true` flag).

Issue #435, opened 2024-03-21
[BUG] Incorrect Posterior Probability with dnUniformInteger and mvRandomIntegerWalk


minghao-du:
**Describe the bug**
When using `dnUniformInteger` and `mvRandomIntegerWalk`, it is not possible to obtain the correct posterior probability.

**To Reproduce**
I generated 1000 samples ($y_i, i=1,...,1000$) from a Bernoulli distribution with $p=0.1$ and conducted inference on $p$.
$$y_i \sim \text{Bernoulli}(p)$$

$$p_\text{int} \sim \text{UniformInteger}(0,10)$$

$$p \coloneqq \frac{p_\text{int}}{10}$$

However, the inferred posterior distribution of $p_\text{int}$ is almost the same as the prior, uniformly distributed between $0$ and $10$, rather than being mostly $1$.
<img width="1171" alt="image" src="https://github.com/revbayes/revbayes/assets/51898298/6465ee51-f366-44ec-8c02-2f65377e5eee">

When inferring directly using a uniform distribution, there is no problem with the inference.
$$y_i \sim \text{Bernoulli}(p)$$

$$p_\text{int} \sim \text{Uniform}(0,10)$$

$$p \coloneqq \frac{p_\text{int}}{10}$$
<img width="1180" alt="image" src="https://github.com/revbayes/revbayes/assets/51898298/e8a43498-bb23-48a9-88c3-8e99d93d2100">

The scripts used and the output are in this file.
[UniInt_bug.zip](https://github.com/revbayes/revbayes/files/14787938/UniInt_bug.zip)

**Expected behavior**
When inferring using a UniformInteger distribution, the inferred $p_\text{int}$ is $1$.

**Computer info**
I am using a Mac, and the version of RevBayes I am using is development (rapture-2388-ge4baa8).

bredelings:
Note that the Posterior and Likelihood are constant.  So they aren't getting updated when p_int changes.  This is the underlying source of the problem.

I tweaked the script to replace the collection of Bernoulli observations with a single Binomial observation, and the problem doesn't go away.  So this could possibly be a problem with mvRandomIntegerWalk( )... but I'm not sure how the move can prevent the likelihood and prior from changing.

bjoelle:
This seems tied to the `model()` call
 * when using `model(uni_int)` as in the example script, `p_estimated` and the Bernouilli distribution do not get integrated into the model (so `uni_int` updates but then nothing else does)
[...]

PR #545, opened 2024-08-27
Type conversion fix

bjoelle:
This PR is designed to fix the type conversion system to avoid issues such as #435 
Basically instead of doing implicit conversions under the hood, all conversions now go through `TypeConversionFunction` and create an actual `DeterministicNode` in the graph

The goal is to switch all types to using the new system (instead of just `Integer` currently) but I would like some feedback first so I'm leaving this as draft for now


bjoelle:
@bredelings it's because the upstream code in `ArgumentRule::fitArgument` also does special handling on constant nodes (i.e. when the object is constant it replaces it instead of creating a new variable, which destroys the original node and so invalidates the deterministic node)
I could change it throughout but my guess is that deterministic nodes are not needed when the underlying value cannot change, and replacing is likely faster in that situation

bredelings:
> @bredelings it's because the upstream code in `ArgumentRule::fitArgument` also does special handling on constant nodes (i.e. when the object is constant it replaces it instead of creating a new variable, which destroys the original node and so invalidates the deterministic node) I could change it throughout but my guess is that deterministic nodes are not needed when the underlying value cannot change, and replacing is likely faster in that situation

Yeah.  Can you put a comment about this?

bredelings:
I wonder how to think about creating new DagNodes on the fly like this.  Can we assume that each conversion will only happen once?  Could we create a second duplicate conversion DagNode, and then leak the first one?

bjoelle:
this is a good question - at the moment nothing seems to be checking whether the conversion already exists, so I think you could indeed end up with a bunch of duplicated conversion nodes, although I think the chances it would actually happen are quite low
I suppose you could do some sort of model graph parsing to check if a node you want already exists and use that one, but that seems quite convoluted - honestly at that point it may be easier to just simplify the type system already

bredelings:
> this is a good question - at the moment nothing seems to be checking whether the conversion already exists, so I think you could indeed end up with a bunch of duplicated conversion nodes, although I think the chances it would actually happen are quite low I suppose you could do some sort of model graph parsing to check if a node you want already exists and use that one, but that seems quite convoluted - honestly at that point it may be easier to just simplify the type system already

In theory we could create a new conversion node for each function that takes the integer node as input and converts it to something.  That's probably fine -- creating and re-using a single conversion node is more like a CSE optimization.

I think as long as we aren't creating new conversion nodes on successive MCMC iterations it is fine for now.

bjoelle:
that sounds like a good idea, I'm not sure where would be the best place - just as an added doc to the conversion function ?

bredelings:
> that sounds like a good idea, I'm not sure where would be the best place - just as an added doc to the conversion function ?

For now, perhaps that is the best place?

There could be more places that we want to add new deterministic nodes in the future.  If so we could add a utility wrapper function and move the docs there, but for now the conversion function is probably fine.

bredelings (approved):

PR #549, opened 2024-09-04
Correctly resetting the branch lengths via treeAssembly after the corresponding moves being rejected

jsigao:
This PR is supposed to fix an issue with the treeAssembly function when updating branch lengths after restoring. This issue leads to incorrect likelihood computation when the branch lengths are specified as a product of a simplex (the relative branch lengths) and a scalar (the tree length), but becomes innocuous when the branch lengths are specified as a vector where each element is specified as a stochastic node (the absolute length of that branch) or when the move schedule is single and the tree is logged every generation. Under the innocuous scenarios, this issue leads to redundant calculations in updating partial likelihoods and P matrices that have not changed (though the slow down should be rather minor, in roughly 5-20% range).

This PR consists of four commits. Commit [a271975](https://github.com/revbayes/revbayes/commit/a27197570ba0243dc0e98bb63d07745e0f696694) is the major one that introduces additional flags and forces the update function to always be called right after a `restore` in `treeAssembly` (i.e., before entering the next iteration regardless of how frequently the tree is logged). This ensures the branch lengths will always be reset after a move on them being rejected and avoids dirty flags being carried over to the next move.

Commit [462ad97](https://github.com/revbayes/revbayes/commit/462ad977ee2514625d31e5316165fcf7b18859ea) avoids some further redundant calculations on updating the P matrices when we know for sure that the corresponding branch lengths are not altered at the corresponding topological moves.

Commit [a7d5010](https://github.com/revbayes/revbayes/commit/a27197570ba0243dc0e98bb63d07745e0f696694) intends to avoid recomputing the partial likelihoods and P matrices of the children nodes involved in a topological moves when the corresponding subtending branch lengths are not altered. In RevBayes (at least for discrete characters) the partial vector of a (non-root) node (i.e., the corresponding subtree) is supposed to represent the conditional likelihoods at the rootward starting point on the subtending branch of that node, so for a node involved in a tree move that only serve as a child (instead of a parent) the partial should always stay the same as the corresponding subtree is unaltered. Accordingly, the children nodes should not need to be flagged dirty. (I am not entirely confident if this also always applies to continuous characters, though I don't see why it wouldn't.)

I don't feel strongly about the above two commits as the speed up is rather minor (dataset-dependent but probably in the range of 1-5%). If anyone feels those redundancies are better kept to avoid unforeseen consequences I would be more than happy to drop them.

Commit [c90c05a](https://github.com/revbayes/revbayes/commit/c90c05a5fa9cfaf879f0d6365d56b404d062cfe5) changes  the expected output of a test; this mismatch is expected given the bug-fix commit. Previously `update` won't be called after `restore` so branch lengths in the tree logged by stochasticVariable monitor may not match their true values. Examining the the previous expected output, we can actually tell the inconsistency between the values of branch lengths in the recorded tree and their values recorded as separate stochastic variables; that inconsistency disappears after the bug fix. This inconsistency should be innocuous for downstream analyses based on the stochastic-variable logs such as the PPS though.


bredelings:
Does #550 complain that the invariants are broken in your non-innocuous scenario?

jsigao:
Thanks for adding that check @bredelings!

No it does not, which I think is to be expected. Previously I was wrong about the possibility that the ancestor nodes may be left clean (because the descendant node is already flagged dirty before being regrafted) while they should have been flagged dirty after a topological move. It appears that in a topological move RevBayes fires tree changes on the corresponding nodes not only at the `setParent` function, but also at `addChild` and `removeChild`, which seems a bit redundant initially but does avoid the original bug possibility I was suspecting. 

The bug reported here (i.e., non-innocuous scenario) was effectively caused by not resetting the branch lengths in the tree according to the branch-length vector after a branch-length move being rejected when the immediately next move is on topology (because the `update` that is needed after the rejected branch-length move wouldn't be evoked until after the `touch` of this next topological move, which already marks the affected object in `treeAssembly` as topology before that `update` is called).

jsigao:
I just revised some of the commits in this PR as I realized that part of the previous fix relies on the assumption that by the first time the `restore` of `treeAssembly` is called, the parent branch-length vector would have been fully updated; that assumption does not hold under some scenario (e.g., when a vector proposal like `mvVectorScale` is applied).

More generally, because `restore` of a given deterministic node may be executed multiple times at each MCMC generation (i.e., move), `update` has to be either executed after no more `restore` will be called in that generation or together with (i.e., right after) each `restore`.  

In the latter case, if there are touched indices then no matter whether it's cleared by `restore` or `update`, the second time they are executed we would have lost all touched indices (as they should have been cleared at the end of the first execution round). That is, to ensure all the value are always correctly updated, we would have to reassign all values when the touched indices vector is empty.

The former case seems to be more desirable, and was how `treeAssembly` worked previously (i.e., before this PR). The problem was that the `update` might not be called until after the `touch` of the next MCMC move, which will either confound the dirty node flagging or result in branch lengths not being correctly reassigned due to the `touch`-related flags in `treeAssembly` being altered by this next move.

I revised the major commit (now [b6fb05c](https://github.com/revbayes/revbayes/commit/b6fb05c1194319fe959f5e6b9fff2b77c8791c08)) to achieve the main goals that 1) the branch lengths will be correctly reset after a branch-length move being rejected (regardless of when update will be executed) and 2) resetting the branch lengths does not add any unnecessary dirty node flags for likelihood computation. This implementation comes with the cost of possibly resetting some branch lengths when they are not altered, but it dissociates branch-length resetting with tree-node-dirty flagging, while keeping the branch lengths in the tree in sync with the branch-length vector.

I dropped the commit on the expected test output as this revised bug fix reverts back to our previous way that the `update` may not be called until the next move, so branch lengths in the tree logged by the stochastic-node monitor may still be off-synced with the branch-length vector. This is the expected behavior of a deterministic node and is innocuous (i.e., the branch lengths in a `treeAssembly`-produced tree logged by the stochastic-node monitor is a side effect and is not supposed to be interpreted).

jsigao:
After thinking twice, I am still quite ambivalent on whether to keep commit [1673a27](https://github.com/revbayes/revbayes/commit/1673a27fcb4e223e7e573352036604da8d9b5749). This is a minor commit in this PR that intends to avoid some redundant likelihood computation by not flagging the children nodes to be dirty; the partials of these children nodes should be unaltered as their descendant subtrees are not changed when they are pruned and regrafted in a tree proposal.

I checked implementations of the likelihood computation of both discrete and continuous characters, as well as their data-augmented versions. I think they are all designed in a way that all the necessary recalculations should be associated with the parent node (and ancestral nodes when we do pruning) but not with the children clades, so as long as the parent is (recursively) flagged dirty then the computation should be correct.

However, although I'm pretty sure that that should be the case for discrete characters under pruning, I could never be too sure about that across all the relevant distribution classes given the complexity (and I have not been able to do any test other than for the `AbstractPhyloCTMCSiteHomogeneous` class).

The upside of making this change is that we on average would save about 1-20% time (owing to flagging four fewer nodes to be dirty) for evaluating each topological move, while the downside could be huge if I overlooked or misunderstood some of the classes so that the parent-flagging logic does not always hold. To avoid doing any harm, I am leaning toward dropping this commit.

Any comments or suggestions would be highly welcome!

ms609:
This does seem fix a couple of substantial bugs, so it'd be great to see this merged into development in some form.  Is there a clear next step to doing so?  Does the implementation of the `debugMCMC` option mean that any unintended consequences will be easier to track down, if they do arise?

bredelings:
I was wondering if this would fix some of those problems -- thanks for checking!  This is next on my list, along with fixing brownian motion.

bredelings:
Part of the issue here seems to be that the way restore and update are called makes it hard to do things correctly.  Allowing restore and update to be called multiple times causes problems, for example.

bredelings (approved):
I am going to approve this.

Probably we need to do more work later to (a) document the problems here in the source code and (b) change how restore/keep/update work.

PR #550, opened 2024-09-05
Check some invariants for AbstractPhyloCTMCSiteHomogeneous

bredelings:
This PR adds some checks that some invariants assumed in the likelihood calculator are actually true.

Currently the invariants are:
1. If a tree node is dirty, then its parent is dirty.
2. If a tree node has a dirty p-matrix, the it has a dirty conditional likelihood.

Actually, for the second one, I am wondering if it is actually the parent node that should be required to be dirty.


jsigao:
Thanks for adding this @bredelings! This all looks correct to me. Some quick tests did not trigger the assertions though, which is to be expected as I tried to describe in my latest comment under #549.

> Actually, for the second one, I am wondering if it is actually the parent node that should be required to be dirty.

I think in RevBayes the conditional-likelihood vector stored for a node (except the root) is computed at the (rootward) starting point of the subtending branch (right before joining the parent node); i.e., it's supposed to be the conditional likelihoods after multiplying the P matrix on the corresponding branch, so if a node has a dirty p-matrix, then it should also have a dirty conditional-likelihood vector.

jsigao (approved):

bredelings:
Thanks for the info about which side of the branch the conditional likelihoods are stored on.

PR #553, opened 2024-09-10
MCMCMC checkpointing fix

### LINKED ISSUE #540: [BUG] Failure to resume MCMCMC from checkpoint
**Describe the bug**
A completed mcmcmc run does not resume from a checkpoint, but starts afresh.
- Log files are wiped and restarted from zero
- Low initial likelihoods indicate that each run is starting from a new randomly-initialized point, rather than the state of the previous chain

**To Reproduce**

Script files are included in
[project675.zip](https://github.com/user-attachments/files/16709054/project675.zip)

Run `source("675_by_k.Rev")` – `by_k.Rev` is `source`d by that script.

The key lines are:

```rb
mymc3 = mcmcmc(
  mymodel, # a simple Mk-style model on morphological characters
  monitors, moves,
  nruns = 2, nchains = 8, tuneHeat = TRUE,
  swapMethod = "both", swapMode = "multiple", combine = "none"

ckpFile = "675_by_k.ckp" 

# The checkpoint file "675_by_k.ckp" is not actually created, so we use:
if (fileExists("675_by_k_run_1_chain_0.ckp")) {
   # If checkpoint files have been created:
  mymc3.initializeFromCheckpoint(ckpFile)
} else {
  # On first run only, perform a burnin step
  mymc3.burnin(generations = 1500, tuningInterval = 300)
}

mymc3.run(rules = stopping_rules, checkpointFile = ckpFile, checkpointInterval = checkFreq)
```

**Expected behavior**
The run should resume from the state that it left off, and append to existing logs

**Computer info**
Win 10
Tested in 1.2.4 and recent-ish build of 1.2.5

bjoelle:
Should fix #540

### DIFF
--- a/src/core/analysis/mcmc/Mcmcmc.cpp
+++ b/src/core/analysis/mcmc/Mcmcmc.cpp
@@ -477,6 +477,7 @@ void Mcmcmc::initializeSamplerFromCheckpoint( void )
         if ( chains[i] != NULL )
         {
             chains[i]->initializeSamplerFromCheckpoint();
+            setCurrentGeneration(chains[i]->getCurrentGeneration());
         }
         
     }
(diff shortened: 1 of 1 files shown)


ms609 (approved):
Looks good – verified working for me.  Thank you!

PR #555, opened 2024-09-11
Fix Gelman-Rubin and stationarity stopping rules

### LINKED ISSUE #543: [BUG] srGelmanRubin stopping rule doesn't stop run
**Describe the bug**
Convergence diagnostic stopping rules seem not to be working

**To Reproduce**

Run the examples from the stopping rules documentation pages.

```rb
# Binomial example: estimate success probability given 7 successes out of 20 trials
r ~ dnExp(10)
p := Probability(ifelse(r < 1, r, 1))
n <- 20
k ~ dnBinomial(n, p)
k.clamp(7)
mymodel = model(k)

moves = VectorMoves()
moves.append( mvSlide(r, delta=0.1, weight=1) )

paramFile = "output/parameters.log"

monitors = VectorMonitors()
monitors.append( mnModel(filename = paramFile, printgen = 100, p) )
monitors.append( mnScreen(printgen = 1000, p) )

# Stop when the potential scale reduction factor falls below 1.01
stopping_rules[1] = srGelmanRubin(1.5, file = paramFile, freq = 1000)
stopping_rules[2] = srMinESS(15, file = paramFile, freq = 1000)

# Create the MCMC object
mymcmc = mcmc(mymodel, monitors, moves, nruns = 2)

# Begin the MCMC run
mymcmc.run(rules = stopping_rules)
```

**Expected behavior**
The run should stop pretty promptly – an ESS of 15 is trivial, as is a PSRF of 1.5.

But it continues almost indefinitely; Tracer confirms an ESS > 2500.

When I set `stopping_rules[1] = srGelmanRubin(5, file = paramFile, freq = 1000)`, the run **did** stop at iteration 1000.

Hypothesis: Are the convergence diagnostics only being triggered at iteration `freq`, rather than every `freq` iterations?

**Computer info**
Win 10, rb1.2.5

davidcerny:
This should fix #543. As a bonus, it should also fix an analogous problem with `srStationarity`.

I tested the values returned by RevBayes against R's `coda::gelman.diag()`; they seem to be intermediate between those obtained with `autoburnin=TRUE` and `autoburnin=FALSE`. (Note that RevBayes does try to automatically calculate burnin.)

The code might not be super elegant – let me know if there are any obvious simplifications.


src/core/analysis/stoppingRule/GelmanRubinStoppingRule.cpp
```diff
@@ -100,13 +98,34 @@ bool GelmanRubinStoppingRule::stop( size_t g )
...
+    {
+        for ( size_t j = 0; j < numReplicates; j++)
         {
-            data[j].setBurnin(maxBurnin);
+            data_exp[i].push_back( data[j][i] );
         }
         
         // conduct the test
-        passed &= grTest.assessConvergence(data);
+        passed &= grTest.assessConvergence(data_exp[i]);
```
ms609, line 128: Why not just return `false` here if `!grTest.assessConvergence(data_exp[i])`?

ms609 (commented):
Looks promising, thanks for following this up!  I'll build it and have a play around when I'm next at my other machine.

bjoelle:
if possible, I would also add a small test here

src/core/analysis/stoppingRule/GelmanRubinStoppingRule.cpp
davidcerny: Hmm, that's a good point. This solution (which was already in the code; I just left it unchanged) is more elegant, since we return the test itself rather than _conditional on_ the test, but it could lead to a lot of unnecessary computation – e.g., if you have a complex model with several hundred parameters and the 1st or 2nd one already fails the test. I'll change it.

davidcerny:
Hmm... a lot of compilers apparently stumble on this bit of code – it will always successfully compile, but often cause a segfault when `GelmanRubinStoppingRule::stop` is actually called in one of the newly added tests:

```
std::vector< std::vector<TraceNumeric> > data;
data.reserve( numReplicates );
for ( size_t i = 0; i < numReplicates; ++i)
{
        // ...
        data[i] = reader.getTraces();
        // ...
}
```

For some reason, macOS seems to be the only exception. It's not clear to me why.

bredelings:
I guess you could run it in a debugger to figure out where it is SEGFAULTing.  Probably there is a real bug that just by chance isn't causing a SEGFAUL on macOS.

davidcerny:
Hmm, but that's exactly what I did – it was gdb that told me it was the `data[i] = reader.getTraces();` line that was problematic. Here is the full backtrace output:

```
#0  0x00007ffff5db0c91 in __memmove_avx_unaligned_erms () from /lib64/libc.so.6
#1  0x000000000041a3fb in std::char_traits<char>::copy (__s1=0x0, __s2=0x33f0ea0 "output/GelmanRubin_test_run_1.log", __n=33)
    at /usr/include/c++/8/bits/char_traits.h:359
#2  0x000000000041f1c6 in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_S_copy (__d=0x0, 
    __s=0x33f0ea0 "output/GelmanRubin_test_run_1.log", __n=33) at /usr/include/c++/8/bits/basic_string.h:344
#3  0x0000000000420df1 in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_assign (this=0x3412900, __str=...)
    at /usr/include/c++/8/bits/basic_string.tcc:272
#4  0x000000000041e97f in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::assign (this=0x3412900, __str=...)
    at /usr/include/c++/8/bits/basic_string.h:1370
#5  0x000000000041cc8d in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::operator= (this=0x3412900, __str=...)
    at /usr/include/c++/8/bits/basic_string.h:699
#6  0x0000000000cb4821 in boost::filesystem::path::operator= (this=0x3412900, p=...) at /software/boost-1.75.0-el8-x86_64/include/boost/filesystem/path.hpp:220
#7  0x0000000001305e90 in RevBayesCore::Trace<double>::operator= (this=0x34128f0) at /home/davidcerny/revbayes-fsr/src/core/analysis/mcmc/output/Trace.h:21
#8  0x0000000001305eff in RevBayesCore::TraceNumeric::operator= (this=0x34128f0) at /home/davidcerny/revbayes-fsr/src/core/analysis/mcmc/output/TraceNumeric.h:8
#9  0x0000000001306056 in std::__copy_move<false, false, std::random_access_iterator_tag>::__copy_m<RevBayesCore::TraceNumeric const*, RevBayesCore::TraceNumeric*>
    (__first=0x342e6d0, __last=0x342e9d0, __result=0x34128f0) at /usr/include/c++/8/bits/stl_algobase.h:324
#10 0x0000000001305c58 in std::__copy_move_a<false, RevBayesCore::TraceNumeric const*, RevBayesCore::TraceNumeric*> (__first=0x342e610, __last=0x342e9d0, 
    __result=0x3412830) at /usr/include/c++/8/bits/stl_algobase.h:386
#11 0x0000000001305add in std::__copy_move_a2<false, __gnu_cxx::__normal_iterator<RevBayesCore::TraceNumeric const*, std::vector<RevBayesCore::TraceNumeric, std::allocator<RevBayesCore::TraceNumeric> > >, __gnu_cxx::__normal_iterator<RevBayesCore::TraceNumeric*, std::vector<RevBayesCore::TraceNumeric, std::allocator<RevBayesCore::TraceNumeric> > > > (__first=..., __last=..., __result=...) at /usr/include/c++/8/bits/stl_algobase.h:422
#12 0x00000000013058a0 in std::copy<__gnu_cxx::__normal_iterator<RevBayesCore::TraceNumeric const*, std::vector<RevBayesCore::TraceNumeric, std::allocator<RevBayesCore::TraceNumeric> > >, __gnu_cxx::__normal_iterator<RevBayesCore::TraceNumeric*, std::vector<RevBayesCore::TraceNumeric, std::allocator<RevBayesCore::TraceNumeric> > > > (__first=..., __last=..., __result=...) at /usr/include/c++/8/bits/stl_algobase.h:455
#13 0x0000000001305521 in std::vector<RevBayesCore::TraceNumeric, std::allocator<RevBayesCore::TraceNumeric> >::operator= (this=0x340e600, __x=...)
    at /usr/include/c++/8/bits/vector.tcc:226
#14 0x0000000001304c12 in RevBayesCore::GelmanRubinStoppingRule::stop (this=0x34179f0, g=1000)
    at /home/davidcerny/revbayes-fsr/src/core/analysis/stoppingRule/GelmanRubinStoppingRule.cpp:83
```

I'm struggling to interpret this, but frames 11 and 12 do look to me like the problem might indeed be compiler-specific. I have no idea why this particular line triggers a segfault, but I did notice that previously, the equivalent line read `std::vector<TraceNumeric> &data = reader.getTraces();` – i.e., `data` was a reference, whereas `data[i]` is an object.

Also, this time I made sure the problem is not stochastic – the segfault happens in an integration test with a fixed seed. You can see the GitHub actions are failing for gcc and clang on Linux but not for Xcode on macOS; in addition, I can confirm that when I build this branch on my laptop using AppleClang 15.0.0.15000309, the segault doesn't happen, either.

bredelings:
I think if you change 
```
std::vector< std::vector<TraceNumeric> > data;
data.reserve( numReplicates );
```
to 
```
std::vector< std::vector<TraceNumeric> > data(numReplicates);
```
then it might work?

After `v.reserve( )`, the size will still be zero.  `v.resize( )` actually changes the size.  The purpose of `reserve` is (I think) to avoid doing multiple memory allocations and copies when you start with size 0 and grow the vector incrementally using `v.push_back( )`.

davidcerny:
Thanks so much for the explanation, Ben! It worked :)

src/core/analysis/stoppingRule/GelmanRubinStoppingRule.cpp
```diff
@@ -104,7 +104,7 @@ bool GelmanRubinStoppingRule::stop( size_t g )
         num_variables_in_rep = data[i].size();
         
         // if we have different numbers of variables in different replicates, something has gone terribly wrong
-        if (i > 1 && data[i].size() != num_variables_in_rep)
+        if (i > 0 && data[i].size() != num_variables_in_rep)
```
ms609, line 107: Line 104 has just defined `num_variables_in_rep = data[i].size();` so I don't think this check can fail.

Perhaps it is intended to come before the previous line, or, equivalently, to compare with `data[i-1].size()`?

src/core/analysis/stoppingRule/GelmanRubinStoppingRule.cpp
```diff
@@ -104,7 +104,7 @@ bool GelmanRubinStoppingRule::stop( size_t g )
...
         {
             throw RbException("The replicates contain different sets of parameters.");
```
ms609, line 109: Is it overkill to also check that the parameter names are equivalent?

src/core/analysis/stoppingRule/StationarityStoppingRule.cpp
```diff
@@ -104,7 +104,7 @@ bool StationarityStoppingRule::stop( size_t g )
         num_variables_in_rep = data[i].size();
         
         // if we have different numbers of variables in different replicates, something has gone terribly wrong
-        if (i > 1 && data[i].size() != num_variables_in_rep)
+        if (i > 0 && data[i].size() != num_variables_in_rep)
```
ms609, line 107:
```suggestion
        if (i > 0 && data[i-1].size() != num_variables_in_rep)
```

src/core/analysis/stoppingRule/GelmanRubinStoppingRule.cpp
ms609: As a secondary thought, defining `num_variables_in_rep` at the start of the loop would allow us to replace the calls to `data[i].size()` in each loop heading with a reference to the variable.

davidcerny: Duh! What was I thinking? Thank you for spotting this!

davidcerny: And yes, this could be solved in a number of ways. `i > 0 && data[i-1].size() != num_variables_in_rep` is one option. Defining `num_variables_in_rep` outside the loop is another, but then the trace-reading code would be duplicated both inside and outside the loop. Maybe the simplest solution is to redefine `num_variables_in_rep` _after_ the check, rather than before.

src/core/analysis/stoppingRule/GelmanRubinStoppingRule.cpp
davidcerny: Maybe not, but I'm not sure it should be done by every individual stopping rule... if anything, I'm wondering if the check is needed at all. Even when manipulating log files by hand, I couldn't design a test case that would violate it.

bjoelle (approved):

Issue #557, opened 2024-09-12
[BUG] TreeScaleProposal can move fossil tips outside of their age range


bjoelle:
There is no check for parents of sampled ancestors in `TreeScaleProposal`, and `TopologyNode::setAge` will automatically adjust the SA age to match the new parent age
this can lead to fossil tips silently falling out of their specified age ranges (possibly tied to the issue in #544 ? although that seemed to affect a different move)

[FBD_move_bug.zip](https://github.com/user-attachments/files/16975185/FBD_move_bug.zip)

Main file is `bug_mcmc.rev`, FBD setup is in `model_FBDP.rev`

Issue #544, opened 2024-08-27
[BUG] Crash due to -inf log likelihoods in `mvCollapseExpandFossilBranch`


davidcerny:
**Describe the bug**
I'm running a fairly large FBD analysis (152 taxa, 931 characters) that sometimes causes RevBayes to crash. This is the stack trace I'm getting from lldb:

```
* thread #1, queue = 'com.apple.main-thread', stop reason = hit program assert
    frame #0: 0x0000000187332a60 libsystem_kernel.dylib`__pthread_kill + 8
    frame #1: 0x000000018736ac20 libsystem_pthread.dylib`pthread_kill + 288
    frame #2: 0x0000000187277a30 libsystem_c.dylib`abort + 180
    frame #3: 0x0000000187276d20 libsystem_c.dylib`__assert_rtn + 284
  * frame #4: 0x000000010110cca8 rb`search_interval(double, double&, double&, (anonymous namespace)::slice_function&, double) + 96
    frame #5: 0x000000010110bbf8 rb`slice_sample_doubling(double, (anonymous namespace)::slice_function&, double, int) + 212
    frame #6: 0x000000010110b940 rb`RevBayesCore::SliceSamplingMove::performMcmcMove(double, double, double) + 172
    frame #7: 0x0000000101110b14 rb`RevBayesCore::AbstractMove::performMcmcStep(double, double, double) + 80
    frame #8: 0x00000001011e29d8 rb`RevBayesCore::Mcmc::nextCycle(bool) + 152
    frame #9: 0x000000010119ea4c rb`RevBayesCore::MonteCarloAnalysis::run(unsigned long, RevBayesCore::RbVector<RevBayesCore::StoppingRule>, unsigned long, boost::filesystem::path const&, unsigned long, bool) + 2876
    frame #10: 0x000000010019ed98 rb`RevLanguage::MonteCarloAnalysis::executeMethod(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, std::__1::vector<RevLanguage::Argument, std::__1::allocator<RevLanguage::Argument>> const&, bool&) + 2028
    frame #11: 0x0000000100da7b20 rb`RevLanguage::MemberProcedure::execute() + 144
    frame #12: 0x0000000100c59a70 rb`RevLanguage::SyntaxFunctionCall::evaluateContent(RevLanguage::Environment&, bool) + 2192
[...]

PR #559, opened 2024-09-12
Stop tree rescaling from changing ages of SAs or their direct parent

bjoelle:
Fixes issue #557 and adds some sanity checks to (hopefully) stop this from happening again


davidcerny:
This is great; thank you so much, Joëlle! I'm wondering if we should try to be proactive and see if other node time proposals could benefit from the same fix/extra condition :

- [`mvNodeRateTimeSlideUniform`](https://github.com/revbayes/revbayes/blob/master/src/core/moves/proposal/tree/NodeRateTimeSlideUniformProposal.cpp)
- [`mvNodeTimeScale`](https://github.com/revbayes/revbayes/blob/master/src/core/moves/proposal/tree/NodeTimeScaleProposal.cpp)
- [`mvNodeTimeSlidePathTruncatedNormal`](https://github.com/revbayes/revbayes/blob/master/src/core/moves/proposal/tree/NodeTimeSlidePathTruncatedNormalProposal.cpp)
- [`mvNodeTimeSlideUniform`](https://github.com/revbayes/revbayes/blob/master/src/core/moves/proposal/tree/NodeTimeSlideUniformProposal.cpp)
- [`mvNodeTimeSlideUniformAgeConstrained`](https://github.com/revbayes/revbayes/blob/master/src/core/moves/proposal/tree/NodeTimeSlideUniformAgeConstrainedProposal.cpp)

I should be able to look into these later today...

davidcerny:
Okay, I added the extra check to all other equivalent node age moves I could find. Two quick notes:

- [`mvNodeTimeSlidePathTruncatedNormal`](https://github.com/revbayes/revbayes/blob/master/src/core/moves/proposal/tree/NodeTimeSlidePathTruncatedNormalProposal.cpp) and [`mvNodeTimeSlideUniformAgeConstrained`](https://github.com/revbayes/revbayes/blob/master/src/core/moves/proposal/tree/NodeTimeSlideUniformAgeConstrainedProposal.cpp) sort of do their own thing. It's not clear to me if they suffer from the same problem, and if so, how to fix them.
- [`mvNodeTimeSlideUniform`](https://github.com/revbayes/revbayes/blob/master/src/core/moves/proposal/tree/NodeTimeSlideUniformProposal.cpp) already includes the extra check, but in a slightly different form: instead of just `isSampledAncestorParent()`, it uses `isSampledAncestorTipOrParent()`. Not sure how significant this difference is.

davidcerny:
I can confirm this fixes the hanging problem described in #557. It doesn't solve #544 – the same dataset/seed combination now triggers the exception newly added to `TopologyNode::setAge()`. I'm wondering if an analysis-terminating exception is really what we want to do here – couldn't we just reject the move that tries to set a node's age to an illegal value?

bjoelle:
@davidcerny we can certainly reject the move instead, but that also has to be done at the move level, `TopologyNode::setAge()` cannot do it - so my view is, if we have to change the move we might just as well do the right thing in the first place
imo raising an exception is the right thing to do - it's not supposed to happen, so either the calling code expects it and deals with it by catching the exception, or we crash because it's better than getting into an invalid state

bjoelle:
with this PR I can now run the script from #544 until iteration 150,000 without issue - @davidcerny can you confirm if this is still an issue ?

bjoelle:
With this PR I can now run the script from issue #544 for 160,000 iterations without issue
@davidcerny can you confirm if this resolves your issue ?

davidcerny:
@bjoelle I can confirm it doesn't, unfortunately. Are you setting `seed(234)` by any chance? When I do that, the analysis throws the following error after 110,000 iterations:

```
   Error:	New age of taxa Gamatavus_antiquus incompatible with age range
   Error:	Problem processing line 250 in file ""/Users/David/Downloads/fbd_mvslice_crash/test.Rev""
```

bjoelle:
thanks for the additional information ! 
it looks like this may happen because the root age has a separate move on it - so when it is being modified and the root is an SA, it tries to update the node age as well and then that fails

I'm not sure what is the best solution for this honestly, because the root age move has no reference to the tree so it cannot check for SAs - I think having a separate move on the root age may just be a bad idea ?

bjoelle:
ok I've disabled the exception for now while we decide what to do about it, and added checks to the root time proposals
if you agree @davidcerny maybe we can merge this first and then look deeper into the root age related issues ?

davidcerny (approved):
Thank you @bjoelle! For what it's worth, after an absurd amount of tinkering with the original script from #544, I managed to find a combination of settings that allowed me to verify what happens to the problematic tip age when the newly added warning is emitted: seed = 234, burnin = 480,000 iterations, sampling = 20,000 iterations, move schedule = "single", frequencies of all monitors (log, screen, file) set to 1 in order to sample after every single attempted move. During the sampling period, the warning was emitted 31 times, always for the same taxon. However, when I checked the age of this taxon in the sampled trees, it was within the specified range every single time (i.e., in all 20,000 samples). So everything works as intended, and as far as I'm concerned, this PR solves #544. :)

I've made a cosmetic change to the warning ("taxa" is plural; the singular is "taxon").

I've pre-emptively approved the PR, but I do have one last question (which I already asked above, but we both understandably got sidetracked): is the condition that you've added:

```
while ( node->isRoot() || node->isTip() || node -> isSampledAncestorParent())
```

perfectly equivalent to the following condition that was already present in `NodeTimeSlideUniformProposal`?

```
if (not node.isRoot() and not node.isTip() and not node.isSampledAncestorTipOrParent())
```

If not, then I guess one is either too lax or too strict. Based on the syntax choices, it looks to me like the latter may have been written by @bredelings – feel free to chime in, Ben! :)

bjoelle:
@davidcerny sorry I had missed that question
I think the condition in `NodeTimeSlideUniformProposal` is slightly redundant - if the node is not a tip then by definition it cannot be a SA tip
but both conditions should give the exact same results yes

PR #560, opened 2024-09-17
Retain cache when switching branches

ms609:
[Adding `build/` to `.gitignore`](https://github.com/revbayes/revbayes/commit/6f2a02d2f93e4790e919c15619739238f1d7ce74) causes the cached build artefacts to be lost when switching branches, meaning that rb must be built from scratch whenever moving to a new branch. 

I believe that these modifications should cause the `build` directory to be retained when switching branches.  The content of the `.gitkeep` file is arbitrary – I felt a brief explanation was more useful than a blank file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,8 @@ src/generated_include_dirs.cmake
 src/*/CMakeLists.txt
 
 # ----------------------------------------------
-build/
+/build/*
+!/build/.gitkeep
 projects/cmake/*
 !projects/cmake/README.txt
 !projects/cmake/*.sh
--- a/build/.gitkeep
+++ b/build/.gitkeep
@@ -1,0 +1,1 @@
+This file ensures that cached build files are retained when switching branches.


bredelings:
Why would `.gitignore` affect building at all?  I feel like I'm missing something here...

Also, do you have `ccache` installed?  It is really helpful...

ms609:
It's quite possible I've misdiagnosed the problem here, but what I think is happening for me is that when I trigger a build in VSCode, object files are built to the `/build` directory.  Subsequent builds then only recompile files as necessary.

When switching branches, however, the contents of the ignored `/build` folder seem not to be copied to the new branch.  Before adding `build/` to the `.gitignore`, I could choose (in GitHub Desktop) to bring the changes with me to the new branch – though of course these many files made actual changes harder to identify.

bredelings:
Ah, OK then.  So VSCode apparently copies the source directory for each branch you work on, instead of modifying a single directory?  In that case, people not using VSCode (i.e. just meson or cmake) would not see this.

Can you change the explanation file to say something like "This file ensures that cached build files are copied by VSCode when switching branches."?  That clarifies that "retained" means "copied" as opposed to "not deleted", and that it relates only to VSCode behavior.

ms609:
Having interrogated the precise behaviour of GitHub Desktop and VSCode further, I think your initial diagnosis was right and that this is not a `.gitignore` issue.  I'll close this PR.

PR #569, opened 2024-10-29
fixing a segfault in Model::getOrderedStochasticNodes

mikeryanmay:
I'm encountering a segfault in an analysis involving a model with many dynamic nodes. The error occurs when the model object is first being created. I've recreated the problem with a toy example (attached). In the toy example, I generate a number (n) of normal random variables. If n = 1000, the analysis runs. If n = 5000, I get a segfault.

I have narrowed down the problem to the function getOrderedStochasticNodes, specifically with the std::set<const DagNode*> visitedNodes object (whose purpose is to maintain a list of all the nodes that have been visited as getOrderedStochasticNodes traverses the model graph). The code occurs on line 327 (on development), where the current node is inserted into visitedNodes. The behavior of std::set::insert should be that if an object is already in the set, it is not inserted again, so this code looks fine as-is. However, when I modify the code (on this fix branch) so that a node is only inserted after explicitly checking that it is not already in the set, the segfault is resolved.

The code on development looks conceptually fine to me, so I am a bit baffled by this behavior. However, I have learned to be suspicious of ordered sets of pointers in other situations. I am not sure why the current code is a problem or why my solution works, so I am open to any comments or feedback before merging into development.

Here is my toy example:
[test.txt](https://github.com/user-attachments/files/17563606/test.txt)


bredelings:
I was able to reproduce the SEGFAULT.  It looks like it is actually caused by a stack overflow in which getOrderedStochasticNodes alternates between looking at parents and looking at children.

bredelings:
I think your patch is the correct fix.

bredelings (approved):

src/core/analysis/Model.cpp
```diff
@@ -303,6 +303,9 @@ void Model::getOrderedStochasticNodes(const DagNode* the_dag_node, std::vector<D
         //The node has been visited before
         //we do nothing
         return;
+    } else {
+        // add myself here for safety reasons
+        visitedNodes.insert( the_dag_node );
```
bredelings, line 308: Maybe add a comment that looks something like:
```
NOTE:  Immediately mark the node as visited in order to prevent any future recursion on this node.
             If we don't do this we can ping-pong back and forth between parents and children if the parents.
             If parents have a lot of children, this can lead to a stack overflow.
```

Although maybe that is too long.

bredelings:
I guess that finding nodes in a different order can change the order of MCMC moves, and that is why the tests are now failing?

src/core/analysis/Model.cpp
bredelings: Also maybe this is clearer out of the else block.

mikeryanmay:
> I think your patch is the correct fix. It doesn't really add a condition -- you could take the `insert` out of the else block. The reason that it works is that it does the insert before any recursive calls.

This makes total sense. Thanks for pointing it out!

mikeryanmay:
> I guess that finding nodes in a different order can change the order of MCMC moves, and that is why the tests are now failing?

Exactly why I don't like ordered sets of pointers! Will work on getting the tests updated.

bredelings:
OK, I added some notes.

Also... maybe we should add your toy example as a test case?

mikeryanmay:
> OK, I added some notes.
> 
> Also... maybe we should add your toy example as a test case?

Will do.

mikeryanmay:
I've added a new test (with lots of stochastic nodes), and attempted to update the outputs for the failing tests following the instructions in tests/README.md (including recompiling rb with -DCONTINUOUS_INTEGRATION=TRUE). However, the tests are still failing. Anyone have any insights?

davidcerny:
> However, the tests are still failing. Anyone have any insights?

@mikeryanmay I think I've run into this before. The JSON files generated by the `Partition` test are architecture-specific; the last few digits differ on x86_64 and ARM64, so it's impossible for a single set of files to pass both checks. Our "solution" is to special-case the test runner so that it ignores the check for macOS+ARM64, but with your expected output, that's the only check that passes.

An inelegant workaround is to take the output from one of the failing builds and make it the expected output.

bredelings:
I was going to compile with -Dopenlibm=true and try to commit the resulting output.  But it seems like openlibm has now been removed from Debian?  Sigh.

Issue #567, opened 2024-10-25
[BUG] `mvSlice()` causes RevBayes to exit during `mcmc` run


ms609:
**Describe the bug**
Replacing `mvScale` with `mvSlice` causes RevBayes to exit unexpectedly during an MCMC run.

**To Reproduce**
- Download [data.txt](https://github.com/user-attachments/files/17519572/data.txt)
- Run the following:
```rb
moves = VectorMoves()

chars <- readDiscreteCharacterData("data.txt")
taxa <- chars.names()
outgroup <- clade(taxa[1])
nTaxa <- chars.size()
nEdge <- 2 * nTaxa - 3

topology ~ dnUniformTopology(taxa, outgroup = outgroup)
moves.append( mvNNI(topology, weight = nEdge/2.0, tuneTarget = 0.26) )
moves.append( mvSPR(topology, weight = nEdge/10.0, tuneTarget = 0.26) )
tree_length ~ dnGamma(shape = 2, rate = 1)
moves.append( mvSlice(tree_length) )

rel_br_lengths ~ dnDirichlet( rep(1.0, nEdge) )
moves.append( mvBetaSimplex(rel_br_lengths, weight = nEdge / 5.0, tuneTarget = 0.26) )
moves.append( mvDirichletSimplex(rel_br_lengths, weight = nEdge / 10.0, tuneTarget = 0.26) )
br_lengths := rel_br_lengths * tree_length

phylogeny := treeAssembly(topology, br_lengths)

m_morph[1] ~ dnPhyloCTMC(
    tree = phylogeny,
    Q = fnJC(5),
    type = "Standard"
)
m_morph[1].clamp(chars)

monitors = [ mnScreen(printgen = 2, tree_length, likelihood = FALSE, posterior = FALSE, prior = FALSE) ]
mymcmc = mcmc(model(phylogeny), monitors, moves)
mymcmc.run(generations = 42)
```

**Expected behavior**
RevBayes exits without displaying an error message or reporting a Seg fault after a few generations.  The run completes successfully when `mvSlice` is replaced with `mvScale`.

**Computer info**
Windows 11 with a recent development build

bredelings:
So mvSlice(x) always checks that PDFs don't change by setting x to its _current_ value.  So, this is likely to be a bug somewhere else that mvSlice complains about.  I added some code to figure out which PDF changes, and get this:
```
probability for m_morph[1] is -12202.64911 but should be -12205.97713:  delta = 3.328013235
```
[...]

Issue #565, opened 2024-10-23
Inference of clock model parameters for continuous data fails with MCMC flatlines, suspect interaction of FBD tree model and dnPhyloBrownianREML()


ixchelgzlzr:
Hello RevBayes developers! 

I am unsure if the issue I am describing below is a bug or an user error (most likely in the Brownian model component), so I thank you so much in advance for any help you can provide!

I am setting up a Total Evidence Dating analysis on a dataset that contains molecular, morphological categorical, and morphological continuous data. 

**Model description:**
I am using an FBD tree model, standard GTR + G and Mk +G models of evolution for the molecular and categorical morphological data, and UCLN clock model for the molecular data. The clock of the categorical morphological data is linked to the molecular clock rates. For the continuous data, I am using a Brownian model, and trying to set up a global rate for the continuous data clock model. As an additional note, I have set up two topological constraints.

**Issue:**
The MCMCs flatline for all the parameters of the continuous characters evolution and clock models., e.g.:

![Screenshot 2024-10-23 at 12 45 09 PM](https://github.com/user-attachments/assets/d3947b38-8a1a-4ab3-b2ce-ef3003bddc31)

I have troubleshooted many options and here is what I have so far:

1) When I run an unrooted/no-time-divergence model, the parameters of the Brownian model are well sampled. ---> the problem seem to be specifically with the clock model for continuous data.
2) I turned off the moves on sampled ancestors and I still get the flatlines, so I don't think the issue is related to having sampled ancestors. 
3) When I don't include the continuous data, the MCMCs mix well ---> Another evidence is the continuous part of the model that isn't working. 
4) The continuous dataset has missing data (i.e. fossils), so I fillied in that missing data, and still got the flatlines --->  missing data might not be the issue. 

In summary, I suspect an interaction between the FBD tree prior and the molecular clock for the continuous data.  Alternatively, I might be setting this model wrong...

**To Reproduce**
I have set up a reduced example of my analysis here:
` https://github.com/ixchelgzlzr/bug_report_TED_solanaceae.git`

There are instructions on how to run the different versions of the analyses in the `Readme` file. 

I have also included some examples of output in `/output_examples`. Please notice I have set up a seed() that replicates the screenshot I pasted above.

**Expected behavior**
Mixing MCMCs for the continuous data parameters.

**Computer info**
I'm using RevBayes version (1.2.5-preview)
Build from remotes/origin/development (rapture-3049-g3bb16a)
[...]

PR #570, opened 2024-10-31
Convert some MCMC checks from #ifdef to options.

bredelings:
For issues #565 and #567 we need to solve problems where PDFs for some distributions are out of date.  Right now this can be partially addressed by compiling with `-DDEBUG_MCMC`, but even that doesn't find all the problems.

This PR tries to make these problems slightly easier for users to look for by (i) changing the compile-time option to a run-time setting and (ii) adding some functionality to the slice sampler to say WHICH variable has an incorrect PDF seems to be incorrect.  It also simplifies the code for `performMcmcMove( )` _a lot_.  The `debugMCMC` setting is an integer level, with `0` having no effect, and higher levels enabling more costly functionality.

For #567, we can now get the following: 
```
$ rb-debug-O --setOption debugMCMC=1 script.Rev
RevBayes version (1.2.5-preview)
Build from check-likelihoods (rapture-3061-ge482e5) on Thu Oct 31 03:10:03 PM EDT 2024

Visit the website www.RevBayes.com for more information about RevBayes.

RevBayes is free software released under the GPL license, version 3. Type 'license()' for details.

To quit RevBayes type 'quit()' or 'q()'.


> source("script.Rev")
   Processing file "script.Rev"
   Successfully read one character matrix from file 'data.txt'
   
   Running MCMC simulation
   This simulation runs 1 independent replicate.
   The simulator uses 5 different moves in a random move schedule with 27.1 moves per iteration
   

Iter        |    tree_length   |    elapsed   |        ETA   |
--------------------------------------------------------------
0           |       1.412281   |   00:00:00   |   --:--:--   |

mvSlice for tree_length: probability is -14239.76608 but should be -14239.76609:  delta = 1.973286999e-06
         cause: probability for m_morph[1] is -12205.97713 but should be -12205.97713:  delta = 1.973286999e-06
Aborted
```

Oddly enough, enabling the checks for Metropolis-Hastings with --setOption `debugMCMC=2`actually suppresses this error.  So there is more work to do.  However, I'd be interested in merging this (or getting feedback) now before continuing.

### DIFF
--- a/src/core/moves/MetropolisHastingsMove.cpp
+++ b/src/core/moves/MetropolisHastingsMove.cpp
@@ -340,48 +340,34 @@ void MetropolisHastingsMove::performMcmcMove( double prHeat, double lHeat, doubl
     // compute the probability of the current value for each node
     for (auto node: views::concat(touched_nodes, affected_nodes))
     {
-        if ( RbMath::isAComputableNumber(ln_prior_ratio) && RbMath::isAComputableNumber(ln_likelihood_ratio) && RbMath::isAComputableNumber(ln_hastings_ratio) )
+        if ( RbMath::isAComputableNumber(ln_prior_ratio) &&
+             RbMath::isAComputableNumber(ln_likelihood_ratio) &&
+             RbMath::isAComputableNumber(ln_hastings_ratio) )
         {
+            double ratio = 0;
+            try {
+                ratio = node->getLnProbabilityRatio();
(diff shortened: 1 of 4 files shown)


ms609 (approved):
Without looking through the implementation in detail, this all looks very sensible and will be very helpful in debugging – whether or not you find opportunities to improve it further in the future.

Issue #571, opened 2024-11-01
[BUG] printing `time()` fails


ms609:
In a recent Windows 10 development build (and 1.2.4), `time("fromBeginning")` (or, as this is the default, `print(time())`) outputs

      Error:       Negative value for Natural

Other options (e.g. `time("seconds")` work successfully.

davidcerny:
Marking this as a Windows-specific issue, as I was unable to reproduce it on macOS or Linux. Based on the code in `Func_time.cpp`, the returned value should never be negative unless boost determined your local time to predate the year 1400!

bredelings:
The code stores the milliseconds in a `long`.  However, the definition of `long` and `long long` differs on Linux, Mac, and Windows.  We should switch to something like `std::in64_t`.

bredelings:
So to make things even trickier, it turns that that on Mac only, `std::uint64_t` not the same as `size_t`.  I think the are both 64-bit unsigned integers, but you can't pass an argument of one type to a function that takes the other type without a type conversion.  This makes that if we have `f(std::uint64_t)` and `f(std::uint32_t)` then calling `f(size_t(0))` is ambiguous :-(

And we need to make the `RevLanguage::Natural( )` constructor accept 32-bit ints, 64-bit ints, and (on Mac) also `size_t`.  Grrr......

I have an idea though...

davidcerny:
> So to make things even trickier, it turns that that on Mac only, `std::uint64_t` not the same as `size_t`. I think the are both 64-bit unsigned integers, but you can't pass an argument of one type to a function that takes the other type without a type conversion. 

That's... terrifying, actually, and I can see how my [recent attempts](https://github.com/revbayes/revbayes/pull/652) to replace `int`s with `size_t`s could have made a bad situation worse. Apologies!

bredelings:
Fixed by #708

PR #572, opened 2024-11-01
Document time

ms609:
This PR explains the effect of each parameter to `time()`.
(I'm guessing `fromBeginning` because it's currently broken – #571)

### DIFF
--- a/help/md/time.md
+++ b/help/md/time.md
@@ -3,8 +3,18 @@ time
 ## title
 Get the time information
 ## description
-Get the current system time in milliseconds.
+Get the current system time.
+
 ## details
+
+"year" reports the current year in the Julian calendar.
+
+"day" returns the index of the day in the year (e.g. Jan 1 = 1; Feb 1 = 32).
+
+"(milli)seconds" returns the number of (milli)seconds that have elapsed since midnight.
+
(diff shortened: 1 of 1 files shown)


bredelings:
Hmm.... yeah, the number of seconds does seem wrong:
```
> t=time()
> t/60/60/24/365.25
   624822.2
```

The term "Julian date" confused me.  I thought that meant "without gregorian calendar corrections", or something like that.

https://www.longpelaexpertise.com.au/toolsJulian.php

bredelings:
This is fun: https://en.wikipedia.org/wiki/Julian_day

ms609:
> 624822.2

I think `fromBeginning` gives milliseconds – throw in another `/1000` and 62.48222 sounds closer to (but not quite!) the number of years since 1970-01-01.

bredelings (approved):

davidcerny:
@ms609 That doesn't work out – following that assumption, you'd need to divide by 10,000 rather than 1000 to get ~62.4 years, and even then the result would be off by >7 years, as you point out.

For some reason, boost's "earliest representable date" (the `min_date_time` constant used in `Func_time.cpp`) is 1400-Jan-01 00:00:00:
https://www.boost.org/doc/libs/1_33_1/doc/html/date_time/gregorian.html
https://stackoverflow.com/a/25084661

`time("fromBeginning")` gives the number of milliseconds elapsed since that date.

PR #574, opened 2024-11-01
Another `time()` doc fix

davidcerny:
`time("fromBeginning")` counts milliseconds since 1400-Jan-01, not 1970-Jan-01.


bredelings:
Thanks!  The 1400 seems right -- any idea why 1400AD would be the start of the epoch??

bredelings (approved):

davidcerny:
No idea! It does seem super arbitrary as a starting date – it predates the adoption of the Gregorian calendar, but not by a whole lot. I guess we'd have to ask the boost team...

PR #575, opened 2024-11-02
Debug mcmc

bredelings:
This change makes `--setOption debugMCMC=1` actually work for Metropolis-Hastings moves.  Previously some of the code to check the node PDFs was actually hiding the recalculation bugs by doing extra recalculation, but only during debugging!

This change also separates logging from debugging.  The logging option is `logMCMC`.  For Metropolis-Hastings moves and the values are:
- `1`  writes out the generation, step within each generation, and move name for each move.
- `2` also writes out posterior, likelihood, prior, and Hastings ratios.
- `3` writes out each changed PDF along with its node name.

Together, these should allow debugging #565 and #567.

### DIFF
--- a/src/core/moves/MetropolisHastingsMove.cpp
+++ b/src/core/moves/MetropolisHastingsMove.cpp
@@ -277,10 +277,10 @@ void MetropolisHastingsMove::performMcmcMove( double prHeat, double lHeat, doubl
     int debugMCMC = RbSettings::userSettings().getDebugMCMC();
     if (logMCMC >= 3)
     {
-	std::cerr<<std::setprecision(10);
-	for(auto& [node,pr]: getNodePrs(nodes, affected_nodes))
-	    std::cerr<<"    BEFORE:   "<<node->getName()<<":  "<<pr<<"\n";
-	std::cerr<<"\n";
+        std::cerr<<std::setprecision(10);
+        for(auto& [node,pr]: getNodePrs(nodes, affected_nodes))
+            std::cerr<<"    BEFORE:   "<<node->getName()<<":  "<<pr<<"\n";
+        std::cerr<<"\n";
     }
 
     if (debugMCMC >= 2)
@@ -383,9 +383,9 @@ void MetropolisHastingsMove::performMcmcMove( double prHeat, double lHeat, doubl
 
     if (logMCMC >= 3)
     {
-	for(auto& [node,pr]: getNodePrs(nodes, affected_nodes))
-	    std::cerr<<"    PROPOSED: "<<node->getName()<<":  "<<pr<<"\n";
-	std::cerr<<"\n";
+        for(auto& [node,pr]: getNodePrs(nodes, affected_nodes))
+            std::cerr<<"    PROPOSED: "<<node->getName()<<":  "<<pr<<"\n";
+        std::cerr<<"\n";
     }
 
     // exponentiate with the chain heat
@@ -436,19 +436,18 @@ void MetropolisHastingsMove::performMcmcMove( double prHeat, double lHeat, doubl
     }
 
 
-
     if (logMCMC >= 3)
     {
-	for(auto& [node,pr]: getNodePrs(nodes, affected_nodes))
-	    std::cerr<<"    FINAL:    "<<node->getName()<<":  "<<pr<<"\n";
-	std::cerr<<"\n";
+        for(auto& [node,pr]: getNodePrs(nodes, affected_nodes))
+            std::cerr<<"    FINAL:    "<<node->getName()<<":  "<<pr<<"\n";
+        std::cerr<<"\n";
     }
 
     if (logMCMC >= 2)
     {
-	std::cerr<<"    log(posterior_ratio) = "<<ln_posterior_ratio<<"  log(likelihood_ratio) = "<<ln_likelihood_ratio<<"   log(prior_ratio) = "<<ln_prior_ratio<<"\n";
-	std::cerr<<"    log(acceptance_ratio) = "<<ln_acceptance_ratio<<"  log(hastings_ratio) = "<<ln_hastings_ratio<<"\n";
-	std::cerr << "  The move was " << (rejected ? "REJECTED." : "ACCEPTED.") << std::endl;
+        std::cerr<<"    log(posterior_ratio) = "<<ln_posterior_ratio<<"  log(likelihood_ratio) = "<<ln_likelihood_ratio<<"   log(prior_ratio) = "<<ln_prior_ratio<<"\n";
+        std::cerr<<"    log(acceptance_ratio) = "<<ln_acceptance_ratio<<"  log(hastings_ratio) = "<<ln_hastings_ratio<<"\n";
+        std::cerr << "  The move was " << (rejected ? "REJECTED." : "ACCEPTED.") << std::endl;
     }
 
     // --------------------------
--- a/src/core/moves/MetropolisHastingsMove.cpp
+++ b/src/core/moves/MetropolisHastingsMove.cpp
@@ -296,7 +296,7 @@ void MetropolisHastingsMove::performMcmcMove( double prHeat, double lHeat, doubl
     int debugMCMC = RbSettings::userSettings().getDebugMCMC();
     if (logMCMC >= 3)
     {
-        std::cerr<<std::setprecision(10);
+        std::cerr<<std::setprecision(11);
         for(auto& [node,pr]: getNodePrs(nodes, affected_nodes))
             std::cerr<<"    BEFORE:   "<<node->getName()<<":  "<<pr<<"\n";
         std::cerr<<"\n";
--- a/src/core/moves/MetropolisHastingsMove.cpp
+++ b/src/core/moves/MetropolisHastingsMove.cpp
@@ -291,18 +291,14 @@ void MetropolisHastingsMove::performMcmcMove( double prHeat, double lHeat, doubl
     if (debugMCMC >= 1)
     {
         // 1. Compute PDFs before proposal, before touch
-        std::map<const DagNode*, double> untouched_before_proposal;
-        for (auto node: views::concat(nodes, affected_nodes))
-            untouched_before_proposal.insert({node, node->getLnProbability()});
+        auto untouched_before_proposal = getNodePrs(nodes, affected_nodes);
 
         // 2. Touch nodes.
         for (auto node: nodes)
             node->touch();
 
         // 3. Compute PDFs before proposal, after touch
-        std::map<const DagNode*, double> touched_before_proposal;
-        for (auto node: views::concat(nodes, affected_nodes))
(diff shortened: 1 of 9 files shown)


bredelings:
For #567 we now get an error WITHOUT the slice move.
For `rb-debug-O script2.Rev --setOption logMCMC=3 --setOption debugMCMC=1`, we get:
```

generation = 0    proposal = 4/27    NNI(topology)
    BEFORE:   topology:  -32.99446143
    BEFORE:   m_morph[1]:  -17757.05043

    PROPOSED: topology:  -32.99446143
    PROPOSED: m_morph[1]:  -17695.89855

    FINAL:    topology:  -32.99446143
    FINAL:    m_morph[1]:  -17695.89855

    log(posterior_ratio) = 61.15188787  log(likelihood_ratio) = 61.15188787   log(prior_ratio) = 0
    log(acceptance_ratio) = 61.15188787  log(hastings_ratio) = 0
  The move was ACCEPTED.

generation = 0    proposal = 5/27    DirichletSimplex(rel_br_lengths)
    BEFORE:   m_morph[1]:  -17695.89855
    BEFORE:   rel_br_lengths:  67.88974314

   Error:	Executing DirichletSimplex(rel_br_lengths)): PDFs not up-to-date before proposal!
       m_morph[1]: -17695.898546 != -17647.352473    diff = -48.546073321
   
   Error:	Problem processing line 32 in file "script2.Rev"
```

bredelings:
For #565 we get:
```
generation = 0    proposal = 3/2483    NodeTimeSlideUniform(timetree)
    BEFORE:   F[1]:  -0.0198026273
    BEFORE:   morph[2]:  -178.4869602
    BEFORE:   timetree:  -788.3465682
    BEFORE:   morph[3]:  -147.4611279
    BEFORE:   seq[3]:  -10820.39915
    BEFORE:   cont_traits:  -376585.4865
    BEFORE:   F[3]:  -2.87356464
    BEFORE:   F[2]:  -1.011600912
    BEFORE:   F[9]:  -0.6931471806
    BEFORE:   F[7]:  -0.6931471806
    BEFORE:   F[8]:  -2.595254707
    BEFORE:   F[6]:  -1.671473303
    BEFORE:   F[10]:  0.8209805521
    BEFORE:   seq[1]:  -14151.73157
    BEFORE:   F[11]:  0.8209805521
    BEFORE:   F[4]:  -1.805004696
    BEFORE:   F[5]:  -1.566530411
    BEFORE:   morph[1]:  -700.8573876
    BEFORE:   seq[2]:  -19465.69126

   Error:	Executing NodeTimeSlideUniform(timetree)): PDFs not up-to-date before proposal!
       cont_traits: -376585.48653 != -376689.977    diff = 104.4904645
   
   Error:	Problem processing line 26 in file "modules/analysis/MCMC_test.Rev"
   Error:	Problem processing line 269 in file "modules/template_backbone_clade_const.Rev"
   Error:	Problem processing line 49 in file "headers/MCMC/model_TED_w_backbone_const090624.Rev"
```

davidcerny (approved):
This is enormously useful – thank you, Ben! The distinction between `debugMCMC` (thoroughness of checks) and `logMCMC` (thoroughness of reporting) makes sense to me, too. I was wondering how the system for determining the position of a proposal within a given iteration (`proposal = 3/2483` etc.) would work under `moveschedule = "single"`, but it does exactly what it should (i.e., `proposal = 1 / 1`). I'm in favor of merging this as-is.

Questions / comments:

(1) Under #567, you suggested that `debugMCMC=2` "could indicate more expensive extra checks", but from looking at the code, that's not the case, and there are just two distinct options: 0 vs. 1 or higher – correct?

(2) Your documentation for the different values of `logMCMC` is extremely useful. Where should we put it to make it as accessible as possible? Should there be a way to look it up from within the program?

src/core/utils/RbSettings.h
```diff
@@ -30,6 +30,7 @@ class RbSettings {
     double                      getTolerance(void) const;                           //!< Retrieve the tolerance for comparing doubles
     bool                        getUseScaling(void) const;                          //!< Retrieve the flag whether we should scale the likelihood in CTMC models
     int                         getDebugMCMC(void) const;                           //!< How much work should we perform to check MCMC?
+    int                         getLogMCMC(void) const;                           //!< How much work should we perform to check MCMC?
```
davidcerny, line 33: Maybe change the comment to clarify the distinction b/w the two

src/core/utils/RbSettings.h
```diff
@@ -41,6 +42,7 @@ class RbSettings {
     void                        setTolerance(double t);                             //!< Set the tolerance for comparing double
     void                        setUseScaling(bool s);                              //!< Set the flag whether we should scale the likelihood in CTMC models
     void                        setDebugMCMC(int d);                                //!< How much work should we perform to check MCMC?
+    void                        setLogMCMC(int d);                                //!< How much work should we perform to check MCMC?
```
davidcerny, line 45: Ditto

src/core/moves/MetropolisHastingsMove.cpp
```diff
@@ -258,58 +258,73 @@ void MetropolisHastingsMove::performHillClimbingMove( double lHeat, double pHeat
...
 
+std::map<const DagNode*, double> getNodePrs(const std::vector<DagNode*>& nodes, const RbOrderedSet<DagNode*>& affected_nodes)
+{
+    std::map<const DagNode*, double> Prs;
+    for(auto node: views::concat(nodes, affected_nodes))
+	Prs.insert({node, node->getLnProbability()});
+    return Prs;
+}
 
+constexpr double rel_err_threshhold = 1.0e-11;
```
bjoelle, line 269: I wonder if these two should not be into settings as well ?
or at least somewhere that is a bit more discoverable than in the middle of a random file ?

src/core/moves/MetropolisHastingsMove.cpp
```diff
@@ -258,58 +258,73 @@ void MetropolisHastingsMove::performHillClimbingMove( double lHeat, double pHeat
...
-
-    if (debugMCMC >= 2)
+    if (logMCMC >= 3)
     {
-	// --------------------------
-	//
-	//     DEBUG (BEGIN)
-	//
-	// --------------------------
+        std::cerr<<std::setprecision(10);
```
bjoelle, line 299: is there a reason we're not using the precision we selected just above ?

bjoelle (approved):
This looks good overall
I agree with David about documenting the levels - I think ideally we want to have this both on the website and within the rb help ?
would also be good to clarify the intention behind each setting (as I understand logMCMC does not impact performance or require additional calculations whereas debugMCMC can) and level (e.g. is level 4 reserved for individual move checks ?)

src/core/utils/RbSettings.h
bredelings: Fixed

src/core/utils/RbSettings.h
bredelings: Fixed

bredelings:
Two issues are what logMCMC and debugMCMC mean, and how to document them in a way that is discoverable.

For documentation, we could put info into setOption and/or getOption.  Currently I think all the options are undocumented.  Maybe we need a separate place to document options?  Additionally, it think we might want to allow some way of finding the `debugMCMC` option by searching the web for `revbayes mcmc problems` or something like that.  Maybe we could add an FAQ about debugging revbayes?

The meaning issue is a bit more tricky.  I think these two knobs should function more like developer features, and will be most useful if they don't have a fully specified interface.  Instead, they are knobs that developers can use to determine when to enable various debug checks and debug logging.  So I'd like the following guarantees:
- if `logMCMC==0`, no additional logging is performed, and no CPU time is spent on logging.
- if `debugMCMC==0`, no additional CPU time is spent on checking MCMC
- `logMCMC=n+1` logs more stuff than `logMCMC=n`, and may spend more CPU time computing the logged items.
- `debugMCMC=n+1` checks more stuff than `debugMCMC=n` and may spend more CPU time computing the checks
- `debugMCMC=1` is a reasonable thing to recommend to non-developers
Beyond that, I think we can decide on a case-by-case basis whether documenting the effects is necessary.  For `logMCMC`, the effects of increasing the level can be discovered simply by setting the option and running `rb`.  For `debugMCMC` it could be more helpful to document what is being checked, since the checks are only discoverable when they fail.

Thoughts?

bredelings:
Also, regarding `debugMCMC=2`, there may be no such checks yet, but (i) it seems good to allow for them in the future, and (ii) developers might gate various checks on this when debugging specific problems, even if the checks don't get committed to the common repo.

(In practice what happened was that at various points of debugging #565 and #567, I had checks gated on `debugMCMC` being 2, 3, or 4.  This was I wanted to enable some checks but not others.  It turned out that some of the checks suppressed the problems I was trying to check for!  But when I removed the code that suppressed the problems, it seemed reasonable to move all the remaining checks to `debugMCMC=1`.  However, in the future we might consider more expensive checks.)

some of the checks actually made the MCMC problems disappear by forcing recalculations of node only when debugging was enabled.  But for the final version I ended up eliminate

ms609 (approved):
Without looking into details of the implementation, I've built rb from this branch and it has immediately identified a problem during MCMC that I spent a couple of hours inferring by indirect means yesterday – so this gets a thumbs up from me.  Now I just need to debug the issue that it's identified...

davidcerny:
(1) Implementing options as map<string,json>: I have to admit that this idea just seems too complicated to me. I like the idea of there being predefined thoroughness levels represented by integers. If we allowed enabling and disabling individual checks, my instinct would probably still be to run all of them just in case, except now I'd be paranoid that maybe there are checks I should have enabled but didn't. I'm not convinced that the increased flexibility we'd buy ourselves by doing this is realistically needed.

(2) Where to put the documentation: for me, a pretty natural place would be the help page we display on invoking the `-h` flag. Currently, it just says this:

```
  --setOption arg       Set an option key=value.
```

I think it should also say (1) what keys are available, (2) what values they can be set to, and (3) what these values mean.

I agree all of this should also go on the website somewhere.

(3) Should we be listing specific things that are done when invoking, say, `--setOption logMCMC=2`, or should we just provide general guidance along the lines of "`logMCMC=n+1` is more detailed but potentially also more time-intensive than `logMCMC=n`"? I'm definitely in favor of the former. To me, the latter seems to anticipate a great deal of future development which I'm not sure is immediately forthcoming. I'd much rather document things in detail as they are now, and either change or add to the documentation if their meaning changes or if new levels are added.

> For `logMCMC`, the effects of increasing the level can be discovered simply by setting the option and running `rb`.

Well, not quite. Unless your script contains an `mvSlice`, `logMCMC=3` and `logMCMC=4` will look exactly the same.

bredelings:
I ended up bailing on adding more options because I think we need this debugging functionality immediately.  We can consider a more extensible setting framework later.  So I just changed all the error thresholds to `1e-11` and all the precision values to `11`.

PR #580, opened 2024-11-05
Fix `print(time)`

### LINKED ISSUE #571: [BUG] printing `time()` fails

ms609:
Implements @bredelings fix for #571.

Confirmed working on Windows; closes #571

### DIFF
--- a/src/revlanguage/functions/basic/Func_time.cpp
+++ b/src/revlanguage/functions/basic/Func_time.cpp
@@ -40,7 +40,7 @@ RevPtr<RevVariable> Func_time::execute( void )
     
     boost::posix_time::ptime t = boost::posix_time::microsec_clock::local_time();
 
-    long time = 0;
+    long std::int64_t = 0;
     if ( option == "year" )
     {
         time = t.date().year();


bredelings:
I guess this confirms what the problem was :-)

bredelings (dismissed):

ms609:
Or perhaps not quite: I notice that on Windows the `time()` prints 67492225, whereas on a few moments earlier on Ubuntu on the same machine (via WSL) the time was 19718262083631 .  Even if Windows users might be considered stuck in the past, that's still quite a discrepancy.  

Is there a case for setting a reference point in time (perhaps the start of the Unix epoch?) rather than using an OS-specific value?

bredelings:
Probably a number of different places should be changed to replace `long` with `std::int64_t`.  For example, I wonder what is the C++ object behind `Natural`?

It should be possible to change the reference point to the Unix epoch by creating the data for Jan 1 1970 and subtracting that.  I looked at the documentation for boost::date_time, and it looked possible.  Long-term we want to use std::chrono though.  But I think we might need c++20 features to do that.

davidcerny:
This doesn't compile for me.

davidcerny:
Now I'm getting the right number of milliseconds since 1970-Jan-01 00:00:00!

ms609:
Me too, on Linux: but Windows is complaining about negative numbers again.  Further investigation underway...

bredelings:
On windows `long` is only 32 bits, whereas on linux and Mac `long` is 64 bits.  So it could be that other places need to use `std::int64`.

``` sh
$ git grep 'long[ ,>)]' | wc -l
1529
```

In theory every one of these places should be replaced with `std::int64_t` or `std::uint64_t`. 

But the cases inside `src/revlanguage/datatypes` might be especially important:
- `basic/Integer.h:    class Integer : public ModelObject<long> {`
- `basic/Natural.h:        Natural(long x);`

bredelings:
Also maybe you could add a test case?  That way the windows tests would run on github.  Although I'm not sure how to check for non-negative values...  Can you write the result of `time() > 0` to a file somehow?  That seems hacky...

ms609:
Yes, I've got Natural and Integer, and now I need to go deeper; RbVector.h is next.  If you're happy for me to indiscriminately replace `long` with `std::int64_t` I'll get zapping!

I thought about adding a test case – perhaps the value of time("fromBeginning") could be computed from `(time("years") - 1970) + time("days") + ...`

bredelings:
I think replacing long everywhere would be good... it looks a bit ugly, but I'm not sure what else to do.

bredelings:
Look like there's a problem in ncl where its using snprintf, and you have to specify `long` or `long long` with that.  Maybe just back out the changes to `ncl`?

Also note that `long long int` is 64 bits on Linux, Mac, and Windows, so if `std::int64_t` doesn't work somewhere, you can probably use `long long` there.

ms609:
Nearly there; @bredelings could you cast an eye at grappar.tab.cpp and see whether you can easily spot what's causing:

[build] ./grammar.tab.c:155:31: error: expected unqualified-id before 'static_cast'
[build] ./grammar.tab.c:412:3: note: in expansion of macro 'YY_CAST'
[build] ./grammar.tab.c:581:3: error: 'yy_state_t' does not name a type; did you mean 'mbstate_t'?

I couldn't easily navigate the macros.

ms609:
Well, the good news is that `time()` now works on Windows!

The bad news is that tests that include `dnCoalescentSkyline` are now failing with memory access errors; it's not obvious that this has been caused by switching from `long` to `int64_t`.  Has the switch introduced a new error, or exposed an existing one?

bredelings:
I looked into the failure using gdb.  Somehow the coalescent distribution has 21 total events, and 5 intervals, with events_per_interval = [1, 1, 1, 2, 1].  Or something like that.  So something broken, but I'd have to chase down where.

For huge changes like this, I think you want to do the changes in smaller groups, and then make sure the tests pass after each group.  Then it is easier to track down why things break.

bredelings:
I was also thinking it might be good to try and fix all the `unsigned long` before the `long` (to the extent possible) so that we make sure not to change the signedness by accident.

I started a new branch `replace-long`.  Feel free to push to that branch if you want.

ms609:
Yes, the smaller groups approach would certainly make it easier to to identify the point at which an error arises.  I started with this approach but following the connections through an unfamiliar codebase quickly became more than I could keep track of – and when it came to it, just sanity checking and correctly implementing the grep replacements across the 483 files ended up being a sizeable job on its own.

Perhaps a feasible approach is to set Y = dnCoalescentSkyline, and X = "everything else".  I'll have a stab at this under #584.

bredelings:
I hadn't thought of that, that's a good idea.  It could be possible to trace where the wrong number of values-per-interval are coming from using the debugger, but I have too many things on my plate right now to do it myself.

bjoelle:
@ms609 should we switch this PR to draft since it's still being worked on ?

davidcerny:
Closing this as superseded by #708.

Issue #581, opened 2024-11-05
[BUG] PDFs not updated


ms609:
**Describe the bug**

PDFs not updated before proposal when conducting MCMC with two morphological datasets

```
Error:       Executing Scaling(tree_length): PDFs not up-to-date before proposal!
    m_morph[2]: -1556.4265621 != -1653.7328371    diff = 97.306275003
    m_morph[1]: -13984.554612 != -13903.385021    diff = -81.16959119
```
**To Reproduce**
- Download [data.txt](https://github.com/user-attachments/files/17519572/data.txt) and [data2.txt](https://github.com/user-attachments/files/17637137/data2.txt)

- Run the following:

```rb
seed(0)
setOption("debugMCMC", 1)
moves = VectorMoves()

chars <- readDiscreteCharacterData("data.txt")
taxa <- chars.names()
outgroup <- clade(taxa[1])
nTaxa <- chars.size()
nEdge <- 2 * nTaxa - 3

topology ~ dnUniformTopology(taxa, outgroup = outgroup)
moves.append( mvNNI(topology, weight = nEdge) )

tree_length ~ dnGamma(shape = 2, rate = 2 / 3)
moves.append( mvScale(tree_length ) )

rel_br_lengths ~ dnDirichlet( rep(1.0, nEdge) )
br_lengths := rel_br_lengths * tree_length
phylogeny := treeAssembly(topology, br_lengths)

m_morph[1] ~ dnPhyloCTMC(
  tree = phylogeny,
  Q = fnJC(4),
  type = "Standard",
  coding = "variable"
)
m_morph[1].clamp(chars)

m_morph[2] ~ dnPhyloCTMC(tree = phylogeny, Q = fnJC(2), type = "Standard")
condChars <- readDiscreteCharacterData("data2.txt")
m_morph[2].clamp(condChars)

monitors[1] = mnScreen(printgen = 1, tree_length, posterior = FALSE, prior = FALSE)
mymc3 = mcmc(model(phylogeny), monitors, moves, nruns = 2, moveschedule = "single")

mymc3.run(generations = 1000)
```

**Computer info**
Latest RevBayes build from debug-mcmc branch

davidcerny:
Possibly a duplicate of #567, having to do with `treeAssembly` not properly restoring branch lengths...

ms609:
This does indeed seem to be fixed by #549 – worth verifying once that PR is merged.

davidcerny:
[...]

Issue #577, opened 2024-11-05
[BUG] Likelihood calculation broken by creating an unrelated array


ms609:
**Describe the bug**
Likelihood calculations for a model are broken when an array that contains certain parameters of the model is defined – even though the array is never used.

**To Reproduce**
- Download [data.txt](https://github.com/user-attachments/files/17519572/data.txt)
- Run the following:

```rb
chars <- readDiscreteCharacterData("data.nex")
taxa <- chars.names()
outgroup <- clade(taxa[1])
nTaxa <- chars.size()
nEdge <- 2 * nTaxa - 3

topology ~ dnUniformTopology(taxa, outgroup = outgroup)
tree_length ~ dnGamma(shape = 2, rate = 2 / 3)
rel_br_lengths ~ dnDirichlet( rep(1.0, nEdge) )
br_lengths := rel_br_lengths * tree_length
phylogeny := treeAssembly(topology, br_lengths)

k <- 4
nc := abs( 1 / ( k - 1) ) # Replace 'k' with '4' and the issue disappears
rates := [ [0, nc, nc, nc],
           [nc, 0, nc, nc],
           [nc, nc, 0, nc],
           [nc, nc, nc, 0]]

Q4 := fnFreeK(rates, FALSE, "eigen")

m_morph[1] ~ dnPhyloCTMC(
  tree = phylogeny,
  Q = Q4,
  type = "Standard",
  coding = "variable"
)
# Remove coding = "variable" and third pr1 goes from nan to -inf

m_morph[1].clamp(chars)

pr1 := m_morph[1].lnProbability()
write("1: ", pr1) # -12345.6
write("2: ", pr1) # -12345.6
rates2 := [ [ 0, nc],
            [ k * nc , 0] ] # Same when replaced k * nc with abs(k / (k - 1))
write("3: ", pr1) # nan
q()
```

**Computer info**
WSL, building from debug-mcmc branch

**Additional context**
Behaviour observed with `-setOption debugMCMC=0` and `=1`.

ms609:
This bug seems to have been fixed for Windows too.

Issue #425, opened 2024-01-26
[BUG] Can't store densities of simple math distribution in a deterministic node


davidcerny:
**Describe the bug**
For simple math distributions like `dnExp` and `dnNorm`, the output of the `.probability()` and `.lnProbability()` methods can't be stored in a deterministic node. Interestingly, it *is* possible to do this for more complex distributions like `dnBDP`.

**To Reproduce**
```
> seed(12345)
> a ~ dnExp(1.0)
> b := a.probability()
> c := a.lnProbability()
> a
   0.07298359
> a.probability()
   0.9296161
> b
   0.9296161
> a.lnProbability()
   -0.07298359
> c
   -0.07298359
> a ~ dnExp(1.0)
> a
   0.11636
> a.probability()
   0.8901547
> b
   0.9296161
> a.lnProbability()
   -0.11636
> c
   -0.07298359
```

**Expected behavior**
In the code above, the output of `a.probability()` and `b` (as well as of `a.lnProbability()` and `c`) should be identical after each redraw.

**Computer info**
macOS 12.7.1, RevBayes 1.2.2, latest version of the `development` branch (rapture-2386-g54c6ba).

ms609:
There's some other odd behaviour that I suspect might be related here.  (Perhaps it's not, and belongs in a separate issue..?)

#524 gives some context.

If I define a deterministic node before running `model.graph()`, then the call to `.graph()` causes RevBayes to crash:

```rb
p2 := m_morph[2].lnProbability()

mymodel = model(phylogeny)
mymodel.graph(dotFile) # crashes
```

If define it after defining the model:

```rb
mymodel = model(phylogeny)
p2 := m_morph[2].lnProbability()

mymodel.graph(dotFile) # works fine

monitors = VectorMonitors()
monitors.append( mnScreen(p1 ) )
```
I see
```
   Error:       Cannot find node with name 'p1' in the model but received a monitor working on it.
```

If I explicitly include the deterministic variable in the model, it seems to work:

```rb
p2 := m_morph[2].lnProbability()
mymodel = model(phylogeny, p2)

monitors = VectorMonitors()
monitors.append( mnScreen(p1 ) )
mcmc(mymodel, monitors, moves) # Runs and displays p2, which changes
```


This is possibly related to another issue I noticed some time ago (but didn't report) whereby variables are not always recognized as part of a model.

```rb
# Everything works fine:
condRates := [ [ 0, u],
               [ 4, 0] ]
[...]

PR #582, opened 2024-11-05
Alternative to dnInverse implementation

ms609:
After a largely fruitless day spent debugging `dnInverse` (motivation: #492; implementation: #562), I've experimented with a less general approach that is simpler to implement.  This PR adds a boolean `invertProbability` parameter to `dnPhyloCTMC`; if `TRUE`, the likelihood computation within that function is inverted (i.e. the reciprocal is provided).

Once other bugs (#577, <strike>#581</strike>) are resolved, it may be easier to debug `dnInverse` – but I'm also open to the possibility of removing that function in favour of the `invertProbability` parameter, which would be trivial to add to other functions if a use case arose.  Our earlier discussion made me wonder how likely `dnInverse` was to be in practice beyond the context of `dnPhyloCTMC`.


src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
```diff
@@ -2077,11 +2080,11 @@ void RevBayesCore::AbstractPhyloCTMCSiteHomogeneous<charType>::tipDrawJointCondi
             double sum = 0.0;
 
             // if the matrix is compressed use the pattern for this site
-            size_t pattern = i;
-            if ( compressed == true )
-            {
-                pattern = site_pattern[i];
-            }
+            // size_t pattern = i;
```
bjoelle, line 2083: is this change intentional ? I don't see why it's needed
either way it would probably be better to just remove those lines than comment out

bredelings:
What is the problem with the `dnInverse( )` implementation?  I thought it was working when we merged it.

ms609:
So did I – but the debugMCMC flag has revealed that the likelihood/PDF is not being properly updated.

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
ms609: This suppresses a compiler warning about an initialized but unused variable.  I'll remove the line completely.

davidcerny:
I'm wondering if it's worth trying to salvage / debug `dnInverse` instead, since a general facility for inference under a conditional is a neat extension of current RevBayes functionality. That said, I fully understand this extension was motivated by a specific research project, and if the goals of that project are more easily accomplished in this way, it's hard to justify the time and effort spent on debugging the more general implementation.

I assume that MCMC runs with the same seed produce different likelihoods when using `dnInv(dnPhyloCTMC(...))` vs. `dnPhyloCTMC(..., invertProbability = TRUE)`, and that the latter runs okay under `--setOption debugMCMC=1`?

ms609:
I agree that it'd be great in principle to retain `dnInverse`. However, debugging it lies beyond my current abilities; and a truly general approach to inference under a conditional might aim to capture conditions such as `x > 5` which can't be tackled with `dnInverse` – perhaps such an approach would make dnInverse redundant.  As it turned out to be rather straightforward to add the `invertProbability` argument, it may be more efficient in terms of developer resources to add that to other `dn` functions as the need arises.  We could even do this now if we can conceive of anticipated use cases that would not be met by the PR as is.

ms609:
In response to your second comment: The former MCMC behaves very differently (i.e. incredibly low proposal acceptance), which I now believe is due to the problem exposed by `debugMCMC = 1`; this problem does not afflict `dnPhyloCTMC(invertP = TRUE)`.

davidcerny:
> That approach would only work if the relevant distribution is modified so that the probability methods return deterministic nodes. But we could modify the distributions so that it works.

That would be amazing, actually, and would also solve #425. IIRC correctly, some distributions (birth-death processes I think?) already support this, so I'll see if I might be able to generalize it...

mikeryanmay:
Just a minor comment: if we decide to resuscitate `dnInverse`, can we consider a different name? If I saw the code `x ~ dnInverse(dnNormal(0,1))`, I would assume that 1/x ~ Normal(0,1), which is different from inverting the probability density of x ~ Normal(0,1).

PR #583, opened 2024-11-06
Document runtime options

davidcerny:
Took a stab at this, but there are some things I'm unhappy about. In particular:

1. I have no idea what we mean by "modules";
2. I have no idea what we mean by "scaling" CTMC likelihood, or by only doing it for every n-th node.


bredelings:
I don't think this is the right place for help on options:
- Options are not primarily a command-line feature.  The `setOption(key, value)` function can be run inside RevBayes; the ability to do `--setOption key=value` was only added recently.
- This info already seems too large for the `--help` message.  But we may add more options in the future, making the option help even larger.

davidcerny:
I don't find this excessive at all – you should see the output of `.iqtree2 -h`, which lists every single command available in that program, gives the full list of all implemented substitution models, etc.

If this is a bad place, so be it – but let's come up with a better one, then. I want to make sure the recent additions don't stay undocumented, with the only guidelines to their use buried in an old PR.

ms609:
Thanks for making this first pass at documentation, @davidcerny .  The first place I'd look for this information is https://revbayes.github.io/documentation/setOption.html.  I'd be in favour of also including the information here – as an inexperienced user, the easier it is to locate this sort of information, the better – or at the very least a link to its eventual location.

bredelings:
I like the idea of putting it in setOption.  Then referencing it from `--help` and from `--getOption` makes sense.

davidcerny:
I've now moved the description of available keys to `?setOption` and referenced it from `-h` and `?getOption` as suggested. Note that the description of `logMCMC=2` anticipates the merging of #592.

ms609 (approved):
Looks great - I appreciate the terse but clear explanations and clear presentation of defaults.

bredelings:
- "modules" don't currently exist.  The idea was that you could define some models/functions in the rev language and load them at startup.  Maybe they will exist later.
- scaling refers to a method of avoiding underflow in the partial likelihoods.  You multiply all the likelihoods by a factor, and then record that factor to include it back into the likelihood at the end.  

I think this is a good place to put the options.

However, the `debugMCMC` and `logMCMC` levels higher than 1 are experimental and may change at any time.  They are not part of the stable interface.  So if we are going to document them, then we need to mark them as experimental/unstable options for development and debugging.

bredelings:
Additionally, life is complicated and one problem that I found is that doing _more checks_ sometimes finds _fewer bugs_ :grimacing:.  Specifically, if we touch nodes to check that their likelihood does not change when recalculated, sometimes the recalculation makes bugs go away.  So it is going to be hard to make guarantees about what the effect of the options is.

davidcerny:
This is really useful, thanks @bredelings! I've made the following changes:

1. Removed the mention of `modules` – I see no point in documenting a nonexistent option;
2. Expanded the description of `useScaling` a little bit to explain the point is to ensure numerical stability;
3. Added a note to `logMCMC` explaining that the option is experimental and the meaning of individual values may frequently change.

I haven't added an equivalent note to `debugMCMC`, because as far as I understand, _right now_, it has no levels higher than 1 (or rather, specifying values higher than 1 doesn't do anything different from specifying 1).

I don't think it's a problem that more thorough checks may find fewer bugs. The documentation does _not_ promise that you'll find more bugs if you pick higher values.

> However, the `debugMCMC` and `logMCMC` levels higher than 1 are experimental and may change at any time.

I mean, it _would_ be great if changing them also involved a change to this documentation, then. Surely a debugging facility with multiple options is more useful if there's a way to learn what these options do, without having to remind yourself of it by searching the source code for every instance of `if (logMCMC >= 1)`, `if (logMCMC >= 2)` etc. every time you want to debug something.

PR #584, opened 2024-11-07
long→int64_t in CoalescentSkyline.cpp

ms609:
Partial implementation of switch from long to int64_t, per #580


ms609:
I've poked around a bit and for this c++ novice a 'bit-by-bit' implementation is too complicated; having got down to RbVector I've not managed to isolate the simplest unit at which `long` can be converted to `int64_t`; as this is a widely used class I think too much of the codebase is embroiled.

bredelings:
Yeah... maybe just wait until someone has time to run the non-incremental version in the debugger and figure out what is not working.

Issue #308, opened 2022-11-17
dnUniformTopologyBranchLength only accepts dnUnif w/ upper bound > 1


davidcerny:
One of my students recently ran into the following issue, which I was able to reproduce for a different dataset (example data attached below):

A uniform distribution can be passed to the `branchLengthDistribution` argument of `dnUniformTopologyBranchLength()` just fine if its upper bound exceeds 1:

```
morpho = readDiscreteCharacterData("example_data.txt")
taxa <- morpho.names()
out_group <- clade("Hesperocyoninae")

# This works
tree ~ dnUniformTopologyBranchLength(taxa, outgroup=out_group, branchLengthDistribution=dnUniform(0, 5))
```

However, when the upper bound is less than 1, RevBayes throws the following error:

```
Error:	Argument or label mismatch for function call.
   Provided call:
   dnUniformTopologyBranchLength (Taxon[]<constant>,
                                  Clade<constant> 'outgroup',
                                  Dist_unifProbability 'branchLengthDistribution' )
   
   Correct usage is:
   dnUniformTopologyBranchLength (Taxon[]<any> taxa,
                                  Clade<any> outgroup,
                                  Distribution__RealPos<any> branchLengthDistribution)
```

The following functionally equivalent solution still works:

```
topology ~ dnUniformTopology(taxa, outgroup=out_group)
for (i in 1:num_branches) {
  branch_lengths[i] ~ dnUniform(0, 0.5)
}
tree := treeAssembly(topology, branch_lengths)
[example_data.txt](https://github.com/revbayes/revbayes/files/10028187/example_data.txt)

**Comment by bredelings** [comment]
That seems wrong.  I guess this is because `0.5` and `5` have different types:

```
> x = dnUniform(0, 5)
> y = dnUniform(0 ,0.5)
> type(x)
   Dist_unif
> type(y)
   Dist_unifProbability
```

Is there a trick to make `0.5` into a RealPos?

**Comment by davidcerny** [comment]
Hmm, so it gets even more complicated. The second solution I showed (relying on `treeAssembly()`) *appears* to work, but the resulting elements of the `branch_lengths` vector are actually of type Probability, which can still lead to issues:

```
> tree_length := sum(branch_lengths)
   Error:	Ambiguous call to function 'sum' with arguments ( Probability[] )
   Potentially matching functions are:
   sum (Real[]<any> x)
   sum (RealPos[]<any> x)
   sum (Integer[]<any> x)
[...]

Issue #585, opened 2024-11-08
[BUG] Variable type changed when used in a calculation


ms609:
**Describe the bug**
Performing an operation on a value changes its type.

**To Reproduce**
```rb
x <- 2
type(x, TRUE) # Natural, Integer, RevObject
rep(2, 2) # [ 2, 2 ]
x / 2 # Should not change x!
type(x, TRUE) # RealPos, Real, RevObject
rep(x, x) # [ 2.0 ]
```

**Computer info**
Win 10, development build

bredelings:
See https://github.com/revbayes/revbayes/pull/545 by @bjoelle.

The fact that variables can have their type changed by a calling function is apparently part of the intended design of RevBayes.  But it makes it hard to reason about types, and causes a variety of problems.

In a lot of cases this behavior could be fixed by adding a type-conversion DAG node like #545 does.  But there's one case that doesn't fix.  Imagine that we have something like
```
x[1] = 1
x[2] = 2
x[3] = 3.5
```
If RevBayes processes these statements one-at-a-time, then after it processes `x[1] = 1` its not yet clear what the type of `x` should be.  If could be `vector of integer` or `vector of real`.  So I think what revbayes does it set it to `vector of integer` (the most contrained type) and then change the type to `vector of real` later.

The main reason that RevBayes has this problem is because of the command interface -- the REPL (read-eval-print-loop).  If RevBayes could analyze the entire program before executing then we could determine a single static type and never change it.  Does that make sense?

bredelings:
Interesting.  In that case we might do a good-enough fix by just eliminating type changes that result from argument type matching, and we can do that by introducing type-conversion nodes like #545.

We should do that.  I  will put it on my list, although the list is not short so I may not get to it immediately.

Issue #414, opened 2023-12-03
basic arithmetic conversion


igoryanovich:
Currently, there seems to be no available documentation on type conversion within Rev. Of course, most of the time this is just fine. But I've just spent a few hours trying to sort out a problem, which ultimately arose because (1-_b_) with _b_ a Probability, resulted in Real rather than RealPos, and this ultimately led to the fn rate matrix function to reject the input matrix with rates. I've found a work-around by converting the result explicitly using Probability(), but this was not overtly documented, as far as I could see. 
  (I have not managed to infer this information from studying the C++ source; granted, I am not very experienced with the overall code. But luckily for me, a reference to Probability() is present [in here](https://applibugs.mathnum.inrae.fr/sites/default/files/2021-07/fichier-document/applibugs.18_06_21.bboussau.pdf)...)

I can see several things that it'd be nice to have: 
(1) perhaps check if automatic conversion by subtraction can go for the most restricted type possible, as is already the behavior with many other operations? 
(2) perhaps add more explicit conversion operators in Rev, specifically RealPos() with the analogous behavior to Probability()? 
(3) add info about Probability() into the open documentation.

PR #587, opened 2024-11-08
Support Real _n_ in `rep(x, n)`

### LINKED ISSUE #586: [Feature Request] Support Reals in rep
`rep(x, n)` currently requires that `n` is a Natural.  If `n` is a RealPos, it behaves as if `n == 1`.

```rb
rep(2, 2) # [ 2, 2 ]
rep(2, 4 / 2) # [ 2 ]
```

By analogy with R, I suggest implementing rep such that the floor of a RealPos is used.

I've not seen any user-facing functions that are documented as accepting multiple types for the same argument – though this does seem to be the case already for `x`, which the documentation stipulates as requiring an Integer, though RealPos are accepted.  (Perhaps this could be noted there?)

ms609:
Closes #586 .

Allows
```rb
rep(1, 4 / 2)
rep(1, 2.0)
```

Previously these returned single-element vectors.


ms609:
This Windows test failure is odd.  When I run the FBD tests locally on Win10 the files are created.  The GHA logs make it look like the files are created but not copied correctly.

bredelings:
Hmm.... it seems like the deeper problem is that RevBayes allows passing a Real to a Natural, but doesn't  convert it.

bredelings:
For example:
```
> _mod
   mod (Natural<any> x,
        Natural<any> y)
> _mod(2,5)
   2
> _mod(2.0,5)
   2
> _mod(4/2,5)
   2
```

bredelings:
Also apparently this work for me:
```
> rep(1, 4/2)
   [ 1, 1 ]
> rep(1, 2.0)
   [ 1, 1 ]
```

ms609:
> Also apparently this work for me:
> 
> ```
> > rep(1, 4/2)
>    [ 1, 1 ]
> > rep(1, 2.0)
>    [ 1, 1 ]
> ```

Could this be because different OSs/architectures handle differently the cast to `int` in `(int)static_cast<const Natural &>( this->args[1].getVariable()->getRevObject() ).getValue();`?  I was surprised that reals all seem to get cast to `1` on my machine.  (I think from memory this was true under WSL as well as Windows?)

ms609:
> I'm also curious what happens if you specify the argument names and if you give real numbers that aren't integers:


My guess is that the different behaviour reflects different cases that may or may not lead to calls to `ArgumentRule::fitArgument`.

ms609:
On development branch, I saw the same output as you did.

davidcerny (commented):
Hmm... to be honest, I don't necessarily see this as a good idea. The number of repetitions can only take natural numbers as meaningful values. In R, `rep(1, 2.5)` does apparently yield a vector of length 2, but in Python, `[1]*2.5` throws a type error (`can't multiply sequence by non-int of type 'float'`), and the latter seems like the correct behavior to me.

I agree that if `rep(1, 4/2)` and `rep(1, 2.0)` yield vectors of length 1 on Windows, that's an issue that's worth fixing, but IMO the fix shouldn't make it possible to also evaluate nonsensical expressions like `rep(1, 2.5)` or `rep(1, 5/2)`. I can also confirm that just like Ben, I'm getting the correct output for the first two of these expressions (i.e., `[ 1, 1 ]`) with the development branch on macOS. Getting rid of the inconsistent behavior on Windows vs. Unix is definitely worthwhile, but I'm not a big fan of rounding down `n` to the nearest integer.

ms609:
If the intention is for users not to need to think about the types of their variables, I think it is important to support Real-typed variables. Insofar as the RevBayes philosophy is for a models to be explicitly specified, I can see a case for throwing an error on non integer Reals, perhaps pointing the user to call `ceiling` or `floor` for themselves.

bredelings:
Hmm... so there are a few cases:
- `rep(x, 4/2)`
- `rep(x,5/2)`
- `rep(x,sqrt(y))`
I kind of feel like the first one should work.  The second one, I'm not sure.  I feel like the third one should not work.

On the other hand, it seems like it would be better to implement a _general_ approach that converts `Real` to `Integer` if this is something good to do, instead of manually adding a second version of each function that takes an `Integer`.

I'm not sure this is the RevBayes philosophy, but moving away from implicit conversions seems like a good idea.  Especially loss type conversions like `5/2 -> 2`.  It has the unfortunately downside of making code more verbose though.

I guess in R all numbers are `Real`, so this doesn't come up.

bredelings:
Another idea would be to introduce a 'divide-and-floor' operator.  In R that is `%/%`.

bredelings:
https://search.r-project.org/CRAN/refmans/numbers/html/div.html\

Part of the issue is the standard `/` operator effectively works like `Integer x Integer -> Real`

davidcerny:
@ms609:

> I think it is important to support Real-typed integers. [...] I can see a case for throwing an error on non integer Reals, perhaps pointing the user to call `ceiling` or `floor` for themselves.

I agree with this 100%. I think maybe the goal of this PR should be to extend the current behavior on Unix (i.e., `rep(1, 4/2)` works but `rep(1, 5/2)` does not) to Windows as well.

@bredelings:

I don't see why `rep(x, sqrt(y))` shouldn't work as long as `sqrt(y)` evaluates to a natural number, which actually is the current behavior on Unix:

```
> y <- 9
> rep(1, sqrt(y))
   [ 1, 1, 1 ]
> y <- 8
> rep(1, sqrt(y))
   Error:	No overloaded function 'rep' matches for arguments ( Natural, RealPos )
```

In general, I think implicit conversions (i.e., weaker typing) can be a good thing as long as they are coupled with evaluation and checking. With this example, RevBayes (on Unix at least) is able to determine that `4/2` or `sqrt(9)` evaluates to a natural number (in which case it performs the appropriate conversion) but `5/2` and `sqrt(8)` do not (in which case it throws a type error). That seems like the right behavior to me. My problem with the type system is that it's inconsistent, and we can't rely on this desirable behavior across the board. For example, while `rep` evaluates (and if need be, converts) its `n` argument, `dnExp` doesn't do the same for its `lambda` argument, so `a <- 3.2 - 1.6; b ~ dnExp(1/a)` fails.

ms609:
> On the other hand, it seems like it would be better to implement a _general_ approach that converts `Real` to `Integer` if this is something good to do, instead of manually adding a second version of each function that takes an `Integer`.

Agreed; ideally this would happen without the user thinking about it.  Presumably we would still need to register both function signatures in `RbRegister_*`.

For the meantime I've updated the code to fail on non-integer `Real` inputs, which I'm testing locally now before I push.  I want to include test cases to check that the 'fail' cases fail as anticipated; is there a mechanism for that?  I think if I just include the failures in `tests/.../replicate.Rev` the script will terminate at the first error.

bredelings:
I don't really like the approach of adding an overload with a Real argument.  I think that if the user wants a lossy type conversion they should use floor.

I think we should fix windows so that 4/2 works but 5/2 does not, the same as on UNIX.  Arguably we can/should add a message when trying to pass a Real to an Integer that this can be accomplished with `floor( )`.

bredelings:
> I don't see why rep(x, sqrt(y)) shouldn't work as long as sqrt(y) evaluates to a natural number, which actually is the current behavior on Unix:

This is the current behavior, and I'd lean against changing it.

However, if `y` is a deterministic node, then probably `sqrt(y)` evaluates to a natural number in some cases and not others. :man_shrugging:

ms609:
#585 brought this to the surface; `rep` failed in a context like:

```rb
x <- 4          # x is an Integer
y <- x / 2      # silently converts x to a Real
rep(0, x)       # fails unexpectedly
```

Even now that I'm somewhat familiar with RevBayes, It certainly didn't strike me as intuitive that adding `floor` would fix this issue – or indeed that type conversion was relevant to the situation.

Perhaps there's a broader question here about how we communicate to the user what RevBayes is doing when types differ from the expected.  As a user, I've found it counterintuitive to have to use `abs()` to convince RevBayes that a `Real` I know to be positive is positive (coercing to `RealPos`); I'd find it similarly counterintuitive to use `floor` on a number I know to be an integer.  Sure, we can make the user do this themselves, and implement whatever checks are necessary – but it feels to me like a more straightforward user experience if RevBayes handles basic validation and supports valid input whatever the type.

bredelings:
```
> x <- 4
> y <- x/2
> rep(0,x)
   [ 0, 0, 0, 0 ]
```

This works?

bredelings:
I think it might be helpful to divide revbayes type issues into at least different classes.
1. crazy things that only revbayes does
2. non-crazy things that only revbayes does
3. things that some languages do that are not crazy

Change the type of `x` when you write `x/2` is in class 1 (rare, crazy).  Treating RealPos and Real as different types is in class 2 (rare, not crazy). Treating an Integer 4 different than a Real 4 is in class 3 (common, not crazy).  

A lot of languages require you to do explicit conversions to change types.  The question isn't if 4 is mathematically an integer.  The question is what bit representation is chosen to represent the 4.  And `Real` and `Integer` use different bit representations for the 4.  So arguably these are not the same thing.

In `C`, despite the fact that these things have different bit representations, you don't need to do explicit casting.  Sometimes this causes trouble.   For the example the following program:
```
#include <iostream>

int main()
{
    double z = 1e38;
    int a = z;
    unsigned b = z;
    long c = z;
    unsigned long d = z;
    std::cout<<"z = "<<z<<"   a = "<<a<<"  b = "<<b<<"  c = "<<c<<"  d = "<<d<<"\n";
}
```
compiles with no errors and produces the following output:
```
z = 1e+38   a = -2147483648  b = 0  c = -9223372036854775808  d = 0
```
Interestingly, when compiled with `-O3`, it gives different output:
```
z = 1e+38   a = 2147483647  b = 4294967295  c = 9223372036854775807  d = 18446744073709551615
```
In order to prevent crazy results like this, many languages force you to explicitly convert numeric types with different bit representations.

ms609:
I agree that it's best to adopt a systematic approach to issues like this.  Since RevBayes likens itself to R (e.g. [here](https://revbayes.github.io/tutorials/intro/)] and many users are likely to be familiar with R syntax, I suspect that many RevBayes users will not be accustomed to thinking about types – besides other benefits, consistent behaviour when a type is 'incorrect' will help users to learn this distinction.  Also useful would be to direct users to clear documentation on how to manage types – for example, R has `as.integer`, so it took me a bit of thinking before I realised that `abs` could serve as `as.RealPos`.

davidcerny:
This is a great example of my [earlier complaint](https://davidcerny.github.io/post/teaching_revbayes/#fn:4) about Rev being strongly but implicitly typed. I think it would be completely fair to make the user think about the bit representation of a given value if the language was _explicitly_ typed, i.e., if it had type annotations:

```
Natural a <- 4
Real b <- 4
rep(1, a) # should work
rep(1, b) # should not
```

But because of design decisions that took place very early on, this is not the case. I guess we might eventually move in that direction a little bit by adding more function-style type casts along the lines of what we already have in `Probability()`: this was suggested in #414, and would help fix #308, for example.

However, for now, I tend to agree with @ms609 that the user is right to expect, say, integer values to behave as integers even if their underlying representation is different. If, in this particular case, this can be accomplished by adding an overload and checking the input, I'd be happy to approve.

bredelings (changes requested):
I don't think that adding a new signature with a Real argument is the right approach.

Also, it seems like supplying `4/2` _already_ converts to an integer, so it also seems like this is unneeded?

bjoelle:
so from what I remember of my experiments with conversions, Rev currently supports converting `Real` to `Integer` if the conversion doesn't change the value (i.e. `2.0 -> 2` is (intended to be) possible but not `2.5 -> 2`)
at the moment this check is based on casting however (see `Real::convertTo`) which may be why there is OS-dependent weirdness

bjoelle:
also for added fun, note that the conversions will treat constant nodes differently than the others
the type changes only happen to constant nodes, other nodes will have a deterministic conversion node created

davidcerny:
Unfortunately, the latest commit doesn't preserve the desirable Unix behavior:

```
> rep(1, 5/2)
   [ 1, 1 ]
> rep(1, 2.5)
   [ 1, 1 ]
> y <- 8    
> rep(1, sqrt(y))
   [ 1, 1 ]
```

davidcerny:
@ms609 This looks very good, and preserves the desirable Unix behavior. I think that to make the tests pass, you just need to update the expected output?

ms609:
I've done so, but the expected output only covers the tests that do not throw an error; once the first error is thrown, the script terminates, without verifying that the error is displayed as intended or that subsequent errors are observed.

ms609:
As an imperfect workaround, I've created separate a test file for each error case; if an error is not thrown, this will overwrite the expected content in the output file.  This doesn't check that an error occurs for the anticipated reason, but is hopefully better than nothing.

davidcerny:
@ms609 Would you maybe like to push that workaround to see if the test runner can handle it? This PR is basically ready to be merged as far as I'm concerned; I'm happy to approve it as soon as it passes the tests.

The idea of a failing test producing a file where there should be none is interesting, and could be exploited elsewhere if it works.

ms609:
Ah! Thanks for spotting that the push hadn't completed – I was pushing through WSL and didn't spot an error message.  Let's see what the test runner makes of the approach...

davidcerny (approved):
Works exactly as intended, and on top of that provides a very nice solution to the problem of testing for cases that are supposed to fail. Approving!

ms609:
Great, thanks.  In `tests/run_integration_tests.sh`, I'm not sure what the purpose of checking for an `XFAIL` file in the test directory root would be – I wonder whether this was originally included in the preceding loop, or whether perhaps this is intended for developer use.  As it's not harmful, I'm inclined to leave it.

@bredelings are you satisfied that this now brings Windows behaviour in line with Unix, whilst retaining failure on inappropriate input?

davidcerny:
This is ready for (re-)review, I think.

bredelings:
> I don't think that adding a new signature with a Real argument is the right approach.
>
> Also, it seems like supplying 4/2 _already_ converts to an integer, so it also seems like this is unneeded?

I'm not sure I ever heard a response to the question here...  did I miss something?

davidcerny:
Since `rep(1, 4 / 2)` returns `[ 1 ]` on Windows according to the issue that this PR is trying to solve, it sounds like whatever conversion is currently taking place is not robust enough...

bredelings:
So shouldn't we find where that conversion is going wrong and fix it?

Additionally, `rep(1, 4 / 2)` returns `[1, 1]` when I run `rb.exe` under wine after compiling it for windows.  So that is confusing :-(

ms609:
`rep(1, 4 / 2)`, and `rep(sqrt(4))`, do in fact return `[1, 1]` on Windows – I hadn't appreciated that these retained the integer type (as `type(4/2)` returns `RealPos`). `rep(1, 5 / 2)` throws a "no overloaded function" error.  

The initial motivation for this change was to handle a case where a number defined as an integer later had its type changed to a Real (#585).   I am in fact finding it difficult to construct an analogous failure case that doesn't rely on this behaviour.  This PR should nonetheless make us robust to equivalent cases, should they arise.

davidcerny:
I guess I'm getting a little confused. Is #586 basically a duplicate of #585 then? I.e. the problem is not with `rep()` at all, but with the unexpected and undesirable change of type?

ms609:
I'm also feeling a bit confused, as I now can't reproduce the original behaviour from the script in #585 in either rb development or rb1.2.4, and I can't see why this would be the case.

davidcerny:
Since #586 has been closed, should we perhaps close this PR as well?

ms609:
I wonder whether it might be worth retaining the addition to documentation, and the test cases?

davidcerny:
That sounds like a good idea. If you want to open a new PR for these additions, I'll be happy to approve it.

PR #590, opened 2024-11-08
Remove useless exceptions

bredelings:
Revbayes throw a lot of exceptions that are immediately caught and ignored.  This PR removes some of these exceptions.  The main benefits are:
1. should be faster
2. it is possible to do `catch throw` in gdb and find where an error comes from.


wrightaprilm (approved):
Looks reasonable.  Tests pass.

PR #592, opened 2024-11-09
Log reason for -Inf and NaN

bredelings:
This PR allows logging a reason why we return -Inf.

```
             if( the_node.getAge() - the_node.getParent().getAge() > 0 )
            {
                return withReason(Double::neginf)<<"Pr(tree)=0: node age "<<the_node.getAge()<<" greater than parent age "<<the_node.getParent().getAge();
            }
```

The reason is only logged if `logMCMC >= 2`.

Issues:
- should we use a separate option?  Or `logMCMC >= 1`?

Changes:
- The message no longer needs to specify "\n".

### DIFF
--- a/src/core/distributions/phylogenetics/tree/AbstractRootedTreeDistribution.cpp
+++ b/src/core/distributions/phylogenetics/tree/AbstractRootedTreeDistribution.cpp
@@ -20,6 +20,7 @@
 #include "Tree.h"
 #include "TypedDagNode.h"
 #include "TypedDistribution.h"
+#include "RbSettings.h"
 
 namespace RevBayesCore { class DagNode; }
 namespace RevBayesCore { template <class valueType> class RbOrderedSet; }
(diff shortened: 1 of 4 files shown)


src/core/utils/RbConstants.h
```diff
@@ -30,8 +33,40 @@ namespace RbConstants {
...
+		std::cerr<<t;
+	    return *this;
+	}
+
+	struct nanWithReason
+	{
+	    template <typename T>
+	    nanWithReason& operator<<(const T&);
+
+	    operator double() const {return nan;}
```
ms609, line 58: Just glancing over this on my tablet so might've missed something, but could you append the \n here (and above) instead of to each message to defeat the annoyance?

bredelings: Good question.  I think if you do `nanWithReasion()<<a<<b<<c` that will put a `\n` after `a`, `b`, and `c`, (three `\n`s) instead of once at the end.

I'm thinking I might store a `std::ostringstream` inside the `struct` instead.  Then we can add a '\n' when we actually print the message.

davidcerny (approved):
Looks good!

PR #594, opened 2024-11-09
Fix mvRootNodeTimeSlideUniform

bredelings:
This PR fixes mvRootNodeTimeSlideUniform by recording the stored age before we stop the proposal by returning `-Inf` when the root is a sampled ancestor.  This prevents undoProposal from setting the wrong age (which we caught because it was outside the age range).

Instead of modifying `TopologyNode::setAge( )` the PR just adds an explanatory note there.


davidcerny (dismissed):
This looks good to me; thank you Ben!

I was wondering about the comment that @bjoelle made on Slack – without the `return` statement, perhaps an attempt to set the age outside the taxon's age range can only be guaranteed to yield an `-Inf` log likelihood if we are sampling from a distribution spanning the age range. If that was the case, removing the `return` statement would be a bad idea, since it could lead to undesirable behavior in analyses with fixed taxon ages.

Fortunately, it _doesn't_ seem to be the case. I happen to have an [analysis](https://github.com/user-attachments/files/17689752/594_test.zip) that's perfect for testing this – all taxon ages are fixed (to the minimum values specified in the taxon file), and the oldest taxon ("Leptocyon_mollis", at 29.0 Ma) tends to become the root (i.e., a sampled ancestor of all the remaining taxa) in a majority of the samples. When I investigated the sampled trees, the age of "Leptocyon_mollis" was always equal to 29.0 Ma as expected. The root age was either also equal to this value (when the taxon represented a sampled ancestor) or greater (when the taxon was subtended by a nonzero-length branch).

That said, when I ran the analysis under `--setOption logMCMC=2`, I noticed that all the `-Inf` log acceptance ratios were coming from `CollapseExpandFossilBranch` or `FNPR`, not `RootTimeSlideUniform`, which I find a bit mystifying. Maybe that's because this branch doesn't detect all instances of `-Inf` log likelihoods, whereas the more thorough checks implemented in `current_stored` do? Then again, I also noticed that while `TopologyNode` would previously flood the screen with the warnings about an attempt to set the age to an illegal value, the modified warning on this branch was never triggered.

This is not in any way bad – the analysis yields sensible results without errors or warnings. I'm just wondering why the attempt to move the age of the root would _not_ yield an `-Inf` log likelihood when the root coincides with a sampled ancestor of a fixed age.

davidcerny:
Hmm... unfortunately, I have to withdraw my approval. This PR brings back issue #544. I'm getting:

```
setAge: new age 243.735 of taxon Lutungutali_sitwensis is outside age range [235.8, 243.5]
Age of taxa Lutungutali_sitwensis incompatible with age rangeAge of taxa Lutungutali_sitwensis incompatible with age rangeAge of taxa Lutungutali_sitwensis incompatible with age range <printed back to back thousands of times>
zsh: abort
```

bredelings:
@bjoelle So, would it be correct to say that the whole point of the `return` statement is to cause the situation where the parent and and tip age for a sampled ancestor are different, leading to a rejection of the proposed tree?

bredelings:
> setAge: new age 243.735 of taxon Lutungutali_sitwensis is outside age range [235.8, 243.5]
Age of taxa Lutungutali_sitwensis incompatible with age rangeAge of taxa Lutungutali_sitwensis incompatible with age rangeAge of taxa Lutungutali_sitwensis incompatible with age range <printed back to back thousands of times>
zsh: abort

I'm running `test1.Rev` from the data set from #544 and I've gotten up to generation 167,000 without seeing any crash.

Can you run this with `--setOption logMCMC=2` and tell me which move caused this?

davidcerny:
Yeah, this issue is annoying, because many iterations can pass before it occurs, and since the test dataset is large, it takes some time. I was "lucky" in that my first attempt crashed almost immediately, but a second one only did after ~26 minutes.

I'll experiment with different random seeds and report back. Maybe if I can trigger the issue early enough, we can add the dataset as a `RootTimeSlideUniform` test.

bredelings:
Yeah, I still haven't seen it yet.  I have been looking on branch `current_stored`, perhaps I should try looking on this branch.

bredelings:
```
generation = 97229    proposal = 1/1    FNPR(fbd_tree)
    log(posterior_ratio) = -117.245  log(likelihood_ratio) = -117.245   log(prior_ratio) = 0
    log(acceptance_ratio) = -117.245  log(hastings_ratio) = 0
  The move was REJECTED.

generation = 97230    proposal = 1/1    CollapseExpandFossilBranch(fbd_tree,root_age)
setAge: new age 245.652 of taxon Lutungutali_sitwensis is outside age range [235.8, 243.5]
Age of taxa Lutungutali_sitwensis incompatible with age range    log(posterior_ratio) = -inf  log(likelihood_ratio) = 0   log(prior_ratio) = -inf
    log(acceptance_ratio) = -inf  log(hastings_ratio) = -0.766566
  The move was REJECTED.

generation = 97231    proposal = 1/1    FNPR(fbd_tree)
Age of taxa Lutungutali_sitwensis incompatible with age rangeAge of taxa Lutungutali_sitwensis incompatible with age range    log(posterior_ratio) = -nan  log(likelihood_ratio) = 0   log(prior_ratio) = -nan
    log(acceptance_ratio) = -nan  log(hastings_ratio) = 0
  The move was REJECTED.

generation = 97232    proposal = 1/1    CollapseExpandFossilBranch(fbd_tree,root_age)
Age of taxa Lutungutali_sitwensis incompatible with age range
```

davidcerny:
Okay, here's the best I can do so far (as the meme goes, "it ain't much, but it's honest work"): with `seed(395848862)` and the weight of `mvRootTimeSlideUniform` increased from 1 to 20, I get the crash at iteration 132914. With `--setOption logMCMC=2`, the output is:

```
generation = 132913    proposal = 1/1    NarrowExchange(fbd_tree)
    log(posterior_ratio) = 0  log(likelihood_ratio) = 0   log(prior_ratio) = 0
    log(acceptance_ratio) = -inf  log(hastings_ratio) = -inf
  The move was REJECTED.

generation = 132914    proposal = 1/1    SliceSampling(alpha)
mvSlice(alpha,stepping_out):  x0 = 1.08167  g(x0) = -23836.2   logy = -23836.8
     L0 = 0.23649   x0 = 1.08167   R0 = 1.23649
```

There is no line saying whether the slice move was accepted or rejected, so I guess the crash must have taken place _within_ the move, before reaching the point in the control flow where the message gets printed. In the entire analysis, there is just one time when `log(prior_ratio) = nan`: in iteration 234, for the `FossilTipTimeSlideUniform` move. (The previous iteration, 233, consisted of a `NodeTimeSlideBeta` move that had a finite negative acceptance ratio.) There are no instances of the log prior ratio being equal to `-nan`.

Here is the modified script and the associated data again, just in case: [fbd_mvslice_crash.zip](https://github.com/user-attachments/files/17690399/fbd_mvslice_crash.zip)

For this analysis, I compiled RB from `fix-root-node-time-slide`, not `current_stored`. When I put back `return;` into `TopologyNode::setAge()` and recompile, the analysis starts to diverge from the original version at iteration 47360, where the log posterior ratio in `CollapseExpandFossilBranch` becomes `-inf` instead of 0, and the order of the subsequent moves changes. The crash no longer occurs (at least not within the first ~370,000 iterations), but perhaps that's just a stochastic result of the changed move schedule rather than a sign the issue was resolved.

Reaching iteration 132914 takes ~10.5 minutes on my M3 Max machine, so it's still impractical for an integration test. I'll keep cycling through different seeds to see if I can find out a better one.

davidcerny:
On `current_stored`, the output is exactly the same.

bredelings:
Was that with JUST `--setOption logMCMC=2` or with `--setOption logMCMC=2 --setOption debugMCMC=1`?

davidcerny:
The former. With the latter, I get a silent quit (not a crash) at iteration 47361.

bredelings:
I removed the changes to `TopologyNode::setAge( )` in favor of adding an explanatory note there.

davidcerny:
Great. I'll recompile and check if this gets rid of the crashes. (FWIW, the earliest instance I've found so far occurs with `seed(936160379)` after ~50,000 iterations.)

davidcerny (approved):
All 8 random seeds that previously resulted in a crash within the first ~120,000 iterations now run without issues for >300,000 iterations, so the problem appears to be resolved. With fixed taxon ages, the age of the root is only being moved when the root doesn't coincide with a sampled ancestor, and kept constant otherwise. This all looks good!

PR #595, opened 2024-11-11
Get rid of `sprintf` in favor of `snprintf`

davidcerny:
In the spirit of cleaning up the code base and getting rid of compiler warnings (see e.g. #589), here's another small change to prevent Xcode from screaming at me...


ms609 (approved):
Ah yes, that's a familiar edit to me...
Is it worth doing this across the entire codebase, or would this induce complication?

davidcerny:
This was the only remaining occurrence of `sprintf` I could find, so we should be good!

PR #596, opened 2024-11-14
Fix problems with brownian motion distributions.

bredelings:
In #565 there is a problem with `phyloBrownianProcessREML` having invalid likelihoods after a proposed move is rejected.  The same occurs with `dnPhyloBrownianMVN` in `tests/test_BM_MVN`.  This PR fixes invalid likelihoods in both cases.  I also proactively addressed possible problems in two other `phyloBrownian` cases.


mikeryanmay:
I was doing a little bit of debugging on #565 and had been able to resolve the flatlining problem by forcing all nodes to be dirty during the likelihood recalculation in `PhyloBrownianProcessREML` but had not narrowed the problem down to an error with `restoreSpecialization`. The solution here is clearly better. It seems like the issue with restoreSpecialization is shared with the other distributions as well, and you've implemented comparable fixes.

I have a minor concern about `PhyloBrownianProcessMVN` though. If I understand correctly, we are juggling old/new states of the VCV, but not old/new states of the inverse VCV. Previously, `restoreSpecialization` wasn't updating the inverse, so subsequent calls to `computeLnProbability` could use the out-of-date inverse. The fix here is to force recomputation of the inverse after `restoreSpecialization`, which makes sense. What I find confusing (and which is unrelated to the current fix, if I understand correctly) is that the new/old juggling is handled differently for the VCV and the inverse VCV, which is a bit counter-intuitive (for me at least), and I'm worried it could be confusing for a naive maintainer. Also, performance-wise, the inverse VCV is usually going to be much more expensive to compute than the VCV, so it's odd that we're caching/restoring the VCV but not the inverse VCV (in fact, we probably only need to be caching the inverse at all, as the only function of the VCV is to get the inverse VCV). Basically, I think this code could be cleaned up/streamlined quite a bit, but I think that's mostly a maintenance/performance concern so maybe not in scope for this pull request.

mikeryanmay (approved):
(I meant to end my previous comment with an approve!)

bredelings:
That is a good point.  I think I'll put a note about that into the file instead of doing that work now.

(FWIW, I started out doing a more substantial rewrite, but it got out of hand and I had to toss it.  Then I started doing incremental changes and this one fixed it. I was going to rewrite more if it was necessary.)

Issue #598, opened 2024-11-15
[BUG] Variables, once clamped, are not logged


ms609:
**Describe the bug**
Clamping a variable causes it to be omitted from log files.

**To Reproduce**
```rb
seed(0)
chars <- readDiscreteCharacterData("simmat/k5obsno5.nex")
taxa <- chars.names()
outgroup <- clade(taxa[1])
nTaxa <- chars.size()
nEdge <- 2 * nTaxa - 3

moves = VectorMoves()
topology ~ dnUniformTopology(taxa, outgroup = outgroup)
moves.append( mvNNI(topology) )
moves.append( mvSPR(topology) )
tree_length ~ dnGamma(shape = 2, rate = 2 / 3)
moves.append( mvScale(tree_length) )

rel_br_lengths ~ dnDirichlet( rep(1.0, nEdge) )
moves.append( mvBetaSimplex(rel_br_lengths) )
moves.append( mvDirichletSimplex(rel_br_lengths) )
br_lengths := rel_br_lengths * tree_length

phylogeny := treeAssembly(topology, br_lengths)

u ~ dnGamma( shape = 1, rate = 1 )
rates := [ [0, u, 1, 1],
           [u, 0, u, 1],
           [1, u, 0, u],
           [1, 1, u, 0]]

Q4prime := fnFreeK(rates)

m_morph[1] ~ dnPhyloCTMC(
    tree = phylogeny,
    Q = Q4prime,
    type = "Standard"
)
m_morph[1].clamp(chars)
m_morph[1].lnProbability()

monitors = VectorMonitors()
monitors.append( mnScreen(printgen = 1, u, tree_length) ) # u is printed to screen
monitors.append( mnFile(filename = "mnFile.log", u, tree_length, printgen = 1) ) # u is printed
monitors.append( mnModel(filename = "mnModel.log", u, printgen = 1,
                         exclude = ["br_lengths", "rel_br_lengths"]) ) # u does not appear

u.clamp(1) # Remove this line and u is logged
mcmc(model(phylogeny), monitors, moves).run(generations = 3)


# Also occurs in mcmcmc:
#mymc3 = mcmcmc(model(phylogeny), monitors, moves, nruns = 2, nchains = 6,
#               moveschedule = "single", swapMethod = "both", 
#               swapMode = "multiple", combine = "mixed").run(generations = 3)
```

In the mcmcmc instance, we see a bonus error:
```
[...]

PR #599, opened 2024-11-15
Distinguish `.clamp()` and `.setValue()`

ms609:
Following on from #598 and a Slack discussion with @bredelings , I now have a better understanding of the distinction between `.clamp()` and `.setValue()`, which I've tried to encapsulate in documentation entries.  Perhaps the `clamp` example is a bit laboured, but I think it's useful to illustrate the distinction between the two methods.


davidcerny:
It might be worth mentioning that `.setValue()` is often used to initialize the MCMC with plausible values. [This tutorial](https://revbayes.github.io/tutorials/divrate/branch_specific.html) has a nice example:

```
speciation_mean ~ dnLoguniform( 1E-6, 1E2)
extinction_mean ~ dnLoguniform( 1E-6, 1E2)
extinction_mean.setValue( speciation_mean / 2.0 )
```

ms609:
I've rewritten the description to emphasize this, and added a link to the tutorial – it's a good idea to refer the user to a real-life example.

help/md/clamp.md
```diff
@@ -0,0 +1,44 @@
...
+# Set the observed value of x
+x.clamp(1)
+
+# Evaluate P(z) at x = X, y = 1
+y.setValue(1)
+z.probability()
+
+# Modify the observed value of x
+x.clamp(2) # equivalent to x.unclamp(); x.clamp(2)
+z.probability()
```
davidcerny, line 32: These lines don't work from me: on calling `z.probability()`, I get:

```
   Error:	Variable of type 'Real' has no method called 'probability'.  You can use '.methods()' to find available methods.
```

I suppose that makes sense – there is no mechanism for RevBayes to figure out what the PDF is for the product of two normally distributed random variables.

help/md/clamp.md
```diff
@@ -0,0 +1,44 @@
...
+
+# Evaluate P(z) at a different value of y
+y.redraw()
+print(y)
+z.probability()
+
+# Because x is clamped, it is invalid to call x.redraw() or mvSlide(y)
+# x will remain constant during mcmc, whereas y will be inferred.
```
davidcerny, line 40: Very nitpicky, but perhaps capitalize "MCMC" here

davidcerny (changes requested):

help/md/clamp.md
ms609: Oof... thanks for catching that.  I've updated (and tested!) the example.

davidcerny (approved):

PR #600, opened 2024-11-15
Prevent modification of clamped value

ms609:
Per #598, this PR throws an exception if the following attempt to modify the value of a clamped node:

- `redraw()`
- `[mc]mcmc()`, via an invalid `Proposal`

Example:

```rb
x ~ dnNorm(1, 1)
y ~ dnNorm(1, 1)
z <- x * y
x.clamp(1)
x.redraw()
# > Error:       Cannot modify the value of a clamped node.
mcmc(model(z), [mnScreen()], [mvScale(x)]).run()
# > Error:       Cannot add the clamped node 'x' to a proposal.
```


ms609:
I've updated #598 with a link to the data. Removing the exception in `Proposal` will cause _u_ to be moved during MCMC.  I thought that idea seemed a little too good to be true!

@bjoelle's `allowClamped` suggestion seems like a good way forwards.

ms609:
Tests now pass ... but it'd be worth someone familiar with the code checking for other proposals I may have missed that need the marking with `allowClamped`.

bjoelle (commented):
you've added the `initialize` function in some proposals but I don't see where this is called in `EventBirthDeathProposal` ?

ms609:
I think EventBirthDeathProposal is an abstract class, so a call to `initialize()` is only required when creating instances of child classes.

bjoelle:
ah ok makes sense, I didn't see that it was abstract

ms609:
Perhaps I should have been more precise: I think it's _functionally_ abstract as it seems to be the base class for ContinuousEBDP and DiscreteEBDP, and not to be called on its own.  I don't think it's formally defined as abstract; perhaps it could / should be?

Issue #360, opened 2023-06-19
[Feature Request] Better documentation for Unix/Linux command line arguments?


mdhall272:
The Unix/Linux command line help for `rb` is as follows:

```
$ rb-mpi --help
Usage: rb [OPTIONS]
       rb [OPTIONS] <file1> [<file2> ...]
Bayesian phylogenetic inference using probabilistic graphical models and an interpreted language

Options:
  -h [ --help ]         Show information on flags.
  -v [ --version ]      Show version and exit.
  -b [ --batch ]        Run in batch mode.
  -j [ --jupyter ]      Run in jupyter mode.
  --args arg            Command line arguments to initialize RevBayes
                        variables.
  --cmd arg             Script and command line arguments to initialize
                        RevBayes variables.
  --file arg            File(s) to source.
  --setOption arg       Set an option key=value.

See http://revbayes.github.io for more information.
```

As far as I can tell this is the limit of the documentation for these flags, anywhere. It took me quite some time to work out that `--args` simply fills a vector called `args` (which is not mentioned in https://revbayes.github.io/documentation/) and that if you do that then `rb [OPTIONS] <file1> [<file2> ...]` does not actually work (because everything after `--args` is used to fill `args`) and you need to specify `--file` instead. I do not need it right now, but I also have no idea how to approach using `--cmd`. It seems like these could use some elaboration.

Apologies if there is actually a tutorial for these somewhere, but I certainly could not find it.

bjoelle:
should be fixed by PR #583

davidcerny:
@bjoelle Hmm, I'm not sure that's the case, unfortunately. That PR elaborates on what one can pass to `--setOption`, but leaves the rest of the help page unchanged. This issue seems to be more about the differences between `--args`, `--cmd`, and `--file`, and how to combine them. I agree this should be better documented, ideally with examples, but to be honest, I'm not very clear on this, either.

Based on my experimentation, `--file` and `--cmd` simply source a script. The only difference I can see in their behavior is that when the interpreter encounters an error in the script (e.g., a nonexistent variable), `--cmd` closes RevBayes but `--file` does not. My impression was that `--cmd` was originally supposed to execute Rev code inline, e.g., `./rb --cmd "x ~ dnNorm(0, 1); print(x)"`, but this currently doesn't work. I haven't found a way to make `--args` do anything useful at all. So I'm re-opening the issue.

davidcerny:
[...]

PR #602, opened 2024-11-16
Document flags and `args`

davidcerny:
This PR attempts to add the extra documentation requested in #360. It does so as follows:

1. In `--help`, it replaces the "Usage" section by a more extensive "Usage examples" section that shows how one can use the `--args`, `--cmd`, and `--file` flags, and how they interact with each other;
2. Adds a help page for `args` illustrating its use and explaining its properties (e.g., the fact that it can hold values of different types). Since `args` is not a type, function, or distribution, adding the help page required some changes to `Parser` and `RbHelpRenderer`.


bredelings:
I like a lot of this, but I would like to keep traditional notation for the `Usage: ` string, and be a bit shorter.  The Usage message here is...  *ahem* ... _extremely_ thorough :-P  

Although you can indeed write `rb file --arg a --arg b --arg c` this is mainly because we really want to support `rb file --arg a b c` and we can't support the second without the first using Boost.ProgramOptions.  So lets just mention the second (recommend) method in the Usage info.

See #603 for something I like better.  If you like it, you could approve it and add on top of it.

davidcerny:
Damn, I guess we should have coordinated better :p

Personally, I don't find the current Usage section very helpful – e.g., am I supposed to actually use square brackets to group arguments together, or is that just part of the notation? I'm afraid the change proposed in #603 may further complicate things instead of clarifying them, and I believe examples work better.

Brevity is a virtue, obviously, but I don't think we should subordinate all other considerations to it. Especially given that the complaint made in #360 is basically "... that's it?". I.e., what we have now is clearly insufficient. My proposed `--help` page still comfortably fits on a single terminal screen, so length seems like a non-issue. That said, I do agree that demonstrating the equivalence of things like `rb --args 1 2` and `rb --args 1 --args 2` is unnecessary, so we can shave 2 lines off the page.

I do like several features in #603, like the helpful error messages thrown when flags are combined or positioned in invalid ways. I also think introducing separate help pages for individual flags is a great idea, and the overload of `renderHelp()` introduced by this PR should make that easy.

bredelings:
> Personally, I don't find the current Usage section very helpful – e.g., am I supposed to actually use square brackets to group arguments together, or is that just part of the notation?

Square brackets indicate that something is optional.  So `<file1> [<file2> ...]` means "at least one file argument, following optionally by more files".

bredelings:
> Brevity is a virtue, obviously, but I don't think we should subordinate all other considerations to it. 

True.  But brevity isn't quite the issue here.  The question is more what the information is for.
```
        "Usage examples:\n\n"
        "   1. rb\n"
        "   2. rb --args 1 2\n"
        "   3. rb script.Rev\n"
        "   4. rb --file script.Rev                                # equivalent to 3\n"
        "   5. rb --cmd script.Rev                                 # equivalent to 3 and 4\n"
        "   6. rb --file script.Rev --args 1 2\n"
        "   7. rb --args 1 2 --file script.Rev                     # equivalent to 6\n"
        "   8. rb --cmd script.Rev 1 2                             # equivalent to 6 and 7\n"
        "   9. rb --file script.Rev script2.Rev\n"
        "  10. rb --file script.Rev --args 1 2 --file script2.Rev\n"
        "  11. rb --file script.Rev script2.Rev --args 1 2         # equivalent to 10\n";
```
T

First, this leaves out `rb --cmd script1.Rev 1 2 --file script2.Rev` and `rb --args 1 2 --file script1.Rev`.  So it doesn't really cover everything.

bredelings:
Oops... pressed the wrong button!

bredelings:
We should also mention somewhere that `--cmd` is designed for use in executable scripts starting with e.g. `#!/usr/bin/rb --cmd`.

davidcerny:
Thank you for that explanation @bredelings! Now it seems obvious :) The nested structure of `[--args <arg1> [<arg2> ...]]` tripped me up a bit, but now I get it just means that specifying more than 1 argument is optional. I put the Usage section back; the current `--help` page looks like this:

Screenshot 2024-11-16 at 18 29 55
<br>
<br>
Some notes:

1. I incorporated the error message about combining `--cmd` and `--args` from #603, but not the one about filenames not being allowed before `--cmd`. For me, `rb script.Rev --cmd script2.Rev 1 2` works.
2. The point about the `--file` flag being useful for calls like `rb --cmd script.Rev 1 2 --file script2.Rev` is a good one, so I added it to the examples. Interestingly, though, `script2.Rev` gets executed before `script.Rev` in this case, which is in contrast to `rb script.Rev --args 1 2 --file script2.Rev`. I found that behavior surprising, so I added a note about it to the `--cmd` flag. However, this might be the level of detail that's better left for a future `?cmd` help page.
3. I agree we should mention that `--cmd` is primarily intended for things like `#!/usr/bin/env rb --cmd script.Rev`, and also that it automatically sets `batch_mode=true`. Again, I think `?cmd` would be a natural place for this. The reason why I haven't attempted to include these individual help pages here is that I'm not comfortable writing them for flags like `batch` and `jupyter`.

bjoelle:
so from what I can tell batch mode primarily means that the rb console will exit if it encounters an error in the script rather than hanging around - I think this would be useful to include
I would also add the info that `cmd` is intended for batch use (e.g. on a cluster) and `file` for interactive use, right now it's not clear at all (and I don't think it should be hidden behind another command)

davidcerny:
Sounds good! Would you maybe like to push a commit to that effect?

src/revlanguage/main.cpp
```diff
@@ -28,14 +28,38 @@ using po::variables_map;
 
 std::string usage()
 {
-    return "Usage: rb [OPTIONS]\n       rb [OPTIONS] <file1> [<file2> ...]\n";
-    // Other usages not mentioned
+    std::string usage =
```
bredelings, line 31: How about adding the other line I suggested?  From my (closed) PR:
```
rb [OPTIONS] --cmd <file> [<arg1> <arg2>...]
```

I was also thinking that maybe using lower-case for `option` like R, and dropping the angle brackets, might reduce visual clutter.  

Additionally, it might actually improve clarity to drop the square brackets around `[<arg2> ...]`.  If the user tries to use `--args` with no arguments, they will get a helpful error messages, so this is probably fine.

Lastly, we should probably indicate that you can use `--file` to add more files after arguments.

This leads to something like

```
Usage: rb [options]
       or: rb [options] file1 [file2 ...] [--args arg1 arg2 ...] [--file file3 ...]
       or: rb [options] --cmd file1 [arg1 arg2 ...] [--file file2 ...]
```
The naming of `file3` seems odd, but I'm not sure what else to write.

src/revlanguage/main.cpp
```diff
@@ -28,14 +28,38 @@ using po::variables_map;
...
+    return usage;
+}
+
+
+std::string usage_examples()
+{
+    std::string usage_examples =
+        "Usage examples:\n"
+        "   1. rb\n"
+        "   2. rb --args 1 2\n"
```
bredelings, line 44: Supplying arguments without a script seems odd.

In general, some of these examples seem like they are intended more to explain the parser than to be used as actual examples.

src/revlanguage/main.cpp
```diff
@@ -53,14 +77,14 @@ variables_map parse_cmd_line(int argc, char* argv[])
...
     // multitoken means that `--args a1 a2 a3` works the same as `--args a1 --args a2 --args a3`
-    ("args",value<std::vector<std::string> >()->multitoken(),"Command line arguments to initialize RevBayes variables.")
+    ("args",value<std::vector<std::string> >()->multitoken(),"Supply command-line arguments to RevBayes. These can be accessed from within the program using the 'args' vector. See ?args for details.")
+    // composing means that --file can occur multiple times
+    ("file",value<std::vector<std::string> >()->composing(),"Source one or more files.")
     // multitoken means that `--args a1 a2 a3` works the same as `--args a1 --args a2 --args a3`
-    ("cmd",value<std::vector<std::string> >()->multitoken(),"Script and command line arguments to initialize RevBayes variables.")
-	// composing means that --file can occur multiple times
-    ("file",value<std::vector<std::string> >()->composing(),"File(s) to source.")
+    ("cmd",value<std::vector<std::string> >()->multitoken(),"Source a file and supply command-line arguments in batch mode. Cannot be combined with the --args flag. If followed by the --file flag, the script sourced via --file is executed first.")
```
bredelings, line 87: The script with "command" should be executed first, I think.  So we should really fix this bug rather than describe it.

src/revlanguage/main.cpp
```diff
@@ -28,14 +28,38 @@ using po::variables_map;
...
+
+
+std::string usage_examples()
+{
+    std::string usage_examples =
+        "Usage examples:\n"
+        "   1. rb\n"
+        "   2. rb --args 1 2\n"
+        "   3. rb script.Rev\n"
+        "   4. rb --file script.Rev                         # equivalent to 3\n"
```
bredelings, line 46: For example, `4. rb --file script.Rev` seems unnecessary, since the recommended form is #3.  It does explain how the parser works, but it actually seems like a counter-example or anti-pattern.

src/revlanguage/main.cpp
```diff
@@ -28,14 +28,38 @@ using po::variables_map;
...
+        "   5. rb --cmd script.Rev                          # equivalent to 3 and 4 but runs in batch mode\n"
+        "   6. rb script.Rev --args 1 2\n"
+        "   7. rb --args 1 2 --file script.Rev              # equivalent to 6\n"
```
bredelings, line 49: For `7. rb --args 1 2 --file script.Rev`, users should actually write `rb script.Rev --args 1 2`.  So I think we should drop this as well.

src/revlanguage/main.cpp
```diff
@@ -28,14 +28,38 @@ using po::variables_map;
...
+        "   8. rb --cmd script.Rev 1 2                      # equivalent to 6 and 7 but runs in batch mode\n"
```
bredelings, line 50: I think the usage examples would be clearer if all the uses of `--cmd` were together, at the end.
- rb --cmd script.Rev
- rb --cmd script.Rev 1 2
- rb --cmd script.Rev 1 2 --file script2.Rev

bredelings (changes requested):
Thanks for the improvements!  I made some suggestions here.  I am probably busy until Thurs, but can look again then.

src/revlanguage/main.cpp
davidcerny: Changed as suggested

src/revlanguage/main.cpp
davidcerny: Removed

src/revlanguage/main.cpp
davidcerny: Done

src/revlanguage/main.cpp
davidcerny: Yeah, it's not likely to be a common use case – I can maybe see it being used in classroom / workshop settings precisely to demonstrate how the `--args` flag works. The `?args` help page relies on this usage for the same reason.

src/revlanguage/main.cpp
davidcerny: I'd like to keep this one, since it was one of the cases mentioned in #360:

> It took me quite some time to work out that `--args` simply fills a vector called `args` [...] and that if you do that then `rb [OPTIONS] <file1> [<file2> ...]` does not actually work (because everything after `--args` is used to fill `args`) and you need to specify `--file` instead.

ms609 (commented):
This looks fantastic and you have found a really elegant way to incorporate this into the web documentation.  And now I'm aware of this functionality, I'm looking forwards to dispensing with the rat's nest of R scripts I was using to accomplish the same ends!

As this PR greatly improves the discoverability and explanation of the behaviour I'm all in favour.  

With discoverability in mind, I also cast my eye over the tutorials page to reflect on where I might have expected to find pointers towards running batches of scripts.  There doesn't seem to be an existing tutorial where this would obviously fit.  I almost wonder whether there's a case for splitting the "Introduction to RevBayes" into two sections: one introducing the Rev 'philosophy' for users new to the software, i.e. the current getting started / syntax / graphical models trio; and a separate section; and a second focussing on the mechanics of running RevBayes, covering Setting Up RB, Using w/ Docker; and a brief tutorial largely corresponding to the new `?args` aimed at users who are not familiar with batch-running scripts?  This idea might be more within the scope of the new discussions board than the current PR, however...

src/revlanguage/main.cpp
bredelings: It seems like the issue here is that
- the Usage string said to put "OPTIONS" before files.
- however, if you write actually do that by writing `rb --args arg1 file`, then this actually does not do what was promised, because `--args` consumes all arguments after it until it hits another words starting with `-`.
- as a result, the user had to try a whole bunch of random things until figuring out something that worked.

The solution they found was to insert `--file` here, yielding something like your `7. rb --args 1 2 --file script.Rev`.  But of course an equally good solution would be to write `rb file --args arg1`, which contradicts the previous Usage string.  So it doesn't seem to me that _specific_ they found is required, so much that:
- it needs to be obvious how to specify a file and an argument
- it needs to be obvious why `rb --args arg1 file` is not the right way to do that.
- the Usage string needs to not imply that you should do things in a way that doesn't work.

That said, although the solution they found doesn't seem to me to be required by #360, it obviously isn't ruled out by it either.

davidcerny: I agree it's not required – it's just a valid use case that someone has previously found useful. For me, the fact the two solutions work equally well is a good enough reason to include them both.

I guess it comes down to whether we want Usage examples to... well, list valid usage examples, or _educate_ users on how to do things _properly_. I think the latter is a job for a tutorial rather than the help page.

src/revlanguage/main.cpp
bredelings: I pushed a fix for that.

src/revlanguage/main.cpp
bredelings: There are certainly a lot of valid usage examples that are not listed.  We don't list `-b` or `-j` or `--setOption`, and we don't list things like `rb file2 --file file3 --cmd file1 arg --file file4 --file 5`.  I'd be in favor of adding examples for `-b`, `-j`, and `--setOption`, but less in favor of listing `rb file2 --file file3 --cmd file1 arg --file file4 --file 5`.

However, we do have different preferences in that I prefer the usage examples to be more information-dense.

@bjoelle Do you have an opinion here?

bjoelle: in general I think it's fine to propose to users one "canonical" way to do things and not list all potential alternatives
on the other hand, I would be in favour of keeping 7. because putting the input file at the very end is a very common Linux/command-line way of doing things so it will likely be the syntax that will come naturally to a lot of people - the only exception I can think of is Rscript, but that one doesn't need a special `--args` flag either

bjoelle: actually if we're thinking about reorganizing the input options at some points Rscript may not be the worst model

bredelings: I do hope we can reorganize the input options, and I like what Rscript does.

Since the changes won't be backwards-compatible, we will have to check how many people rely on doing things like `rb script1 script2`.  With the Rscript approach, that would have to be changed to one of:
- put `source(script2)` at the end of `script1`, `rb  script1`
- `cat script1 script2 | rb`
- `rb -e 'source(script1)' -e 'source(script2)'`

bredelings: Of course, reorganizing the input options is a separate project.

wrightaprilm: `rb script1 script2`

I had never even heard of doing this.

davidcerny: Hence this PR :)

ms609 (commented):
I've just had a stab at using the args option by creating a Windows shortcut with the target
`<path to>rb.exe --args 5 4 3 --file "rb/MyScript.Rev"`

RevBayes launches the correct script, which starts `x = args[1]`, but I see the error "Variable args does not exist".

I'm not quite sure whether this is an issue with my usage (I think I've followed the new docs, but it's probably most likely I've missed something...), or whether the issue lies somewhere else?

davidcerny:
@ms609 Hmm... I'm worried this is yet another example of RevBayes behaving differently (and undesirably) on Windows compared to Unix, because that exact command -- `<path to>/rb --args 5 4 3 --file "rb/MyScript.Rev"` -- works just fine on macOS. I'm using this as `MyScript.Rev`:

```
x = args[1]
y = args[2]
z = args[3]

print( (x + y)^z )
```

And RevBayes is duly printing 729 for me as it is supposed to.

ms609:
Ah! I'd included a `clear()` command at the top of my script to aid debugging.  This cleared the `args` parameter.  That's interesting; I'd expected the `args` variable to be read-only, with a fixed value throughout a session.  I wonder whether it's worth spelling out that it's just an ordinary variable on the args manual page.

davidcerny:
Hmm, but the man page already points out that `args` is not a reserved keyword, and shows that you can clear it from your workspace? Feel free to commit a change if you find it unclear / not explicit enough.

In general, this is why it's better to include `clear()` at the end of a script rather than at the beginning...

ms609:
Doh, I missed that, thanks.  🤦🏾 

Thanks for the `clear()` tip.  Putting it at the end makes sense in general, but on this occasion I need to work further with the objects created with the script.

bredelings:
So part of the concern is that I thought the Examples section wasn't sufficiently information-dense.  I also don't think that explaining all the different ways you can specify file arguments is very helpful -- it primarily points to how bad the design of the command-line interface is.  The other part of my concern was that it didn't seem like explaining my concerns was helping reach a consensus.

It seemed like Joelle agreed that we didn't need 12 examples, but she did want to keep the specific example I mentioned.  (Maybe example #7?)  So, okay, keep example #7, but remove some of the others.  There are a few ways to do this.

davidcerny:
> The other part of my concern was that it didn't seem like explaining my concerns was helping reach a consensus.

Sounds like a fair assessment. For me, information density cannot be an overriding priority when we have users telling us the current super-dense help page is not detailed enough. It's not the end of the world if there are 12 examples. Eliminating genuine redundancies (i.e., different ways of doing the same thing) would only take us down to 10.

But yes, no consensus is being reached, so I'll probably close this once the new and hopefully better documented interface replaces the current one.

bredelings:
@davidcerny Are you OK with closing this now in favor of #803?

davidcerny:
@bredelings Yes, I'm closing this. I still like the idea of having a help page for the `args` vector, and of making the necessary changes to the help infrastructure to create one. But that can be a new PR, opened after #803 is merged.

PR #603, opened 2024-11-16
Alternative improved usage message

bredelings:
This PR:
- hides `--help` info on `--file`, which is only used to handle position arguments
- improves the usage message to cover --args and --cmd
- actually gives an error message when --args and --cmd are both used, instead of crashing.
- also gives an error messages when --cmd and position arguments (i.e. "file") are both used.

``` sh
$ ./rb -h
Usage: rb [OPTIONS]
       rb [OPTIONS] <file1> [<file2> ...] [--args <arg1> [<arg2> ...]]
       rb [OPTIONS] --cmd <file> [<arg1> <arg2>...]

Bayesian phylogenetic inference using probabilistic graphical models and an interpreted language

Options:
  -h [ --help ]         Show information on flags.
  -v [ --version ]      Show version and exit.
  -b [ --batch ]        Run in batch mode.
  -j [ --jupyter ]      Run in jupyter mode.
  --setOption arg       Set an option key=value. See ?setOption for the list of
                        available keys and their associated values.
  --args arg            Supply command line arguments.
  --cmd arg             Source a file and supply command line arguments.

```

``` sh
$ ./rb --args A --cmd B
Usage: rb [OPTIONS]
       rb [OPTIONS] <file1> [<file2> ...] [--args <arg1> [<arg2> ...]]
       rb [OPTIONS] --cmd <file> [<arg1> <arg2>...]

Error: received both --args and --cmd, but only one is allowed.
```

``` sh
$ ./rb file1 --cmd file2 arg1
Usage: rb [OPTIONS]
       rb [OPTIONS] <file1> [<file2> ...] [--args <arg1> [<arg2> ...]]
       rb [OPTIONS] --cmd <file> [<arg1> <arg2>...]

Error: filenames not allowed before --cmd.
```

### DIFF
--- a/src/revlanguage/main.cpp
+++ b/src/revlanguage/main.cpp
@@ -28,8 +28,12 @@ using po::variables_map;
 
 std::string usage()
 {
-    return "Usage: rb [OPTIONS]\n       rb [OPTIONS] <file1> [<file2> ...]\n";
-    // Other usages not mentioned
+    std::ostringstream o;
+    o<<"Usage: rb [OPTIONS]\n";
+    o<<"       rb [OPTIONS] <file1> [<file2> ...] [--args <arg1> [<arg2> ...]]\n";
+    o<<"       rb [OPTIONS] --cmd <file> [<arg1> <arg2>...]\n";
+
+    return o.str();
 }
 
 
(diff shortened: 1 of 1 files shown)


bredelings:
Actually, perhaps there was a use case for `--file` in doing something like `rb --cmd script arg1 arg2 arg3 --file script2`.  So I'm going to close this.

Issue #523, opened 2024-07-24
[Feature Request] Implement a `captureOutput` function


davidcerny:
**Context:**

I wanted to write an integration test to go with PR #522, but couldn't find a way to print the operator summary to a file. This is because `write()`/`print()` expects a Rev object, and `.operatorSummary()` only returns `NULL`.

Normally, this isn't much of a problem, since we can just redirect stdout to a file:

```
rb_command="source(\"test_MC3/scripts/Test_MC3.Rev\");"
echo $rb_command | ../projects/cmake/build/rb > Test_MC3.txt
```

Job schedulers like Slurm even do this automatically. However, this can't be done from within a Rev script, so it's not an option for validation tests.

**Proposed resolution**

Implement a function similar to R's `capture.output()` that can redirect `std::cout` calls to a stringstream.

bredelings:
I'd prefer to just modify the testing script to capture the standard output.  We would need to chop off the first few lines that display the version number, or perhaps add an option to suppress them.

It also looks like R's `capture.output( )` takes an unevaluated expression as the first argument, and then executes the expression in a modified environment.  I'm not sure we can implement that as a normal RevBayes function -- the revbayes function would just receive the return value of the expression.  In order to modify where the output goes, I think we'd need to modify the interpreter.

davidcerny:
Between #604 and #612, the issue that motivated this feature request has now been solved.

PR #604, opened 2024-11-20
Make stdout and stderr accessible to the test runner

davidcerny:
This PR modifies the test runner and adds to the expected output of existing tests to make it possible to check not only output files, but also the standard output (e.g., operator summaries printed to the screen; see #522, #523) and the standard error (e.g., tests that are supposed to fail; see #587).

It accomplishes this by redirecting both stdout and stderr to a file and further processing it. Specifically, it deletes everything up to and including the first occurrence of the string `   Processing file`. This is to get rid of the version and commit info, but also to account for the differences between Unix and Windows GitHub Actions: what in the former looks simply like

```
> source("scripts/Test_topologyProb.Rev")
   Processing file "scripts/Test_topologyProb.Rev"
```

looks like this in the latter:

```
> source("Z:\\home\\runner\\work\\revbayes\\revbayes\\tests\\test_topologyProb\\scripts\\Test_topologyProb.Rev")
   Processing file "Z:\home\runner\work\revbayes\revbayes\tests\test_topologyProb\scripts\Test_topologyProb.Rev"
```

Some notes:

1. Many of our test scripts contained the `mnScreen` monitor. This previously had no effect, but now it has become undesirable, since the `elapsed` and `ETA` columns are not reproducible. I have deleted it from all scripts.
2. Previously, a test failure could be due to a segfault, error (= any other nonzero exit code), or output mismatch. Now, it can only be due to a segfault or output mismatch. Since we now check the contents of stderr, all former errors should cause output mismatches. The two solutions should be strictly equivalent.
3. Random fun stuff: apparently, `nan` log likelihoods are unsigned on macOS and (?emulated) Windows but signed on Linux. I had to special-case the GitHub Action workflow to account for this.


bjoelle (approved):
looks good to me, nice work David !

ms609 (commented):
This looks great, and much more powerful and resilient than my previous hacky attempt.  

I've tried pulling this through to #587 but notice that there's some inconsistent behavior in the FBD test between clang-13 (fails) and mac-xcode-15 (passes).  Not sure why that's not the case in this PR.

tests/test_skygrid/output_expected/mcmc_skygrid.errout
```diff
@@ -0,0 +1,17 @@
+Problem when computing waiting time.
+Problem when computing waiting time.
```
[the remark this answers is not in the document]
davidcerny: Well, since the test passes, it would seem that it is indeed consistent across platforms and compilers, as ugly as it looks. :)

I've never used these popgen functions myself, but my guess is that this is no more worrying than the `Could not compute lnProb for <...>` warnings when RevBayes is trying to find a viable initial tree.

tests/test_pruning/output_expected/prune_trees.errout
```diff
@@ -0,0 +1,6 @@
+   Processing file "data/unpruned.trees"
+
+Progress:
+0---------------25---------------50---------------75--------------100
```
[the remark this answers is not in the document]
davidcerny: It doesn't seem to depend on terminal width settings. Merely resizing a terminal window certainly makes no difference, and even running RevBayes with `--setOption linewidth=320` (double the default value) still produces the same output.

davidcerny:
> I've tried pulling this through to https://github.com/revbayes/revbayes/pull/587 but notice that there's some inconsistent behavior in the FBD test between clang-13 (fails) and mac-xcode-15 (passes). Not sure why that's not the case in this PR.

It looks like you fixed the misspelling in the name of the `FBDP_fossil_tip_age_uncetrainty.Rev` script on your branch, whereas this branch still used the old misspelled name. Because the name of the `.errout` file is determined by the name of the corresponding `.Rev` file, the name of your output (`FBDP_fossil_tip_age_uncertainty.errout`) did not match the misspelled name of the expected output (`FBDP_fossil_tip_age_uncetrainty.errout`). The macOS GitHub Action wasn't affected by this, because it's special-cased to ignore the FBD tests altogether.

I've now fixed the spelling on this branch too, so this should no longer be a problem. If there are no objections, I'm going to merge this PR soon.

ms609:
Ah, well spotted!  Then this looks good to merge to me 👍🏾

.github/workflows/build.yml
```diff
@@ -252,7 +252,14 @@ jobs:
         # On Mac/ARM, the JSON output disagrees in the 16th and 17th digit
         #  with the JSON output on Linux/Intel and Mac/Intel.
         if [ ${{ matrix.arch }} = "mac-arm64" ] ; then
-          find tests -name '*.json' | xargs rm
+           find tests -name '*.json' | xargs rm
+        fi
+
+        # Another ugly hack: remove the *.errout file generated for the CBDSP
```
bredelings, line 258: It might be better if there is at least one architecture that we do NOT remove files for.
Right now this would be Linux.
So it would be slightly better if we remove the Mac and Windows ones, IMHO.
No strong feelings though.

.github/workflows/build.yml
```diff
@@ -252,7 +252,14 @@ jobs:
...
+        # test on Linux. Apparently, NaN log likelihoods are unsigned on macOS
```
bredelings, line 259: As far as I know, all platforms allow both positive and negative NaNs.  Are the unsigned NaNs actually positive NaNs?

Is there some known feature of Mac/Windows and not Linux that results in a different sign here?  Or do we just observe different signs in practice?

.github/workflows/build.yml
davidcerny: Hmm, good point, I didn't think about consistency. I disabled the Linux check because in this case, it was Linux that behaved as an outlier. No strong feelings on this either, though.

tests/run_integration_tests.sh
```diff
@@ -100,17 +100,16 @@ for t in test_*; do
     res=0
     # run the test scripts
     for f in scripts/*.[Rr]ev ; do
-        ${rb_exec} -b $f # print output so we can see any error messages
+        mkdir output
+        tmp0=${f#scripts/}
+        tmp1=${tmp0%.[Rr]ev}
+        ${rb_exec} -b $f &> output/${tmp1}.errout # print output so we can see any error messages
```
bredelings, line 106: What is the reasoning for combining the stderr and the stdout into a single file?
It seems better to separate these files.
That would allow checking one but not the other.

tests/run_integration_tests.sh
```diff
@@ -150,6 +149,29 @@ while [  $i -lt ${#tests[@]} ]; do
...
+            # Use a temporary intermediate file to make this work w/ both GNU and BSD sed
+            sed '1,/   Processing file/d' output/${tmp1}.errout > output/${tmp1}.errout.tmp
+            mv output/${tmp1}.errout.tmp output/${tmp1}.errout
+            
+            # Also delete the final line of failing tests, which reprints the path to the script
+            # that differs between Windows and Unix (has no effect if the line is absent)
+            sed '/   Error:\tProblem processing/d' output/${tmp1}.errout > output/${tmp1}.errout.tmp
+            mv output/${tmp1}.errout.tmp output/${tmp1}.errout
+            
+            # The dos2unix utility does not always catch OS-specific differences in path separators
```
bredelings, line 168: I think dos2unix only addresses line feeds by changing char(10) char(13) to char(10).

.github/workflows/build.yml
davidcerny: We just observe different signs in practice. This line from the standard output of the CBDSP test when run on macOS oe (emulated) Windows

```
Could not compute lnProb for node 'timetree': lnProb = nan
```

reads

```
Could not compute lnProb for node 'timetree': lnProb = -nan
```

on Linux.

tests/run_integration_tests.sh
davidcerny: Is that sort of flexibility really needed?

.github/workflows/build.yml
bredelings: Hmm... in that case, case we remove the Mac and Windows files instead?

davidcerny: Done.

tests/run_integration_tests.sh
bredelings: If you want to check that the error is written to the right stream (i.e. stderr and not stdout), then it is needed.  If you look at other test suites like this, they typically do not combine the stderr and the stdout.

But I guess writing stderr and stdout separately loses the interleaving, which might also be important.  Ideally we could write `name.stderr`, `name.stdout`, and `name.errout`.  But this turns out to be a bit tricky.  So I think this is fine for now.

tests/run_integration_tests.sh
bredelings: Can you fix the comment?  Right now it sounds like dos2unix sometimes fixes pathnames, which it doesn't.

bredelings (approved):

PR #605, opened 2024-11-21
Report values in error message

ms609:
This PR makes it simpler to debug a mismatch between the number of branches and the number of branch lengths by including each value in the text of the error message.

### DIFF
--- a/src/core/functions/phylogenetics/TreeAssemblyFunction.cpp
+++ b/src/core/functions/phylogenetics/TreeAssemblyFunction.cpp
@@ -37,7 +37,7 @@ TreeAssemblyFunction::TreeAssemblyFunction(const TypedDagNode<Tree> *t, const Ty
 
     if (tau->getValue().getNumberOfNodes() - 1 != brlen->getValue().size())
     {
-        throw(RbException(std::string("Number of branches (") + std::to_string(tau->getValue().getNumberOfNodes() - 1) + ") does not match the number of branch lengths (" + std::to_string(brlen->getValue().size()) + ")"));
+        throw RbException() << "Number of branches (" << (tau->getValue().getNumberOfNodes() - 1) << ") does not match the number of branch lengths (" << brlen->getValue().size() << ")";
     }
 
     // add the lambda parameter as a parent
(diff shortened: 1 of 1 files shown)


src/core/functions/phylogenetics/TreeAssemblyFunction.cpp
```diff
@@ -37,7 +37,7 @@ TreeAssemblyFunction::TreeAssemblyFunction(const TypedDagNode<Tree> *t, const Ty
 
     if (tau->getValue().getNumberOfNodes() - 1 != brlen->getValue().size())
     {
-        throw(RbException("Number of branches does not match the number of branch lengths"));
+        throw(RbException(std::string("Number of branches (") + std::to_string(tau->getValue().getNumberOfNodes() - 1) + ") does not match the number of branch lengths (" + std::to_string(brlen->getValue().size()) + ")"));
```
bredelings, line 40: Adding the values here is good.  However, RbException now has the ability to act like a stream, so you should use this instead.  So something like 

````
    throw RbException()<<"Number of branches "<<tau->getValue().getNumberOfNodes - 1 <<" does not match ..."
```

bredelings (approved):

src/core/functions/phylogenetics/TreeAssemblyFunction.cpp
ms609: Done.  I copied the syntax from an existing file; besides consistency, would there be an advantage to updating all other occurrences to use the stream syntax?

bredelings: Consistency and readability would be the main reasons.  If the command looks like `throw RbException(message)`, then that is fine.  I think we would mainly want to convert cases (eventually) if they use something like `std::to_string( )` instead of a stream.

bredelings: Or maybe if they do something like `throw RbException("..." + string + "....")`.

davidcerny (approved):
Looks useful – thanks!

PR #606, opened 2024-11-21
Document simplex moves

ms609:
In the spirit of "something is better than nothing", this PR proposes some skeletal documentation to moves pertaining to simplexes.

A documentation page exists for `mvDPPValueBetaSimplex` but I can't find the corresponding source code; this is listed at https://revbayes.github.io/tutorials/mcmc/moves.html but never explained.

https://revbayes.github.io/tutorials/mcmc/moves.html also contains moves (e.g. `mvSimplex`) that don't obviously exist.

### DIFF
--- a/help/md/Simplex.md
+++ b/help/md/Simplex.md
@@ -1,9 +1,23 @@
 ## name
 Simplex
 ## title
+Simplex
 ## description
+A simplex is a vector of elements that sum to 1.
 ## details
 ## authors
 ## see_also
+
+Moves that operate on Simplexes:
+- mvBetaSimplex
+- mvDirichletSimplex
+- mvDPPValueBetaSimplex
+- mvElementSwapSimplex
+
 ## example
+```rb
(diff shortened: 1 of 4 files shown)


davidcerny:
Hmm, that's strange. Since that move indeed appears to not exist, I'd be in favor of removing it from both the tutorial and `RbHelpDatabase`.

davidcerny:
@ms609 Would you like to merge this as-is, or do you think it requires further work? I'd be happy to approve once the nonexistent move is removed.

ms609:
I've deleted `mvDPPValueBetaSimplex.md`.  Further work is welcome, but I'm afraid I can't offer to provide it – so I would suggest merging as is for the time being.  @bredelings, if your ingenious idea of documenting functions as an introduction to RevBayes and GitHub comes with a list of functions whose documentation needs improving as part of, this could be added.

davidcerny (approved):

PR #613, opened 2024-11-21
Don't reject Metropolis-Hastings proposals because one PDF changes from NaN or Zero to something positive and finite.

bredelings:
Previously we always rejected any move where a stochastic node probability changed from zero or NaN to something finite.  But such moves should be fine.  This might possibly allow MCMC to solve some constraints by wandering around until PDFs that are 0 change to something non-zero.

### DIFF
--- a/src/core/dag/StochasticNode.h
+++ b/src/core/dag/StochasticNode.h
@@ -486,7 +486,15 @@ double RevBayesCore::StochasticNode<valueType>::getLnProbabilityRatio( void )
 template<class valueType>
 double RevBayesCore::StochasticNode<valueType>::getPrevLnProbability( void ) const
 {
-    // 1. If the node is not affected/touched, then the probability is the same for the current and previous state.
+    /*
+     * NOTE: If there is no previous probability then we could do a few things:
+     *         (1) throw an exception (current done).
+     *         (2) return an optional<double> to indicate if there is a previous probability or not.
+     *         (3) return the current probability.  This would make the method non-const.
+     *       Right now we do (1).
+     */
+
+    // 1. If the node is not affected/touched, then throw an exception.
     if (not stored_ln_prob)
         throw RbException()<<"getPrevLnProbability: no previous probability!";
 
--- a/src/core/dag/StochasticNode.h
+++ b/src/core/dag/StochasticNode.h
@@ -33,6 +33,7 @@ namespace RevBayesCore {
         virtual const TypedDistribution<valueType>&         getDistribution(void) const;
         void                                                getIntegratedParents(RbOrderedSet<DagNode *>& ip) const;
         virtual double                                      getLnProbability(void);
+        virtual double                                      getPrevLnProbability(void) const;
         virtual double                                      getLnProbabilityRatio(void);
         virtual std::vector<double>                         getMixtureLikelihoods(bool log=true) const;
         virtual std::vector<double>                         getMixtureProbabilities(void) const;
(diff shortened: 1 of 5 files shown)


src/core/moves/MetropolisHastingsMove.cpp
```diff
@@ -370,36 +370,50 @@ void MetropolisHastingsMove::performMcmcMove( double prHeat, double lHeat, doubl
...
+                zero_or_nan_to_finite = true;
+            }
             else
-                ln_prior_ratio += ratio;
+                ratio = current - prev;
         }
+        catch (const RbException &e)
+        {
+            if ( e.getExceptionType() != RbException::MATH_ERROR )
+                throw;
```
bjoelle, line 406: shouldn't that be throw e ? so we keep the original error

bjoelle (commented):
I don't see anything wrong with this change but I'm not sure I understand the rationale behind it - if the previous likelihood was `-Inf` or `Nan` we should not have accepted that move in the first place, right ?

src/core/moves/MetropolisHastingsMove.cpp
```diff
@@ -370,36 +370,38 @@ void MetropolisHastingsMove::performMcmcMove( double prHeat, double lHeat, doubl
     for (auto node: touched_nodes)
         node->touch();
 
+    bool fail_probability = not std::isfinite(ln_hastings_ratio);
+
```
davidcerny, line 374: It's not clear to me why `Rb::isAComputableNumber` has been replaced by `std::isfinite` for the Hastings ratio but not the likelihood or prior ratios

src/core/dag/StochasticNode.h
```diff
@@ -482,6 +483,23 @@ double RevBayesCore::StochasticNode<valueType>::getLnProbabilityRatio( void )
 }
 
 
+template<class valueType>
+double RevBayesCore::StochasticNode<valueType>::getPrevLnProbability( void ) const
+{
+    // 1. If the node is not affected/touched, then the probability is the same for the current and previous state.
+    if (not stored_ln_prob)
+        throw RbException()<<"getPrevLnProbability: no previous probability!";
```
davidcerny, line 491: The comment here seems to contradict the error message. Surely having no previous probability is different from the previous probability being equal to the current probability?

src/core/moves/MetropolisHastingsMove.cpp
davidcerny: Never mind, this was done by a subsequent commit

src/core/moves/MetropolisHastingsMove.cpp
bredelings: It actually does.  This is special syntax to re-throw the exception we just caught.  It is generally safer because it re-uses the existing exception.

src/core/dag/StochasticNode.h
bredelings: Yeah, I forgot to modify the comment after copy/pasting the code.  Fixed.

src/core/moves/MetropolisHastingsMove.cpp
```diff
@@ -389,7 +391,10 @@ void MetropolisHastingsMove::performMcmcMove( double prHeat, double lHeat, doubl
             // Compute the current lnProbability.
             double current = node->getLnProbability();
 
-            ratio = current - prev;
+            if (std::isfinite(current) and (std::isnan(prev) or prev == RbConstants::Double::neginf))
+                zero_or_nan_to_finite = true;
```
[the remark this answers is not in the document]
bredelings: No, I'm intentionally *not* allowing the case when current is `+Inf`.  We don't have a sensible story for what to do if any of the probabilities are `+Inf`.

src/core/moves/MetropolisHastingsMove.cpp
bjoelle: ok thanks :)

davidcerny (approved):

PR #614, opened 2024-11-22
Expand description, with reference

ms609:
I've read through Neal (2003) (thanks for the reference!) for a better understanding of what a Slice move is doing. I think it's worth linking to that reference here, and have updated the documentation to give a slightly more complete overview.

### DIFF
--- a/help/md/mvSlice.md
+++ b/help/md/mvSlice.md
@@ -1,14 +1,25 @@
 ## name
-Slice move
+mvSlice
 ## title
 Propose a slice move
 ## description
-Instead of using a fixed move size, `mvSlice` determines the size of a move proposal based on the current shape of the likelihood function.
-This allows small moves to be proposed in certain parts of parameter space,
-and large moves in other parts of the space, as appropriate.
+`mvSlice` proposes a new value for a variable based on the current shape of its likelihood function.
+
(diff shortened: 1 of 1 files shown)


bredelings:
Hmm.... I think the mvslice documentation should mention:
- sampling from the (x,y) coordinates of area under the PDF curve
- resampling y|x by setting y=uniform(0,1)*f(x)
- resampling x by finding a horizintal slice through the function at height y.
- how stepping out uses the window
- how doubling uses the window
Both stepping out and doubling randomly place an interval [L,R] of with W around the current point x.  Stepping out randomly extends the window by W on the left or right sides until the density f(x) is less than y.  Double random extends the interval by W*(2^i) for the i-th extension.

For the stepping out strategy, the move always accepts.  For the double strategy, the move usually accepts, but has a chance of rejecting.

ms609:
These sound like excellent things to add to the documentation – it'd be great if you could incorporate these details.

bredelings:
I'll put it on my list :-)

bjoelle:
@bredelings do you want to keep this PR open, or should we merge the existing changes and do another update later ?

hoehna (approved):
Seems like a good improvement

PR #616, opened 2024-11-22
Try unbreaking macOS ARM tests.

bredelings:
The macOS ARM tests were failing because we manually installed `pkg-config`, but `meson` now seems to depend on an alternative implementation called `pkgconf` that provides the same links.

Just stop manually installing `pkg-config` and let meson handle its own dependencies.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -136,7 +136,7 @@ jobs:
     - name: Install (macOS)
       if: runner.os == 'macOS'
       run: |
-          brew install pkg-config boost ccache coreutils meson openlibm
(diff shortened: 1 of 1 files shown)


ms609 (commented):
Seems to work – well diagnosed!  Odd that this seemed to be passing much of the time and only occasionally failing.

bredelings:
Weird!  Previously I guess I just tried re-running the tests without looking too closely, so I didn't realize that this was an old failure until you pointed it out.

Today a whole bunch of tests failed in a row, and rerunning the tests doesn't seem to help.

Do you want to approve?

davidcerny (approved):

PR #617, opened 2024-11-22
Allow adding, subtracting, and multiplying distributions by a scalar

bredelings:
The PR fixes a bug that prevented transforms with parameters from working with MCMC.  We need to change the transforms to have the form `f(parameters,x)` instead just `f(x)`.  For example, in the transformation `f(x) = lambda*x`, `lambda` is a parameter, so we need to represent it as `f(params,x) = params[0]*x`.  This allows us to call `swapParameterInternal` on the DAG nodes for the parameters.  Otherwise, when we duplicate the DAG to form the model DAG, we will point to the old DAG nodes for the parameters, and the parameters won't get updated.  I could split this out if that makes it easier to review.

This PR also allows adding, subtracting, and multiplying distributions by a scalar.  The goal here is to make it easier to represent distributions on the date of a fossil, so that you can write e.g:
```
t2 ~ t - dnExponential(lambda)
t2.clamp(observed_date)
```
instead of
```
e ~ dnExponential(lambda)
e.clamp(t - observed_date)
```
which requires solving for the value of the exponential variable that corresponds to the observed date.

Technically, you could also write:
```
t2 ~ dnExponential(lambda) |> tnScale(-1) |> tnShift(t)
t2.clamp(observed_data)
```
But that is also not very readable.

**reviewing**: Considering reviewing this PR by looking at the separate patches below.  The big bug-fix for transformed distributions is in `Fix how transformed distributions work`, and smaller fixes are in `Add overrides to make sure... ` and `Also update setValue( )`.


bjoelle (commented):
overall I think the different classes could use some doc - it took me a while to figure out what the difference was between sub1 and sub2 for instance

I am also wondering if we need several versions of the transforms for Real, RealPos, etc - I thought the point of the implicit conversions was that we could just define for Real and it would adjust

it also seems as if the logit can be coded as a combination of the others ? so it wouldn't need its own class, unless there is a difference I'm not seeing

bredelings:
@bjoelle Thanks for taking a look.

> overall I think the different classes could use some doc - it took me a while to figure out what the difference was between sub1 and sub2 for instance

Like some code-level documentation?  That would make sense.  There isn't really a user-level function to document...

> I am also wondering if we need several versions of the transforms for Real, RealPos, etc - I thought the point of the implicit conversions was that we could just define for Real and it would adjust

Using Real for everything *does* run, but it doesn't give tight enough bounds on the result.  For example, it doesn't encode the fact that RealPos + RealPos is a RealPos.  And if you don't know the result is a RealPos, then you can do certain things to it.

> it also seems as if the logit can be coded as a combination of the others ? so it wouldn't need its own class, unless there is a difference I'm not seeing

`logit(x) = log(x/(1-x))` -- you could decompose this into `log(x)` and `x/1-x`, but I don't see a way to do `x/1-x`.

bredelings:
There is indeed a lot of duplicated code here.  I think it is worth getting the return type right.  But its possible I can still decrease the amount of code duplication.  Let me see...

bjoelle (approved):
I think you have reversed the description for Transform_Add and Transform_Mul, otherwise looks good to me

PR #618, opened 2024-11-25
Change the `makeBifurcating()` method

davidcerny:
**Problem**

The `makeBifurcating()` member function of class `Tree` is fractally wrong:

1. The name isn't specific enough. All it can do is (1) suppress outdegree-1 nodes ("knuckles" / sampled ancestors) and (2) resolve a trifurcation at the root. These two operations are not guaranteed to make a tree bifurcating.
2. The description of the first argument is "Do we want to bifurcate only nodes with degree 1, or all nodes with degree different from 2?". The terminology here is wrong; it should refer to nodes with _outdegree_ 1. Tips are also of degree 1, but not of outdegree 1.
3. More importantly, this description makes a pretty explicit promise of being able to resolve multifurcations (= nodes with an outdegree that is neither 1 nor 2), which – again – the method can't do.
4. If you look at the implementation, the method actually tries to resolve a trifurcation at the root regardless of whether this argument is `true` or `false`, so the description is doubly wrong in that sense – there is no option you can pick if you really just want to suppress outdegree-1 nodes.
5. The first argument is called `fossils_only`. This is downright confusing; the name seems to be completely unrelated to the description. I guess something like `sampled_ancestors_only` was intended, but not all fossils are sampled ancestors, and not all sampled ancestors are fossils (think epidemiology).
6. Finally and most importantly, if you track the boolean set by `fossils_only` through `Tree` all the way to `TopologyNode`, it turns out that it does something completely different: it decides whether an outdegree-1 node should be replaced by a bifurcation in which one of the children is subtended by a zero-length branch (if `true`), or just removed altogether (if `false`).

**Proposed solution**

This PR:

1. Renames the `makeBifurcating` (`RlTree`, `TopologyNode`) and `makeInternalNodesBifurcating` (`Tree`) methods to `suppressOutdegreeOneNodes`. It's a mouthful, but hopefully less ambiguous.
2. Renames their `fossils_only` (`RlTree`) and `as_fossils` (`Tree`, `TopologyNode`) arguments to `replace`, and uses a plain-language description as well as source code comments with some ASCII art to explain what it does.
3. Removes the call to `Tree::makeRootBifurcating` from the `suppressOutdegreeOneNodes` method of `RlTree` – now that it's been renamed, it wouldn't make sense to keep it. If you have a trifurcation at the root and want to get rid of it, you can still do that by using the `reroot` method and setting `make_bifurcating=TRUE`.
4. Removes an unused argument from `Tree::makeRootBifurcating`, and makes it throw a more informative error if the root has more than 3 children.

**Notes**

This PR lays the groundwork for the introduction of a new `Tree` member function that randomly resolves multifurcations. See the [`dev_supertree`](https://github.com/revbayes/revbayes/tree/dev_supertree) branch for preliminary work on this.


src/core/datatypes/trees/Tree.cpp
```diff
@@ -2185,6 +2162,29 @@ void Tree::setTaxonObject(const std::string& current_name, const Taxon& new_taxo
 }
 
 
+void Tree::suppressOutdegreeOneNodes(bool reindex, bool replace)
+{
+    // If reindex is false, then suppressOutdegreeOneNodes will either
```
bjoelle, line 2167: I don't get this - what is the point of having a function argument that will crash the code ?

src/core/datatypes/trees/TopologyNode.cpp
```diff
@@ -2012,48 +1978,95 @@ void TopologyNode::setUseAges(bool tf, bool recursive)
...
-    sampled_ancestor_tip = tf;
-
-    // Only tips can have the sampled_ancestor_tip flag set.
-    assert(not sampled_ancestor_tip or isTip());
-}
-
 
-void TopologyNode::setSpeciesName(std::string const &n)
-{
+    if ( isTip() == false )
```
bjoelle, line 1990: I don't see the point of this check when the next thing we do is checking the number of children

src/core/datatypes/trees/TopologyNode.cpp
```diff
@@ -2012,48 +1978,95 @@ void TopologyNode::setUseAges(bool tf, bool recursive)
...
+            }
 
+        //}
 
-void TopologyNode::setTree(Tree *t)
-{
+        // call this function recursively for all its children
+        for (size_t i=0; i<getNumberOfChildren(); ++i)
+        {
+            getChild( i ).suppressOutdegreeOneNodes( replace );
```
bjoelle, line 2062: does that work if replace = false ? since we've just removed ourselves from the tree, can we still recurse ?

src/core/datatypes/trees/Tree.cpp
davidcerny: Great point. It looks like the only place where that argument can ever be set to `false` is in `Tree::reroot`, and it's not super clear to me why that method calls `suppressOutdegreeOneNodes` anyway. I'm keeping the call in there in case some behavior depends on it, but eliminating the `reindex` argument.

src/core/datatypes/trees/TopologyNode.cpp
davidcerny: Good catch! Indeed we cannot; we get a segfault if we try. I've fixed this now – the recursion only takes place when `replace = true`. When it's false, the removal of outdegree-1 nodes is instead handled by a combination of iteration and recursion at the level of `Tree::suppressOutdegreeOneNodes`. I've added an integration test to make sure that we can handle trees with multiple outdegree-1 nodes (including potentially at the root) under both `replace = true` and `replace = false`.

davidcerny:
Okay, this should be much more robust @bjoelle. I'm still not totally happy with the current methods – `removeDegree2Node`, `removeNodeIfDegree2`, and `removeRootIfDegree2` are named in an inconsistent way (my gripe is that the root of a fully resolved clock tree is also of degree 2, so the terminology is not as precise as it could be), and probably redundant to some degree. Even if they are not redundant, it might be worth refactoring them. But maybe that's a problem for another day.

bjoelle (approved):
looks good now

PR #627, opened 2024-12-05
Dev beagle

bredelings:
I think this is the actual BEAGLE branch.


davidcerny:
Was this not the right branch either? Or were you just checking how many merge conflicts there were?

bredelings:
This one actually was the right branch.  If I create the PR, then that makes it hard for me to review it :-P

PR #628, opened 2024-12-07
Allow ignoring specified data elements.

bredelings:
This PR adds the ability to ignore data at specific clamped nodes.  This works as a method on model objects that looks like
```
mymodel.ignoreData(phySeq, morph_data)
```
The PR also removes the `underPrior` flag from `mcmc.run(...)` and other Monte Carlo analyses.  To replace that, we add the method
```
mymodel.ignoreAllData()
```

Changes:
- If a vector is passed to `model.ignoreData( )`, then we try to ignore its elements.  This works for vectors of vectors.
- The dot graph for `model.graph(filename)` is tweaked so that ignored nodes are grey:
- Also rename `prior_only` to `ignore_data` and move it to StochasticNode

Errors/warnings:
- Errors are thrown if an ignored node is not clamped.
- A warning is given if an ignored node is not part of the model.
- There is currently NOT an error for ignoring a node with children, because vector elements `y[1]` have the vector `y` as a child.


bredelings:
Screenshot From 2024-12-07 11-35-16

davidcerny:
Removing the `underPrior` option doesn't seem like a good idea to me; it's an unnecessary breaking change in something that sees a lot of use. I've counted 13 occurrences in tutorials alone, not to mention integration tests and documentation.

bredelings:
Hmm... good point.  I only looked in `tests/`.  Technically we could modify the tutorials, of course.  We could also print out an error message for `underPrior` saying what to do instead.  I can split these into two changes so they can be discussed separately, if need be.

Some points in favor of removing `underPrior`
- people can incorrectly assume that it doesn't remove fossil data
- right now, `underPrior=FALSE` and `underPrior=TRUE` both modify the model object and override any changes that are made by `model.ignoreData( )`
- as @mikeryanmay said (I think), it would be better to make these modifications in the model object, not the mcmc object.  So the current `underPrior` is kind of in the wrong place.

@bjoelle @mikeryanmay @trayc7 ?

bredelings:
It looks like `underPrior=TRUE` is mentioned in:
- cont_traits/modules/simple_OU_exercise_1.md
- cont_traits/scripts/mcmc_BM_prior.Rev
- cont_traits/scripts/mcmc_OU_prior.Rev
- tutorials/dating/nodedate.md   -- Here it seems to be used incorrectly!
- divrate/modules/simple_Yule_parameter_estimation.md
- divrate/scripts/mcmc_Yule_prior.Rev
- mcmc_old/index.md
- mcmc_troubleshooting/Priors.md
- mcmc_troubleshooting/scripts/priors.Rev
- multispecies_coalescent/index.md
- quick_tips/index.md

mikeryanmay:
Overall, I like the model.ignoreAllData() and model.ignoreData() solution. It provides the required flexibility without adding a hacky argument to model(). Also, I like methods in general, so I like to see this :).

> right now, underPrior=FALSE and underPrior=TRUE both modify the model object and override any changes that are made by model.ignoreData( )

To me, this is a very compelling argument for getting rid of underPrior. My main concern is just backward compatibility: removing the argument entirely can invalidate existing scripts. I know it would cause problems for a couple of my own supplemental code archives---perhaps this reflects badly on me because I haven't provided exact commits the scripts are supposed to work with, but also many revbayes users will just be using a downloaded executable and won't want to/know how to use that commit information anyway. 

A proposed solution: retain the underPrior argument but effectively disable it. Throw an error/warning if the user specifies underPrior=TRUE that points them to model.ignoreAllData(). This way, at least people using older scripts will get some feedback about what they need to do to.

bredelings:
Retaining the `underPrior` argument and making it issue an error or warning would be relatively easy.

I'm trying to collect github repos that store scripts published in papers so that I can use them as extended test cases. Here's what I see:
```
analyses/biome_shift/code/run_inf.Rev:ch.run(n_gen, underPrior=under_prior)
analyses/zipping/code/run_quick_hym_387taxa.Rev:ch.run(n_gens, underPrior=under_prior)
analyses/hym_biogeo_dec/code/run_quick_hym_387taxa.Rev:ch.run(n_gens, underPrior=under_prior)
analyses/biogeo_silversword/scripts/run_model_g0.Rev:mymcmc.run(n_gen, underPrior=under_prior)
analyses/biogeo_silversword/scripts/run_model_g4.Rev:mymcmc.run(n_gen, underPrior=under_prior)
analyses/biogeo_silversword/scripts/run_model_g1.Rev:mymcmc.run(n_gen, underPrior=under_prior)
analyses/vib_div/code/run_quick.Rev:ch.run(n_gens, underPrior=under_prior)
283/run_quick_hym_387taxa.Rev:ch.run(n_gens, underPrior=under_prior)
```
In all cases `under_prior` is defined to be `false`.

In these cases, it looks like it would be beneficial to accept the argument and then issue a warning or error.

davidcerny:
> A proposed solution: retain the underPrior argument but effectively disable it. Throw an error/warning if the user specifies underPrior=TRUE that points them to model.ignoreAllData(). This way, at least people using older scripts will get some feedback about what they need to do to.

That sounds like a reasonable compromise.

bredelings:
I'll implement that on Thursday then.

bredelings:
OK, so now specifying `underPrior=FALSE` print the following warning and continues:

```
   Warning! The `underPrior` argument to `run` was removed in version 1.2.5 and now has no effect.
            You can accomplish the same effect as `underPrior=TRUE` with `model.ignoreAllData()`.
            See help on `model` for more information.
```

Specifying `underPrior=TRUE` prints the warning, and additional throws an exception:

```
   Error:	run: argument `underPrior=TRUE` has no effect.
```

mikeryanmay:
I like all of these changes. I just added a slightly more complex test for model.ignoreData() which also includes a vector of vector of ignored variables, just to ensure that it works for slightly more complicated nesting. I'll approve this pull request if we're ok with this test (and it passes).

bredelings:
Nice, yeah, I'm OK with that.  It looks good.

davidcerny:
Finally got the chance to test this on two datasets, with very good results:

- For the dataset from #515, leaving the `phySeq` variable unclamped and running the analysis with the `suppressCharacterData=TRUE` flag proposed in that PR produced the same likelihoods, but the former option was ~7 times slower. Running the analysis after modifying the model with `.ignoreData(phySeq)` also yields the same likelihoods and is just as fast as `suppressCharacterData=TRUE`, so that's great.
- For @wrightaprilm's dataset from #623, leaving all formerly clamped variables unclamped yielded different likelihoods as well as longer runtimes than setting `underPrior=TRUE`. This reflects the fact that for each unclamped node, the likelihood of the initial draw factors into the overall likelihood of the model instead of being simply set to 1. Running the analysis after modifying the model with `.ignoreAllData()` produces the same likelihoods as `underPrior=TRUE`, and is just as fast.

For consistency's sake alone, I'd like to add an `.errout` file to the expected output of the new integration test, since all other tests have them. I also think, though, that the `.errout` mechanism can genuinely come in handy here: it'd be nice to add a test that tries to ignore a constant variable or a non-clamped stochastic variable, and fails with a non-empty stderr.

PR #632, opened 2024-12-16
Add stdout/stderr test.

bredelings:
Add ignore.errout test output.

Adding a test to check that we get the right errors would indeed be useful.  Right now I have some commented-out lines in the test that could be added in to check errors.  However, it seems that RevBayes quits running the script at the first error which means that you can't test multiple errors per script.  I plan to address this after 1.2.5 as part of changing the options to match Rscript.

### DIFF
--- a/tests/test_basics/output_expected/ignore.errout
+++ b/tests/test_basics/output_expected/ignore.errout
@@ -1,0 +1,5 @@
+   
+   Running MCMC simulation
+   This simulation runs 1 independent replicate.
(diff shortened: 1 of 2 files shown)


davidcerny (approved):
A simple solution would be to just have multiple scripts, one for each failing case (in which case, it might be worth grouping them together in a separate directory rather than `test_basics`). But I don't insist on this; it can wait until the functionality you propose is added.

PR #641, opened 2025-01-15
Random resolution of multifurcations and MBL time-scaling

davidcerny:
This PR represents joint work with @laumul and adds two features to the code base:

(1) A new member function of classes `Tree` and `TopologyNode` for randomly resolving multifurcations. It can handle both `BranchLengthTree` and `TimeTree` objects, and multifurcations at the root as well as other internal nodes – new tests have been added to demonstrate this. The code is heavily based on `UniformTopologyDistribution::simulateClade()` for branch length trees and two methods of `UniformSerialSampledTimeTreeDistribution` for time trees (as acknowledged in source code comments). This sort of code duplication may not be the most elegant, but I wanted to avoid circularity – using distributions that are themselves defined over `Tree` objects for `Tree` member functions.

(2) A new function called `fnMinBLTimeScaling` that takes a tree without branch lengths*, or with branch lengths in units of expected change, and time-scales it based on the ages of its tips using the minimum branch-length approach of [Laurin (2004)](https://academic.oup.com/sysbio/article-abstract/53/4/594/1646435). It should be more or less equivalent to `timePaleoPhy(..., type="mbl")` from [`paleotree`](https://github.com/dwbapst/paleotree), though I haven't tested this extensively, and achieving strict equivalence was not the main objective, since the function is mostly just intended to generate plausible starting trees for complex FBD analyses. (In particular, FBD analyses with backbone constraints, which just [won't initialize](https://github.com/revbayes/revbayes/issues/384) unless a compatible starting tree is supplied).

Help page and integration tests are included.

\* This refers to their Newick/Nexus representations. When read into RevBayes, such trees are represented as phylograms with all branch lengths equal to 0.


src/core/datatypes/trees/TopologyNode.cpp
```diff
@@ -1903,6 +2063,48 @@ void TopologyNode::setParent(TopologyNode* p, bool recompute_branch_length)
 }
 
 
+void TopologyNode::setParentAge(double minbl)
```
bjoelle, line 2066: I don't find the name of this function very clear - suggestion `scaleAgesFromTaxonAgesMBL`

help/md/fnMinBLTimeScaling.md
```diff
@@ -0,0 +1,52 @@
...
+The age of each internal node is based on the age of the oldest tip descended
+from it. However, if t0 is the oldest tip descended from a given node (denoted
+x) and also the oldest tip descended from that node's parent (denoted y), then
+setting t(x) = t(y) = t(t0) would produce zero-length branches y->x and x->t0.
+We avoid this by requiring every branch to be no shorter than some user-supplied
+constant. This has the effect of shifting node ages deeper into the past.
+
+Conceptually, the undated tree would usually correspond either to a bare
+topology (a tree without branch lengths) or a tree with branch lengths in units
+of expected change; in practice, both `BrachLengthTree` and `TimeTree` arguments
```
bjoelle, line 17: typo ? BranchLengthTree

src/core/datatypes/trees/TopologyNode.h
```diff
@@ -158,13 +158,15 @@ namespace RevBayesCore {
...
         void                                        setBranchLength(double b, bool flag_dirty=true);                                    //!< Set the length of the branch leading to this node.
         void                                        setIndex(size_t idx);                                                               //!< Set the index of the node
 
         void                                        setName(const std::string& n);                                                      //!< Set the name of this node
         void                                        setNumberOfShiftEvents(size_t n);                                                   //!< Set the number of shift events for stochastic character maps
         void                                        setParent(TopologyNode* p, bool recompute_branch_length = true);                    //!< Sets the node's parent
+        void                                        setParentAge(double minbl);                                                         //!< Sets the age of the parent to the age of the oldest of its children + minbl
```
bjoelle, line 169: this seems misleading - the function actually sets the age of the node, not the parent ?

src/core/datatypes/trees/TopologyNode.cpp
```diff
@@ -1762,6 +1764,163 @@ void TopologyNode::renameNodeParameter(const std::string &old_name, const std::s
...
+                    prnt->setAge( current_time );
+                    active_children.push_back( prnt );
+                    
+                    // add the newly created parent to the list of the children of the current node
+                    addChild( prnt );
+                    prnt->setParent( this );
+                }
+                
+            }
+            else
```
bjoelle, line 1876: I believe that you can have `use_ages = F` even if there is branch length information, but the function doesn't seem to account for it

src/core/datatypes/trees/TopologyNode.cpp
davidcerny: Good catch; the name is a leftover from an earlier implementation attempt. Changed as suggested.

help/md/fnMinBLTimeScaling.md
davidcerny: Fixed.

src/core/datatypes/trees/TopologyNode.h
davidcerny: See above; fixed.

src/core/datatypes/trees/TopologyNode.cpp
davidcerny: That's true: when the method was applied to a `BranchLengthTree`, the newly inserted branches had no branch lengths, which caused trouble when processing the resulting trees using `ape` (the missing branch lengths were represented as `NaN`).

I've now changed the code to give them a length of 0. This ensures that the original branch lengths and the original root-to-tip distances are both unchanged. The behavior for a `BranchLengthTree` is now exactly in line with how `ape::multi2di` works.

I also noticed it was possible for the branch subtending the multifurcating node to have its original length removed. This has been fixed as well.

Issue #642, opened 2025-01-16
[BUG] Stopping rules seemingly not triggered when GelmanRubin included


ms609:
**Describe the bug**

I'm afraid I don't have time to fully interrogate this at the moment, but wanted to flag that there still seems to be an issue with the `srGelmanRubin` stopping rule: when such a rule is specified, analyses seem not to terminate, even if other rules are triggered.

**To Reproduce**

Here's the relevant part of a script:
```
mymc3 = mcmcmc(
  mymodel, monitors, moves,
  nruns = 2,
  nchains = 8,
  tuneHeat = TRUE,
  swapMethod = "both",
  swapMode = "multiple",
  combine = "none"
)
checkFreq = 50
mymc3.initializeFromCheckpoint(ckpFile)

stopping_rules[1] = srMaxTime(2, "hours")
stopping_rules[2] = srMinESS(1000, paramFile, checkFreq)
stopping_rules[3] = srGelmanRubin(1.01, paramFile, checkFreq)

mymc3.run(rules = stopping_rules, checkpointFile = ckpFile, checkpointInterval = checkFreq)
```

I'm running the script in serial.

A full repo is available at https://github.com/neo-trans/07200_by_n_kv/ – the relevant script there is `[mc3serial.Rev](https://github.com/neo-trans/07200_by_n_kv/blob/main/mc3serial.Rev)`.

**Expected behavior**

I'm running [mc3serial.Rev](https://github.com/neo-trans/07200_by_n_kv/blob/main/mc3serial.Rev) via slurm.  My R analysis reports a psrf of 1.001273 and ess of 1906.  The slurm jobs are continuing until their container times out (after 2h15).  I'm expecting them to (i) stop after `checkFreq` generations, if both the ESS and PSRF thresholds have been hit; and (i) stop after 2 hours, whether or not a run has not hit the other thresholds.

It be useful for RevBayes to print out its calculation of the ESS and PSRF statistics, to make it easier for a user to diagnose the behaviour (and to report this information in log files); I'm not aware of a way to do this from the RevBayes interface.

**Computer info**
RevBayes 1.2.5 release, on Linux HPC

davidcerny:
[...]

Issue #542, opened 2024-08-23
[BUG] ETA displays enormous number


ms609:
**Describe the bug**
When a number of generations is not specified in an MCMC[MC] run, the ETA displays something that presumably corresponds to TIME_MAX - RUN_TIME

**To Reproduce**
```rb
	# Create a simple model (unclamped)
	a ~ dnExponential(1)
	mymodel = model(a)
	
	# Create a move vector and a monitor vector
	moves[1] = mvScale(a, lambda=1.0, weight=1.0)
  
  monitors = VectorMonitors()

  monitors.append( mnModel(filename = "output/mnModel.log", printgen = 100) )
  monitors.append( mnFile(filename = "output/mnFile.log", printgen = 100) )
  monitors.append( mnModel(filename = "output/params.log", printgen = 10, stochasticOnly = TRUE) )
  monitors.append( mnScreen(printgen = 1000, prior = FALSE, posterior = FALSE) )

	# Create an mcmcmc object
	mymc3= mcmcmc( mymodel, monitors, moves, nruns = 2, nchains=4, deltaHeat=5)
	

  checkFreq = 100
  stopping_rules[1] = srMaxIteration(100000)
  stopping_rules[2] = srMaxTime(1, "hours")
  
  mymc3.run(rules = stopping_rules)
  ```

**Expected behavior**
- Option 1: Display `???` or `>24:00:00` to indicate completion time cannot be estimated
- Option 2: Use `maxIter` / `maxTime` stopping rules to provide bound on ETA

**Screenshots**
![image](https://github.com/user-attachments/assets/7e2bc80d-d74c-49a5-be38-5ff44f64ec42)

**Computer info**
WIn 10, Rb 1.2.5

PR #645, opened 2025-01-20
Print convergence statistics during MCMC runs with stopping rules

davidcerny:
[Both](https://github.com/revbayes/revbayes/issues/642) @ms609 and [myself](https://github.com/revbayes/revbayes/issues/543#issuecomment-2307404153) have argued that it would be good to have the MCMC stopping rules print their values to the screen every time they are checked. That's what MrBayes does with its ASDSF-based stopping rule, for example. This PR provides that functionality. The changes only have an effect when the analysis includes at least one stopping rule other than `srMaxIter`. For example, this Rev code:

```
stopping_rules[1] = srMaxTime(15, "hours")
stopping_rules[2] = srMinESS(50, myOutputFile, 10000)
stopping_rules[3] = srGelmanRubin(1.01, myOutputFile, 10000)
stopping_rules[4] = srMaxIteration(200000)
stopping_rules[5] = srGeweke(prob=0.001, file=myOutputFile, freq=10000)
stopping_rules[6] = srStationarity(prob=0.01, file=myOutputFile, freq=10000)

mymcmc = mcmc(mymodel, monitors, moves, moveschedule="random", nruns=2)
mymcmc.run(stopping_rules)
```

now produces the following output:

```
10000       |       -45.4796   |       -39.7758   |       -5.70378   |       60.74352   |      -2.367357   |   00:00:01   |   00:00:00   |
   
   Elapsed time: 00:00:10; remaining time: 14:59:50
   Minimum effective sample size (ESS): 1.557837
   Maximum value of the Gelman-Rubin statistic (PSRF): 1.004275
   The Geweke test statistic is < 0.0005 or > 0.9995 in 20/50 comparisons
   The CI of a single-chain mean excludes the overall mean in 0/50 comparisons
   
11000       |       -45.3797   |       -39.6759   |       -5.70378   |       47.20355   |      -2.271312   |   00:00:20   |   00:00:00   |
12000       |       -44.1768   |        -38.473   |       -5.70378   |       31.68508   |      -2.219538   |   00:00:20   |   00:00:00   |
13000       |       -48.8873   |       -43.1835   |       -5.70378   |       41.72365   |      -2.362557   |   00:00:20   |   00:00:00   |
14000       |       -44.0703   |       -38.3665   |       -5.70378   |       36.58749   |      -2.211202   |   00:00:20   |   00:00:00   |
15000       |       -45.8181   |       -40.1144   |       -5.70378   |       44.50335   |      -2.243187   |   00:00:20   |   00:00:00   |
16000       |       -45.3434   |       -39.6396   |       -5.70378   |        22.1649   |      -2.144946   |   00:00:20   |   00:00:00   |
17000       |        -43.887   |       -38.1832   |       -5.70378   |       27.74028   |      -2.129627   |   00:00:20   |   00:00:00   |
18000       |       -43.7073   |       -38.0036   |       -5.70378   |       21.33286   |      -2.086433   |   00:00:20   |   00:00:00   |
19000       |       -46.5554   |       -40.8516   |       -5.70378   |       47.71708   |      -2.376249   |   00:00:20   |   00:00:00   |


Iter        |      Posterior   |     Likelihood   |          Prior   |              A   |         lalpha   |    elapsed   |        ETA   |
------------------------------------------------------------------------------------------------------------------------------------------
20000       |       -43.9887   |       -38.2849   |       -5.70378   |        41.5126   |      -2.271926   |   00:00:20   |   00:00:00   |
   
   Elapsed time: 00:00:32; remaining time: 14:59:28
   Minimum effective sample size (ESS): 1.557837
   Maximum value of the Gelman-Rubin statistic (PSRF): 1.004275
   The Geweke test statistic is < 0.0005 or > 0.9995 in 20/50 comparisons
   The CI of a single-chain mean excludes the overall mean in 0/50 comparisons
   
21000       |       -45.2438   |       -39.5401   |       -5.70378   |       53.01919   |      -2.318033   |   00:00:42   |   00:00:00   |
```

The messages get written to the standard output. I originally wanted to make them part of the `mnScreen` monitor, and possibly other monitors as well. That would have the added advantage of allowing us to exploit the convergence statistics to calculate the ETA, and consequently eliminating the [absurdly large values](https://github.com/revbayes/revbayes/issues/542) we currently get when the maximum number of iterations is left unspecified. However, that proved to be difficult, because monitors operate on `model` objects rather than `mcmc` objects. Perhaps someone else will find a solution to this – my hope is that the new member function added to the `StoppingRule` class, which exposes the value of the convergence statistic (instead of just checking whether or not it meets the threshold), will make it easier.

Over the course of working on this, I noticed other problems related to stopping rules, so the PR also does the following:
1. Provide more detailed and accurate source code comments;
2. Re-enable several previously disabled member functions of class `TraceNumeric` for calculating convergence statistics – these were commented out with a note calling them "broken", probably in reference to #543;
3. Remove `src/core/analysis/mcmc/convergence/HeidelbergerWelchTest.cpp` and its corresponding header file. These were clearly created by copying the Geweke test code with the intention of eventually implementing the Heidelberger-Welch test in its place, but no further work was done on them, resulting in a misnamed duplicate of an existing test.


ms609:
This looks fab, thank you for your work on this!

I would find it useful to see also the threshold value for a `stopping_rule`, e.g.

     Maximum value of the Gelman-Rubin statistic (PSRF): 1.004275 < 1.01

This would be particularly useful if the value may differ from that specified in `stopping_rules` (for example if overwritten by a checkpoint file), but could also catch cases where a user's script includes bugs and is not setting the value they anticipated.

davidcerny:
That's a great idea. The current code already does this for `srMaxTime`:

```
Elapsed time: 00:00:32; remaining time: 14:59:28
```

and for `srGeweke`:

```
The Geweke test statistic is < 0.0005 or > 0.9995 in 20/50 comparisons
```

(the bounds of the confidence interval here are equal to [alpha/2, 1 − alpha/2], where alpha is the user-specified threshold). I'll make sure to also add this info to the remaining rules.

davidcerny:
Alright, the messages now look as follows:

```
Elapsed time: 00:00:18 (target: 15:00:00)
Minimum effective sample size (ESS): 0.931898 (target: > 50)
Maximum value of the Gelman-Rubin statistic (PSRF): 1.001753 (target: < 1.01)
The Geweke test statistic is < 0.0005 or > 0.9995 in 20/50 comparisons (target: 0)
The CI of a single-chain mean excludes the overall mean at p = 0.01 in 0/50 comparisons (target: 0)
```

Feel free to approve or provide further feedback @ms609

ms609:
Looks good!

My first thought on reading the Geweke line was "But 0 lies outside the range 0.005-0.995".  I quickly realised what was meant, but it might be clearer to end the line `(target: 0/50)`, or to use the same syntax as other lines, _viz_ `Number of comparisons in which  Geweke test statistic is < 0.0005 or > 0.9995: 20/50 (target: 0)`

src/core/analysis/mcmc/convergence/EssTest.h
```diff
@@ -9,10 +9,10 @@ class TraceNumeric;
     /**
      * @brief ESS test statistic for assessing convergence.
      *
-     * The ESS test statistic computes the effective sample size (ESS) for the given chain.
-     * We detect failure of convergence if the ESS is lower than k (usually 200).
-     *
-     *
+     * The ESS test statistic computes the effective sample size (ESS) for a given chain.
+     * We detect failure of convergence if the ESS is lower than k (often 200, but 625 may
+     * be more appropriate: see Guimaraes Fabreti & Hoehna 2022, Methods Ecol. Evol.
```
ms609, line 14: :+1:

src/core/analysis/stoppingRule/GelmanRubinStoppingRule.h
```diff
@@ -10,10 +10,12 @@ namespace RevBayesCore {
...
      *
-     * This stopping rule returns true when the the variance of samples between runs 
+     * This stopping rule returns true when the variance of samples between runs 
      * is approximately as large as the variance within runs has been reached.
-     * This rule is most useful if you want to check for convergence of continuous parameter
-     * between runs.
+     * This rule is most useful if you want to check for convergence of continuous parameters
+     * between runs. Although GelmanRubinTest.cpp contains a function for applying
+     * this test to a single chain (by splitting it into batches, and pretending that these
+     * represent different chains), we do not allow this here and require at least 2 chains.
```
ms609, line 18: :+1:

ms609 (approved):

davidcerny:
Excellent point as always! This is what the final messages look like:

```
Elapsed time: 00:00:02 (target: 15:00:00)
Minimum effective sample size (ESS): 0.504188 (target: > 50)
Maximum value of the Gelman-Rubin statistic (PSRF): 1.024429 (target: < 1.01)
Comparisons in which the Geweke test statistic is < 0.0005 or > 0.9995: 20/50 (target: 0/50)
Comparisons in which the CI of a single-chain mean excludes the overall mean at p = 0.01: 0/50 (target: 0/50)
```

I also noticed that in the unlikely but possible case of only having the `srMaxTime` rule, but also in the much more likely case of having `srMaxTime` and `srMaxIteration` as the only two rules, the `checkNow` variable would be set to `true` in every iteration, potentially spamming the standard output with hundreds of millions of lines. This is now fixed.

davidcerny:
> As `srMaxIteration` is somewhat redundant to `mcmc.run(generations = n)` I have been known to use only `srMaxTime`.

Note that the `generations` argument is converted to a `srMaxIteration` rule under the hood (see [`src/revlanguage/analysis/mcmc/RlMonteCarloAnalysis.cpp`](https://github.com/revbayes/revbayes/blob/a55579f5b4a9ae03622c323daa5d2157f71eb124/src/revlanguage/analysis/mcmc/RlMonteCarloAnalysis.cpp#L125)), so you still have two stopping rules in that situation. The case I described as unlikely but possible would be something like

```
mcmc.run(rules = [ srMaxTime(15, "hours") ])
```

> would perhaps some message at the start of a run be a useful indicator in this case that the stopping rules have been interpreted by the model as expected?

Maybe. Currently, `srMaxIteration` is the one rule that never prints anything, but it'd be easy to make it output a message in the same "elapsed (target)" format that's used by `srMaxTime`. I mostly didn't do this because it seemed redundant – an iteration counter is part of all the standard monitors (`mnFile`, `mnModel`, `mnScreen`), and any realistic analysis would use at least one of those. Similarly, we could change the printing condition to something like `if (checkNow || gen == 0)`, which would display the status of all the rules at the beginning of a run.

Then again, it's a question of just how far we're willing to go to try and save the users from themselves by printing all their settings back at them. 😛 Some other programs have a `verbosity` setting for this; maybe we could consider implementing something similar.

Let me know what you think. I can open a new PR for the changes described above, or you can experiment a bit, see what you like best, and open one yourself.

Issue #531, opened 2024-08-14
[BUG] Checkpointing doesn't work for FBD trees w/ nonzero offset from the present


davidcerny:
**Describe the bug**
Since #484, `BirthDeathSamplingTreatmentProcess::setValue()` has called `TreeUtilities::startingTreeInitializer()` to restore fossil age ranges when restarting FBD analyses from a checkpoint. Unfortunately, this solution only works for trees with extant tips. If a tree is entirely extinct (i.e., has a nonzero offset from the present), the restart attempt will fail, since the youngest tip of the tree stored in the checkpoint file is assumed to be of age 0. As a result, the tree will fail the check performed by `startingTreeInitializer()`, because the ages of its tips fall outside of specified bounds.

**To reproduce**
Run the Rev script from [this ZIP archive](https://github.com/user-attachments/files/16615970/fbd_test.zip). One absolute path has to be changed. 

**Expected behavior**
We should be able to restart FBD analyses from a checkpoint file without failing the check.

**Computer info**
macOS 14.5 (Sonoma), ARM64 architecture, RevBayes 1.2.5-preview, latest version of the `development` branch (f2a0610).

**Attempted solution**
In principle, the value of the offset _is_ recoverable, and we could modify `setValue()` by inserting something like this before the call to `startingTreeInitializer()`:

```c_cpp
for (size_t i = 0; i < taxa.size(); i++)
{
    double new_age = v->getTipNodeWithName( taxa[i].getName() ).getAge() + some_offset;
    v->getTipNodeWithName( taxa[i].getName() ).setAge( new_age );
}
```

The problem is figuring out the value of `some_offset`. It's equal to the age of the youngest tip sampled in the checkpointed iteration, but tip ages are only stored in the `.log` file rather than the `.state` file. The `.state` file does, however, store both the tree (the `fbd_tree` column in the example above) and the process age (the `origin_time` column in the example above; it can be either the age of the origin or of the root, depending on how BDSTP is set up). So the offset can also be computed by subtracting the root age of the stored tree (plus the length of the root edge if there is one) from the process age. I tried the following:

```
some_offset = getOriginAge() - v->getRoot().getAge();
```

but that doesn't work; `getOriginAge()` doesn't return the value stored in the checkpoint file (approximately 250.756 in the example above) but just a random draw from the prior. BDSTP also has a private member called `offset`, but that one is simply initialized from the `taxa` vector: in the example above, it is set to 66.2, while the true offset is 66.05169.

[...]

Issue #314, opened 2022-12-15
[BUG] Stoachstic variables when restarting from Checkpoint


wrightaprilm:
**Describe the bug**
A user contacted me with this bug. I'm attaching sample code and data. When restarting from checkpoint, it appears stochastic variables do not re-initialize correctly. 

**To Reproduce**

Code attached. 

**Expected behavior**

These should re-initialize in trace. You can see from the mnScreen output that they do not.

**Computer info**
Which OS are you using? Which version of RevBayes? This should be the first line you see when you start RevBayes.
[rerevbayesv1_1_1vsrevbayesv1_2_1queries.zip](https://github.com/revbayes/revbayes/files/10238946/rerevbayesv1_1_1vsrevbayesv1_2_1queries.zip)

Problem occurring on Windows and Mac. User reported issue on v1.1.1, v1.2.0, and v1.2.1. I continue to have the issue on the latest development.

bredelings:
Do we really support re-initializing from a trace?  I was thinking that maybe this is a historical artifact, and people really should be reinitializing from checkpoints instead.

If that's the case, perhaps the initializeFromTrace( ) function should be removed?

bredelings:
That said, I just tried this on the leave-ancestors-alone branch, and it seems to work.  Unlike checkpoints, branch lengths for the tree only have about 6 decimal places.

hoehna:
Yes, that's a historical artifact.

bjoelle:
closed as unsupported - won't fix

PR #646, opened 2025-01-22
MPI checkpointing fix

davidcerny:
This PR cherry-picks two commits from @jsigao's [`dev_jsigao`](https://github.com/revbayes/revbayes/tree/dev_jsigao)branch, which fix issue #184 with MPI checkpointing – one of our longest-standing and most reported problems.

@jsigao kindly allowed me to open this PR, but pointed out that his original issue which had motivated these two commits had not been completely solved by them. That makes sense – we know the checkpointing system is fragile and suffers from [other limitations](https://github.com/revbayes/revbayes/issues/531). To test the impact of this fix, I ran a modified version of one of the analyses from [Pol et al. (2024)](https://onlinelibrary.wiley.com/doi/10.1111/cla.12583): simple inference of a non-clock tree from a morphological matrix. I used MPI to distribute 4 MCMC runs across 4 cores, saving checkpoint files every 1000 iterations. I ran the analysis from scratch until it reached 25,000 iterations, then I terminated it, restarted it from the checkpoint, and ran it for another 25,000 iterations:

[MPI_test.zip](https://github.com/user-attachments/files/18512163/MPI_test.zip)

From the attached files and Tracer screenshots below, everything seems to have worked perfectly fine: all 4 runs were correctly restored, and for none of the monitored parameters is there an observable difference between the pre-restart and post-restart portions of the trace. So while @jsigao's changes don't solve all of our checkpointing problems, they solve a large class of them, which is why I think they are worth merging.

Screenshot 2025-01-22 at 14 05 44
Screenshot 2025-01-22 at 14 05 20
Screenshot 2025-01-22 at 14 06 40
Screenshot 2025-01-22 at 16 12 00


bjoelle (approved):
thanks for porting these changes !
I'm wondering if we should also have a null check in `MonteCarloAnalysis::initializeFromTrace` l381 since it seems like it would suffer from the same issue ?

davidcerny:
That would make sense, though based on #314, I'm wondering if we should perhaps instead remove `initializeFromTrace` in its entirety. Any thoughts on this?

bjoelle:
ah yes, it seems this function is not supported anymore and doesn't work anyway, so I would agree with removing it altogether

jsigao:
@davidcerny Thanks for taking care of this! It's great to know that these commits (which appeared to be unsuccessful attempts for fixing the problem I was encountering back then) actually do fix some related checkpointing problems.

(This PR looks good to me and I have nothing to add here. What I'll try to describe briefly below may be relevant to a separate case where checkpointing is used for an MC3 analysis. Not sure if this comment belongs here so please let me know if I should move this to somewhere else.)

I thought that it would be great if I could recall and describe what was the remaining issue I had with checkpointing that wasn't fixed by these two commits [unfortunately my notes (and possibly some unpushed changes) about this were lost due to a hard drive failure later that year]. After some attempt to recall what was the issue, it occurred to me that, although some of my unresolved problems back then probably should have been solved by the more recent `treeAssembly` related fixes, there is one remaining issue with checkpointing when used together with MC^3.

Specifically, under MC^3, each chain has its own associated checkpoint files automatically named as "\*chain_0\*", "\*chain_1\*", etc. The heats but not the chain indices are swapped during an MC^3 run. At the beginning of an MC^3 run, before any successful swap attempt, chain_0 does correspond to the cold chain; however, after a swap between the cold chain and the coldest heated chain, chain_0 would then correspond to the coldest heated chain instead. That is, as the checkpointing system does not record the heat (or the heat rank) associated with each chain, after some chain swaps there is no way to know which cold or heated chain that the "\*chain_0\*" checkpoint files are actually checkpointing.

When an MC^3 run is resumed from checkpoint files, looks like it always assumes that the "\*chain_0\*" files correspond to the cold chain, "\*chain_1\*" files correspond to the coldest heated chain, and so on. Therefore, the previous mapping between the (uncheckpointed) chain heats and other (checkpointed) chain-specific properties (e.g., variable values) would likely be distorted. In that case, the run would resume but likely behave differently from the original run.

davidcerny:
@jsigao Thank you so much for digging around to find out what the original issue was! It's very useful to know that while your changes solve the checkpointing problem for MPI, they do not solve it for MC^3 (which, of course, is one of the most common use cases for MPI).

Would you mind opening an [issue](https://github.com/revbayes/revbayes/issues) for this, just so that we don't lose sight of it once this PR is closed? Just copying the last two paragraphs of your comment here would be perfectly fine!

PR #647, opened 2025-01-23
Add cosine function

bredelings:
This just adds a cosine function.


src/revlanguage/functions/math/Func_cos.cpp
```diff
@@ -0,0 +1,105 @@
+#include "Func_cos.h"
+
+#include "Real.h"
+#include "RealPos.h"
```
bredelings, line 4: Why are we including RealPos here?

sigibrock (approved):

PR #658, opened 2025-01-31
Remove empty touch( ), keep( ), and  restore( ) specializations.

bredelings:
Very few functions specialize touch( ), keep( ), and restore( ).
These functions do specialize them, but then just call the parent class function which (a) would happen anyway without the specialization and (b) also doesn't do anything.
Let's remove these specializations so that its clear which few functions actually specialize touch/keep/restore.

I have also tacked on an unrelated change that fixes the name of rb-help2yml for the meson build.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -163,7 +163,7 @@ subdir('tests')
 
 if get_option('help2yml')
 
-  executable(rb_name+'help2yml',
+  executable(rb_name+'-help2yml',
              ['src/help2yml/main.cpp'],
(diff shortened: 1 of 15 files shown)


davidcerny (approved):
Looks good to me – a nice change that makes for cleaner code. It's a shame that none of the changed functions seems to have an integration test, which would help demonstrate that their behavior stays the same. But I can't imagine how getting rid of the empty specializations could possibly change anything.

PR #659, opened 2025-02-03
Changes to build files to streamline help infrastructure

brpetrucci:
Here, we make two sets of changes. They both refer to the generation of `src/core/help/RbHelpDatabase.cpp` from markdown files in `help/md/`. 

1. Moving `generate_help.sh` from `projects/meson/` to `projects/`. The script is not particular to meson builds, and it would be useful to make it so one does not need to rebuild every time one needs to update the help database.
2. Modifying `projects/cmake/build.sh` to always update the help database, and generate the `rb-help2yml`. It does not seem like doing so adds much time to the build, and it will eliminate unnecessary flags.


projects/cmake/build.sh
```diff
@@ -188,14 +182,11 @@ mv GitVersion.cpp ../../src/revlanguage/utils/
...
-then
-    (
-        cd ../../src
-        echo "Generating help database"
-        perl ../help/md2help.pl ../help/md/*.md > core/help/RbHelpDatabase.cpp
-    )
-fi
+(
+cd ../../src
+echo "Generating help database"
```
bredelings, line 187: Here can we just call the generate_help.sh script?

projects/cmake/build.sh
```diff
@@ -160,10 +158,6 @@ if [ "$static_boost" = "true" ] ; then
     cmake_args="-DSTATIC_BOOST=ON $cmake_args"
 fi
 
-if [ "$help2yml" = "true" ] ; then
-    cmake_args="-DHELP=ON $cmake_args"
```
bredelings, line 164: Deleting this line means that we will NOT build rb-help2yml, right?

projects/cmake/build.sh
brpetrucci: Sure! Will change

projects/cmake/build.sh
brpetrucci: My mistake. Will revert it

bredelings (approved):
Looks good!

PR #652, opened 2025-01-26
Prevent loss of integer precision in MC^3

davidcerny:
This should fix #651.

It'd be great to add a test for this, but the analysis where this issue popped up keeps crashing even after it is fixed – now for a different reason that no longer has to do with MC^3.

Issue #651, opened 2025-01-26
MC^3 may attempt to swap non-existent chains


davidcerny:
@ms609's dataset [950_ns_ki](https://github.com/mk-nt/950_ns_ki) throws an interesting error that apparently stems from the loss of precision that occurs when converting from `size_t` to `int`. In this example, the script specifies 8 chains per run, which are indexed from 0 to 7. However, because of this issue, the function `Mcmcmc::swapNeighborChains()` at one point picks a non-existent chain with an index of 8 for a swap, producing an invalid access error:

> I rebuilt the latest development RevBayes on my HPC and ran an mpi run with [950_ns_ki](https://github.com/mk-nt/950_ns_ki).  I saw:
> 
> ```
> RevBayes version (1.2.5-preview)
> Build from development (rapture-3215-g906131) on Wed  4 Dec 15:45:35 GMT 2024
> MPI enabled: using 128 processes.
> 
> Visit the website www.RevBayes.com for more information about RevBayes.
> 
> RevBayes is free software released under the GPL license, version 3. Type 'license()' for details.
> 
> To quit RevBayes type 'quit()' or 'q()'.
> 
> 
> Unknown user setting with key 'collapseSampledAncestors'.
> [ ... repeated many times (!) ... ]
> Unknown user setting with key 'collapseSampledAncestors'.
> source("mcmcmc.Rev")
>    Processing file "mcmcmc.Rev"
> Timestamp:      2024    339     59397
>    Successfully read one character matrix from file 'project950.neo.nex'
>    Successfully read one character matrix from file 'project950.trans.nex'
>    Processing file "ns_ki.Rev"
> 5       characters in partition with    2       states
> 1       characters in partition with    3       states
>    Processing of file "ns_ki.Rev" completed
>   - Created MCMCMC object
> 
>    Running burn-in phase of Monte Carlo sampler for 3000 iterations.
>    This simulation runs 2 independent replicates.
>    The MCMCMC simulator runs 1 cold chain and 7 heated chains.
>    The simulator uses 7 different moves in a random move schedule with 14.875 moves per iteration
> 
> 
> Progress:
> 
> 0---------------25---------------50---------------75--------------100
> ***********************************************[login2:3972680:0:3972680] Caught signal 11 (Segmentation fault: add     ress not mapped to object at address 0x8)
> [login2:3972690:0:3972690] Caught signal 11 (Segmentation fault: address not mapped to object at address 0x8)
[...]

Issue #609, opened 2024-11-21
[BUG] PDFs not up to date


ms609:
**Describe the bug**
debugMCMC  reports that PDFs not up to date during an MCMC run.

**To Reproduce**
- Check out https://github.com/mk-nt/950_ns_t_ki
- Launch `mcmcmc.Rev`.  This contains `setOption("debugMCMC", 1)`, and sets a seed.
![image](https://github.com/user-attachments/assets/c784f4a8-c45b-4ffb-975d-e6d9d7965491)

If I run without the `debugMCMC` flag then the run crashes.

https://github.com/mk-nt/950_ns_ki crashes during the burnin phase (with or without debugMCMC = 1) – potentially with the same cause?

**Computer info**
Win 10, latest Dev build.

**Additional info**
It appears that this has surfaced recently; these scripts seem to run okay on a version of development I built a month or so ago.

bredelings:
Hmm... what kind of crash is it?  Is it a SEGFAULT or is there an error message?

ms609:
On Windows, when I run rb as a separate process, it closes with no message.  On the Linux HPC it segfaults in MPI mode, and stops with "Aborted" in non-mpi.

bredelings:
OK, so at least its clearly triggering some kind of error condition in non-mpi mode on Linux.
I don't see a crash, but I do see this:
```

Iter        |     Likelihood   |       rate_neo   |   root_freqs..      |root_freqs..   |    elapsed   |        ETA   |
-----------------------------------------------------------------------------------------------------------------------
0           |        -47.084   |     0.07518095   |       0.156231      |    0.843769   |   00:00:00   |   --:--:--   |
300         |       -51.0988   |      0.2241027   |        0.88862      |     0.11138   |   00:00:03   |   --:--:--   |
600         |       -46.0547   |     0.02301069   |       0.418351      |    0.581649   |   00:00:05   |   5124095576030431:00:11   |
900         |        -48.192   |        1.20194   |       0.267106      |    0.732894   |   00:00:08   |   5124095576030431:00:08   |
1200        |       -49.8724   |      0.6096102   |      0.0982725      |    0.901727   |   00:00:11   |   5124095576030431:00:05   |
```

The ETA values appear rather wild

davidcerny:
@bredelings Those are due to #542 – we don't have a good way of estimating ETA when using stopping rules other than a fixed number of iterations.

ms609:
[...]

PR #661, opened 2025-02-07
Prevent NaN heats in MC^3

davidcerny:
This PR makes the following changes to `Mcmcmc.cpp`:

1. Addresses @bredelings's comment under PR #652 (which was probably merged prematurely) and avoids subtraction in a `for`-loop that checks if `ht[i+1]` is valid;
2. Solves issue #651 (and partially #609), which ultimately stems from the fact that the interpolation system we are using to correct lower-than-allowed chain heats can, under certain circumstances, produce `nan` values, which then screw up everything else.

It's still not clear to me why the problem with `nan` heats only appeared on certain platforms, but I am confident that the interpolation system now does what the corresponding source code comment says it should do. I have verified that this PR solves the issue on macOS(ARM64) 15.3 with clang 16 (where it was previously giving a bus error) and Red Hate Enterprise Unix 8.4 with gcc 10.2 (where it was previously causing a segfault), but I would be grateful if @ms609 could check that it works on his setup as well.

I could add a pared down version of @ms609's analysis as an integration test, but it would be a demanding one – it would require running 16 (!) MCMC chains of ~10,000 moves each.

### DIFF
--- a/src/core/analysis/mcmc/Mcmcmc.cpp
+++ b/src/core/analysis/mcmc/Mcmcmc.cpp
@@ -1143,7 +1143,7 @@ void Mcmcmc::setHeatsInitial(const std::vector<double>& ht)
     }
 
     // 3. Check that the heats are sorted with the largest heat first (== smallest temperature first).
-    for(size_t i=0;i<ht.size()-1;i++)
+    for(size_t i=0; i+1 < ht.size(); i++)
         if (ht[i] < ht[i+1])
             throw RbException()<<"Heat "<<i+1<<" ("<<ht[i]<<") should be greater than heat "<<i+2<<" ("<<ht[i+1]<<")";
 
--- a/src/core/analysis/mcmc/Mcmcmc.cpp
+++ b/src/core/analysis/mcmc/Mcmcmc.cpp
@@ -1865,9 +1865,27 @@ void Mcmcmc::tune( void )
 //            std::cout << "chain_heats[" << hotterChainIdx << "]=" << chain_heats[hotterChainIdx] << std::endl;
         }
         
-        // if the heat of a given hot chain is smaller than the minimum bound
-        // interpolate this heat and the heats of all the hotter chains
-        // to fall between the lowest heat that is greater than the minimum bound and the minimum bound
+        /* If the heat of a given hot chain is smaller than the minimum bound, we want to linearly interpolate
+         * this heat and the heats of all the hotter chains to fall between the lowest heat that is greater
+         * than the minimum bound and the minimum bound.
+         * The formula for linearly interpolating between (x1,y1) and (x2,y2) is y(x) = y2 - (y2-y1)*[ (x-x1)/(x2-x1) ].
+         *
+         * Let "heat(i)" be a notational shortcut for chain_heats[ chainForHeatIndex(i) ], and let k be the heat
+         * index of the coldest of such "bad" chains. Then, its heat can be calculated as
+         *
+         * log(heat(k)) = log(heat(k-1)) - [ log(heat(k-1)) - log(heatMinBound) ]*[ (k - (k-1)) / (num_chains - (k-1)) ]
+         *              = log(heat(k-1)) - [ log(heat(k-1)) - log(heatMinBound) ]*[ 1 / (num_chains - (k-1)) ]
+         *
+         * Let us further denote heat(k-1) as "lowestGreaterThanMin", and k-1 as "m". Then,
+         *
+         * log(heat(k)) = log(heatMinBound) + [ log(lowestGreaterThanMin) - log(heatMinBound) ]*[ 1 / (num_chains - m) ]
+         *
+         * Finally, let us denote the second addend on the right-hand side of the equation above by log(rho). Then,
+         *
+         *       rho = (lowestGreaterThanMin / heatMinBound)^[ 1 / (num_chains - m) ],
+         *   heat(k) = lowestGreaterThanMin / rho
+         * heat(k+1) = lowestGreaterThanMin / rho^2, etc.
+         */
         
         std::vector<double> greaterThanMin;
         std::vector<size_t> badChainIdx;
--- a/src/core/analysis/mcmc/Mcmcmc.cpp
+++ b/src/core/analysis/mcmc/Mcmcmc.cpp
@@ -1870,14 +1870,31 @@ void Mcmcmc::tune( void )
         // to fall between the lowest heat that is greater than the minimum bound and the minimum bound
         if (j < num_chains)
         {
-            double rho = pow(chain_heats[colderChainIdx] / heatMinBound, 1.0 / (num_chains - j));
-            size_t k = j;
+            std::cout << "The following should be sorted from largest to smallest:" << std::endl;
+            for (size_t l = 0; l < chain_heats.size(); ++l)
+            {
+                if (l == j)
+                {
+                    std::cout << chain_heats[ chainForHeatIndex(l) ] << " <--- You are here" << std::endl;
+                }
+                else
+                {
+                    std::cout << chain_heats[ chainForHeatIndex(l) ] << std::endl;
+                }
+            }
+            
+            double rho = pow(chain_heats[ chainForHeatIndex(j) ] / heatMinBound, 1.0 / (num_chains - j + 1));
+            std::cout << "Current rho is: " << rho << std::endl;
+            size_t k = j + 1;
             
             for (; k < num_chains; ++k)
             {
-                chain_heats[hotterChainIdx] = chain_heats[colderChainIdx] / pow(rho, k + 1 - j);
+                std::cout << "Attempting to set the heat of chain " << k << " to " << (chain_heats[ chainForHeatIndex(j) ] / pow(rho, k + 1 - j)) << std::endl;
+                chain_heats[ chainForHeatIndex(k) ] = chain_heats[ chainForHeatIndex(j) ] / pow(rho, k + 1 - j);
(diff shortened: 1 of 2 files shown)


ms609:
Wonderful, thanks for this!

- [x] `950_ns_ki` now completes burnin (previous error at ~75% burnin) and runs happily for >8k generations
- [x] `950_ns_t_ki` runs happily for >3000 generations (previous error at ~500)

so this looks like it's addressed the presenting issues.  Thank you!

davidcerny:
Oh wow, I didn't expect this to fix the issue with `950_ns_t_ki` as well – nice!

src/core/analysis/mcmc/Mcmcmc.cpp
```diff
@@ -1868,15 +1868,38 @@ void Mcmcmc::tune( void )
...
+            else
+            {
+                badChainIdx.push_back(l);
+            }
+        }
+        
+        // std::cout << "We have " << badChainIdx.size() << " bad chain(s)" << std::endl;
+        if ( badChainIdx.size() != 0 )
+        {
+            double lowestGreaterThanMin = *std::min_element( greaterThanMin.begin(), greaterThanMin.end() );
```
bjoelle, line 1892: does that work if all chains are bad ? is that a possibility ?

bredelings: Do we know which operation produce the NaN, and what its arguments where?

Also, I am trying to follow the interpolation math.  It seems a lot easier on the log scale.  Could we add a 1-line comment?

davidcerny: > Do we know which operation produce the NaN, and what its arguments where?

Yes – I described it in detail in a [comment](https://github.com/revbayes/revbayes/issues/651#issuecomment-2642082750) on the original issue, but probably should have done it here instead. The `nan` was coming from the `rho` constant, because when the initially proposed value of `chain_heats[colderChainIdx]` was negative (which can occasionally happen), we ended up trying to take the square root of a negative number.

Adding a comment is a good idea; will do.

bredelings:
```
            // We want to linearly interpolate the log(heat) between (j-1, log(chain_heats[colderChainidx])) and (num_chains-1, log(heatMinBound))
            // The formula for linearly interpolating between (x1,y1) and (x2,y2) is y(x) = y1 + (y2-y1)*[ (x-x1)/(x2-x1) ]
            //
            // Therefore log(chain_heats[k]) = log(chain_heats[colderChainIdx]) + (log(heatMinBound) - log(chain_heats[colderChainIdx]))
            //                                         * (k - (j-1))/((num_chains-1) - (j-1))

```

bredelings: We could also implement it this way.

davidcerny: Have you actually tested that on `950_ns_ki`? I'm not convinced the original code with `chain_heats[colderChainidx]` was always guaranteed to pick the lowest heat greater than the minimum.

bredelings: If we want to avoid having the hottest chain heat equal to `heatMinBound`, we can accomplish this simply by changing `x2` to `num_chains`.

Hmm.... interesting that the `chain_heats[colderChainIdx]` could be negative.  That's unfortunate.

src/core/analysis/mcmc/Mcmcmc.cpp
```diff
@@ -1865,18 +1865,59 @@ void Mcmcmc::tune( void )
...
-        if (j < num_chains)
+        /* If the heat of a given hot chain is smaller than the minimum bound, we want to linearly interpolate
+         * this heat and the heats of all the hotter chains to fall between the lowest heat that is greater
+         * than the minimum bound and the minimum bound.
+         * The formula for linearly interpolating between (x1,y1) and (x2,y2) is y(x) = y2 - (y2-y1)*[ (x-x1)/(x2-x1) ].
+         *
+         * Let "heat(i)" be a notational shortcut for chain_heats[ chainForHeatIndex(i) ], and let k be the heat
+         * index of the coldest of such "bad" chains. Then, its heat can be calculated as
+         *
+         * log(heat(k)) = log(heat(k-1)) - [ log(heat(k-1)) - log(heatMinBound) ]*[ (k - (k-1)) / (num_chains - (k-1)) ]
```
bredelings, line 1876: I think `(k - (k-1))` should be `(k - (j-1))`, where `j-1` is the has the lowest heat greater than the min.  Also `(num_chains - (k-1))` should be `(num_chains - (j-1))`.

davidcerny: When you sort your chains from coldest (index `0`) to hottest (index `num_chains - 1`), then if the chain with index `k` is the "coldest bad chain" as stated in the comment, it follows that the chain with index `k-1` is the chain with the lowest heat greater than the min. This is why you can simplify `(k - (k - 1))` to just 1, which is what we do in the calculation of the `rho` constant. That part is not newly introduced by my PR, it was already there and it works as intended. My contribution is a more reliable look-up of the chains with indices `k` and `k-1`.

davidcerny: Hmm, no, with a bit of ~~pencil-and-paper~~ Wacom-and-S Pen math, I've convinced myself that the current formula is correct, and that not making a distinction between `j` and `k` is key to understanding why the exponent of `rho` simplifies to `1 / (num_chains - (k-1))`:

Notes_250207_151437

bredelings: I guess I am confused because `log(heat(k)) = log(heat(k-1)) - [ log(heat(k-1)) - log(heatMinBound) ]*[ (k - (k-1)) / (num_chains - (k-1)) ]` isn't an interpolation formula anymore.

davidcerny: I'm also increasingly confused...

My best guess is that we're really talking about the same thing, just in different notations. You prefer a general formula for an arbitrary "bad" chain, so your `k` ranges from your `j` to `num_chains - 1`. In the code (and the comment I added to explain the code), `k` instead denotes the heat index of one particular "bad" chain (the coldest one), and is used to calculate an intermediate variable `rho` (= the ratio of the heats of two chains with adjacent heat indices). The heats of all "bad" chains are then expressed in terms of powers of `rho`. The math is the same.

But again, I'm not 100% sure I understand your feedback. I think what would help me understand it is if you could clarify if you are objecting to (1) the code, (2) the comment, or (3) both; and if you think that things just need to be renamed for clarity, or that the interpolation math itself is not right.

bredelings: OK, I looked more, and I think I understand the comment now.  I was confused because the formula that you wrote was very similar to what I wrote, but you were using it to derive the `rho`  and compute `heat(k)` from `heat(k-1)` instead of to compute `heat(k)` directly for each `k`.  I guess both ideas use similar triangles (like your sketch), but because you wrote "interpolating between", I expected you to compute f(k) for k _between_ two extremes.

Hmm... I guess I am still uncertain.  In the interpolation formula, `x1` and `x2` represent constants, whereas `x` represents a variable.  Is the `k` in your formula a variable, or a constant?  At first I thought it was a variable, but `num_chains - (k-1)` seems like it only makes sense if `k` is a constant.

src/core/analysis/mcmc/Mcmcmc.cpp
```diff
@@ -1865,18 +1865,59 @@ void Mcmcmc::tune( void )
...
+         *              = log(heat(k-1)) - [ log(heat(k-1)) - log(heatMinBound) ]*[ 1 / (num_chains - (k-1)) ]
+         *
+         * Let us further denote heat(k-1) as "lowestGreaterThanMin", and k-1 as "m". Then,
+         *
+         * log(heat(k)) = log(lowestGreaterThanMin) - [ log(lowestGreaterThanMin) - log(heatMinBound) ]*[ 1 / (num_chains - m) ]
```
bredelings, line 1881: Is `k` a variable here?  If so, it seems like there should be a `k` on the right-hand-side.  So then maybe `k` is not a variable?

src/core/analysis/mcmc/Mcmcmc.cpp
```diff
@@ -1865,18 +1865,59 @@ void Mcmcmc::tune( void )
...
+        if ( badChainIdx.size() != 0 )
+        {
+            double lowestGreaterThanMin = *std::min_element( greaterThanMin.begin(), greaterThanMin.end() );
+            size_t k = *std::min_element( badChainIdx.begin(), badChainIdx.end() );
+            size_t m = k - 1;
+            
+            double rho = pow(lowestGreaterThanMin / heatMinBound, 1.0 / (num_chains - m));
+            // std::cout << "Current rho is: " << rho << std::endl;
             
             for (; k < num_chains; ++k)
```
bredelings, line 1917: Here `k` seems to be a variable...

davidcerny:
`k` is a constant, as I was trying to say in my previous comment:

> `k` [...] denotes the heat index of one particular "bad" chain (the coldest one)

From this, it follows that the heat index of the second coldest "bad" chain will be `k+1`, that of the third coldest "bad" chain will be `k+2`, etc. It is true that the loop at line 1917 that you point out simply increments `k` directly instead of defining a new variable and letting it vary over an appropriate range, which might be seen as an abuse of notation. But the math does work out, in that each heat index is assigned the correct heat (`lowestGreaterThanMin / rho` for `k`, `lowestGreaterThanMin / rho^2` for `k+1`, etc.).

bredelings (approved):

Issue #428, opened 2024-02-12
[BUG] MPI version does not run the specified number of independent replicates


davidcerny:
**Describe the bug**
The `nruns` argument of the `mcmc()` function is supposed to specify the number of independent replicates. This works fine with the single-thread version of RevBayes, but if the analysis is executed using the MPI version, some of the replicates are apparently initiated with the same random seed. This makes them draw exactly the same samples, producing redundant output and reducing the number of independent replicates below that specified by the user.

**To reproduce**
[This zipped directory](https://github.com/revbayes/revbayes/files/14252328/tmp.zip) contains a script specifying 10 independent runs and all the data required to run it. One absolute path has to be changed in the script. When I execute it using the MPI version of RevBayes with one core per run (`mpirun -np 10 ./rb-mpi`), the analysis will indeed start with 10 runs, but of these, no more than 3 to 7 will actually be independent of one another.

**Expected behavior**
The replicates should all be independent of one another.

**Computer info**
macOS 14.3 (Sonoma), RevBayes 1.2.2, latest version of the `development` branch (rapture-2386-g54c6ba) and Red Hat Enterprise Linux 8.4 (Ootpa), RevBayes 1.2.2, `development` branch, commit rapture-2364-ga6a40c.

davidcerny:
Unfortunately, yes.

davidcerny:
Fixed by #662 – a big thank you to @bjoelle for getting to the bottom of this!

PR #662, opened 2025-02-07
Use different seeds for replicates in MPI run

bjoelle:
This is an attempt to fix issue #428 by explicitly setting seeds instead of doing draws
I have tested it with the issue script and I obtain 10 separate log files with no identical output

### DIFF
--- a/src/core/analysis/MonteCarloAnalysis.cpp
+++ b/src/core/analysis/MonteCarloAnalysis.cpp
@@ -517,8 +517,7 @@ void MonteCarloAnalysis::resetReplicates( void )
     size_t replicate_start = size_t(floor( (double(pid-active_PID) / num_processes ) * replicates ) ) + active_PID;
     
     RandomNumberGenerator *rng = GLOBAL_RNG;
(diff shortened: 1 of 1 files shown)


davidcerny:
Oh wow, thanks so much for this – your change looks so much simpler than I assumed it would have to be! Will give this a try later today.

bredelings (approved):

davidcerny (approved):
I've also tested this using the analysis from the original issue (with 10 independent runs), performing 3 trials with user-set seeds and another 3 trials with auto-generated seeds on both macOS and Linux. I can confirm the issue is fully fixed by this PR.

Interestingly, it looks like the issue has already gotten less severe between v1.2.3 and now. In that version, the number of unique initial states would be lower than the specified number of runs, and the runs with the same initial state would keep sampling the same values throughout (e.g., 10 runs, 4 unique initial states, 4 unique traces). On the current tip of `development`, there are still fewer initial states than runs, but the runs with the same initial state diverge from one another, so their outputs end up being different (10 runs, 4 unique initial states, 10 unique traces). This PR provides a definitive fix that ensures everything works as expected (i.e., 10 runs, 10 unique initial states, 10 unique traces).

PR #666, opened 2025-02-11
Document DPP moves

davidcerny:
At my request, PR #606 deleted the Markdown help file for `mvDPPValueBetaSimplex`, because neither @ms609 nor myself could find the source code for it. The help file was empty, so no great harm was caused by this, but the move does in fact exist. The reason why we couldn't find any occurrence of the string "DPPValueBetaSimplex" in the `src` directory is that it is derived from a template (`DPPTableValueUpdate.h`). This PR (1) adds a brief note to the move register to clarify this, and (2) provides help files for all three moves deriving from `DPPTableValueUpdate` (`mvDPPValueBetaSimplex`, `mvDPPValueScaling`, `mvDPPValueSliding`).

### DIFF
--- a/help/md/mvDPPValueBetaSimplex.md
+++ b/help/md/mvDPPValueBetaSimplex.md
@@ -1,0 +1,31 @@
+## name
+mvDPPValueBetaSimplex
+## title
+Beta simplex move applied to individual categories of a Dirichlet process mixture
+## description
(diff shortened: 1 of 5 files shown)


ms609 (approved):
👍 Well done for tracking this down!

PR #667, opened 2025-02-13
vectorFlatten.md fnFreeK.md, fnReadVCF.md, dnUniformInteger.md help files

raymondcast18:
- [x] vectorFlatten.md
- [x] fnFreeK.md
- [x] fnReadVCF.md
- [x] dnUniformInteger.md
- Added error mssage to VCFReader.cpp


raymondcast18:
Need to verify readVCF functionality

help/md/readVCF.md
```diff
@@ -1,9 +1,16 @@
...
+Read VCF file into revbayes
 ## details
+readVCF accepts two arguments to read in a VCF file. The first argument
+specifies the relative or absolute file path to desired VCF file. The second
+specifies type of data to be constructed (default binary). This function
+only allows for 0, 1, and . characters in the VCF file.
 ## authors
 ## see_also
 ## example
+    x <- readVCF("path/to/VCF/file", "DNA")
```
bredelings, line 15: Maybe the "DNA" here should be "binary".

help/md/dnUniformInteger.md
```diff
@@ -1,9 +1,23 @@
...
 dnUniformInteger
 ## title
+Uniform Integer Distribution
 ## description
+This function creates a stochastic node drawing a random integer from a uniform distribution.
 ## details
+This function will randomly draw an integer from a uniform distribution
+from a minimum and maximum integer set in the first and second arguments.
+dnUniformInteger must be defined as a stochastic node. This function can also
+be called using the alias 'dnUnifInt'.
```
davidcerny, line 11: Several points here:

- This is just stylistics, but when describing the support of a distribution, we can say "from _x_ to _y_" or "between _x_ and _y_". You say "from _x_ and _y_", which is a combination of the two that doesn't quite work.
- It's not `dnUniformInteger` itself that "must be defined as a stochastic node", but rather a draw from it. This distinction between distributions and draws from distributions is quite important. For example, we can define a workspace object corresponding to a distribution (using the `=` assignment operator), and then draw from it and assign the result to a stochastic node (using the `~` operator).

    ```
    my_dist = dnUniformInteger(0, 100)   # the distribution is assigned to a workspace object
    var ~ my_dist                        # a *draw* is assigned to a stochastic node
    ```

- However, it is true of _all_ distributions that draws therefrom are treated as stochastic nodes, so perhaps it's a bit superfluous to include this information in the first place.

help/md/dnUniformInteger.md
```diff
@@ -1,9 +1,23 @@
...
 ## authors
+Sebastion Hoehna
 ## see_also
+dnNormal
+dnExponential
```
davidcerny, line 16: Surely `dnUniform` would be more relevant here?

help/md/dnUniformInteger.md
```diff
@@ -1,9 +1,23 @@
...
 ## example
+    # Create stochastic node
+    x ~ dnUniformInteger(1, 10)
+    # See what x was assigned
+    x
+    5
```
davidcerny, line 22: It won't always be 5. If you want the user to be able to reproduce the example (i.e., draw a particular number), make sure to set the seed first.

help/md/fnFreeK.md
```diff
@@ -1,9 +1,31 @@
 ## name
 fnFreeK
 ## title
+Free K Rate Matrix
 ## description
+This function generates and returns a free rates matrix. 
 ## details
+This function accepts both RealPos[] or RealPos[][] as the first argument to automatically
```
davidcerny, line 8: The information on acceptable argument types is already given to us by the auto-generated portion of the help file. Here, it might be better to say that the first argument can be either a vector or a matrix of nonnegative real numbers, and explain the difference between these two kinds of input – i.e., when to supply a vector (and what the function does with it) and when to supply a matrix (and what the function does with it).

help/md/fnFreeK.md
```diff
@@ -1,9 +1,31 @@
...
+generate a rate matrix with corresponding substitution rates, returning a rate matrix object.
+Users can specify if matrix should be normalized in the second argument using a boolean 
+variable (default TRUE). Lastly users can specify what matrix exponential method to
+use (default eigen) with a string. Possible options include:
+scalingAndSquaring
+scalingAndSquaringPade
+scalingAndSquaringTaylor
+uniformization
+eigen
```
davidcerny, line 17: Maybe put hyphens (`-`) before the individual items to turn this into an actual list

help/md/readVCF.md
```diff
@@ -1,9 +1,16 @@
 ## name
 readVCF
 ## title
+Read VCF
 ## description
+Read VCF file into revbayes
```
davidcerny, line 6: Capitalize RevBayes as appropriate

help/md/readVCF.md
```diff
@@ -1,9 +1,16 @@
...
 ## details
+readVCF accepts two arguments to read in a VCF file. The first argument
```
davidcerny, line 8: Can we explain what the VCF acronym stands for?

davidcerny (changes requested):

help/md/dnUniformInteger.md
bredelings: I agree on the first two points.  For the last point, one complicating factor is that for some distributions we have two versions, where one creates a stochastic node and one does not.  For example `rnorm( )` and `dnNormal( )`.

davidcerny: True – it would be more accurate to say that everything that starts with `dn` creates a stochastic node. Still, it seems to me like that information should either be included in every help file (for clarity), or in none of them (for brevity). Reminding the user of it in some help files but not others is a bit inconsistent...

help/md/dnUniformInteger.md
davidcerny: I don't think this particular piece of feedback has been addressed

help/md/dnUniformInteger.md
```diff
@@ -1,9 +1,20 @@
...
+from a minimum to maximum integer set in the first and second arguments.
+This function can alsonbe called using the alias 'dnUnifInt'.
 ## authors
+Sebastion Hoehna
 ## see_also
+dnNormal
+dnExponential
 ## example
+    # Create and assign stochastic node
+    # To obtain a new value for x the distribution will need to be called again
```
davidcerny, line 18: Or one could just write `x.redraw()`... I don't think it's necessary to include the middle line here.

help/md/fnFreeK.md
```diff
@@ -1,9 +1,52 @@
...
 ## description
+This function generates and returns a free rates matrix. 
 ## details
+This function accepts both a vector or a matrix of non-negative, real numbers in the 
+first argument to automatically generate a rate matrix with corresponding substitution 
+rates, returning a rate matrix object. The function will fill rates in the matrix from
+left to right as provided in the first argument, skipping the diagonal when using a vector
+as input. For this reason,using a vector of lengths 2 to 5 will create a 2-by-2 rate
+matrix but a vector of length 6 will create a 3-by-3 rate matrix as fnFreeK will have 
+enough values to fill the matrix. Using a matrix to create in fnFreeK will create a
```
davidcerny, line 14: Thank you for documenting this behavior! I actually had no idea this was how `fnFreeK` worked, and find it a bit disconcerting. A 2-by-2 matrix has 2 off-diagonal entries that the user can set, while a 3-by-3 one has 6, so what happens if we supply a vector of length 3, 4, or 5? Apparently, everything but the first two elements will be ignored. I can see several ways to deal with this:

- Rewrite the function so that it throws an error if the vector is not of the right length (i.e., _n_*(_n_ − 1) for some _n_);
- Keep the current behavior but warn the user that the last _k_ elements will be ignored (similar to what R does when it has to recycle a vector).

I understand that these changes go beyond just documenting existing behavior, though, so they are probably outside the scope of this PR. Still, for the time being, it would be nice to at least warn the users here in the documentation that if the vector is not of the right length, the extra elements will be ignored. (The current wording already implies this, but we should make it explicit.)

help/md/fnFreeK.md
```diff
@@ -1,9 +1,52 @@
...
+first argument to automatically generate a rate matrix with corresponding substitution 
+rates, returning a rate matrix object. The function will fill rates in the matrix from
+left to right as provided in the first argument, skipping the diagonal when using a vector
+as input. For this reason,using a vector of lengths 2 to 5 will create a 2-by-2 rate
+matrix but a vector of length 6 will create a 3-by-3 rate matrix as fnFreeK will have 
+enough values to fill the matrix. Using a matrix to create in fnFreeK will create a 
+rate matrix object with rates filled in their respective position in the provided matrix.
+Users can specify if matrix should be normalized in the second argument using a boolean 
+variable (default TRUE). Lastly users can specify what matrix exponential method to
+use (default eigen) with a string. Possible options include:
```
davidcerny, line 18: Just two minor typos to fix in this paragraph: (1) in the phrase "Using a matrix to create in fnFreeK", there are either some missing words or the "in" doesn't belong there; and (2) there should be a space after the comma following "For this reason".

help/md/readVCF.md
```diff
@@ -1,9 +1,17 @@
 ## name
 readVCF
 ## title
+Read VCF
 ## description
+Read VCF file into RevBayes
 ## details
+readVCF reads in a file that is in Variant Call Formati (VCF), accepting two
```
davidcerny, line 8: Typo: "Formati" should read "Format"

help/md/readVCF.md
```diff
@@ -1,9 +1,17 @@
...
+ arguments. The first argument specifies the relative or absolute 
+file path to desired VCF file. The second specifies type of data
+to be constructed (default binary). This function
+only allows for 0, 1, and . characters in the VCF file.
 ## authors
 ## see_also
 ## example
+    x <- readVCF("path/to/VCF/file", "binary")
 ## references
```
davidcerny, line 17: We could cite [Danecek et al. (2011)](https://academic.oup.com/bioinformatics/article/27/15/2156/402296) here...

davidcerny (changes requested):

help/md/dnUniformInteger.md
raymondcast18: Sorry about that! Thought I did ill fix it!

help/md/dnUniformInteger.md
raymondcast18: sounds good!

help/md/fnFreeK.md
raymondcast18: Ill make a note and bring it up in my meeting with ben! possibly could be addressed in a future PR

help/md/fnFreeK.md
davidcerny: Apologies, maybe I wasn't totally clear about what changes I was requesting. It was specifically this phrase, which I don't think is grammatical:

> Using a matrix to create in fnFreeK will create a rate matrix object with [...]

We could instead say something like:

- "Using a matrix in fnFreeK will create a rate matrix object with [...]"
- "Supplying a matrix to fnFreeK will create a rate matrix object with [...]"

Also, this is super minor, but if possible, let's remove the extra whitespace here (line 35):

```
               ↓
assigned using  their respective
```

help/md/dnUniformInteger.md
```diff
@@ -1,9 +1,19 @@
 ## name
 dnUniformInteger
 ## title
+Uniform Integer Distribution
 ## description
+This distribution creates a stochastic node drawing a random integer from a uniform distribution.
 ## details
+This distribution will randomly draw an integer using uniform distribution
+from a minimum to maximum integer set in the first and second arguments.
+This function can alsonbe called using the alias 'dnUnifInt'.
```
davidcerny, line 10: Typo: `alsonbe` should read `also be`.

davidcerny:
Sorry for the delay in reviewing this @raymondcast18! I've provided a few last suggestions, but overall this is looking great.

davidcerny (approved):

Issue #650, opened 2025-01-25
[BUG] possible bug in resuming MCMCMC from checkpoint


jsigao:
After group efforts from @davidcerny @bjoelle @bredelings and others, checkpointing now seems to be fully functional for various scenarios. It occurs to me that there may be another issue that affects checkpointing for being used together with MC^3. 

**Describe the bug**
Specifically, under MC^3, each chain has its own associated checkpoint files automatically named as "\*chain_0\*", "\*chain_1\*", etc. The heats but not the chain indices are swapped during an MC^3 run. At the beginning of an MC^3 run, before any successful swap attempt, chain_0 does correspond to the cold chain; however, after a swap between the cold chain and the coldest heated chain, chain_0 would then correspond to the coldest heated chain instead. That is, as the checkpointing system does not record the heat (or the heat rank) associated with each chain, after some chain swaps there is no way to know which cold or heated chain that the "\*chain_0\*" checkpoint files are actually checkpointing.

When an MC^3 run is resumed from checkpoint files, looks like it always assumes that the "\*chain_0\*" files correspond to the cold chain, "\*chain_1\*" files correspond to the coldest heated chain, and so on. Therefore, the previous mapping between the (uncheckpointed) chain heats and other (checkpointed) chain-specific properties (e.g., variable values) would likely be distorted. In that case, the run would resume but likely behave differently from the original run.

**To Reproduce**
Unfortunately I don't have a good minimum reproducible example now for this issue. It's a bit tricky to find a good reproducible example that can clearly expose this issue because it will not trigger an error or cause a crash. Instead, resuming an MC^3 analysis from checkpoint would appear to be issue-free while the problem would likely go unnoticed. Example Rev scripts that specify MC^3 analyses with checkpointing can be found in this previous [issue](https://github.com/revbayes/revbayes/issues/540). 

**Expected behavior**
When resume an MC^3 analysis from checkpoint, the mapping between chain heats and other chain-specific properties should remain undistorted. 

**Potential fixes**
When the heats are pre-specified and thus do not change throughout the MC^3 analysis, it seems that this issue could be solved by also swapping the corresponding `checkpoint_file_name` when swap two chains. In that case, for example, the "\*chain_0\*" checkpoint file names would always associate with the cold chain.

[...]

Issue #525, opened 2024-07-26
[BUG] Some iterations not logged by `mnModel` w/ restarted MC^3


davidcerny:
**Describe the bug**
When an MCMCMC analysis is restarted and an `mnModel` is used to write a log file, the monitor correctly logs iterations at the specified frequency for the first run, but starts to stochastically omit certain iterations during the subsequent runs. In contrast, the `mnScreen` monitor behaves as expected.

**To reproduce**
Run the following Rev code:

```
seed(123)

p ~ dnUniform(0, 1)
n <- 20
k ~ dnBinomial(n, p)
k.clamp(7)
mymodel = model(k)

moves = VectorMoves()
moves.append( mvSlide(p, delta=0.1, tune=TRUE, weight=1) )

monitors = VectorMonitors()
monitors.append( mnScreen(printgen=20) )
monitors.append( mnModel(filename="output/MC3.log", printgen=20) )

myMcmcmc = mcmcmc(mymodel, monitors, moves, nchains=4, deltaHeat=5)

# Run a short analysis
myMcmcmc.run(generations=500, tuningInterval=50)

# Run the MCMCMC again
myMcmcmc.run(generations=500, tuningInterval=50)
```

This is what `MC3.log` looks like:

```
Iteration	Posterior	Likelihood	Prior	p
0	-6.773369	-6.773369	0	0.6964692
20	-1.712716	-1.712716	0	0.3278227
40	-2.048599	-2.048599	0	0.2640655
60	-3.031421	-3.031421	0	0.532082
80	-3.56648	-3.56648	0	0.1700475
100	-1.690883	-1.690883	0	0.347661
120	-4.187337	-4.187337	0	0.5979775
140	-1.704615	-1.704615	0	0.3679634
160	-2.152482	-2.152482	0	0.2531715
180	-1.706428	-1.706428	0	0.369102
200	-2.376409	-2.376409	0	0.2338294
220	-3.211032	-3.211032	0	0.1850844
240	-2.046377	-2.046377	0	0.4427209
260	-1.788944	-1.788944	0	0.3981595
280	-1.865908	-1.865908	0	0.4146249
300	-1.730537	-1.730537	0	0.3203002
320	-1.693003	-1.693003	0	0.3573531
340	-2.769988	-2.769988	0	0.2076967
360	-2.485443	-2.485443	0	0.4897146
380	-1.699998	-1.699998	0	0.3646803
400	-2.193082	-2.193082	0	0.4605915
[...]

PR #670, opened 2025-02-17
MC^3 checkpointing fix

davidcerny:
This PR fixes issue #650, basically by implementing @jsigao's ideas and suggestions:

- Instead of setting the names of the checkpoint files only once at the beginning of each analysis, it dynamically resets them every time the files are to be written. This ensures that the checkpoint file with the suffix `_chain_0` always stores the state of the cold chain, the file with the suffix `_chain_1` stores the state of the coldest of the heated chains, etc.
- In addition to the three usual checkpoint files (storing the state, generation number, and move statistics of every chain), it also produces a fourth file with the suffix `_heat`, which – as the name implies – stores the heat of a given chain. This ensures we can properly restore even those analyses where the heats are tuned.

Some notes:

- As noted by @jsigao, the issue this PR is intended to solve is rather subtle in the sense that nothing crashes, no error is thrown, and the results might not even look wrong in any way. I have therefore added two lines to the checkpoint-writing code (`Mcmc.cpp`, [line 213](https://github.com/revbayes/revbayes/blob/9a23fd539a4aeb9dac54e057c2ff0aaf0fcc41b0/src/core/analysis/mcmc/Mcmc.cpp#L213)) and the checkpoint resumption code (`Mcmcmc.cpp`, [line 557](https://github.com/revbayes/revbayes/blob/9a23fd539a4aeb9dac54e057c2ff0aaf0fcc41b0/src/core/analysis/mcmc/Mcmcmc.cpp#L557)) to print the heat of a given chain upon saving its state and upon restoring it. They are commented out, but can be uncommented to demonstrate the new code has the intended effect.
- I have also added an integration test based on @ms609's dataset [950_ns_ki](https://github.com/mk-nt/950_ns_ki). It simultaneously tests for the ability to tune heats and resume MC^3 analyses from a checkpoint, so it should cover issues #540, #650, and #651. The test is a little wonky, so the expected output is specific to Linux. I have special-cased the test runner to account for this.


brpetrucci (approved):
Looks good!

src/core/analysis/mcmc/Mcmcmc.cpp
```diff
@@ -239,9 +239,48 @@ void Mcmcmc::checkpoint( void ) const
...
+        if ( chains[ chainForHeatIndex(i) ] != NULL )
         {
-            chains[i]->checkpoint();
+            // get the preliminary checkpoint file
+            path f = chains[ chainForHeatIndex(i) ]->getCheckpointFile();
+            
+            std::string path_string = f.string();
+            std::string test_string = "_chain_";
+            path chain_file_name;
+
```
jsigao, line 250: I wonder how robust is the chain index replacing mechanism to the initial user specified checkpoint file name (i.e., the value provided to the `checkpointFile` argument of the `run` function). For example, if checkpointFile contains '_chain_' or multiple "."s (or no extension at all), would the search-and-replace way of specifying the dynamic chain-specific checkpoint file name still work as intended? I know those are all edge cases, but as we have no control on the user-specified `checkpointFile` argument, it would be nice if the implementation could make fewer assumptions. It seems that things would be easier if we could somehow just store the base file name (without the chain index part appended) and then append that part when looping over the chains here.

src/core/analysis/mcmc/Mcmcmc.cpp
```diff
@@ -476,8 +515,46 @@ void Mcmcmc::initializeSamplerFromCheckpoint( void )
...
+                    std::vector<std::string> key_value;
+                    StringUtilities::stringSplit(line_heat, " = ", key_value);
+                    heat_pars.insert( std::pair<std::string, std::string>(key_value[0], key_value[1]) );
+                }
+                
+            }
+            
+            chain_heat = StringUtilities::asDoubleNumber( heat_pars["heat"] );
+            chains[i]->setChainPosteriorHeat(chain_heat);
+
```
jsigao, line 555: Perhaps we need a `in_file_heat.close();` here?

jsigao (commented):
Thanks @davidcerny for taking care of this issue! It looks all good to me in general. I just left a couple minor comments regarding some potential edge cases. Otherwise it seems great.

bredelings:
I have been thinking about this, and I think maybe a better solution would be to write out the vector of heats.  Then we could restore the checkpoint files like we do now (if I'm understanding correctly), but we would NOT assume that chain 0 is the cold chain.  Of course both ways can work, but this way comes closer to restoring the data structures the same way they were before we checkpointed.  What do you think?

bredelings:
Another thing I have been thinking about is writing out checkpointing information in JSON.  The nlohmann::json library that we are using for JSON makes this pretty easy and idiomatic to do in C++.  We would then get to use the JSON library to do parsing for us.  It also gives us the ability to save/restore complex objects.  And it is human readable/writable as well as machine-readable/writable.

We don't need to write everything in JSON at the start.  But lines like "heat = <number>" could switch to `{"heat": number}`.

With JSON, its pretty easy to write complex objects, so it would be easy to write the vector of heats to JSON and parse it, and in theory, if the info for each chain was a JSON object, we could create a single JSON object that represents the states of all chains, and write that to a single file.  That might be a more long-term goal of course.

src/core/analysis/mcmc/Mcmcmc.cpp
davidcerny: You are totally right on all counts. I've given this some further thought and found an easy way to circumvent the problem, which gets rid of the whole index replacing mechanism and works with arbitrary file names. We simply use the `Mcmcmc::setCheckpointFile()` to set a `base_checkpoint_file_name`, generate chain-specific file names by appending to it, and then use `Mcmc::setCheckpointFile()` to set the checkpoint file names of the individual chains to what we have generated. It works just as well, but it's much more robust, and makes for cleaner and more concise code.

src/core/analysis/mcmc/Mcmcmc.cpp
davidcerny: Indeed; good catch! Fixed.

davidcerny:
@bredelings That's right – if we use the `_heats` files to write out and restore the heats of individual chains anyway, it's no longer necessary to require that chain 0 be the cold one, chain 1 be the coldest heated chain, etc. To me, it just seems... cleaner that way? Otherwise, the indices would be pretty much meaningless. In any case, I think that getting rid of the whole index replacement thing makes this question much less important: it doesn't make much of a difference if we iterate over the chains like this:

```cpp
if ( chains[ chainForHeatIndex(i) ] != NULL )
{
    path chain_file_name = appendToStem( base_checkpoint_file_name, "_chain_" + std::to_string(i) );
    ...
}
```

Or like this:

```cpp
if ( chains[ i ] != NULL )
{
    path chain_file_name = appendToStem( base_checkpoint_file_name, "_chain_" + std::to_string(i) );
    ...
}
```

As for the JSON idea, it sounds promising (though I would say that our current checkpoint files already are both machine- and human-readable), but if it involves an overhaul of the whole checkpointing system, then it's probably outside the scope of this patch.

jsigao:
Looks good to me now!

Regarding whether keeping the chain indices as they were in the previous checkpointed run when restoring from the checkpoint files, the only case occurs to me that this could matter is, when randomly picking two chains to propose a swap, the chains picked could be different from those would have been picked without terminating the previous run. However, this is related to a broader question: do we intend to have the checkpoint-and-restore function behaving like the run has never been terminated and then resumed?

That is, if we initially run an analysis for N generations (either it's completed or terminated prematurely at that generation), and then we resume the analysis from the checkpoint files for running one more generation, do we care if the sample of that generation is different from what would have been sampled at the N+1th generation if the analysis had been kept running without terminating at the Nth generation?

It's hard to think about a real case where this would actually matter, so perhaps we would not want to make any effort to achieve that behavior. At any rate, that behavior likely has not been achieved by the current checkpointing system as the status of the random number generator is not checkpointed. (Therefore, even if the resumed analysis runs under the same seed as the initial analysis, the random number generated at a given generation during the resumed run would be different from the one it would have been generated if the analysis had been kept running.)

Anyway, I think that's certainly beyond the scope of this PR, so I'm approving it now. Thanks again @davidcerny !

bredelings:
> However, this is related to a broader question: do we intend to have the checkpoint-and-restore function behaving like the run has never been terminated and then resumed?

I would say "yes". 

It may not be possible to fully achieve that today, but I think we should stay as close as possible, and not get further away.

If I understand the PR correctly, checkpointing-and-restoring will have the effect of moving the cold chain to index 0.  So it is not a no-op.

As far as JSON, I think the only  JSON changes that would be in scope for this PR would be to make the _new_ information use JSON.  That would just be the heats if I understand correctly.  However, if you don't want to deal with that, we could do it later.

davidcerny:
@jsigao Thank you! I was going to test the impact of re-indexing the chains by conducting the following analyses, all with the same random seed:

1. Run an uninterrupted analysis for 2000 generations;
2. Use the current code (with re-indexing) to run an analysis for 1000 generations, save a checkpoint, restore, and run for another 1000 generations;
3. Drop re-indexing from the code, and proceed as in case (2).

I wanted to see if (1) and (3) would potentially yield identical log files, but I found that we failed to correctly append to the log files in the first place. You can actually see that in the log files I included as the expected output of the integration test added by this PR – it seems to be the same issue I described in #525.

It's not clear to me whether it's a MC^3 bug or a monitor bug (it doesn't affect `mnScreen`, so perhaps the latter), but until it is fixed, MC^3 checkpointing can't really be said to be fixed. I think I'm going to convert this PR to a draft.

davidcerny:
Thanks to @bredelings and his fix for #525, I have now been able to run the experiment I described in my previous comment. As Ben and @jsigao both suspected, even if we prevent the chains from being re-indexed upon resuming an analysis from a checkpoint, this is not sufficient to make the resulting output identical to that of an analysis that has never been interrupted in the first place – checkpointing the state of the random number generator would probably be needed to achieve that. However, after thinking about the issue further and considering Jiansi's argument about picking chains to be swapped, I now agree that

> It may not be possible to fully achieve that today, but I think we should stay as close as possible, and not get further away.

I have therefore made the corresponding change. With this, I think the PR can be re-reviewed if needed, and hopefully merged.

Issue #665, opened 2025-02-10
[BUG] Power posterior log progress output stops at step 3


ms609:
**Describe the bug**
When running a power posterior analysis, only the first three steps appear in the output log.

**To Reproduce**
Execute `marginal.Rev` in https://github.com/neo-trans/692_by_nt_ki

```rb
burninGen = 20000
runGen = 1200
steps = 51

ppFile = "692_by_nt_ki.pp"

neo <- readDiscreteCharacterData("project692.neo.nex")
trans <- readDiscreteCharacterData("project692.trans.nex")
source("by_nt_ki.Rev") # Model setup

power_p = powerPosterior(mymodel, monitors, moves, ppFile, cats = steps)
power_p.burnin(generations = burninGen, tuningInterval = 1000)
print("  - Commencing powerPosterior run at", time("year"), time("day"),
      time("seconds"))
power_p.run(generations = runGen)
print("  - Completed powerPosterior run at", time("year"), time("day"),
      time("seconds"))
```

**Expected behavior**

I see:
![Image](https://github.com/user-attachments/assets/4bb46e48-d216-4c94-9a40-f94d7fe8dbdf)

I expect to see the subsequent steps (`4/52`, ...) included in the log.

I encounter this behaviour in a range of analyses, with output always stopping at step 3.

**Computer info**
Linux HPC, recent development build

davidcerny:
I guess you ran this analysis on 16 cores? When I re-run it on 8 cores, I see 6 steps; and on 4 cores, 13 steps. So it looks like RevBayes reports steps from 1 to `floor((# of power posteriors) / (# of processes))`.

At a more granular level, what's actually happening is that each step corresponds to the completion of `(# of processes)` power posteriors. E.g., in the 8-core case:

```
                  |                      Power posteriors completed
                  |----------------------------------------------------------------------
Step              | core 1 | core 2 | core 3 | core 4 | core 5 | core 6 | core 7 | core 8
------------------|--------|--------|--------|--------|--------|--------|--------|-------
1                 |    0   |    6   |   13   |   19   |   26   |   32   |   39   |   45
2                 |    1   |    7   |   14   |   20   |   27   |   33   |   40   |   46
3                 |    2   |    8   |   15   |   21   |   28   |   34   |   41   |   47
4                 |    3   |    9   |   16   |   22   |   29   |   35   |   42   |   48
5                 |    4   |   10   |   17   |   23   |   30   |   36   |   43   |   49
[...]

PR #673, opened 2025-02-19
Better documentation and progress monitoring for power posterior analyses

davidcerny:
This PR fixes the cosmetic, but still annoying issue #665. Currently, this is how we report the progress of a power posterior analysis with 50 stones on 8 cores:

```
Running power posterior analysis ...
Step  1 / 50		*********************************
Step  2 / 50		*********************************
Step  3 / 50		*********************************
Step  4 / 50		*********************************
Step  5 / 50		*********************************
Step  6 / 50		*********************************
```

This only really works for single-core analyses, where (# of steps) = (# of stones). Note that there are just 6 steps, even though two of the 8 cores actually have 7 consecutive stones assigned to them (so there are 7 rounds of sampling involved in total). This PR changes the above to the following:

```
Running power posterior analysis ...
Steps  1--8  / 50		*********************************
Steps  9--16 / 50		*********************************
Steps 17--24 / 50		*********************************
Steps 25--32 / 50		*********************************
Steps 33--40 / 50		*********************************
Steps 41--48 / 50		*********************************
Steps 49--50 / 50		*********************************
```

I've tested this for various combinations of stone and core numbers, and it's always worked as intended.


davidcerny:
Actually, I did fail to account for an edge case: when (# of stones) %% (# of processes) = 1, we don't want the final line to be something like `Steps 49--49 / 49`, but just `Step     49 / 49`. This is now fixed.

ms609 (commented):
Looks promising, and is running as expected on single and multicore test 
runs.

Intriguingly, if I ask for 128 steps, the output reports steps n/12**9**, both in the single-core and multi-core instances.  Lower values (e.g. 51 steps) are reported as expected (i.e. n/51).  And the expected number of `.pp` files are created. 

This issue occurred before this patch – I'd noticed it on development too – so possibly it deserves a separate issue – but this might be a natural place to address it.

src/core/analysis/PowerPosteriorAnalysis.cpp
```diff
@@ -239,18 +242,73 @@ void PowerPosteriorAnalysis::runStone(size_t idx, size_t gen, double burnin_frac
...
+                {
+                    std::cout << " ";
+                }
+                std::cout << lower_bound;
+            }
         }
-        std::cout << (idx+1) << " / " << powers.size();
-        std::cout << "\t\t";
+    
+        std::cout << " / " << powers.size() << "\t\t";
```
ms609, line 311: The off-by-one error could be fixed with:
```suggestion
        std::cout << " / " << (powers.size() - 1) << "\t\t";
```

Does `powers.size()` also need a `- 1` appending where it is used above?

ms609: On closer inspection, I see that 129 steps _are_ in fact being performed – I missed the file ending `_stone_0.pp`. 
Perhaps the off by one error is occurring where `powers` is defined?

davidcerny: @ms609 That's right – this is not a problem with reporting the progress of the analysis, but with setting it up in the first place. What goes into the `cats` argument of the `powerPosterior()` argument is actually not the number of stones / powers, but the largest index using 0-based indexing. So specifying `cats=127` produces stones indexed 0 through 127, i.e., 128 stones in total.

I've always found this strange, because unlike C++, Rev uses 1-based indexing, and even if that were not the case, it would still make more sense to specify the actual number rather than the largest index. It would be trivial to fix this, but I'm wary of making the change unilaterally, because it would silently change the behavior of existing scripts, potentially including published ones. Feedback by @bredelings @mlandis @hoehna would be much appreciated.

ms609: I hear the hesitation about changing behaviour (and potentially changing - slightly - existing scripts).  Such a change would, however, bring the behaviour of the software in line with the [documentation](https://revbayes.github.io/documentation/powerPosterior.html), which reads (for `cats`): 
> The number of categories if no powers are specified

davidcerny:
@ms609 As a compromise solution, I've kept the current behavior but changed the description of the `cats` to make it less misleading. I have also added a fairly detailed help page for `powerPosterior()` that further explains exactly how the settings relate to the analysis output. Let me know what you think!

ms609:
The detailed documentation is a welcome addition – thank you!

I don't find the "count from zero" `cats` formulation particularly intuitive.  I do see the trade-off between backward compatibility and future usability, but given other breaking changes that have been made in the last update, I'm not sure that backward compatibility need necessarily trump user-friendliness here – particularly as the behaviour will only have a modest effect on previous scripts (i.e. it won't break anything, just subtly alter the precision of power posteriors – which are in any case not the most precise of tools).  So my preference would be to modify the behaviour of `cats` to match what was originally documented.  Indeed, it may be that doing so makes previous scripts work as their authors intended!

Another approach would be to deprecate the `cats` parameter (whilst leaving it in the codebase to support previous scripts) and replace it with a `ncats` parameter with the more obvious interpretation of "number of categories" with the "count from 1" implementation.

davidcerny:
I like the idea of having two alternative arguments of which one is deprecated, but it seems like our approach to deprecation is to throw an error message that either instructs the user what to do instead (e.g., for `underPrior=true`) or explains why a particular feature is no longer available (e.g., for `dnFBDSP`). So doing this would not ensure continued support for pre-existing scripts. Even if we only threw a warning rather than an error, we'd still have to think about how far we're willing to go to maintain two distinct ways of doing things. For example, do we want `cats=127` to produce files named `*_stone_0.out`, ..., `*_stone_127.out`, and `ncats=128` to produce files named `*_stone_1.out`, ..., `*_stone_128.out`?

Though I also found the 0-based indexing counter-intuitive at first, I changed my mind somewhat when working on the help page – it makes sense when interpreted as the number of stones to run _in addition to_ the posterior, and in the context of calculating the beta values, where we want to iterate from _n_ to 0, not from _n_ to 1. (The latter would cause us to neglect the final stone corresponding to the prior.) That said, I'm not opposed to making the change; I would just like to proceed on the basis of a broader consensus rather than the preferences of one or two developers. @bjoelle @bredelings @brpetrucci @mikeryanmay, would you care to chime in?

davidcerny:
Since there doesn't seem to be a deeper reason behind the decision to use 0-based indexing (as per communication with @mlandis and @hoehna), and since @ms609 and @bjoelle are in favor of making the change, I have now done so. From now on, specifying (for example) `cats=128` sets up the requested number of analyses, with output files labeled *_stone_1.out`, ..., `*_stone_128.out`. As before, the first of these contains samples the posterior, and the last contains samples from the prior. The help page, description of the `cats` argument, and integration tests have all been changed accordingly.

ms609 (approved):
Without scrutinizing every line in detail, this looks great, and the improvements to the doc are welcome too.

ms609:
I notice that the `runOneStone()` method isn't currently listed as a method at https://revbayes.github.io/documentation/powerPosterior.html; does it need registering?

davidcerny:
@ms609 That just seems to be an instance of the website being badly out of date, because calling `.methods()` on a `powerPosterior` object _will_ show `.runOneStone()` as one of the available methods. My understanding is that individual methods don't have to be separately registered; the relevant list is retrieved automatically.

Incidentally, the many additions [you made](https://github.com/revbayes/revbayes/pull/506/) to the `mcmc` documentation some time ago don't show up on the website, either, which goes to show that the relevant portions haven't been rebuilt in a long while. I'll bring this up during the hackathon – unless you want to do it yourself!

Issue #563, opened 2024-10-10
[BUG] readPoMoCountFile does not load data


seidels:
**Bug description**
I am unable to load data from a count file using the readPoMoCountFile function in RevBayes v1.2.4. The function does not raise an error, but the data object is not created. The issue occurs with both the [provided tutorial data ](https://revbayes.github.io/tutorials/pomos/) and a custom count file.

**To Reproduce**
I have been following the PoMo tutorial and used the following command:

`data <- readPoMoCountFile(file="data/great_apes_1000.cf", virtualPopulationSize=12, numStates=4)
`

Console output:
_POSITION: 41275799
State: 6,0,0,0
stringvalue: 6,0,0,0
1 1e-08 1e-08 1e-08 3.34898e-07 2.14335e-05 0.000244141 0.00137174 ...
std::bad_alloc_

Even though the function does not call an error - when I check the data object, it seems not to have been assigned.
`data`
Console output:
_Missing Variable: Variable data does not exist_

To make sure that this is not a problem of the provided count file, I created my own little count file for testing, that is much shorter [my_toy_example.txt](https://github.com/user-attachments/files/17324874/my_toy_example.txt):

`data <- readPoMoCountFile(file="data/my_toy_example.txt", virtualPopulationSize=3, numStates=4)
`
Again, the data object is not created:

_POSITION: 1
State: 6,0,0,0
stringvalue: 6,0,0,0
1 1e-08 1e-08 1e-08 0.00137174 0.0877915 0.00137174 0.0877915 0.00137174 0.0877915 1e-08 1e-08 1e-08 1e-08 1e-08 1e-08  
   std::bad_alloc_

`data`
 _Missing Variable: Variable data does not exist_

To verify that this isn't an installation issue, I tested loading data using the readDiscreteCharacterData function in the [substitution model tutorial](https://revbayes.github.io/tutorials/ctmc/), and it worked as expected:
`data <- readDiscreteCharacterData("data/primates_and_galeopterus_cytb.nex")
`
_Successfully read one character matrix from file 'data/primates_and_galeopterus_cytb.nex'_ 

**Computer info**
I am using MacOS Sonoma, version 14.6.1 and revBayes v.1.2.4.

seidels:
That works! Following the tutorial, I get another error when calling the PoMo substitution model function:
```
# rate matrix
Q := fnReversiblePoMo4N(N,pi,rho,phi)
```
Error:	No function named 'fnReversiblePoMo4N'

[...]

PR #674, opened 2025-02-20
Revamped tutorial testing

brpetrucci:
Changed `tests/run_integration_tests.sh` to make tutorial tests more flexible. Originally, one had to write a `test.sh` script for each tutorial (and there were scripts written for the fbd, dating, and ctmc tutorial only), which only allowed for one script per test as it was written. To make this more flexible, we changed the integration tests script so that it runs each script in a file `tests.txt` in a tutorial directory. In this way, we can ensure tutorial developers can specify which scripts must be tested, so that scripts that should not be run (e.g. when they are just written to set up variables for a master script) will not be accidentally added to the tests. To test this, I also reviewed the fbd, dating and ctmc tutorials to make sure the html and scripts matched, deleted `test.sh`, and added `tests.txt` (all on the website side, see the website repo). I updated the `revbayes.github.io` submodule in the `tests` directory to make sure this was working.

### DIFF
--- a/tests/run_integration_tests.sh
+++ b/tests/run_integration_tests.sh
@@ -52,30 +52,40 @@ fi
 tests=()
 status=()
 
-for t in revbayes.github.io/tutorials/*/test.sh; do
-    testname=`echo $t | cut -d '/' -f 2-3`
-    dirname=`echo $t | cut -d '/' -f 1-3`
-    
-    cd $dirname
+for t in revbayes.github.io/tutorials/*/tests.txt; do
+    testname=`echo $t | cut -d '/' -f 2-3`;
+    dirname=`echo $t | cut -d '/' -f 1-3`; 
+	
 
+    cd $dirname
     tests+=($testname)
 
     printf "\n\n#### Running test: $testname\n\n"
-    sh test.sh
-    res="$?"
-    if [ $res = 1 ]; then
-        res="error: $f"
-        break
-    elif [ $res = 139 ]; then
-        res="segfault: $f"
-        break
-    elif [ $res != 0 ]; then
-        res="error $res: $f"
-        break
-    fi
-    if [ $res != 0 ] ; then
-        echo "${dirname}/test.sh ==> error $res"
-    fi
+    
+	for script in $(cat tests.txt); 
+	do
+		(
+		cd scripts
+		sed 's/generations=[0-9]*/generations=1/g' "$script" > "cp_$script"
+		)
+        ${rb_exec} -b scripts/cp_$script
+    	res="$?"
+    	if [ $res = 1 ]; then
+        	res="error: $f"
+        	break
+    	elif [ $res = 139 ]; then
+        	res="segfault: $f"
+        	break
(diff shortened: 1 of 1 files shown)


tests/run_integration_tests.sh
```diff
@@ -52,30 +52,40 @@ fi
...
-        res="error $res: $f"
-        break
-    fi
-    if [ $res != 0 ] ; then
-        echo "${dirname}/test.sh ==> error $res"
-    fi
+    
+	for script in $(cat tests.txt); 
+	do
+		(
```
bredelings, line 67: The indentation for this section seems messed up.  Could it be tabs versus spaces again?

brpetrucci: Fixed!

davidcerny:
That sounds like a useful simplification, thanks! Just a quick comment about this:

> I also reviewed the fbd, dating and ctmc tutorials to make sure the html and scripts matched

Going forward, this might not always be the case. E.g., issue #563 stemmed from the fact that the code in the markdown diverged from the code in the scripts – the latter ran fine but the former did not. It might be good to be able to match the two in some automated way. @bjoelle pointed out that the FBD tutorials already had a system for doing this by pulling the code from a script into the markdown, but (1) it's not clear to me what the system is, and (2) it sounds like we might more often want to go in the opposite direction (i.e., make a script from the markdown).

brpetrucci:
Yes! I agree we need a more permanent solution. The next thing in my POSE todo list is to delve deeper into the tutorials that pull code directly from the script, and modify the three tutorials I listed in this pull request to do so. I'm not sure I agree we'd want to pull code from the markdown--as long as we're sure that code works (i.e. by including the scripts into the integration tests), pulling from the scripts to the markdown sounds like a good solution. Maybe the opposite could also work, but I don't see why it'd be better, the end result is still the same since the scripts are the units being tested.

davidcerny:
It's just that people sometimes seem to update the markdown without updating the attached scripts, plus the users might primarily follow the step-by-step guidelines in the html rather than download the scripts and directly edit those. In the long term, we definitely want to ensure that the two always match, in which case the direction doesn't matter, but when we run into a discrepancy, the markdown should probably be the template for editing the scripts, rather than _vice versa_.

brpetrucci:
Well, I'm generally against having code in the markdown that is not in scripts, because that code cannot be tested. E.g. I added one script to the ctmc and two to the dating tutorial because these scripts were mentioned in the markdown but were not available anywhere to be tested. 

I agree that the intention is for the user to follow the markdown instead of downloading the script, but if the markdown and script agree I don't see why the direction matters specifically, as you said. 

About your last point, I guess I agree, but either way in an ideal future where all tutorials pull code directly from scripts that are tested every time the revbayes and website repos are updated, where the code was originally pulled from won't matter too much. I tend to prefer the markdown when making sure the markdown and scripts match, but if for a given mismatch the script is better (e.g. the script puts higher weight into a move we know has trouble mixing), I don't think it's a problem to change the markdown.

davidcerny (approved):

PR #676, opened 2025-02-25
Ensuring integration tests can run tutorial checkpoint tests

brpetrucci:
Minor change to `tests/run_integration_tests.sh` to ensure that tutorial tests using checkpointing (e.g. `ctmc`) can be run without a hitch. It simply changes the `checkpointInterval` argument (if there is one) to 1 in the copy script used for the test run. Also updated git submodule.

### DIFF
--- a/tests/run_integration_tests.sh
+++ b/tests/run_integration_tests.sh
@@ -67,6 +67,7 @@ for t in revbayes.github.io/tutorials/*/tests.txt; do
         (
         cd scripts
         sed 's/generations=[0-9]*/generations=1/g' "$script" > "cp_$script"
+        sed -i '' 's/checkpointInterval=[0-9]*/checkpointInterval=1/g' "cp_$script"
         )
         ${rb_exec} -b scripts/cp_$script
         res="$?"
--- a/tests/run_integration_tests.sh
+++ b/tests/run_integration_tests.sh
@@ -66,8 +66,7 @@ for t in revbayes.github.io/tutorials/*/tests.txt; do
     do
         (
         cd scripts
-        sed 's/generations=[0-9]*/generations=1/g' "$script" > "cp_$script"
-        sed -i '' 's/checkpointInterval=[0-9]*/checkpointInterval=1/g' "cp_$script"
+        sed 's/generations=[0-9]*/generations=1/g' "$script" | sed 's/checkpointInterval=[0-9]*/checkpointInterval=1/g'  > "cp_$script"
         )
         ${rb_exec} -b scripts/cp_$script
         res="$?"


davidcerny:
I don't think I quite understand what this PR does, and the motivation behind it. What was wrong with the original scripts? Are you saying that we already maintain copies of tutorial scripts that are specifically intended for testing? If so, why not just change those directly?

brpetrucci:
`run_integration_tests.sh` has always altered the number of generations before running those scripts for testing, it was just a task relegated to the `test.sh` script in each tutorial directory. We changed the tutorial testing structure ([PR 47 in the website repo](https://github.com/revbayes/revbayes.github.io/pull/47)) so that now `run_integration_tests.sh` does that directly. The reason it must be done there and not just have the script itself have `generations` or `checkpointInterval` equal to 1 is 1. We want the scripts to be an analysis that can be run as-is, so we want to "recommend" a reasonable number of generations, and 2. We are now making it a goal to have the tutorials pull code directly from scripts, so that we can catch tutorial errors if a certain rb update breaks one of the tutorial scripts, and don't want the html to show `generations=1`. So it's just a simple way to run the tests for just 1 gen (to make sure it runs), making sure the code runs, and then the copy files are deleted right after.

davidcerny:
Makes sense; thanks for the explanation! My only concern is that running a script for just 1 generation is a very permissive test, so much so that it might not be super useful. A lot of our integration tests are meant to catch issues that were reported to only kick in after a certain number of generations. With checkpointing, this would seem to be especially important – what exactly are we trying to test for?

1. The ability to _write_ a checkpoint file?
2. The ability to _resume_ from a checkpoint file?
3. The ability to resume _properly_ (e.g., by making sure that the post-resumption parameter traces are not flat)?

Testing for (1) without (2) seems meaningless, and ideally we'd want to get at (3).

bredelings:
Overall, this looks nice.  Thanks!

In mcmc_GTR_gamma_Inv.Rev, it looks like we now have two `seq ~ dnPhyloCTMC( )` lines, one with `pInv` and one without.  Was that intentional or maybe a mistake?

Hmm... why do we bother changing the checkpointing interval at all?  My impression was that the tutorial tests just check that the commands ran without error, and did not actually assess checkpointing.  Does the mcmc command complain if the checkpointing interval is larger than the number of generations?

brpetrucci:
David: 

(3) is what we're testing, so in `revbayes.github.io/tutorials/ctmc/scripts/asides.Rev`, we run the mcmc with checkpointing, then initialize from checkpoint, then run again. We could run for more generations, but it's already the slowest test of the 3 tutorials currently being tested by `run_integration_tests.sh`, and I'm not sure how many generations we could do (plus, since the standard is to run every other test for 1 generation, I'm not sure how to exclude just the checkpointing tests from that). There's a limit on how much we can test with tutorials--I believe the soundest philosophy is to just make sure all the code in the tutorials run, and test deeper specific RevBayes features separately with integration tests.

Ben:

It's intentional--the html has a section with `pInv` and one without, so we need a line for each.

We have a line in `asides.Rev` where we initialize from checkpoint. If the checkpoint interval is higher than the number of generations we run, a `.state` file will not be created and the initialize command will fail.

bredelings (approved):

PR #678, opened 2025-02-28
Don't treat deterministic nodes as constants

bredelings:
In some cases we have to write
```
tmp := f(x)
y := g(tmp)
```
instead of the more natural
```
y := g(f(x))
```
Apparently this is because we sometimes convert deterministic nodes into constants even when they contain variables.  

This change reverts that behavior and replaces deterministic nodes by constants ONLY when they contain to variables.


bjoelle:
so if I understand correctly, this change is needed because RB cannot correctly identify whether a function call is nested and so cannot set `dynamic = true` in that situation, is that correct ?
it might be useful to update the function doc and add that info

bredelings:
Not exactly.
- we already create constant nodes if the expression has no variables, so I'm not really sure we need another check?
- if there are other situations where it is OK to create a constant node, it seems at best an optimization.  But its not clear that the optimization is needed.  And if we make a mistake, then we get incorrect behavior.  So it seems safer to not do that.
- I didn't see any special logic to determine whether or not to set the `dynamic` flag for `SyntaxFunctionCall`.  Instead it seems like the argument has a default value of `false`, and lots of calls to `SyntaxFunctionCall::evaluateContent( )` just don't set the value.
- Arguments to method calls default to being turned into constants.  BUT you can make them deterministic by doing `tmp := f(x); y = object.g(tmp)`.  This is actually the use case I am interested in fixing.
- I looked into the default flag a bit further, and it seems that all the syntax elements receive a `dynamic` flag, but it doesn't have a fixed meaning.  For example, `SyntaxVariable` interprets the `dynamic` flag to mean that after looking up a variable we should not return that variable, but a new variable with the same value:
```
 * If the parameter dynamic == true, then we need a dynamic evaluation.
 * The method behaves the same except that control variables need to return
 * clones of themselves (temporary variables) rather than themselves,
 * so that they are not included in the DAG.
*/
...

    // get a copy if this is a workspace variable
    if ( dynamic == true )
    {
        // Check whether we have a control variable and make a clone in that case
        if ( the_var->isWorkspaceVariable() )
        {
            the_var = new RevVariable( the_var->getRevObject().clone() );
            the_var->setWorkspaceVariableState( true );
        }
        
    }
```

I'm not sure if this should go into a comment, and if so, how much of the information should go into the comment.

bjoelle:
ok this is way more complex than I thought... I agree that the code comments are not the right place to put this, maybe we can just leave it here then

bjoelle (approved):

bredelings:
Given the complexity I think I'm going to do a little more thinking on this...

PR #684, opened 2025-03-07
Fix broken overrides part1

bredelings:
GCC 15 has some warnings that identified functions that were supposed to be virtual overrides, but were not because of slightly different const-ness.

This corrects the first batch, which could be fixed without re-architecting other stuff.


bjoelle (approved):
we should probably add the override keyword as well

PR #687, opened 2025-03-08
Use std::shared_ptr<Environment> to hold Environments.

bredelings:
This fixes a crash with user functions by ensuring that we don't delete Environments that are still in use.  We replace `Environment*` references with `std::shared_ptr<Environment>`, which is a flexible reference-counted pointer that destroys its contents only when the last reference is removed.

When executing a `UserFunction` that is a deterministic node, we create a copy of the `UserFunction` and its `Environment` so that we can re-evaluate the function later.  Previously the  `UserFunction` referenced the `Environment` using a raw pointer.  However, the `Environment` that it pointed to could be deleted.  Then when we called `UserFunction::update()`, the program would crash in strange ways.  Using `std::shared_ptr<Environment>` ensures that the `UserFunction`'s Environment is not deleted until the `UserFunction` node is deleted.

This is implemented by rather tediously replacing some `Environment*` or `Environment&` method arguments with `const std::shared_ptr<Environment>&` in revlanguage/parser/Syntax*.{h,cpp} files.  Then we have to ensure that we have shared_ptrs to pass to those methods, which is also rather tedious.  The result touches a lot of files, but is pretty much completely mechanical.

We have to avoid reference cycles with `std::shared_ptr`.  There are two kinds of references.  
1. An Environment contains references to "children" that represent namespaces.  These namespaces do not need to have a pointer to the parent Environment, so that is OK.

2. In contrast, when we call a function, we create an execution Environment that only contains local variables, and uses a pointer to a parent Environment to handle references to external variables.  However, in this case the parent Environment has no reference to its descendant, so this does not cause a reference loop.

### DIFF
--- a/src/revlanguage/workspace/Environment.h
+++ b/src/revlanguage/workspace/Environment.h
@@ -32,7 +32,7 @@ class RevObject;
      *     Environment will not be destroyed before the stack frame Environment.
      *
      * (b) Environment objects that represent a namespace do not contain a pointer to the parent
-     *     Environment.  Instead, the parent Environment holds a references to the namespace.
+     *     Environment.  Instead, the parent Environment holds a reference to the namespace.
      *     This ensures that the namespace Environment will not be destroyed before the
      *     enclosing Environment.
      *
--- a/src/revlanguage/workspace/Environment.h
+++ b/src/revlanguage/workspace/Environment.h
@@ -26,15 +26,15 @@ class RevObject;
      * evaluation or execution environment, such as the stack frame of a function call
      * or a namespace.
      *
-     * (a) Environment's that represent stack frames only define local variables and
+     * (a) Environment objects that represent stack frames only define local variables and
      *     contain a pointer to the enclosing (parent) environment to look up external variables.
-     *     Such environments contains a reference to the parent environment so that the parent
-     *     environment will not be destroyed before the stack frame environment.
+     *     Such Environments contain a reference to the parent Environment so that the parent
+     *     Environment will not be destroyed before the stack frame Environment.
      *
-     * (b) Environment's that represent a namespace do not contain a pointer to the parent
-     *     Environment.  Instead, the parent environment holds a references to the namespace.
-     *     This ensures that the namespace environment will not be destroyed before the
-     *     enclosing environment.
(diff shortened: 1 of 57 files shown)


davidcerny:
Can we add a test for this, based on a case that was previously failing?

bredelings:
Well, kind of.  What happens is that `test_basics/scripts/basics.Rev` fails with the `unnamed-deterministic-nodes` branch (PR #678) unless this fix is included.  I ended up submitting it separately because it is so much larger than that other branch.

src/revlanguage/workspace/Environment.h
```diff
@@ -23,9 +23,19 @@ class RevObject;
...
+     *
+     * (a) Environment's that represent stack frames only define local variables and
+     *     contain a pointer to the enclosing (parent) environment to look up external variables.
+     *     Such environments contains a reference to the parent environment so that the parent
+     *     environment will not be destroyed before the stack frame environment.
+     *
+     * (b) Environment's that represent a namespace do not contain a pointer to the parent
+     *     Environment.  Instead, the parent environment holds a references to the namespace.
+     *     This ensures that the namespace environment will not be destroyed before the
+     *     enclosing environment.
```
davidcerny, line 37: Some typos here... "Environment's" should read "Environments" (lines 29, 34), "environments contains" should read "environments contain" (line 31), the word "Environment" seems to be randomly capitalized (line 35), and "a references" should read "a reference" (line 35).

That said, thank you for explaining the logic behind this! I feel like the finer nuances of Rev scoping always escaped me, so documenting how it's implemented helps a lot.

bredelings: I fixed some typos.

You can read "Environment" as "an instance of the Environment class" and "environment" as "a mapping from names to variables".

davidcerny: Thanks! At the risk of being super annoying, the phrase "a references" still occurs on line 35. I believe it should be singular (one reference per namespace).

bredelings: No worries, that's easy enough to fix.

bredelings: Fixed.

PR #688, opened 2025-03-10
Fix rate age beta shift

bredelings:
Fix `mvRateAgeBetaShift`.  The posterior ratio wasn't computed  correctly, so this move would accept proposals that have a posterior of `-inf`.


davidcerny:
For the sake of consistency with other integration tests, can you please add an `.errout` file to the expected output?

bredelings:
OK.  Done.

mikeryanmay:
I pushed a few minor changes to comments in RateAgeBetaShift. Everything looks good to me.

I also ran some longer chains (based on the FBD_dating_ex3.Rev script) with and without the RateAgeBetaShift move to confirm the posterior is the same, basically to check whether the Jacobian/etc. was correct for this proposal. Seems like it's all working as intended.

davidcerny:
I'd like to try including the move in an FBD analysis with fixed fossil ages, just to make sure it doesn't silently change those when there is no explicit likelihood penalty for doing so. Hoping to get around to it by the end of the day!

davidcerny (approved):
I have now included the move in FBD analyses similar to those I described in my reviews of #594. The ages of fossil taxa are never set to incorrect values, regardless of whether they are treated as fixed or sampled from uniform priors, and regardless of whether the taxa in question represent tips or sampled ancestors. Great work!

PR #689, opened 2025-03-11
New tests for the revised coalescent classes

hoehna:
I reduced and modified the tests from the tutorial about the demographic inference using coalescent models. This now performs the integration tests to make sure these keep working.


bredelings:
It looks like the output files are different on Mac-ARM and Mac-Intel :-(

bredelings:
Some other tests give different results on Linux and Mac.  One approach is to only check the results on one platform, and on the other platform to check that it runs and does not crash.  We have been checking on Linux, since you don't need special hardware to access it.  You can also download the linux output files from the bottom of the test page if the linux tests fail.  I see that the Mac-Intel outputfiles are downloadable right now.

Previously, only a few tests (mostly FBD) were different on different architectures though.

@davidcerny and @bjoelle may know more about this.

davidcerny:
That's exactly right – when encountering this problem before, my approach was to take the output from the GitHub Actions artifacts, make it the new expected output, and special-case the test runner so that it wouldn't attempt to run the test on macOS+ARM.

bredelings (approved):

PR #691, opened 2025-03-14
DRAFT PR for merging dev_traitfig

bredelings:
DRAFT PR for merging dev_traitfig

I created this mostly so that it would be clearer what files are modified.

One issue with this branch is that it introduces 3 new functions and 1 new distribution, but doesn't create any new help files.


bredelings:
This branch is not read for merging. (@mlandis)

PR #696, opened 2025-03-16
Fix state space when subsetting by state space

hoehna:
This should be resetting state labels when setting partition by state size. Thus, it fixes the issue that the state space still looks larger.


davidcerny:
Not sure I understand what problem this is intended to solve – could you give an example of a case that's currently failing @hoehna?

basanta33 (approved):

hoehna:
[partition_state_space.zip](https://github.com/user-attachments/files/19274883/partition_state_space.zip)

Attached is an example that was crashing and according to @basanta33 is now working.

bredelings:
But no tests were added to make sure that it _stays_ working.

Can you perhaps give other people a little bit more time to look at these changes in the future?  2.5 hours does not seem like much time.

PR #700, opened 2025-03-17
Clarify `write()` documentation

ms609:
The main motivation of this PR is to make it clear that the default value for `separator` is `\t`: currently this is printed as whitespace:

image

I've also taken the opportunity to clarify some of the parameter descriptions.


bredelings:
It seems like we should be print a quoted version of the default value.  So if the separator was `,`, we should print `","`.  We could also escape the string, so that if the separate is tab, we would print `"\t"`.

davidcerny:
So... this now passes the tests, but is it still worth merging? I don't see a world of difference between the new and old argument descriptions.

bredelings:
Maybe not.  I could go either way. :shrug:

PR #705, opened 2025-03-17
Free binary

sigibrock:
updated the fnFreeBinary function. Unsure if I translated the fnFreeK example properly. Please advise!


davidcerny:
What happened to the commit history here?

PR #706, opened 2025-03-17
updated fnFreeBinary help file

sigibrock:
Updated the fnFreeBinary function. Unsure if I translated the fnFreeK example properly. Please advise!


help/md/fnFreeBinary.md
```diff
@@ -1,9 +1,42 @@
...
+ 
+ It takes in two arguments: 
+    (1) transtion_rates (tr) - A real number that represents the rate of transition between states.
+    (2) rescaled - A boolean vlue that indicates whether or not the matrix should be normalized. Takes on TRUE by default.
+
+0 and 1 represent our 2 states:
+Q = [[-q_{01}, q_{01}],[q_{10}, -q_{10}]]
+
 ## authors
+Sigournie Brock
```
bredelings, line 21: I think this is supposed to be the person who wrote the fnFreeBinary function?

help/md/fnFreeBinary.md
```diff
@@ -1,9 +1,42 @@
...
+NUM_STATES = 2
+NUM_RATES = NUM_STATES * (NUM_STATES-1)
+for ( i in 1:NUM_RATES ) {
+    rate[i] := mu
+}
+
+Q := fnFreeBinary(rate, rescaled=false)
+
 ## references
+[text](https://revbayes.github.io/tutorials/morph_ase/ase.html)
```
bredelings, line 42: There is a specific format for references.  Generally they only reference papers, and use YAML format.
Maybe we should reconsider...

help/md/fnFreeBinary.md
davidcerny, line 21: I think the convention we are using is to list the authors of the code, rather the authors of the help file. Often, it's not clear who wrote the code, in which case it's okay to leave the entry empty.

help/md/fnFreeBinary.md
```diff
@@ -1,9 +1,42 @@
...
+rate_pr := phylogeny.treeLength() / 10
+mu ~ dnExp(rate_pr)
+
+moves.append( mvScale(mu, lambda=1, weight=2.0) )
+
+NUM_STATES = 2
+NUM_RATES = NUM_STATES * (NUM_STATES-1)
+for ( i in 1:NUM_RATES ) {
+    rate[i] := mu
+}
```
davidcerny, line 37: This will work just fine, but it's perhaps unnecessarily complicated? Instead of these 9 lines, we could just write something like:

```
rate[1] ~ dnExp(rate_pr)
rate[2] ~ dnExp(rate_pr)
```

davidcerny (changes requested):

help/md/fnFreeBinary.md
davidcerny, line 37: This will work just fine, but it's perhaps unnecessarily complicated? Instead of these 5 lines, we could just write:

```
rate := rep(mu, 2)
```

davidcerny:
Very nice work! My last comment would be that it may be worth telling the user what happens if `transition_rates` is of length 1 (the single element will get recycled), or of length greater than 2 (everything after the first two elements will be ignored). But I don't insist on this.

davidcerny (approved):

PR #707, opened 2025-03-18
Adding documentation to 'fnDiscretizeDistribution' and 'fnDiscretizeG…

basanta33:
Added documentation to the function to discretize distributions.

### DIFF
--- a/help/md/fnDiscretizeDistribution.md
+++ b/help/md/fnDiscretizeDistribution.md
@@ -1,9 +1,27 @@
 ## name
 fnDiscretizeDistribution
 ## title
+Discretize a Continuous Distribution
 ## description
+`fnDiscretizeDistribution` transforms a continuous probability distribution into a discrete one by dividing it into a specified number of categories.
+
 ## details
+This function takes as two arguments: a continuous probability distribution and a specified number of categories (`num_cats`).
+It then discretizes the distribution into `num_cats` bins, ensuring that each bin represents an equal probability mass.
+
+Note: The values obtained from this function are not automatically normalized. If normalized values are desired, one can use `normalize()` function.
+
 ## authors
 ## see_also
+fnDiscretizeGamma
+
+fnDiscretizeBeta
+
 ## example
+    # Using a Normal distribution to discretize it into 5 categories
+    discrete_values := fnDiscretizeDistribution( dnNormal( 0.0, 1.0 ), 5 )
+
+    # print the discretized values to the screen
+    discrete_values
+
 ## references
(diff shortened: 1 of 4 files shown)


help/md/fnDiscretizeDistribution.md
```diff
@@ -1,9 +1,27 @@
...
+Discretize a Continuous Distribution
 ## description
+`fnDiscretizeDistribution` transforms a continuous probability distribution into a discrete one by dividing it into a specified number of categories.
+
 ## details
+This function takes as two arguments: a continuous probability distribution and a specified number of categories (`num_cats`).
+It then discretizes the distribution into `num_cats` bins, ensuring that each bin represents an equal probability mass.
+
+Note: The values obtained from this function are not automatically normalized. If normalized values are desired, one can use `normalize()` function.
+
```
davidcerny, line 13: I'm wondering if we should keep this note, because `normalize()` only works for positive reals. For instance, you can't apply it to your own example (discretizing a normal distribution).

help/md/fnDiscretizeDistribution.md
```diff
@@ -1,9 +1,27 @@
 ## name
 fnDiscretizeDistribution
 ## title
+Discretize a Continuous Distribution
 ## description
+`fnDiscretizeDistribution` transforms a continuous probability distribution into a discrete one by dividing it into a specified number of categories.
+
 ## details
+This function takes as two arguments: a continuous probability distribution and a specified number of categories (`num_cats`).
+It then discretizes the distribution into `num_cats` bins, ensuring that each bin represents an equal probability mass.
```
bredelings, line 10: I guess the function yields a sequence of values that approximate the distribution when given equal probability weights?

Maybe "assuming" that each bin represents an equal probability mass?

I'm assuming that the value for each bin represents the median of the bin?

help/md/fnDiscretizeDistribution.md
basanta33: I removed it. Didn't realize that it didn't even work for my own example.

help/md/fnDiscretizeDistribution.md
basanta33: Isn't it the mean?
In fnDiscretizeGamma(), the default is the mean and I thought it was similar.

davidcerny: The code is a little opaque, but based on this line, it does seem like we are using the median:

https://github.com/revbayes/revbayes/blob/a55579f5b4a9ae03622c323daa5d2157f71eb124/src/core/functions/distribution/DiscretizeDistributionFunction.cpp#L105

Using the mean would correspond to `(i+1)/n_cats`, I think.

basanta33:
> The number of categories into which this distribution should be discretized.

I have it updated in the latest commit.

davidcerny (approved):

Issue #698, opened 2025-03-17
[BUG] `write()`ing a large integer causes crash


ms609:
**Describe the bug**
`write()` crashes when attempting to print large integers (presumably larger than int_max?)

**To Reproduce**
`write(1234567890)`: outputs `1234567890` ✔ 
`write(1234567890.1)`: outputs `1.234568e+09` ✔ 
`write(12345678901)`: crashes ‼ 
`write(12345678901.1)`: outputs `1.234568e+10` ✔ 

**Expected behavior**
Print successfully, or fail gracefully with error

**Computer info**
Windows 10, latest devel build

davidcerny:
Can't reproduce this on macOS – I suspect this is yet another problem caused by the fact that `long` is 64-bit on Unix but 32-bit on Windows (see also #571).

bredelings:
Fixed by #708

PR #708, opened 2025-03-18
Fix windows long

bredelings:
On windows `long int` is 32 bits, whereas on Mac and Linux it is 64 bits.  The PR solves this problem by replacing `unsigned long` with `std::uint64_t` and replacing `long` with `std::int64_t`.  The new types are 64 bits everywhere.


bredelings:
This was challenging to get compiling on all three architectures.  However, now that it compiles, for some reason I am not seeing the test failures that were present last time!

davidcerny:
I was going to ask what was the magic trick behind making the skyline coalescent tests work! :)

Would you like to add one or two tests based on #571 and/or #698?

bredelings:
I added a test based on #698, that `write(large integer)` gives the same answer on all architectures.

I haven't added a test for `print(time())`, partly because the answer is not reproducible.

davidcerny:
> I haven't added a test for `print(time())`, partly because the answer is not reproducible.

Obviously! I failed to think of that, sorry.

Frustratingly, it looks like Windows is producing an empty `arch.errout` file, so the problem is not solved quite yet.

bredelings:
I think I have fixed the new test now.  There was a problem with the parser.

davidcerny (approved):

bredelings:
Fixes #698 
Fixes #571

PR #709, opened 2025-03-19
Fix out of range proposal

bredelings:
This fixes problems running the model-averaging tutorial with debug logging that were discovered in #664.  An MCMC proposal can propose an out-of-range value (e.g. a vector index of `-2`) which then throws an exception.  We should reject this proposal instead of  stopping the MCMC or SEGFAULTing.  We implement this by marking exceptions for out-of-range errors as `MATH_ERROR`, since that is the only type of exception that we recover from during MCMC.

One additional thing that we might consider doing is to only ignore out-of-range errors in `f(x)` when the proposed value of `x` has probability 0.  While we could check that at least one of the changed variables `x` has `Pr(x|....)=0`, it would be difficult to check that `x` is responsible for the exception.  But we could allow recovering from exceptions only if at least one of the  changed pdfs has probability 0.  If we take this route, we probably need to separate out-of-range exceptions from MATH_ERROR exceptions -- MATH_ERROR exceptions are really for cases like `alpha=0.000001` in `fnDiscreteGamma( )` where the approximation breaks down.

For this tutorial, the crashes only occured if `logMCMC=3` was being specified.  But investigating this revealed all sorts of underlying problems with the infrastructure for handling exceptions.

- when a bad vector index is `-2` don't say `18446744073709551614` in the error message
- avoid SEGFAULTS when an exception occurs in statements like `this->value = f(x)`.
- don't crash in MultiSpeciesCoalescent if we look up the tree node for a non-existance species.
- don't crash when adding `...` to a line that is 52 characters log.
- etc.

The logMCMC output also shows some differences in the ~5th decimal place of the likelihood between Mac/ARM and Intel.  I'm going to investigate that separately.


src/core/functions/vector/VectorIndexOperator.h
```diff
@@ -123,19 +103,25 @@ RevBayesCore::VectorIndexOperator<valueType>* RevBayesCore::VectorIndexOperator<
...
-    size_t idx = size_t(index->getValue());
+    auto idx = index->getValue();
 
-    if ( idx < 1 || idx > v.size() )
+    if ( idx < 1 || idx > v->size() )
     {
-        std::stringstream ss_err;
-        ss_err << "Index out of bounds: The vector of size " << v.size() << " does not have an element for index " << idx << ".";
-        throw RbException(ss_err.str());
+        // This is a MATH_ERROR to avoid stopping MCMC when we propose an out-of-range index during MCMC.
```
bjoelle, line 118: I'm not sure about that - ideally we would like out of range moves to crash so we can catch them and fix them, no ?

bredelings: The situation here is that we have a random integer that is then used to index a vector.  If we use a general MCMC proposal on the random integer, it will naturally visit negative numbers and numbers larger than the vector.  I'm not that means that the move is broken and needs to be fixed?

bredelings: More generally, it makes sense to me that downstream variables can be undefined for values of upstream variables that have probability 0.  When we propose values that have probability 0, this is not a bug, it is something that should be rejected.

bredelings: One additional idea would be make range errors silent only when an upstream node has probability zero.  We could also print the error (without quitting) if `debugMCMC>=1`.

I'm not quite sure how to check which nodes are upstream of each other.  Ideally we would consider the nodes in topological order, so that upstream nodes are always visited first.  If that was the case then when we see an exception we could decide to continue if the log(Pr) is already -Inf, and throw an exception otherwise.

bredelings: I think we already don't warn if a move (somehow) always proposes values that have probability 0.

One thing we could do is suppress range errors only if at least one node has probability 0 in the new state.  That isn't as strict as requiring that it be an _upstream_ node, but it narrows the situation where we are silent about undefined probabilities somewhat.

I'm still a bit unclear about why this test case worked originally...

src/core/moves/MetropolisHastingsMove.cpp
```diff
@@ -341,8 +344,6 @@ void MetropolisHastingsMove::performMcmcMove( double prHeat, double lHeat, doubl
     // compute the probability of the current value for each node
     for (auto node: views::concat(touched_nodes, affected_nodes))
     {
-        if (fail_probability) break;
```
bredelings, line 344: Removing this line means that we need to deal with probabilities that throw an exception.

I removed this because I thought it was fragile, since it relies on the order of the nodes, which is not fully determined.

bredelings: I suspect that this was necessary to match previous behavior, but I can't quite remember...

bredelings:
Lastly, its possible that we could have a variable `y` whose range is `[0 .. x]` where `x` is random.  In this case, I think trying to enforce that we never propose values of `y` outside the (varying) range, is the wrong approach.  Instead, we should allow the proposals, but reject them.

src/core/functions/vector/VectorIndexOperator.h
mikeryanmay: Gracefully rejecting out-of-bounds moves seems appropriate to me. Of course a move that knows about/respects boundaries would be better, but in either case the result should still be an ergodic Markov chain. Note that even our scalar slide proposals do not respect boundaries, so it is possible, for example, to propose a negative proportion-of-invariant-sites parameter---should the program give a warning or crash in this case?

If the proposal is always being rejected because it only ever proposes out-of-bounds values, then that is of course a problem. It may be hard for a user to notice this even if they are doing MCMC diagnosis because proposals on discrete parameters can have acceptance rates of 0 if the posterior probability is 1. I suppose this could be revealed if the user runs multiple chains that never change value but have different values from each other. In any case, the onus is on us to ensure that proposals are ergodic (this one should be), and the onus is on the user to ensure that their MCMC is working.

mikeryanmay:
> Lastly, its possible that we could have a variable `y` whose range is `[0 .. x]` where `x` is random. In this case, I think trying to enforce that we never propose values of `y` outside the (varying) range, is the wrong approach. Instead, we should allow the proposals, but reject them.

I agree with this in principle, but am having a hard time imagining an actual scenario in revbayes where this would arise. Is it even possible?

bredelings:
OK, so after listening to both @bjoelle and @mikeryanmay, it seems reasonable to "gracefully reject" moves when they subscript a vector by a random integer and the random integer is out of range.

However, to somewhat address @bjoelle's concern, I modified the MCMC logging so that if a move is rejected because there's an out-of-range exception, we log the exception.

Are you OK if we merge this now @bjoelle?

bjoelle (approved):

Issue #668, opened 2025-02-14
[BUG] dnUniformTopology ignores negative clade constraints


Levi-Raskin:
**Describe the bug**
I am able to implement positive clade constraints with dnUniformTopology, but negative clade constraints are ignored: 
clade(XXX, negative = TRUE)

**To Reproduce**

seqDat <- readDiscreteCharacterData("XX/FAM000008.faRN_cleaned") //uploaded as TXT file for github

[FAM000008.txt](https://github.com/user-attachments/files/18803101/FAM000008.txt)

taxa <- seqDat.taxa()
coleopteraTaxaN = clade("Archostemata", "Ripiphoridae", "PogonusNCBI", negative = TRUE)
phylogeny ~ dnUniformTopology(taxa, constraints=[coleopteraTaxaN])

The output of both the distribution and the mcmc ignores this constraint. Here is an example draw:

   (((Mengenillidia[&index=5],(Neohemiptera[&index=13],(Neuropterida[&index=8],Papilionoidea[&index=9])[&index=15])[&index=16])[&index=17],((((Bombycidae[&index=10],Aphidiformes[&index=14])[&index=18],Mengenilla[&index=6])[&index=19],Xenos[&index=7])[&index=20],Culicidae[&index=12])[&index=21])[&index=22],((PogonusNCBI[&index=4],(Ripiphoridae[&index=3],Archostemata[&index=2])[&index=23])[&index=24],Drosophilidae[&index=11])[&index=25],Anoplura[&index=1])[&index=26];

**Expected behavior**
I expected the negative clade constraint to exclude proposals that contain a monophyletic coleoptera clade, but instead, it is proposing trees with a monophyletic clade.

**Computer info**
MacOS Sequoia 15.1
RevBays version 1.2.5

Thanks so much! I may be doing something wrong here/misinterpreting what negative clade constraints mean.

Levi-Raskin:
This is resolved, but due to xcode 16.2 not allowing files to be referenced in xcode projects, I am unable to submit this fix to the revbayes project. If someone else would be able to do that, that would be lovely! If I can set xcode up such that I can submit this fix to the revbayes github I will.

Here is the fix:
In /revbayes/src/core/distributions/phylogenetics/tree/UniformTopologyDistribution.cpp

line 267
was: 
`std::vector<Clade> sorted_clades = constraints;`
Now should be:
`std::vector<Clade> sorted_clades = {};
    for(int i = 0; i < constraints.size(); i++)
[...]

PR #711, opened 2025-03-20
BUG #668 dnUniformTopology ignores neg clade const

Levi-Raskin:
Solution to:
https://github.com/revbayes/revbayes/issues/668

dnUniformTopology ignored negative clade constraints and, instead, treated them as positive clade constraints. It was a simple fix of 1) not adding negative clade constraints to sorted_clade and 2) checking to ensure that the negative clades are not in the returned topology.


davidcerny:
Excellent work! Let's add an integration test to make sure this doesn't pop up again in the future.

davidcerny:
It looks like the test script sets up a negative constraint, but then doesn't actually use it...

tests/test_negCladeConst/scripts/negativeCladeConstraint.Rev
```diff
@@ -0,0 +1,133 @@
...
+        strepsipteraList.erase(k)
+    }
+}
+
+strepsipteraTaxa = clade(strepsipteraList)
+coleopteraTaxa = clade(coleopteraList)
+coleopteraTaxaN = clade(coleopteraList, negative = TRUE)
+strepColeop = clade(strepsipteraTaxa, coleopteraTaxa)
+
+c = v(strepColeop)
```
davidcerny, line 84: It looks like the test script sets up a negative constraint, but then doesn't actually use it...

tests/test_negCladeConst/scripts/negativeCladeConstraint.Rev
```diff
@@ -0,0 +1,133 @@
...
+
+phylogeny ~ dnUniformTopology(taxa, constraints=c)
```
davidcerny, line 86: It looks like the test script sets up a negative constraint, but then doesn't actually use it...

davidcerny, line 86: It looks like the test script sets up a negative constraint, but then doesn't actually use it...

davidcerny (approved):

PR #719, opened 2025-04-12
Fix deterministic index

bredelings:
In revbayes script you can write:
```
i ~ dnCategorical([0.1,0.2,0.3,0.4])
for(k in 1:10)
{
  x[k] ~ dnNormal(0,1)
}
i.setValue(1)
j := i*2
y := x[i]
z := x[j]
w := x[2*i]
```

Currently, when revbayes sees a node like `x[i]` it creates a node name like `x[1]` based on the initial value of `i`.  If that node already exists, then revbayes will just use that instead of creating a deterministic node for `x[i]`.  This is OK if `i` is constant, but produces unexpected behavior when `i` is not a constant.

This PR fixes this by only short-circuiting to an existing node like `x[1]` when the DagNode for the index is _not_ deterministic or stochastic.  Additionally, the PR tries to compute a more sensible name for the node `x[i]` that does not just use the initial value of `i` to compute the name.

---
I moved a companion fix to #721


src/revlanguage/functions/argumentrules/ArgumentRule.cpp
```diff
@@ -425,17 +425,16 @@ double ArgumentRule::isArgumentValid( Argument &arg, bool once) const
...
-        const TypeSpec& req_arg_type_spec = *it;
         if ( the_var->getRevObject().isType( req_arg_type_spec ) )
         {
             return 0.0;
         }
             
         double penalty = -1;
         // make sure that we only perform type casting when the variable will not be part of a model graph
-        if ( once == true || the_var->getRevObject().isConstant() == true )
+        if ( the_var->getRevObject().isConstant() == true )
```
bredelings, line 437: This is the line that changes the type-conversion rules.

src/revlanguage/parser/SyntaxIndexOperation.cpp
```diff
@@ -230,25 +230,31 @@ RevPtr<RevVariable> SyntaxIndexOperation::evaluateLHSContent( const std::shared_
...
+    else if (not indexVar->getName().empty())
+        index_name = indexVar->getName();
+
+    std::string identifier = theParentVar->getName() + "[" + index_name + "]";
+
     // test whether we want to directly assess the variable or if we want to assess subelement of this container
-    // if this variable already exists in the workspace
-    if ( env->existsVariable( identifier ) )
+    // if this variable already exists in the workspace and the index cannot change
+    if ( env->existsVariable( identifier ) and constant_index)
```
bredelings, line 251: This is the line that changes how `x[i]` is handled.

tests/test_deterministic_index/scripts/test.Rev
```diff
@@ -0,0 +1,34 @@
...
+{
+  x[k] ~ dnNormal(0,1)
+}
+
+j := i*2
+y := x[i]
+z := x[j]
+w := x[2*i]
+
+model(i).graph("output/model.dot")
```
davidcerny, line 13: Should `model.dot` be included in the expected output?

davidcerny (approved):

thauffe:
Thanks for your feedback David!

Do you see different values for the model_indicator in your log file? I run it only for a few generations but the value stayed constant.

tests/test_deterministic_index/scripts/test.Rev
bredelings: I originally wanted it to be.  That is one way to ensure that the dependencies between `x[i]` and `i` are tracked.

But it turns out that names for the graph nodes are generated from pointers if the node is unnamed.  So the `model.dot` always changes even if the seed is constant.

In the future I would like to fix this, and then we can put `model.dot` in the expected output.  I will file an issue...

bredelings:
Interesting!  Thanks for exploring this!

PR #721, opened 2025-04-12
Only type-convert variables based on value if they are constant

bredelings:
If you type
```
i ~ dnUniformInteger(1,3)
y := x[i]
```
then `y` will not be updated when `i` changes.  The reason is that
- `x[i]` wants `i` to be a `Natural`, but `i` is an `Integer`, and revbayes cannot convert `Integer` to `Natural`, so the attempt to create the `x[i]` will fail.
- HOWEVER, revbayes will then look for a `[]` method on `x`.  This method also wants a `Natural` tried converting the type *based on the value of the variable*, and so it succeeds in converting `i` to a `Natural` for the specific value of `i=1`.  This will succeed, but will _not_ create a deterministic node.

Using different type-conversion rules in different circumstances is a recipe for disaster in this case because it creates a silent failure, where revbayes appears to succeed, but actually succeeds at doing something else.

At the moment, eliminating value-based type-conversion causes too many failures.  So, the solution here is to use the value-based type-conversion rules only when all the arguments are constants.  Then we will get an error for the code above instead of a silent failure.

---
**Only 1 line of this large PR has any functional change**.  The rest is cleanup/simplification.   Let me know if I should split it into 2 PRs.

Fixing this issue revealed a few hidden bugs where tests were doing:
```
mnModel(filename = "output/multipleRules_test.log", printgen = 100, p)
```
It turns out that `mnModel( )` doesn't take arguments like `p`, but somehow the value-based type-conversion rules were accepting this and then ignoring the `p`.


src/revlanguage/functions/argumentrules/ArgumentRule.cpp
```diff
@@ -425,17 +433,16 @@ double ArgumentRule::isArgumentValid( Argument &arg, bool once) const
...
-        const TypeSpec& req_arg_type_spec = *it;
         if ( the_var->getRevObject().isType( req_arg_type_spec ) )
         {
             return 0.0;
         }
             
         double penalty = -1;
         // make sure that we only perform type casting when the variable will not be part of a model graph
-        if ( once == true || the_var->getRevObject().isConstant() == true )
+        if ( the_var->getRevObject().isConstant() == true )
```
bredelings, line 445: This is the line that actually fixes the bug.

davidcerny:
Could you please clarify the relationship between #719 and this PR a little bit? I assume both are necessary to fix the issue from #664, and neither is sufficient on its own? Does one represent a prerequisite for the other?

bredelings:
Technically, #719 is all that is needed if you don't make any mistakes about the types.  For example, with #719, then
```
i ~ dnUniformNatural(1,3)
y := x[i]
```
works.  Indexing by a deterministic node only works when the index has type `Natural`.

This PR is needed so that you get an error message when an `Integer` index is used, instead of silently creating a node for `x[i]` that doesn't depend on `i`.  Does that clear it up?

bredelings:
Also, #664 is not fixed yet.  Fixing `x[i]` is one more step towards the fix -- possibly there is a bug in the #664 script, and possibly there are more revbayes bugs.

davidcerny:
> This PR is needed so that you get an error message when an `Integer` index is used, instead of silently creating a node for `x[i]` that doesn't depend on `i`. Does that clear it up?

It does, thank you!

bjoelle:
so I may be misremembering completely but I thought that `Integer` **was** convertible to `Natural` ? (as can be seen in `Integer::convertTo`) is that only in some cases ?

bredelings:
The way it works is that there is `double Integer::isConvertibleTo( const TypeSpec& type, bool once ) const` to check whether the conversion is allowed, and then `RevObject* Integer::convertTo( const TypeSpec& type ) const` to do the actual conversion.

The code to convert Integer to Natural looks like this:

```
    if ( once && type == Natural::getClassTypeSpec() && dag_node->getValue() >= 0 )
    {
        return 0.1;
    }
```

The `once` argument basically means that we aren't converting `Integer` to `Natural`, but a _specific_ `Integer` (i.e. `Integer 1`) to `Natural`.  This is the "relaxed type conversion rules" that I mentioned above.

So, I would say:
- the `Integer` _type_ isn't convertible to `Natural`
- but specific `Integer` _values_ can be converted to `Natural` values.

 The fix means that we only use the value-based type conversion for constants.

bredelings:
After the cleanups in this patch, the only consumer of `once` is the variables `<type>::isConvertibleTo( )` functions.

So we could rename `once` to `by_value` everywhere it occurs.

davidcerny (approved):
This seems to work fine with the rjMCMC script I mentioned in #719. It's certainly prudent to wait for more people to give it a try and see if it breaks anything, but for now, I'm approving.

Incidentally, I was wondering if the change might make it possible to come up with a workaround to a long-term pet peeve of mine, described in issue #308. When dealing with a draw from a uniform distribution, RevBayes decides what its type should be based on the values of the distribution's bounds. So when `x ~ dnUniform(0, 0.5)`, `x` will always be of type `Probability`, and even writing `x ~ dnUniform(0, abs(0.5))` doesn't really help: the `abs()` function does convert 0.5 from a `Probability` to a `RealPos`, but the distribution only cares about the values of the bounds – not their types. I thought maybe making the upper bound a stochastic node rather than a constant would prevent the value-based type conversion from taking place and allow me to circumvent the issue. To my surprise, it does, but the changes made in this PR are not actually required for it to work.

> I hope to do more simplification in the future...

I realize this goes against the sentiment you're expressing here, but I'm wondering if it's possible to have a mechanism that would allow us to opt out of value-based type conversion altogether (i.e., even for constants)...

bredelings:
Ultimately, it would be nice to eliminate value-based simplification.  Currently, removing it breaks this code in `test_basics/scripts/basics.Rev`:

```
function Natural fac(Natural i) {
   if (i > 1) {
      return i * fac(i-1)
   } else {
      return 1
   }
}
```

What happens is that if you call `fac(4)`, then it successfully calls the function.  But when evaluating the function, we then call `fac(i-1)`.  Although `i` is a `Natural`, `i-1` is *not* a `Natural`.  One solution would be to require writing `fac(Natural(i-1))`...

PR #722, opened 2025-04-14
Document TreeTrace methods

ms609:
Write up discussion at #716


help/md/readTreeTrace.md
```diff
@@ -6,6 +6,17 @@ Function to read in a tree trace, usually produced as the output of an MCMC.
 Reads trees (Nexus or Newick accepted) from a file or folder containing a set of trees and saves them in one object. 
 ## details
 Either a file name or a directory must be provided as argument. If a folder is provided, all the files that contain trees in that directory are read in the same object.
+
+`tree_trace = readTreeTrace(nruns = 1)` returns a `TreeTrace` object.
+`tree_trace = readTreeTrace(nruns > 1)` returns a `TreeTrace[]` object, in which `trace[n]` corresponds to the trace of run $n$.
```
davidcerny, line 11: This is just nitpicking, but I'd suggest changing the code in these lines to

```
tree_trace = readTreeTrace(..., nruns = 1)
tree_trace = readTreeTrace(..., nruns > 1)
```

This is to avoid the impression that we can call `readTreeTrace()` with just the `nruns` argument: we cannot, since `file` has no default. 

Otherwise, this is a very nice and helpful addition to the documentation. Thank you for writing it up!

davidcerny (approved):

PR #723, opened 2025-04-14
Document readTrace methods

ms609:
Incorporating discussion at #716.

`trace.setBurnin()` seems to have no effect, but perhaps I've misunderstood how it is to be used?


davidcerny:
I didn't get a chance to review #702, but since you've already merged it, we should change the example so that it reads

```
trace = readTrace(filename, burnin=burnin)
```

rather than

```
trace <- readTrace(filename, burnin=burnin)[1]
```

The latter would actually extract the first (i.e., iteration) column from the file, which is almost never what we want to do. Also, while `<-` and `=` are basically equivalent, a parameter trace is more appropriately treated as a workspace object rather than a constant node in the model DAG.

ms609:
Good shout; updated.

davidcerny:
Let me just take a quick look at `.setBurnin()` before we merge this, just to see what it's intended to do and if it needs to be fixed...

help/md/readTrace.md
```diff
@@ -1,16 +1,31 @@
...
 Read an MCMC log file.
+## description
+Reads parameter values from a log file,  usually produced as the output of an MCMC or MCMCMC run.
 ## details
+
 Read an MCMC log file with field delimited by `separator`.
 Then drop the first `burnin` iterations if `burnin` is an integer,
 or the fraction `burnin` of iterations if `burnin` if a Real number.
-Then we keep every `n`th entry if the `thinning` is `n`.
+Then we keep every $n$th entry if the `thinning` is $n$.
```
bredelings, line 12: Sadly we don't handle LaTeX, so I think the `$` should show up in the output if you type `?readTrace`

davidcerny:
It took me longer than I care to admit that there is no issue with `.setBurnin()`: its seeming lack of effect actually comes from the fact that `WorkspaceVector` does not allow it to modify its elements. See #726.

Based on this, we might want to further modify the example. E.g.,

```python
traces = readTrace(filename, burnin=burnin)
# Get the posterior trace (2nd column in the log file)
posterior = traces[2]
posterior.summarize()
```

davidcerny:
I took the liberty of adding `Trace` and `TreeTrace` to our type register, which allowed me to create separate help files for them. This, in turn, made it possible to distinguish between documenting the properties of a function and documenting the properties of its output, as we already do for, e.g., `dnPhyloCTMC()` vs. `AbstractHomologousDiscreteCharacterData`.

PR #727, opened 2025-04-17
Fix logging when running Mcmcmc twice.

bredelings:
It looks like current practice is to call `monitors[i].openStream()` on the monitors for every chain -- including hot chains.  However, we only call `monitors[i].closeStream()` on the monitors for the active chain.

Currently all chains in an Mcmcmc analysis have their own duplicate version of each monitor.  For an analysis with 4 chains, we thus open the log file 4 times. The log file is open from 4 different fstream objects, and we write to the file sometimes through one fstream object, and sometimes through another object, as different chains become active.  The fact that this works at all boggles the mind.  Perhaps this explains why we seek to the end of the file before writing?

But apparently it seems to work.  However, after we stop the MCMCMC and restart it, writing to the file sometimes sends the bits directly into the ether (see #525). They just vanish, and the file is unchanged even though I can watch the write (and the flush()) happening in the debugger.  So probably something is corrupting the fstream objects.

SOLUTION: Apparently if we close all the streams after the analysis ends, then the second analysis logs things correctly.

### DIFF
--- a/src/core/analysis/mcmc/Mcmc.cpp
+++ b/src/core/analysis/mcmc/Mcmc.cpp
@@ -330,34 +330,6 @@ void Mcmc::disableScreenMonitor( bool all, size_t rep )
 
 
 /**
- * Finish the monitors which will close the output streams.
- */
-void Mcmc::finishMonitors( size_t n_reps, MonteCarloAnalysisOptions::TraceCombinationTypes tc )
-{
-    
-    // iterate over all monitors
-    for (size_t i=0; i<monitors.size(); ++i)
-    {
-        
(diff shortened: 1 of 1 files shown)


davidcerny (approved):
Works perfectly both for the original script from #525, which simply involves two consecutive runs, and for the slightly more complex case of the [proposed integration test](https://github.com/revbayes/revbayes/blob/b083a67b7ce6c83996ecce88a02a01b4a42ce5a7/tests/test_MC3/scripts/Test_heat_tuning_and_checkpointing.Rev) from #670, which involves restoring the analysis from a checkpoint.

Thanks so much, Ben!

Issue #679, opened 2025-03-06
[BUG]Occasional MCMC Error during Combined-Evidence Analysis


minghao-du:
**Describe the bug**
During Combined-Evidence analysis and the Fossilized Birth-Death process for analysis of extant taxa and fossil specimens, the MCMC chain sometimes crashes and outputs an error message `Error:	Accepted move 'VectorScaling' with with posterior ratio of -1312.76 and Hastings ratio of 1311.44.` . The error encountered appears intermittently and stops the inference, possibly due to issues with specific proposal moves.

**To Reproduce**
The data, code and log files are as follows.
`rb scripts/Rev/empirical/FBD_Model.Rev --args ME 54 0.5 uniform SFBD 150000 False 9`

[Dating_Bug.zip](https://github.com/user-attachments/files/19106531/Dating_Bug.zip)

When restarting the analysis from the checkpoint file, the run may either resume normally or repeat the error, suggesting that the issue is potentially related to certain proposed parameter values.

**Expected behavior**
The inference should run normally without encountering errors or crashes.

**Screenshots**

![Image](https://github.com/user-attachments/assets/5aaa8bba-3133-4778-9ab1-3ea079260f5d)

**Computer info**
OS: CentOS Linux release 7.5.
RevBayes: RevBayes version (1.2.6-preview) Build from development (rapture-3426-gbca173)

bredelings:
See #728

minghao-du:
Thank you for the information you provided.

I believe the issue arises from the very large Hastings ratio, which might have led to the very small posterior ratio. The question is, why is such a large Hastings ratio occurring?

I am not entirely sure about how the VectorScaling (`mvVectorScale`) works, but if it were `mvScale`, such a large Hastings ratio should not occur. The same should apply to `mvVectorScale`.

Taking `mvScale` as an example: I’m not entirely clear on what the `lambda` represents—whether it's a fine-tuning parameter. If `lambda` is indeed a fine-tuning parameter, then the Hastings ratio (in its log form) should be equal to _**lambda * (u - 0.5)**_, where _u ~ U(0, 1)_. To prevent large Hastings ratios, a range limit for `lambda`, such as `lambda` being less than 10, could be enforced in the code.

Similarly, for `mvVectorScale`, there should also be a defined range for the fine-tuning parameter inside the method to avoid such large Hastings ratios.

[...]

PR #728, opened 2025-04-18
Allow MCMC moves with low prob ratio and high Jacobian/Hastings ratio

bredelings:
Apparently we quit an MCMC run with an error if we propose and accept a state with a probability that is 1000 log units worse than the current state.  However, in some cases, this can occur with a Hastings ratio which is 1000 log units better, and so we do accept this.  This occurs with mvScaleVector -- see #679.

The solution here is to print a warning if logMCMC or debugMCMC is enabled, and continue the analysis.


davidcerny (approved):
Given the discussion under #679, this makes sense to me. While we may want to know when the two factors of the MH ratio give such wildly different assessments of a proposed state, as long as their product is greater than 1, the proposal should still be accepted.

Issue #725, opened 2025-04-15
[BUG] PDFs not up to date (infinite)


ms609:
**Describe the bug**
PDFs not up to date before proposal in MCMCMC run

       Error:       Executing NNI(topology): PDFs not up-to-date before proposal!

**To Reproduce**

Data: [k5obs4.txt](https://github.com/user-attachments/files/19753067/k5obs4.txt)

```rb
seed(0)
setOption("debugMCMC", 1)

chars <- readDiscreteCharacterData("k5obs4.txt")
taxa <- chars.names()
nTaxa <- chars.size()
nEdge <- 2 * nTaxa - 3

topology ~ dnUniformTopology(taxa)
br_lengths ~ dnDirichlet( rep(1.0, nEdge) )
phylogeny := treeAssembly(topology, br_lengths)

m_morph[1] ~ dnPhyloCTMC(tree = phylogeny, Q = fnJC(4), type = "Standard")
m_morph[1].clamp(chars)

# Completes OK:
# mcmc(model(phylogeny), [mnScreen(printgen = 1)], [ mvNNI(topology) ]).run(generations = 3)

# PDFs not up-to-date before proposal
mcmcmc(model(phylogeny), [mnScreen(printgen = 1)], [ mvNNI(topology) ]).run(generations = 3)
```

**Screenshots**
```
   Processing file "rb/debug-pdf.Rev"
   Successfully read one character matrix from file 'simmat/k5obs4.nex'

   Running MCMC simulation
   This simulation runs 1 independent replicate.
   The MCMCMC simulator runs 1 cold chain and 3 heated chains.
   The simulator uses 1 different moves in a random move schedule with 1 moves per iteration


Iter        |      Posterior   |     Likelihood   |          Prior   |    elapsed   |        ETA   |
----------------------------------------------------------------------------------------------------
0           |       -17683.1   |         -17718   |        34.8953   |   00:00:00   |   --:--:--   |
   Error:       Executing NNI(topology): PDFs not up-to-date before proposal!
       m_morph[1]: -17717.958332 != -inf    diff = inf

```

**Computer info**
Windows, branch #721 

**Additional context**
Similar issues were fixed with the treeAssembly fixes (e.g. #609) but this one seems to have persisted (or re-emerged on #721??).

bredelings:
I'm looking at this, and it appears that there is some issue with with the caching of conditional likelihoods inside AbstractPhyloCTMCSiteHomogeneous.  Still looking...

[...]

PR #729, opened 2025-04-18
Fix caching bugs in phyloCTMC

bredelings:
This fixes some bugs in dnPhyloCTMC that were discovered in #725.  In some places we were clearing the partial likelihoods, but not marking them dirty.

Additionally, we were marking all partial likelihoods clean in keepSpecialization( ) and restoreSpecialization( ).  We shouldn't be marking partial likelihoods clean for no reason in keepSpecialization( ), but in order to make that work I also needed to change restoreSpecialization( ) so that we save and restore the dirty flags instead of just marking everything clean.


src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
```diff
@@ -176,7 +176,7 @@ namespace RevBayesCore {
 
         // members
         double                                                              lnProb = 0;
-        double                                                              storedLnProb = 0;
+        std::optional<double>                                               storedLnProb;
```
bredelings, line 179: Use an optional, because this only exists when we have a previous state.

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
```diff
@@ -934,9 +937,9 @@ double RevBayesCore::AbstractPhyloCTMCSiteHomogeneous<charType>::computeLnProbab
             size_t right_index = right.getIndex();
             fillLikelihoodVector( right, right_index );
 
+            dirty_nodes[root_index] = false;
```
bredelings, line 940: Previously we weren't marking the partial likelihoods at the root as clean when we computed them.

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
```diff
@@ -950,9 +953,9 @@ double RevBayesCore::AbstractPhyloCTMCSiteHomogeneous<charType>::computeLnProbab
             size_t middleIndex = middle.getIndex();
             fillLikelihoodVector( middle, middleIndex );
 
+            dirty_nodes[root_index] = false;
```
bredelings, line 956: Same here.

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
```diff
@@ -2340,30 +2343,22 @@ void RevBayesCore::AbstractPhyloCTMCSiteHomogeneous<charType>::keepSpecializatio
...
-    // reset the ln probability
-    this->storedLnProb = this->lnProb;
+    // we don't have a previous state anymore.
+    this->storedLnProb = {};
 
     // reset all flags
-    for (std::vector<bool>::iterator it = this->dirty_nodes.begin(); it != this->dirty_nodes.end(); ++it)
-    {
-        (*it) = false;
-    }
```
bredelings, line 2350: Stop marking all partial likelihoods as up-to-date.

bredelings: The current values in dirty_nodes should already be right, we shouldn't change them all to `false`.

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
```diff
@@ -2340,30 +2343,22 @@ void RevBayesCore::AbstractPhyloCTMCSiteHomogeneous<charType>::keepSpecializatio
...
+    prev_dirty_nodes = {};
+    prev_pmat_dirty_nodes = {};
 
     for (std::vector<bool>::iterator it = this->changed_nodes.begin(); it != this->changed_nodes.end(); ++it)
     {
         (*it) = false;
     }
     
-    for (std::vector<bool>::iterator it = this->pmat_dirty_nodes.begin(); it != this->pmat_dirty_nodes.end(); ++it)
```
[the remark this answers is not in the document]
bredelings: The current values in pmat_dirty_nodes should be right, don't overwrite them.

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
```diff
@@ -2654,7 +2649,10 @@ void RevBayesCore::AbstractPhyloCTMCSiteHomogeneous<charType>::resizeLikelihoodV
         {
             partialLikelihoods[i] = 0.0;
         }
-
+        for(auto&& dirty_node: dirty_nodes)
+        {
+            dirty_node = true;
+        }
```
bredelings, line 2655: Make sure to mark all the partial likelihoods as invalid, since we just zeroed them all!

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
```diff
@@ -2686,48 +2688,41 @@ void RevBayesCore::AbstractPhyloCTMCSiteHomogeneous<charType>::resizeLikelihoodV
...
-    this->lnProb = this->storedLnProb;
-
     // reset the flags
-    for (std::vector<bool>::iterator it = dirty_nodes.begin(); it != dirty_nodes.end(); ++it)
-    {
-        (*it) = false;
-    }
+    lnProb = *storedLnProb;
+    storedLnProb = {};
+    dirty_nodes = *prev_dirty_nodes;
```
bredelings, line 2697: Restore dirty_nodes from the saved copy, instead of simply marking them all partial likelihoods up-to-date.

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
```diff
@@ -2686,48 +2688,41 @@ void RevBayesCore::AbstractPhyloCTMCSiteHomogeneous<charType>::resizeLikelihoodV
...
     }
-    
-    for (std::vector<bool>::iterator it = pmat_dirty_nodes.begin(); it != pmat_dirty_nodes.end(); ++it)
-    {
-        (*it) = false;
-    }
-    
+
+    // reset the flags
+    pmat_dirty_nodes = *prev_pmat_dirty_nodes;
```
bredelings, line 2713: Restore pmatrix dirty flags from saved copy, instead of marking them all up-to-date.

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
```diff
@@ -4090,11 +4085,13 @@ void RevBayesCore::AbstractPhyloCTMCSiteHomogeneous<charType>::swapParameterInte
...
 void RevBayesCore::AbstractPhyloCTMCSiteHomogeneous<charType>::touchSpecialization( const DagNode* affecter, bool touch_all )
 {
-
     if ( touched == false )
     {
         touched = true;
         this->storedLnProb = this->lnProb;
+
+        prev_dirty_nodes = dirty_nodes;
+        prev_pmat_dirty_nodes = pmat_dirty_nodes;
```
bredelings, line 4094: When we touch this distribution, save copies of the lnProb, the dirty_nodes vector, and the pmat_dirty_nodes vector, so we can restore them later.

src/core/distributions/phylogenetics/substitution/PhyloCTMCClado.h
```diff
@@ -73,7 +73,7 @@ namespace RevBayesCore {
 
 
     private:
-        virtual void                                            simulate(const TopologyNode& node, std::vector< DiscreteTaxonData< charType > > &t, const std::vector<size_t> &perSiteRates);
+        virtual void                                            simulateClado(const TopologyNode& node, std::vector< DiscreteTaxonData< charType > > &t, const std::vector<size_t> &perSiteRates);
```
bredelings, line 76: Don't call this `simulate` since we already have a virtual function called `simulate`.

davidcerny (commented):
This is definitely above my pay grade, but to the limited extent I can understand them, the changes look good to me. I just have a few questions:

(1) It looks like we are marking things as dirty that we weren't marking as dirty before, meaning that more things need to be recomputed. Have you noticed any performance differences?

(2) Is this sufficient to fix #725? @ms609's script now seems to run fine, but I noticed you have another branch in the works called [`refactor-phyloctmc-caching2`](https://github.com/revbayes/revbayes/tree/refactor-phyloctmc-caching2)...

(3) It's interesting that the only instance in which the changes affect a regular MCMC (as opposed to MCMCMC) analysis is the BDSTP integration test. Do you have any idea why that might be the case? Does it potentially mean that the results of FBD analyses performed with the old caching system are not correct?

davidcerny (approved):

Issue #717, opened 2025-04-11
[Feature Request] `readTrace()` should take an `nruns` argument


davidcerny:
There is a small but annoying inconsistency between the way we handle traces of scalar parameters and the way we handle tree traces, which I just realized when answering discussion topic #716.

If we conduct multiple MCMC runs, we can just give the `mnModel` and `mnFile` monitors the base names of the files in which we want to store the corresponding traces, and run-specific suffixes will be automatically appended to them:

```
monitors.append( mnModel(filename="out.log", printgen=100) )
monitors.append( mnFile(filename="out.trees", printgen=100, tree) )

analysis = mcmc(mymodel, monitors, moves, nruns=2)

# This will create the following files:
# out_run_1.log, out_run_2.log, out_run_1.trees, out_run_2.trees
```

For tree traces, this also works in reverse, and `readTreeTrace()` will automatically read in all the individual tree trace files if we give it their base name and the number of runs:

```
> tree_trace = readTreeTrace("out.trees", nruns=2)
   Processing file "out_run_1.trees"

Progress:
0---------------25---------------50---------------75--------------100
********************************************************************

   Processing file "out_run_2.trees"

Progress:
0---------------25---------------50---------------75--------------100
********************************************************************

> type(tree_trace)
   TraceTree[]
```

However, for `readTrace()`, that's not the case: there is no `nruns` argument, and if you try supplying the base name alone to the `file` argument, the function will dutifully inform you that there is no trace file of that name:

```
> trace = readTrace("out.log")
   Error:	Could not read file named "out.log" in directory named "" because neither the directory nor the file exist
```

Instead, you need to write:

```
trace1 = readTrace("out_run_1.log")
trace2 = readTrace("out_run_2.log")
```

For consistency's sake, `readTrace()` should behave like `readTreeTrace()` and take an `nruns` argument, based on which it can decide whether the return type should be `Trace[]` or `Trace[][]`.

ms609:
Ah, you beat me to it – it was when I was preparing this same request yesterday that I realised I didn't understand how the traces worked, so thought I'd get my head around #716 first!  Thanks for filing.

davidcerny:
Closed by #736.

PR #736, opened 2025-04-22
Add an `nruns` argument to `readTrace()`

davidcerny:
This PR implements feature request #717, i.e., it makes `readTrace()` (our function for reading traces of scalar parameters) behave more like `readTreeTrace()` (our function for reading tree traces). Specifically, `readTrace()` now takes an `nruns` argument, which is set to 1 by default and determines whether the function should return a vector of traces (`Trace[]`) corresponding to a single trace file, or a vector of vectors of traces (`Trace[][]`) corresponding to multiple trace files (one for each run).

As I was making this change, I noticed a few other inconsistencies. Both `readTrace()` and `readTreeTrace()` are supposed to handle passing directory paths to the `file` argument, with the idea that the function then attempts to read all the files in the directory. However, this currently fails for both, in different ways. `readTrace()` does not return anything:

```sh
> a = readTrace("/Users/David/Downloads/traces/")
   Processing file "/Users/David/Downloads/traces/Nonclock_run_1.log"
   Processing file "/Users/David/Downloads/traces/Nonclock_run_2.log"
   vector
> a
   Missing Variable:	Variable a does not exist
```

and `readTreeTrace()` only returns the first file:

```sh
> b = readTreeTrace("/Users/David/Downloads/tree_traces/")
   Processing file "/Users/David/Downloads/tree_traces/Nonclock_run_1.trees"

Progress:
0---------------25---------------50---------------75--------------100
********************************************************************

   Processing file "/Users/David/Downloads/tree_traces/Nonclock_run_2.trees"

Progress:
0---------------25---------------50---------------75--------------100
********************************************************************

> b.size()    # should be 2
   381
> type(b)     # should be TraceTree[]
   TraceTree
```

I've fixed this issue as well: in the use cases above, `readTrace()` will now return a `Trace[][]` object as expected, and `readTreeTrace()` will similarly return a `TraceTree[]` object.

Finally, unlike `readTrace()`, `readTreeTrace()` needlessly duplicated its "file not found" error messages:

```sh
> a = readTreeTrace("strom/Nonclock.trees")
   Error:	Could not find filename: strom/Nonclock.trees
   Could not read file named "Nonclock.trees" in directory named "strom" because neither the directory nor the file exist
```

This, too, has been fixed.


ms609 (commented):
I've not tried to break this, but these changes seem great, thanks for catching the additional issues, and for updating the documentation to match!

Issue #672, opened 2025-02-19
[Feature Request] Display current values in stopping rule notification when resuming run from checkpoint


ms609:
I've been finding the more visible information about stopping rules (#657) very useful.

When resuming a run from a checkpoint, it would be convenient if the message at the start of the run also displayed the current values of stopping rule targets – perhaps simply by mirroring the format of the during-the-run messages:

Current:
![Image](https://github.com/user-attachments/assets/4fd0f082-2d2e-441d-bb51-f0e1ba2a9de5)

During-the-run:
![Image](https://github.com/user-attachments/assets/ea2db3b5-1aa0-4829-85c6-094f2cb3448e)

This would surface this information without having to wait for a number of generations to elapse.  (And would also be useful on MPI runs that don't currently display during-the-run information – #671)

davidcerny:
Closed by #739.

PR #739, opened 2025-04-25
Display current stopping rule values when resuming from a checkpoint

davidcerny:
This PR implements feature request #672, i.e., it displays the current values of stopping rule convergence statistics upon resuming an analysis from a checkpoint:

```
   Appending to previous MCMC simulation of 1000 iterations
   This simulation runs 4 independent replicates.
   The simulator uses 1 different moves in a random move schedule with 1 moves per iteration
   
   Stopping rules:
       Target value for effective sample size (ESS): > 50
       Target value for the Gelman-Rubin statistic (PSRF): < 1.01
   
   Current values:    
       Minimum effective sample size (ESS): 9.636138
       Maximum value of the Gelman-Rubin statistic (PSRF): 1.013099
```

Note that to make this visually pleasing, I had to change the order in which certain operations are performed within `MonteCarloAnalysis::run`: we now reset the stopping rules before writing out monitor headers, rather than the other way around. However, I don't see why that should lead to any problems.


ms609 (approved):
Looks good to me!  Thanks!

PR #745, opened 2025-05-02
Pseudodata

bredelings:
RevBayes currently needs hacky and unreadable ways to add uniform bounds on fossils:
```
    F[i] ~ dnUniform(t[i] - b_i, t[i] - a_i)
    F[i].clamp( 0 )
```
By enforcing `0 \in (t[i]-b[i], t[i]-a[i])`, we end up enforcing that `t[i] \in (a[i],b[i])`.

It might seem simpler to simply write
```
   t[i] ~ dnUniform(a[i],b[i])
```
but that isn't write either, because `t[i]` already has a prior -- we can't add a second prior~

What's really happening here is that we have some unspecific fossil data (represented) by `F[i]`.  We don't know what the distribution of `F[i]` is -- nor do we know exactly what `F[i]` is either.  But we know that function `Pr(F[i] | t[i])` is 1 if `t[i]` is in `(a[i],b[i])` and 0 otherwise.  

We propose to express this kind of thing by writing:
```
   F[i] ~ dnPseudo(t[i])                         # there's some fossil data that gives information at t[i]
   F[i].clamp( pdUniform(a[i], b[i]) )   # this fossil data imposes a likelihood function that is 1 between a[i] and b[i] and 0 otherwise
```

### DIFF
--- a/src/core/datatypes/PseudoData.h
+++ b/src/core/datatypes/PseudoData.h
@@ -32,7 +32,7 @@ namespace RevBayesCore
     template <typename T>
     struct PseudoData: public Cloneable
     {
-        virtual double operator()(const T&) = 0;
+        virtual double operator()(const T&) const = 0;
         virtual ~PseudoData() {};
 
         PseudoData* clone() const = 0;
@@ -50,7 +50,7 @@ namespace RevBayesCore
     struct PseudoDataNothing: public PseudoData<T>
     {
         PseudoDataNothing* clone() const {return new PseudoDataNothing(*this);}
-        double operator()(const T&) {return 0;}
+        double operator()(const T&) const {return 0;}
     };
 
     // Whatever we observed, the data has decreasing probability outside [a,b]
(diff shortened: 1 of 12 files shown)


src/core/help/RbHelpDatabase.cpp
```diff
@@ -1447,6 +1447,89 @@ mymcmc.run(generations=200000))");
...
+        For example pdAbove(a) && pdBelow(b) = pdBetween(a,b)
+
+  * pd1 || pd2: adds the likelihoods.
+
+  * pd + d: shifts the likelihood curve right by d.  
+        For example, pdLogNormal(0,1) + 10 requires that x > 10.
+
+  * pd - d: shifts the likelihood curve left by d.
+
+  * d - pd: computes pd(d-x).)");
```
davidcerny, line 1502: This is exceptionally clear and does a great job of explaining when the user might want to use `dnPseudo()`, and why – thank you!

I only have two minor suggestions:

- When you say that this is known as "virtual evidence" in the statistical literature, it might be worth providing a citation;
- It's better to provide a Markdown file than to edit `RbHelpDatabase` directly – as I'm sure you know, that way we can get the help page to appear both in the RevBayes console and on the website.

bredelings:
This is superseded by #939

PR #751, opened 2025-05-05
Update dlls

bredelings:
Apparently the released version of RevStudio wouldn't start.  The problem was that the DLLs included weren't compatible with each other.  This PR solves this by updating the DLL versions.  We also need to avoid trying to find DLL files for libraries that are now header-only.

The TreeNode::resolveMultifurcation algorithm also suddenly started failing tests.  There was an issue with the coalescence algorithm for TimeTrees when not all children have the same age.  In this case a coalescent event could be scheduled for a time when there was only one active child, and the code would try and remove 2 children from a vector containing only 1 child.  This is solved by sorting the children and not generating a coalescence time for the youngest child.  We also generate only n-2 times now.  Sorting the children also makes it easier to determine which of the children are extant at a given time.  At each coalescent event we include any children that have become active before that coalescent event.

For BranchLengthTrees, the old code generated n-1 coalescent events so that the current node would have only 1 child.  The resulting "knuckle" was then removed later.  I have changed this so that we generate only n-2 coalescent events.  Now we don't generate any knuckles, so we don't have to remove them. 

Additional issues encountered and fixed along the way:
- the new coalescence tests are very slow and very numerous.  They come from a tutorial, so just run 1 iteration like we do with other tutorials.
- printing operator summaries made the *.errout differ if the code was run with assertions enabled, because one assertion computed a likelihood, and one of the operator summaries counted likelihood evaluations.  So I removed this printing.
- AutocorrelatedEventDistribution used `-1` to indicate lack of autocorrelation.  So we need to avoid subscripting by `-1`
- Elliptical slice sampling asserted that recomputing the likelihood should be exactly identical. I changed this to assert a log error of < 1.0e-9

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -178,7 +178,7 @@ jobs:
       env:
         USE_MPI: ${{ matrix.mpi }}
       run: |
-        ${GITHUB_WORKSPACE}/projects/meson/make_winroot.sh -ccache true -gtk true
+        ${GITHUB_WORKSPACE}/projects/make_winroot.sh -ccache true -gtk true
         ( cd ${GITHUB_WORKSPACE}/projects/meson/ ; ./generate.sh )
         
     - name: Configure and build
--- a/projects/make_winroot.sh
+++ b/projects/make_winroot.sh
@@ -96,8 +96,8 @@ cd ${SYSROOT}
 # libwinpthread-git-9.0.0.6090.ad98746a-1
 # We can use the libwinpthread-1 from the cross-compiler instead of download it.
 
-PKGS="boost-1.88.0-1
-boost-libs-1.88.0-1
+PKGS="boost-1.88.0-2
+boost-libs-1.88.0-2
 dlfcn-1.4.2-1
 "
 
--- a/projects/meson/make_winroot.sh
+++ b/projects/meson/make_winroot.sh
@@ -96,44 +96,47 @@ cd ${SYSROOT}
 # libwinpthread-git-9.0.0.6090.ad98746a-1
 # We can use the libwinpthread-1 from the cross-compiler instead of download it.
 
-PKGS="boost-1.83.0-2
-dlfcn-1.4.1-1
+PKGS="boost-1.88.0-1
+boost-libs-1.88.0-1
+dlfcn-1.4.2-1
 "
 
 if [ "${gtk}" = "true" ] ; then
     PKGS="$PKGS
-    atk-2.50.0-1
-    brotli-1.1.0-1
+    atk-2.56.2-1
+    brotli-1.1.0-4
     bzip2-1.0.8-3
-    cairo-1.18.0-1
-    expat-2.5.0-1
-    fontconfig-2.14.2-1
-    freetype-2.13.2-1
-    fribidi-1.0.13-1
-    gdk-pixbuf2-2.42.12-1
+    cairo-1.18.4-1
+    expat-2.7.1-2
+    fontconfig-2.16.2-1
+    freetype-2.13.3-1
+    fribidi-1.0.16-1
+    gdk-pixbuf2-2.42.12-4
     gettext-0.22.4-3
-    glib2-2.78.2-1
-    graphite2-1.3.14-2
+    glib2-2.84.1-2
+    graphite2-1.3.14-3
     gtk2-2.24.33-6
-    harfbuzz-8.3.0-1
-    iconv-1.17-4
-    libiconv-1.17-4
-    jasper-4.0.0-1
+    harfbuzz-11.2.0-1
+    iconv-1.18-1
+    libiconv-1.18-1
+    jasper-4.2.5-2
+    jbigkit-2.1-5
+    lerc-4.0.0-1
     libdatrie-0.2.13-3
-    libdeflate-1.19-1
-    libffi-3.4.4-1
-    libjpeg-turbo-3.0.1-1
-    libpng-1.6.40-1
+    libdeflate-1.23-1
+    libffi-3.4.8-1
+    libjpeg-turbo-3.1.0-1
+    libpng-1.6.48-1
     libthai-0.1.29-3
-    libtiff-4.6.0-1
-    libwebp-1.3.2-1
+    libtiff-4.7.0-1
+    libwebp-1.5.0-1
     lzo2-2.10-2
-    pango-1.50.14-4
-    pcre2-10.42-1
-    pixman-0.42.2-1
-    xz-5.4.5-1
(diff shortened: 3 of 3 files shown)


davidcerny:
(1) The refactored `.resolveMultifurcations()` method works correctly in all of my scripts that make use of it, so that's great!

(2) That said, I wonder if the refactoring was necessary for that variant of the method which operates on `BranchLengthTree` objects, since it was just the variant operating on `TimeTree`s that was causing the Windows tests to fail. I assume this was done simply for the sake of consistency between the two?

(3) As the comments in the original code clearly stated, I pieced the `.resolveMultifurcations()` method together from bits of code I found in other places: specifically, [`UniformTopologyDistribution`](https://github.com/revbayes/revbayes/blob/master/src/core/distributions/phylogenetics/tree/UniformTopologyDistribution.cpp) (for the `BranchLengthTree` variant) and [`UniformSerialSampledTimeTreeDistribution`](https://github.com/revbayes/revbayes/blob/master/src/core/distributions/phylogenetics/tree/UniformSerialSampledTimeTreeDistribution.cpp) (for the `TimeTree` variant). So if the former required changes, we might want to change the latter too. In the case of the time tree code, this might actually be necessary: if the issue identified by the Windows test ultimately stems from the fact that in the presence of extinct tips, a "coalescent"(*) event can occur when there is just 1 active node, then `UniformSerialSampledTimeTreeDistribution::simulateCoalescentAges()` likely suffers from the same problem, and should be refactored as well. But then the issue arises of whether the refactored version still follows the original math. For uniform time trees with arbitrary tip ages, the math was introduced by [Ronquist et al. (2012)](https://watermark.silverchair.com/sys058.pdf), I believe. I know that on Slack, you said

> I don't think I changed the overall way the coalescence was done -- i.e. for TimeTrees it still draws a uniform coalescence time for the children

so hopefully this won't be that big of a deal.

(4) The ability to run just one test at a time sounds like a neat addition, but it's not clear to me when and how it's intended to be used. Is the idea that if we run into a similar problem in the future, we can modify `build.yml` on the bug-fix branch so that it calls `run_integration_tests.sh` with this flag? That way, GitHub Actions won't take forever with every new commit, and once the test passes, the change to `build.yml` can be reverted?

(5) That's unfortunate about the operator summaries – do I understand correctly it was the elliptical slice sampling move that counted likelihood evaluations? If so, I wonder if it might be better to just make it not report this, instead of getting rid of the operator summaries wholesale...

(6) 10^(−9) looks pretty arbitrary. Is there any deeper reason why this particular value was chosen, rather than, say, `RbSettings::userSettings().getTolerance()`?

(*) I don't like this terminology and its extensive use in the new code, because I feel like it can engender confusion with the multispecies coalescent, which has nothing to do with what's happening here. Then again, the relevant member function of `UniformSerialSampledTimeTreeDistribution` is also named `simulateCoalescentAges()`, so it's probably okay.

bredelings:
Hi David, thanks for the response.  I guess the tests are mysteriously passing again now, so this is no longer such an urgent issue.  Since I am in Crete at a workshop from May 9-May 24 I'll just respond briefly.

> (2) That said, I wonder if the refactoring was necessary for that variant of the method which operates on BranchLengthTree objects, since it was just the variant operating on TimeTrees that was causing the Windows tests to fail. I assume this was done simply for the sake of consistency between the two?

Until I did this change also, the tests would hang.

> (4) The ability to run just one test at a time sounds like a neat addition, but it's not clear to me when and how it's intended to be used. Is the idea that if we run into a similar problem in the future, we can modify build.yml on the bug-fix branch so that it calls run_integration_tests.sh with this flag? That way, GitHub Actions won't take forever with every new commit, and once the test passes, the change to build.yml can be reverted?

The intention was that while running the tests locally, I could run just the nonBifurcating test (which takes 30 seconds) instead of the whole test suite, (which takes like 30 minutes).

> (3) As the comments in the original code clearly stated, I pieced the .resolveMultifurcations() method together from bits of code I found in other places: specifically, [UniformTopologyDistribution](https://github.com/revbayes/revbayes/blob/master/src/core/distributions/phylogenetics/tree/UniformTopologyDistribution.cpp) (for the BranchLengthTree variant) and [UniformSerialSampledTimeTreeDistribution](https://github.com/revbayes/revbayes/blob/master/src/core/distributions/phylogenetics/tree/UniformSerialSampledTimeTreeDistribution.cpp) (for the TimeTree variant). So if the former required changes, we might want to change the latter too. In the case of the time tree code, this might actually be necessary: if the issue identified by the Windows test ultimately stems from the fact that in the presence of extinct tips, a "coalescent"(*) event can occur when there is just 1 active node, then UniformSerialSampledTimeTreeDistribution::simulateCoalescentAges() likely suffers from the same problem, and should be refactored as well. But then the issue arises of whether the refactored version still follows the original math. For uniform time trees with arbitrary tip ages, the math was introduced by [Ronquist et al. (2012)](https://watermark.silverchair.com/sys058.pdf), I believe. I know that on Slack, you said

My apologies, I just didn't have time to check how similar the code you wrote was to the other code since I was fixing my slides in order to teach soon. 

If indeed the tests are again mysteriously passing, then I could wait until I get back to the states and split it into multiple separate PRs, which we could review separately.

davidcerny:
@bredelings I definitely agree this can wait until you're back from Crete – I hope you are enjoying your time there!

Although the problems with the Windows build came and went (which I still find hard to explain), it does seem like they revealed a tricky edge case that could rear its head again in the future, so I think the changes you made are worth merging. Splitting this into multiple PRs (DLLs / infrastructure / "coalescence" algorithms, maybe?) might be a good idea.

> The intention was that while running the tests locally, I could run just the nonBifurcating test (which takes 30 seconds) instead of the whole test suite, (which takes like 30 minutes).

I see. I guess I never really saw the need for that when trying to run an individual test locally, since it's relatively easy to just do, say,

```
cd tests
../projects/meson/build/rb ./test_nonBifurcating/scripts/Test_resolveMultifurcations.Rev &> ./output/Test_resolveMultifurcations.errout
```

and examine the output. But obviously, different people may prefer different workflows, so it might still be a useful addition.

> My apologies, I just didn't have time to check how similar the code you wrote was to the other code since I was fixing my slides in order to teach soon.

Apologies, I definitely didn't mean for that to come across as accusatory! It was just an observation that since the method and the distribution are based on duplicated code, if one is changed, then the other should be changed as well. 

I've been thinking about the question I asked (i.e., if your new algorithm still draws from the same distribution). Maybe there are ways to test it using simulations. E.g., comparing the distributions of the likelihoods of 10,000 random draws...

davidcerny:
@bredelings I'm happy to approve that!

PR #758, opened 2025-05-16
Make sure `writeNexus()` can handle `BranchLengthTree` vectors

### LINKED ISSUE #756: [Feature Request]
Hello, 
the writeNexus function can take this type of data as input:
AbstractHomologousDiscreteCharacterData|ContinuousCharacterData|Tree|Tree[]|TimeTree[]<any> data

as a result, if one wants to write branchLengthTrees, they get written without their branch lengths. It would be nice to be able to write branchLengthTrees with their branch lengths!

Example code, for testing:
numTrees = 10

for (i in 1:50) {
  taxa[i] <- taxon("T_"+i)
}

# Get some useful variables from the data.
n_taxa <- taxa.size()
n_branches <- 2 * n_taxa - 3

##############
# Tree model #
##############

out_group = clade(taxa[1])
lambda=10.0

for (t in 1:numTrees) {

  trees[t] ~ dnUniformTopologyBranchLength(taxa, outgroup=out_group, branchLengthDistribution=dnExponential(lambda))
  print(trees[t])
}

writeNexus("tmp", trees)

davidcerny:
Fixes #756. I also noticed that when we print multiple trees (of type `BranchLengthTree` or `TimeTree` – it doesn't matter) to the same Nexus file, an extraneous semicolon is appended after each Newick string. While `ape::read.nexus()` can handle this, it is enough to confuse FigTree. I've therefore fixed this as well.


Boussau (approved):
The changes look fine to me but I have not tested them.

Issue #757, opened 2025-05-16
A method nbranches for topologies


Boussau:
Hello, 

A BranchLengthTree object as generated by dnUniformTopology or by dnUniformTopologyBranchLength does not contain an "nbranches" method. I think it would be useful to have. One benefit would be to simplify a bit the tutorials: not everybody will remember that an unrooted tree has 2n-3 nodes ; instead nbranches could be obtained from the topology.

davidcerny:
Closed by #759.

Issue #710, opened 2025-03-20
Documentation for methods


ms609:
Currently, the only documentation that exists for methods is the auto-generated description of parameters and default values available in the RevBayes console via `?func.method()`, as revealed to the user by a note in  `?func`.  However:

1. This information is unavailable via the web documentation.
2. There is no scope to provide examples, explanation, etc of the method, as for the corresponding function.
3. There is no mechanism for giving detailed explanation of the parameters – either to expound the terse description provided of their function; or to detail their default values (e.g. for `pp.run()`, it would be useful to explain that "Under the default `NULL` value, `preburninGenerations` will be set to equal to the value of the `generations` parameter").

A workaround would be to include this information at the `?func` reference page – but this is unsatisfactory in that (i) it would involve duplicating some information from the auto-generated method page, which may later become outdated; (ii) it risks cluttering up the main manual page, and making information more difficult to find.

An intuitive solution could be to generate methods documentation from `func.method.md`; on the web, these could be linked either in the main reference index list (drawback: this is already rather long, making it difficult to navigate); or by hyperlinking the list of methods that occur on `func`'s reference webpage.

PR #759, opened 2025-05-16
Add an `.nbranches()` method to `Tree`

davidcerny:
Addresses #757.


ms609 (commented):
This is a very useful method to have – most of my scripts start by calculating this manually, so it will be nice to be able to simplify.

Will this implementation work correctly for both rooted and unrooted trees?  Is the method self-explanatory enough not to need test cases or documentation?

davidcerny:
I think so. [Trees](https://en.wikipedia.org/wiki/Tree_(graph_theory)) are [by definition](https://mathworld.wolfram.com/Tree.html) connected graphs with $n$ vertices and $n − 1$ edges, so the method will always work: for rooted trees, unrooted trees, trees with multifurcations, trees with degree-2 nodes representing sampled ancestors, etc.

ms609:
That's grand – I occasionally encounter issues with this in R due to how (or whether) a root edge is represented in the tree structure, so thought it worth flagging up.

davidcerny:
@ms609 A-ha! Thank you for making me think about an edge case (pun intended) I originally failed to consider. :) My graph-theoretical answer was apparently a bit facile, or at least incomplete. In RevBayes, branches are treated as attributes of their end nodes, so strictly speaking, we always have as many branches as we have nodes – not one fewer. It's why this:

```
tree.branchLength( tree.nnodes() )
```

is always guaranteed to work, even with unrooted trees where the interpretation of the result is unclear (in the rooted case, it represents the length of the root edge, just like you suggested). It's where the final "`:0.000000;`" comes from when a tree is written out as a Newick string. It's also true that in some cases (e.g., conditioning a birth-death simulator on the origin age rather than the root age), this edge can have a nonzero length.

I still think it makes sense for `.nbranches()` to return `nodes.size() - 1`, for two reasons:

(1) If we were to include the (possibly meaningless) "root" edge in the total branch count, `.nbranches()` would just be an alias of `.nnodes()`, and therefore redundant.

(2) If the goal is to facilitate iterating over things like branch lengths or branch rates, then we need the method to return one less the number of nodes. Even in those time trees where the root edge has a nonzero duration, we still don't assign it its own branch rate in an uncorrelated relaxed clock model, for example: we consider character evolution to start at the root.

In principle, I agree that this choice of ignoring the "root" edge is worth documenting, but in the absence of a solution to #710, I'm not sure where to do it.

Issue #731, opened 2025-04-21
[BUG] Nexus file not parsed when closing comma not provided


ms609:
**Describe the bug**
A missing final comma in a STATELABELS Nexus block causes `readDiscreteCharacterData` to fail

**To Reproduce**

Create a file with the below contents, and read using
`chars <- readDiscreteCharacterData(file)`

```
#NEXUS
BEGIN TAXA;
  DIMENSIONS NTAX=5;
  TAXLABELS
    'A'
    'B'
    'C'
    'D'
    'E'
  ;
ENDBLOCK;

BEGIN CHARACTERS;
  DIMENSIONS NCHAR=3;
  FORMAT DATATYPE=STANDARD GAP=- MISSING=? SYMBOLS="0123456789";
  CHARLABELS
    [1] 'Cephalization'
    [2] 'Proboscis invaginable'
    [3] 'Degree to which the introvert can be invaginated'
  ;
  STATELABELS
    1
    'Proboscis'
    'Distinct head'
    ,
    2
    'Absent'
    'Present'
    ,
    3
    'Minimal'
    'Extensive'
    [no comma is present here]
  ;
  MATRIX
    'A'	00-
    'B'	0??
    'C'	010
    'D'	011
    'E'	010
  ;
ENDBLOCK;
```

**Expected behavior**
The matrix should be read successfully
Instead I see

```
 Error reading file "file.nex"
   No data matrix was read.
   Error:   Invalid character number (MATRIX) found in STATELABELS command (either out of range or not interpretable as an integer)
```

Matrices downloaded from MorphoBank do not contain this final comma; it would be useful to support these files without requiring manual intervention from the user.

**Computer info**
Win 10, rb devel

davidcerny:
Fixed by #760.

PR #760, opened 2025-05-17
Fix parsing of the `STATELABELS` block in Nexus files

davidcerny:
This PR makes a small change to one of the NCL library files that fixes issue #731. It's not a particularly elegant solution (we  now check for semicolons twice: both before and after moving to the next token), but it does get the job done.

### DIFF
--- a/src/libs/ncl/nxscharactersblock.cpp
+++ b/src/libs/ncl/nxscharactersblock.cpp
@@ -4549,9 +4549,11 @@ void NxsCharactersBlock::HandleStatelabels(
 	charStates.clear();
 	for (;;)
 		{
-		token.GetNextToken();
-		if (token.Equals(";"))
-			break;
+        if (token.Equals(";"))
+            break;
+        token.GetNextToken();
+        if (token.Equals(";"))
+            break;
         int n = -1;
         try {
             n = token.GetToken().ConvertToInt();
--- a/tests/data/statelabels_end_comma.nex
+++ b/tests/data/statelabels_end_comma.nex
@@ -1,0 +1,42 @@
+#NEXUS
+BEGIN TAXA;
+  DIMENSIONS NTAX=5;
+  TAXLABELS
+    'A'
+    'B'
+    'C'
+    'D'
+    'E'
+  ;
+ENDBLOCK;
+
+BEGIN CHARACTERS;
+  DIMENSIONS NCHAR=3;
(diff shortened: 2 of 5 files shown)


ms609 (commented):
Looks like a neat fix to me; is it worth propagating this to the NCL library itself (via https://github.com/mtholder/ncl?)?

davidcerny:
@ms609 Interestingly, there's no need for that, as a [commit](https://github.com/mtholder/ncl/commit/5e2cd01bc70c299cbefe5f3ca1c40308813f99e1) was made to the main branch of that repo 2 years ago with an equivalent fix. So if anything, this PR just helps our version of NCL keep up with the upstream repo.

ms609:
In this case, is there a case for updating our copy of the NCL library to the latest version (or the mtholder version – I wasn't quite clear on the relationship of the fork to the canonical library)?

davidcerny:
> In this case, is there a case for updating our copy of the NCL library to the latest version (or the mtholder version – I wasn't quite clear on the relationship of the fork to the canonical library)?

That was my first thought too, but our version has apparently diverged from the canonical library quite a bit. It's not immediately obvious, because most of the changes took place more than 6 years ago in the [old ("pre-rapture") repo](https://github.com/revbayes/revbayes.archive), and many of them are just aesthetic, like adding whitespace or the use of snake_case instead of camelCase in variable names. However, the differences also include alternative solutions to the same problems (e.g., compare commit [`f517830`](https://github.com/mtholder/ncl/commit/f5178302abddc24b8d8cb17cb149a9b646f579b3) in the upstream repo to the [corresponding line](https://github.com/revbayes/revbayes/blob/44f4234743660f554948a78400e4bc41d7707a14/src/libs/ncl/nxstaxablock.cpp#L279) in our version), so a wholesale update of our NCL copy seems outside the scope of this PR. That said,

> could we use their solution to this specific issue?

I agree 100% this is a good idea; we should not cause the two versions to diverge more than they already have. I've pushed a commit that replaces my original fix to the `STATELABELS` comma issue by the solution proposed by @mbrazeau in commit [`5e2cd01`](https://github.com/mtholder/ncl/commit/5e2cd01bc70c299cbefe5f3ca1c40308813f99e1) to the upstream repo.

ms609 (approved):
Grand; I'll bear the divergence in mind if I encounter future parsing errors.  Thanks for sorting this out.

Issue #732, opened 2025-04-21
[BUG] ASSUMPTIONS Nexus block causes file read to fail


ms609:
**Describe the bug**
Nexus files that contain a `TYPESET` assumption cannot be read

**To Reproduce**
Read a text file containing the content below using `readDiscreteCharacterData()`

```NEXUS
#NEXUS
BEGIN TAXA;
  DIMENSIONS NTAX=5;
  TAXLABELS
    'A'
    'B'
    'C'
    'D'
    'E'
  ;
ENDBLOCK;

BEGIN CHARACTERS;
  DIMENSIONS NCHAR=3;
  FORMAT DATATYPE=STANDARD GAP=- MISSING=? SYMBOLS="0123456789";
  CHARLABELS
    [1] 'Cephalization'
    [2] 'Proboscis invaginable'
    [3] 'Degree to which the introvert can be invaginated'
  ;
  STATELABELS
    1
    'Proboscis'
    'Distinct head'
    ,
    2
    'Absent'
    'Present'
    ,
    3
    'Minimal'
    'Extensive'
    [no comma is present here]
  ;
  MATRIX
    'A'	00-
    'B'	0??
    'C'	010
    'D'	011
    'E'	010
  ;
ENDBLOCK;
BEGIN ASSUMPTIONS;
  TYPESET * UNTITLED = unord: 1 - 3;
ENDBLOCK;
```

I see:
```
   Error reading file "constant-node.nex"
   No data matrix was read.
   Error:       You used a non-model object ( type=NULL ) to create a constant node. Only model objects can be used for constant nodes.
```

**Expected behavior**
The Assumptions block should not affect the reading of the characters.

If an error is anticipated, the error message should communicate the nature of the issue more clearly to the user.

**Computer info**
Win 10, rb devel

davidcerny:
Fixed by #764.

PR #764, opened 2025-05-22
Allow reading Nexus files with an `ASSUMPTIONS` block

davidcerny:
This PR fixes issue #732. It's not an ideal solution – the relevant code in [`NclReader.cpp`](https://github.com/revbayes/revbayes/blob/master/src/core/io/NclReader.cpp) actually suggests that RevBayes should be able to parse the `ASSUMPTIONS` block and automatically create partitions corresponding to the character sets specified by `TYPESET`. To the best of my knowledge, this never worked, and this PR doesn't make it work, either – in effect, it just causes RevBayes to ignore the contents of the `ASSUMPTIONS` block. However, this still seems preferable to the current behavior, which is to throw an error and fail to read the data.


davidcerny:
> I wonder how hard this would be to get working...

Yes, I've been wondering about that too. Technical issues aside, one potential problem is that unlike PAUP* or MrBayes, RevBayes currently enforces a rather strict separation of data and analysis settings. Parsing an `ASSUMPTIONS` block could potentially erode that separation: for example, reading in a Nexus data file as a vector of partitions would prevent us from re-partitioning the data in a different way in our Rev script. Maybe that's okay, but it might be worth further thought.

> If we are going to ignore the ASSUMPTIONS block, maybe we should emit a note saying something to effect of "I see an ASSUMPTIONS block and I'm going to ignore it"?

Great idea! I've added a note to that effect.

PR #683, opened 2025-03-07
more move help files

PhyloevoTi:
more move help files

PR #766, opened 2025-05-26
Fix updownslidemove

hoehna:
This is a bug fix for the `mvUpDownSlide`. The recent changes affected the argument label, but internally that was not changed correctly and broke scripts. Now I have added this move to one of the EBD tests to prevent future failures.


davidcerny (approved):
Oops, apologies – it was me who inadvertently broke the move in #683! Approving.

PR #769, opened 2025-05-29
Allow running just one test.

bredelings:
This addresses two issues with the current testing script:
- there's no way to run only _some_ of the tests, you can only run _all_ the tests
- if a test directory contains several scripts, there is no progress indicator

This was an attempt to mitigate problems where the new coalescent tests take like 20 minutes to run. 
- I sometimes want to run a single test using the test script, instead of by hand, in order to ensure that I'm using the same logic that the test script is using to generate the *.errout files.  
- the coalescent test directory contains many test scripts, but some are slower than others.  At the moment, nothing indicates when individual test scripts in a directory complete.


bredelings:
Canceling this for the moment.

PR #771, opened 2025-05-29
Fix resolve multifurcations crash

bredelings:
Right now its possible for the code that resolves multifurcations for time trees to crash.  Since we are testing this, that makes all tests fail.

Its not clear why the test was passing, then stopped, then started passing again.  However, a lot of RevBayes code creates a `std::vector<TopologyNode*>` and then sorts it.  Such code is not really reproducible, but might give stable results for a while, until some dependency is changed.

In any case, there is a bug in the code.  It assigns each child a coalescence time Uniformly along the branch length.  But when the children have different ages, its possible to create a coalescent event for a time that ends up having only one active node.

The proposed solution sorts the children with the youngest child (which has the longest branch) first.  If we have n children, we need n-2 coalescent events.  We ignore the first and last child when generating coalescence times.  Ignoring the first child ensures that we can't have a situation where the first m coalescence times are all younger than child[m].

Ignoring the last child ensures that the coalescent times are actually uniform.  Previously if we had n children, we drew `n-1` coalescence times, but only used the youngest `n-2` of them.  I think that would skew the times towards the present, and make them no longer uniform.


bredelings:
I took a look at UniformSerialSampledTimeTreeDistribution.cpp, and I think it is very nearly correct.  The only mistake is that the taxa are not sorted by age. It correctly uses n-2 ages if the root age is given, and n-1 ages if not.

I added a fix to sort the taxon ages.

bredelings:
> Hmm, no, they don't prove it, because they can't. I was wrong to think that the "uniform tree prior" is uniform because it induces a uniform prior on labeled histories. Instead, it's just called that because it places a uniform prior on node heights. 

Oh, I see.  That seems important.  I feel like this is a stupid distribution. (Yes, that is not a carefully reasoned statement :eyeroll:)  Also, I note that it is never used in any of the tutorials or tests.

> In principle, we should be able to validate your implementation against the one in MrBayes, but I don't know if MB makes it possible to evaluate the density of a fixed tree...

Interesting.  I guess you are thinking to validate the simulator by checking if the approximate simulated densities match the calculated densites from MrBayes?

> In UniformSerialSampledTimeTreeDistribution.cpp::buildSerialSampledRandomBinaryTree, we are using two separate TopologyNode vectors (active_nodes and extinct_nodes) instead of a single TopologyNode vector (active_children) and an index (used_children). Are these two solutions equivalent? If so, would it make sense to use the same one in both cases?

Yes, the solutions should be equivalent -- the new one has fewer moving parts, so I feel like its easier to check and maintain.   Maybe it would make sense to synonymize the implementations?  Its just that it would probably take >= 3 hours, and it doesn't seem like UniformSerialSampledTimeTreeDistribution is broken.  It seems like UniformSerialSampledTimeTreeDistribution has no users.  And I have other branches that would make better use of those 3 hours...

Alternatively, we could add a note that to UniformSerialSampledTimeTreeDistribution that there is a possibly better solution in TopologyNode.cc, and leave it at that for now.

> It also looks like you made a single change to the code for resolving multifurcations in BranchLengthTrees (i.e., the one we call if use_ages == false): the while loop now requires children.size() > 2 instead of children.size() >= 2. Was this needed to prevent the test from hanging? If so, do we need to make an equivalent change in [UniformTopologyDistribution::simulateClade](https://github.com/revbayes/revbayes/blob/44f4234743660f554948a78400e4bc41d7707a14/src/core/distributions/phylogenetics/tree/UniformTopologyDistribution.cpp#L177), where this code was lifted from?

Good question -- I don't think so.  A key difference is that `UniformTopologyDistribution::simulateClade` takes a vector of nodes that have no parent, so we still need to create the final parent that when we have only 2 children left.

In contrast, when resolving multifurcations for BranchLengthTrees, the children all already have a parent, so if there are only 2 children, there is in fact nothing to do.  The code from before was creating an additional node with 2 children -- which made the original parent into a knuckle -- and then calling code to delete knuckles.  So realizing that we were done when we got down to 2 children meant that we didn't have to create this additional parent and then delete it.  Possibly deleting the knuckle-removing code fixed the hang, although I don't know exactly how.

davidcerny:
> I feel like UniformSerialSampledTimeTree is a stupid distribution.

Our very own @mikeryanmay would probably agree with you on that, given what he wrote about it in a [2021 paper of his](https://academic.oup.com/sysbio/article/70/6/1232/6184877)!

> Also, I note that it is never used in any of the tutorials or tests.

Unfortunately, that's more of a demonstration that there are still gaps in our test suite's coverage than that the distribution has no users. Again, Mike R. May used it in a published paper, and [so did I](https://onlinelibrary.wiley.com/doi/full/10.1111/cla.12583) (although the results only made it into the [supplement](https://onlinelibrary.wiley.com/action/downloadSupplement?doi=10.1111%2Fcla.12583&file=cla12583-sup-0001-Supinfo.pdf)). I also remember pointing @rachelwarnock to `dnUniformSerialSampledTimeTree` some time ago on Slack in response to her asking whether we have an equivalent to the uniform time tree prior in MrBayes.

The uniform time tree prior was very popular in the early days of fossil-based tip-dating – say, between 2012–2016. It's been mostly abandoned in favor of FBD, due in no small part to the issues demonstrated in Mike's paper, but I remember having a discussion with @acapomorphic and @hoehna  where Sebastian recommended it as a possible check on FBD results when the latter look suspect.

Tl;dr, we should continue to support it.

> Interesting. I guess you are thinking to validate the simulator by checking if the approximate simulated densities match the calculated densites from MrBayes?

Frankly, I don't know what I was thinking when I wrote that... I must be tired. For some reason I was concerned with making sure both MrBayes and RevBayes return the same density for the same tree, but given that the changes you made here only affect how trees are simulated (`simulateCoalescentAges`) and not how they are evaluated (`computeLnProbability`), that wouldn't prove anything even if it could be done. Feel free to just ignore that.

> Alternatively, we could add a note that to UniformSerialSampledTimeTreeDistribution that there is a possibly better solution in TopologyNode.c[pp], and leave it at that for now.

That sounds like a great solution to me – certainly preferable to having you spend 3+ hours refactoring something that's not broken! If you can add a note like that, I'll be happy to approve the PR.

> Good question -- I don't think so. A key difference is that `UniformTopologyDistribution::simulateClade` takes a vector of nodes that have no parent, so we still need to create the final parent that when we have only 2 children left.
>
> In contrast, when resolving multifurcations for BranchLengthTrees, the children all already have a parent, so if there are only 2 children, there is in fact nothing to do. The code from before was creating an additional node with 2 children -- which made the original parent into a knuckle -- and then calling code to delete knuckles. So realizing that we were done when we got down to 2 children meant that we didn't have to create this additional parent and then delete it. Possibly deleting the knuckle-removing code fixed the hang, although I don't know exactly how.

Thank you so much for explaining this! Would it be possible to add a brief note to this effect to the corresponding place in `TopologyNode.cpp`? Just something like "we use > 2 rather than >= 2 because here the children are guaranteed to have a parent"?

davidcerny:
I guess that after learning all this about the distribution and how it's implemented, I might as well open a PR to document it...

PR #772, opened 2025-05-30
Run tutorial tests using the script from the website repo

bredelings:
I added some new tutorial tests.  However, the updated logic to change them to run only 1 generation was only in the `run_tutorial_tests.sh` script.

Change the `run_integration_tests.sh` script to run the tutorial tests via the `run_tutorial_tests.sh` script.  This has the benefit that:
- the code to run the tutorial tests is only in one place
- we use the same code to run the tutorial tests no matter whether we are running them from the revbayes repo or the website repo.


bredelings:
Actually I'm just going to add this minor change to the fix-multifurcations-crash branch.

Issue #767, opened 2025-05-28
[BUG] Error on posteriorPredictiveSimulation() in the latest version.


curiosusJR:
**Describe the bug**
Get 
```
Problem in Posterior Predictive Simulation:
Cannot modify the value of a clamped node.
```
in a `posteriorPredictiveSimulation()` run.

This error only occurs in the latest version (v 1.3.0). 
I have tested v1.2.5-preview5 and no error occurs.

**To Reproduce**
A test script edited from the tutorials in the revbayes website: 
```
### Read in sequence data for both genes
data = readDiscreteCharacterData("data.nex")

# Get some useful variables from the data. We need these later on.
num_taxa <- data.ntaxa()
num_branches <- 2 * num_taxa - 3
taxa <- data.taxa()

moves    = VectorMoves()
monitors = VectorMonitors()

# create a constant variable for the rate matrix
Q <- fnJC(4)

# Prior distribution on the tree topology
topology ~ dnUniformTopologyBranchLength(taxa,  dnExponential(10.0))
moves.append( mvSPR(topology, weight=num_taxa/10) )
TL := topology.treeLength()

# the sequence evolution model
seq ~ dnPhyloCTMC(tree=topology, Q=Q, type="DNA")
seq.clamp(data)

mymodel = model(Q)

# add monitors
monitors.append( mnScreen(printgen=1, TL) )
monitors.append( mnStochasticVariable(filename="output/posterior.var") )

# run the analysis
mymcmc = mcmc(mymodel, moves, monitors)
mymcmc.run(generations=1)

trace = readStochasticVariableTrace("output/posterior.var", delimiter=TAB)
pps = posteriorPredictiveSimulation(mymodel, directory="output/sims", trace)
pps.run()

q()
```

The `data.nex` could be any MSA file, like: 
```
#NEXUS

BEGIN DATA;
DIMENSIONS  NTAX=4 NCHAR=7;
FORMAT DATATYPE=DNA GAP=- MISSING=?;
MATRIX

seq1        ATGACCT
seq2        ATGACCA
seq3        ATGACGT
seq4        ATGACTT
;

END;
```

Get output in screen: 
```

RevBayes version (1.3.0)
Build from master (rapture-3959-g44f423) on Wed May 28 02:26:42 PM CST 2025
MPI enabled: using 1 processes.

Visit the website www.RevBayes.com for more information about RevBayes.

RevBayes is free software released under the GPL license, version 3. Type 'license()' for details.

[...]

PR #774, opened 2025-05-31
Allow redrawing values for clamped nodes when necessary

davidcerny:
This PR is intended to fix issue #767 by making sure that posterior predictive simulations can run (and determine what they should simulate by looking for clamped nodes in the model) while still throwing an error for code such as the following:

```r
x ~ dnNorm(1, 1)
x.clamp(1)
x.redraw()
```

I was surprised to find out that my [original idea](https://github.com/revbayes/revbayes/issues/767#issuecomment-2923845795):

> maybe we could add a boolean member variable to `StochasticNode`, check for it in `redraw()`, let `clamp()` set it to `false`, but provide a dedicated setter that could override it and that we could employ in `PosteriorPredictiveSimulation::run()`

was already mostly implemented: `StochasticNode` has a protected member variable called `ignore_redraw`, which is set to `false` by default and can be toggled by a setter called `.setIgnoreRedraw()`. However, we don't employ this setter anywhere except when implementing the `.redraw()` method for (continuous) stochastic nodes, where it is always set to `false` before a redraw is attempted.

Here, I have repurposed the `ignore_redraw` variable to make sure that: (1) it is always set to `true` by `.clamp()` and to `false` by `.unclamp()`; (2) it is _not_ automatically set to `false` when the user tries to call `.redraw()` directly from Rev; and (3) `PowerPosteriorSimulation` sets it to `false` for any clamped node it finds.

Note that the following still works:

```r
x ~ dnNorm(1, 1)
x.setValue(1)
x.redraw()
```

### DIFF
--- a/src/core/analysis/PosteriorPredictiveSimulation.cpp
+++ b/src/core/analysis/PosteriorPredictiveSimulation.cpp
@@ -131,6 +131,8 @@ void RevBayesCore::PosteriorPredictiveSimulation::run( int thinning )
             
             if ( the_node->isClamped() == true )
             {
+                the_node->setIgnoreRedraw(false);
+                
                 // check if the PP simulation must condition on sampled tip states
                 auto& ptr = the_node->getDistribution();
                 if (condition_on_tips == true && typeid( &ptr ) == typeid(StateDependentSpeciationExtinctionProcess))
--- a/src/core/dag/DagNode.cpp
+++ b/src/core/dag/DagNode.cpp
@@ -1022,18 +1022,28 @@ void DagNode::setHidden(bool tf)
 }
 
 
+void DagNode::setIgnoreData(bool tf)
+{
+    throw RbException()<<"Error: can't ignore data at node '"<<getName()<<"' because it is not a stochastic node!";
+}
+
+
+void DagNode::setIgnoreRedraw(bool tf)
+{
+    // nothing to do in the base class
+}
+
+
 void DagNode::setIntegrationIndex( size_t i )
 {
     // nothing to do in the base class
 }
 
 
-
 void DagNode::setName(std::string const &n)
 {
     // set the internal value
     name = n;
-
 }
 
 
@@ -1054,12 +1064,6 @@ void DagNode::setParentNamePrefix(const std::string &p)
 
 }
 
-
-void DagNode::setIgnoreData(bool tf)
-{
-    throw RbException()<<"Error: can't ignore data at node '"<<getName()<<"' because it is not a stochastic node!";
-
-}
 
 void DagNode::setVisitFlag(bool tf, const size_t flagType)
 {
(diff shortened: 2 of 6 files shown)


bredelings:
I kind of don't like this approach.  I don't think it is a good idea to introduce (or use) additional state variables like "allow redraw".  This goes from 2 states (clamped/unclamped) to 4 states (clamped-redraw, clamped-noredraw, unclamped-redraw,unclamped-noredraw).

I think we should instead allow making a copy of the model with all variables unclamped.

davidcerny:
> I think we should instead allow making a copy of the model with all variables unclamped.

As pointed out in #767, that won't work. The posterior predictive simulation wouldn't know what to simulate.

bredelings:
Good point.  The simple alternative would be to simulate every that isn't in the trace.  I left some wider thoughts on #767.

davidcerny:
I'm going to close this PR, since I agree it's not a very elegant solution to the problem. But your comments make me wonder if we should go further and get rid of the `ignore_redraw` variable altogether @bredelings. We check for its value in `StochasticNode::redraw()`, but when do we actually expect the check to fail? Currently (i.e., sans the changes proposed in this PR), the only operation that sets it to `true` is `.setValue()`. But `.setValue()` doesn't prevent a node's value from being moved during the MCMC (that would be counterproductive, since we mostly use it to set starting values), nor does it prevent us from redrawing it outside the MCMC (since a call to `.redraw()` resets `ignore_redraw` to `false`).

For what it's worth, `ignore_redraw` was originally added in [this commit](https://github.com/revbayes/revbayes.archive/commit/cfe1e36adbc422a3cdeb42275dd3daf0b9ff275d) by @mlandis, but I certainly wouldn't expect him to remember the exact reason for doing so 12 years later! 😛

davidcerny:
@bredelings So no, removing `ignore_redraw` breaks the coalescent tests. My guess is that it's because these make heavy use of `.setValue()`, which doesn't work properly without `ignore_redraw`. It's interesting, though, because there are other tests that include `.setValue()` (e.g., [`test_DEC`](https://github.com/revbayes/revbayes/blob/master/tests/test_DEC/scripts/mcmc_DEC.Rev), [`test_HSMRF_BDP`](https://github.com/revbayes/revbayes/blob/master/tests/test_HSMRF_BDP/scripts/mcmc_HSMRF_BDP.Rev), [`test_deterministic_index`](https://github.com/revbayes/revbayes/blob/master/tests/test_deterministic_index/scripts/test.Rev)), and these don't fail when `ignore_redraw` is removed. So we're either lucky that we just recently added the coalescent tests (because without them, the subtle changes in the behavior of `.setValue()` would go undetected), or the coalescent code is unusually fragile and potentially problematic...

Given the difference between `.setValue()` and `.clamp()`, I would say that we essentially already have 4 different states: fixed-redrawable, fixed-nonredrawable, random-redrawable, and (in theory, I guess) random-nonredrawable.

PR #778, opened 2025-06-07
Fix pps

bredelings:
Fix the posterior predictive simulation by copying the model, recording which nodes were clamped, and then unclamping them.

Additionally, 
- print a warning message if any unclamped nodes do not have a trace to provide posterior samples.
- unnest some of the `PosteriorPredictiveSimulation::run( )` code by one level.
- make a StochasticNodeBase so that we can dynamic_cast `DagNode*` to `StochasticNodeBase*`.  Alternatively, we could give `DagNode` an `unclamp()` method that only does something for `StochasticNode`s

Although unnesting some code makes it look like more lines have changed, the PR does not make major changes and is not very complicated.  I think the only interesting part is the decision to put the `unclamp()` method only on StochasticNode instead of on DagNode.  This requires the new `StochasticNodeBase` class, because we can't cast to `StochasticNode<T>` directly when we don't know the type `T`.


bredelings:
Fixes #767

davidcerny (approved):
This all looks great to me. The code for copying the model and iterating over the nodes of the copy is clear and readable, and solves the issue in an elegant way. The new check for untraced unclamped nodes is a welcome addition, too.

I guess the fact that the test runner couldn't handle subdirectories within `output_expected` was the reason why we didn't have PPS integration tests until now. Thank you for solving this as well!

PR #796, opened 2025-07-25
added nanoseconds option in time()

prilau:
I added an option in time() function to allow benchmarking with nanosecond-level precision.


bredelings (changes requested):
This looks good.

Would you be able to add a small test?  I would recommend adding `tests/test_basics/scripts/time.Rev` and making it do a few calls to time like `time()`, `time("millseconds")`, and `time("nanoseconds")`.  You wouldn't need to add anything to `test_basics/expected_output/` since the output will be different each time the test is run.

PR #798, opened 2025-07-29
Reject tree moves if there are no nodes for them to operate on

### LINKED ISSUE #795: [BUG] mcmc stuck in fossil only inference (with/out data)
**Describe the bug**
I run a fossil only inference using FBD with rho = 0.0 to model observing fossils during an interval [t_1, t_2], but knowing the taxa survive after t_2 plus there is no special sampling effort at t_2. In many cases, the mcmc gets stuck, be it in the burnin/tuning interval or the main run. This is the case running both with data, under prior (using `.ignoreAllData`) or using not clamping any nodes in the first place (see example below)

**To Reproduce**

Minimal running example of the effect - gets stuck after approx. 50 steps on my machine. No further outputs to the console, no syscalls, and CPU usage remains at 100 %.

[fossils_cont_rho0_1.csv](https://github.com/user-attachments/files/21389084/fossils_cont_rho0_1.csv)

```
seed(1)
taxa <- readTaxonData( "fossils_cont_rho0_1.csv")

moves = VectorMoves()

origin_time <- 2.0
psi ~ dnExponential(lambda = 1.0)
moves.append( mvScale(psi, weight=15.0) )

fbd_tree ~ dnFBDP(origin=origin_time, lambda=1.0, mu=0.3, psi=psi, rho=0.0, taxa=taxa)

moves.append( mvFNPR(fbd_tree, weight=200.0) )
moves.append( mvCollapseExpandFossilBranch(fbd_tree, origin_time, weight=300.0) )
moves.append( mvNodeTimeScale(fbd_tree, weight = 20.0))

phy_morpho ~ dnPhyloCTMC(tree = fbd_tree, branchRates = 0.1, Q = fnJC(2), type = "Standard", coding = "all")
mymodel = model(phy_morpho)

monitors = VectorMonitors()
monitors.append( mnScreen(printgen=10) )
monitors.append( mnModel(filename="num.log", printgen=10) )
monitors.append( mnFile(filename = "trees.log", printgen = 10, fbd_tree))

mymcmc = mcmc(mymodel, monitors, moves, nruns = 2, ntries = 10000)
mymcmc.run(generations=2000)
q()
```

**Expected behavior**
mcmc progresses with corresponding output to the console, or an error is thrown flagging the underlying problem.

**Computer info**
Running on WSL2 (Ubuntu 24.04.2 LTS), revBayes 1.3.0

davidcerny:
This is intended to close #795.


src/core/moves/proposal/tree/NodeRateTimeSlideUniformProposal.cpp
```diff
@@ -117,14 +117,31 @@ double NodeRateTimeSlideUniformProposal::doProposal( void )
...
+    }
 
     // pick a random node which is not the root, a tip, or the parent of a SA
     TopologyNode* node;
     do {
         double u = rng->uniform01();
         size_t index = size_t( std::floor(tau.getNumberOfNodes() * u) );
         node = &tau.getNode(index);
-    } while ( node->isRoot() || node->isTip() || node -> isSampledAncestorParent() );
+    } while ( node->isRoot() || node->isTip() || node->isSampledAncestorParent() );
```
bredelings, line 144: If we've gone to all the work to put the eligible nodes into a vector, then we can just select an index from the vector instead of doing this weird while loop.

davidcerny: Makes sense!

bredelings:
I think it might be good to put the code that selects random nodes that meet certain conditions into a separate function, and then just call that function in the various places that want it.  There are few different condition types, so we could:
- provide one variant with flags saying which nodes should be included/excluded.
- provide one variant with a lambda function that specifies which nodes are OK
- provide a few variants of the same idea. possibly they could call a single backend with flags or lambda condition.

bredelings:
For large trees, putting all the eligible nodes into a vector take O(N) time, but randomly choosing a node takes O(1) time.

In theory then, we could try the loop for (say) 10 iterations, and then fall back to putting things in the vector, in order to get the best of both worlds.

But this could be a premature optimization - possibly the O(N) time is fine, and so we shouldn't make things more complicated until there's evidence that its worth it.  :man_shrugging:

src/core/datatypes/trees/Tree.cpp
```diff
@@ -1585,6 +1587,48 @@ void Tree::collapseSampledAncestors()
...
+TopologyNode* Tree::pickRandomInternalNode(RandomNumberGenerator* rng) const
+{
+    TopologyNode* node;
+    int counter = 0;
+    
+    do {
+        double u = rng->uniform01();
+        size_t node_idx = size_t( std::floor(getNumberOfNodes() * u) );
+        node = (TopologyNode*)&getNode(node_idx);
+        counter++;
```
bredelings, line 1601: We could simplify this further by just making it a for loop that counts to 10, and then after we sample a node we just `return` immediately if it satisfies the condition:

```
if ( not (node->isRoot() || node->isTip() || node->isSampledAncestorParent() ))) return node;
```

Then we could also remove the `if (counter == 10)` condition and unnest the code inside it.

Issue #791, opened 2025-07-17
[BUG] Reading in a newick tree with `readTrees` drops all the attributes


bredelings:
```
> t = readTrees("bears.mcc.tre")[1]
   Attempting to read the contents of file "bears.mcc.tre"
   Successfully read file
> t
  
   ((((((((((Ailuropoda_melanoleuca[&index=20]:5.838853,Indarctos_punjabiensis[&index=19]:0.000000)[&index=21]:0.914040,Indarctos_arctoides[&index=18]:0.000000)[&index=22]:1.497574,Indarctos_vireti[&index=17]:0.000000)[&index=23]:3.931463,((Ailurarctos_lufengensis[&index=16]:0.766227,Agriarctos_spp[&index=15]:0.000000)[&index=24]:4.013926,Kretzoiarctos_beatrix[&index=14]:0.000000)[&index=25]:0.772060)[&index=26]:3.155917,Ursavus_primaevus[&index=13]:0.000000)[&index=27]:1.077981,Ursavus_brevirhinus[&index=12]:0.000000)[&index=28]:1.687663,(((((Ursus_arctos[&index=11]:1.281101,Ursus_maritimus[&index=10]:1.281101)[&index=29]:1.543555,Ursus_abstrusus[&index=9]:0.000000)[&index=30]:1.014655,(Helarctos_malayanus[&index=8]:2.880817,Ursus_thibetanus[&index=7]:2.880817)[&index=31]:0.958494)[&index=32]:1.940361,Melursus_ursinus[&index=6]:5.779672)[&index=33]:3.386237,(Arctodus_simus[&index=5]:2.876760,Tremarctos_ornatus[&index=4]:4.478447)[&index=34]:4.687462)[&index=35]:8.937582)[&index=36]:3.110199,Ballusia_elmensis[&index=3]:0.000000)[&index=37]:7.257118,Zaragocyon_daamsi[&index=2]:6.985548)[&index=38]:8.187390,Parictis_montanus[&index=1]:0.000000)[&index=39]:0.000000;

```

[bears.mcc.tre.gz](https://github.com/user-attachments/files/21299551/bears.mcc.tre.gz)

bredelings:
This seems to be because trees are always read as "nexus" instead of "newick", and then `NclReader::translateNclSimpleTreeToBranchLengthTree( )` ignores the `unprocessedComments` field:

```
(gdb) p *nTree.allNodes[0]
$17 = {
  scratch = 0x0,
[...]

PR #799, opened 2025-07-29
Newick escaping

raymondcast18:
added proper character escaping to newick strings

### DIFF
--- a/src/core/datatypes/trees/TopologyNode.cpp
+++ b/src/core/datatypes/trees/TopologyNode.cpp
@@ -407,7 +407,7 @@ std::string newickEscape(const std::string in_label)
         char c = in_label[i];
 
         // first check for any illegal characters
-        if (c == '\'' || c == ' ' || c == '(' || c == ')' ||
+        if (c == '\'' || c == '(' || c == ')' ||
             c == '[' || c ==']' || c == ':' || c == ';' || c ==',') {
             needs_quoting = true;
             // check for single quote
--- a/src/core/datatypes/trees/TopologyNode.cpp
+++ b/src/core/datatypes/trees/TopologyNode.cpp
@@ -395,6 +395,49 @@ void TopologyNode::addNodeParameters(std::string const &n, const std::vector<std
     }
 }
 
+/**
+ * This function will convert a raw string to a newick sting with escaped characters.
+ */
+std::string newickEscape(const std::string in_label)
+{
+    bool needs_quoting = false;
+    std::string escaped;
+
+    for (size_t i = 0; i < in_label.size(); ++i) {
+        char c = in_label[i];
+
+        // first check for any illegal characters
+        if (c == '\'' || c == ' ' || c == '(' || c == ')' ||
+            c == '[' || c ==']' || c == ':' || c == ';' || c ==',') {
+            needs_quoting = true;
+            // check for single quote
+            if (c == '\'') {
+                escaped += "''";
+            }
+            else {
+                escaped += c;
+            }
+        }
+
+        else{
+            escaped += c;
+        }
+    }
+
+    //return quoted string
+    if (needs_quoting) {
+        return "'" + escaped + "'";
+    } 
+    // convert blanks to underscores
+    else{
+        for (char& c : escaped) {
+            if (c == ' ') {
+                c = '_';
+            }
+        }
+        return escaped;
+    }
+}
 
 std::ostream& TopologyNode::buildNewick( std::ostream& o, bool simmap = false)
 {
@@ -421,7 +464,7 @@ std::ostream& TopologyNode::buildNewick( std::ostream& o, bool simmap = false)
 
     // 2. Write out the node name is there is any.
     if (children.size() < 2)
-        o << taxon.getName();
+        o << newickEscape(taxon.getName());
 
     // 3. Write out node comments if there are any and (simmap == false)
     if ( ( node_comments.size() > 0 or RbSettings::userSettings().getPrintNodeIndex() == true ) && simmap == false )
@@ -2287,4 +2330,5 @@ std::pair<double,double> getStartEndAge(const RevBayesCore::TopologyNode& node)
 
     return {start_age, end_age};
 }
+
 


bredelings:
This looks good and all the tests pass.

Before its merged, it probably needs a test that checks that some newick strings are correctly escaped.

src/core/datatypes/trees/TopologyNode.cpp
```diff
@@ -395,6 +395,49 @@ void TopologyNode::addNodeParameters(std::string const &n, const std::vector<std
...
+{
+    bool needs_quoting = false;
+    std::string escaped;
+
+    for (size_t i = 0; i < in_label.size(); ++i) {
+        char c = in_label[i];
+
+        // first check for any illegal characters
+        if (c == '\'' || c == '(' || c == ')' ||
+            c == '[' || c ==']' || c == ':' || c == ';' || c ==',') {
```
bredelings, line 411: It occurs to me that we also need to check for `_`, since it will be interpreted as a space if we don't quote.

davidcerny: Why would that be the case? It looks like we break a huge number of tests when we add a check for underscores.

bredelings: If you don't check for underscore, the `a b` and `a_b` are both written as `a_b`.  Then when you read them back in, they both end up as `a b`.  So, not quoting underscores means that when you write and re-read, you convert any underscores to spaces.

One possibility is that RevBayes might be reading Newick incorrectly.  If so, then the changes here may require corresponding  changes to how Newick is read.

bredelings:
@raymondcast18 and I found the revbayes code to read Newick trees in `core/io/NewickConverter.cpp` and it is kind of a hack.  I have written similar hacks before, so no criticism implied.  But we can definitely improve it.

Fixing this should fix some of the breakage caused by writing Newick trees _correctly_.

src/core/datatypes/trees/TopologyNode.cpp
bredelings: The current resolution is to NOT treat unquoted underscores as a space.

davidcerny: @bredelings What about the other direction? You mentioned back in August that

> If you don't check for underscore, the `a b` and `a_b` are both written as `a_b`

It seems to me that quoting the name that contains a space is preferable to automatically replacing that space by an underscore.

davidcerny:
@bredelings 

> One test that I am unsure about is `tests/test_nonBifurcating/output_expected/branch_length_tree_resolve_then_suppress.nex`. The newick looks different, but I am not sure if the new newick is incorrect or just different.

Hmm... it's complicated. The old newick had the root in the same position as the original, unresolved tree (`tests/data/initial_tree.newick`), whereas the new newick has it in a different position. But... does that matter? When we read in the unresolved tree as a non-clock tree (= branch length tree), we automatically unroot it. But then we apply the `.resolveMultifurcations()` method to it with the `resolveRoot` argument set to `TRUE`, which roots it again, and it just so happens to be in a different position than where the root was previously.

It looks to me like this is a mistake that I made in the test script: the `resolveRoot` argument should be set to `FALSE`, since we want to get a branch length tree back, and we are even explicitly marking its newick as unrooted (`[&U]`). Indeed, that's what we do in the converse case of suppressing outdegree-1 nodes first, and resolving multifurcations second (cf. `tests/test_nonBifurcating/output_expected/branch_length_tree_suppress_then_resolve.nex`).

davidcerny:
By the way, does this PR have any bearing on issue #791?

bredelings:
This doesn't fix #791.  If I take the tree from that issue and convert it to newick, then actually the attributes are loaded.  I made the comment-splitter for Newick not split on comments inside `{...}`, so that should work.

It is only when we load as NEXUS that the attributes are lost.

bredelings:
I think this PR is ready to review now.

@ms609 If you want to check that it can load any problematic newick files that you have, then we could fix any errors that you find and maybe make them into tests or something.

bredelings:
Yeah, I saw that.  One question is where exactly the branch length gets set to zero, so that we can remove that.  Using gdb/lldb could help track this down.

This PR makes that issue visible, but doesn't really cause the issue, so it could be fixed afterward.  I'd also like to try fixing the root branch length being zero, but that seems more complex, so I definitely want to do that separately/afterward.

davidcerny:
Pretty sure it's happening here:

https://github.com/revbayes/revbayes/blob/db9aca3168a4c71199b28d6b2738917c1de79da7/src/core/datatypes/trees/TopologyNode.cpp#L1911-L1913

Maybe we could check whether the `children` (which represent pre-existing node) have the branch length attribute, and only set it for the newly created `prnt` node if they do?

bredelings:
Hmm... it seems like you could have (say) a multifurcation at the root, where the root node has no branch length, but the children of the root all have branch lengths.  Am I understanding this correctly?

In that case the code you added would NOT set a zero branch length -- but it seems like it should.  So maybe it would work better to set a zero branch length when all _children_ have branch lengths (i.e. the branch length of all children is not NaN).  If some children have branch lengths but some do not then I think either choice is OK, but I would lean toward not setting the branch length to zero in that case.

davidcerny:
That makes sense. Feel free to rewrite that commit – i'm not entirely happy with this solution anyway, because as a result of repeated rooting, `branch_length_tree_all_resolved.nex‎` still has _two_ zero branch lengths in a tree that's otherwise devoid of branch lengths.

bredelings:
Hmm... for the moment, lets drop that from this PR.  It really is logically independent.  If we have a completely clean solution we could include it, but if there is any complexity it should probably be disentangled by going in a separate PR.

I'd like to do a second wave of tree cleanups that involve at least:
- not setting the root branch length to 0 in recomputeBranchLengths
- defaulting TopologyNode::use_ages to `false`

and maybe more.  I initially tried to do those at the same time as this PR but the interactions were getting too complex, so I decided it was best to get this in first, and then address the other issue.

So thematically a full solution to this problem might fit there, where we try to eliminate ALL unnecessary :0.0000 branch lengths.

bredelings:
Thanks for fixing the nonBifurcating test!

src/core/datatypes/trees/TopologyNode.cpp
davidcerny: I guess I can answer my own question here:

- When reading a Newick tree with unquoted spaces in tip labels:

    ```
    (((Lemur_catta[&index=5]:53.464817,Galago senegalensis[&index=4]:53.464817)[&index=6]:2.326349,Tarsius_syrichta[&index=3]:55.791166)[&index=7]:22.397696,(Pan_paniscus[&index=2]:67.692728,Alouatta_palliata[&index=1]:67.692728)[&index=8]:10.496134)[&index=9]:0.000000;
    ```
    
    RevBayes fails to read the file and throws an informative error. For comparison:
    - `ape::read.tree()` reads in the tree but removes the space (reducing the name to `Galagosenegalensis`)
    - FigTree handles the space just fine but mistakenly includes the `index` annotation in the tip labels. (This doesn't happen with Nexus trees, only with plain Newick.)
    
- When reading a Newick tree with quoted spaces in tip labels:

    ```
    (((Lemur_catta[&index=5]:53.464817,'Galago senegalensis'[&index=4]:53.464817)[&index=6]:2.326349,Tarsius_syrichta[&index=3]:55.791166)[&index=7]:22.397696,(Pan_paniscus[&index=2]:67.692728,Alouatta_palliata[&index=1]:67.692728)[&index=8]:10.496134)[&index=9]:0.000000;
    ```
    
    RevBayes reads in the tree without issues. When listing the tip labels using the `.taxa()` method, the single quotes are not displayed. However, when the tree is printed to a Nexus file, the single quotes are retained, so the space is properly escaped. For comparison:
    
    - `ape::read.tree()` reads in the tree but displays the single quotes when listing the tip labels using `$tip.label` (showing the name as `'Galago senegalensis'`)
    - FigTree reads in the tree without issues and doesn't display the single quotes around the name, though it still includes the annotation in the label.
    
This seems fine. It would be cool if we could automatically escape unquoted spaces, like FigTree can, but then again, those violate the Newick standard (or at least [Gary Olsen's interpretation](https://phylipweb.github.io/phylip/newick_doc.html) thereof), so it's not essential.

src/core/io/NewickConverter.cpp
```diff
@@ -27,65 +33,706 @@ NewickConverter::~NewickConverter()
...
+    };
+
+    // 3. Extract the current line
+    size_t lineEnd = input.find('\n', lineStart);
+    if (lineEnd == std::string::npos) lineEnd = input.length();
+    auto current = processLine(lineStart, lineEnd, pos, true);
+
+    // 4. Build Output
+    std::string result = "Newick error at line " + std::to_string(line) + ", column " + std::to_string(visualColumn) + ":\n";
+
```
davidcerny, line 123: The word "column" can be a little confusing here – most people probably wouldn't think of a Newick string as a table. "Character" might be preferable.

bredelings: Interesting.  I think "Line" and "column" are the standard way of referring to positions in a text file, for example in error messages.  With fixed-with fonts, characters do fit in a grid.

I asked Gemini about this, and there can actually be a slight difference between "column 10" and "character 10" if the line contains tabs, since a tab is one character but spans 8 columns.  (I chose 8 since that seems to be what my terminal does.)

Gcc shows this:
```
a.cc:4:2: error: ‘df’ does not name a type
    4 | ;df
      |  ^~
```
so it avoids naming line and column at all.

We are currently doing this:

```
   Error:	Newick error at line 1, column 4:
   | (a b:1.0, c:1.0);
        ^
```

davidcerny: Good to know!

> I asked Gemini about this, and there can actually be a slight difference between "column 10" and "character 10" if the line contains tabs, since a tab is one character but spans 8 columns.

That seems to support my point then, no?

bredelings: I was thinking the opposite.  Like, if you have an error at the b in the line `a\tb`, then the normal thing would be to print `a        b` and say that the error occurs in column 10.

bredelings: This is what gcc does:
```
a.cc:1:9: error: ‘asdf’ does not name a type
    1 |         asdf;
      |         ^~~~
a.cc:2:9: error: ‘lwn’ does not name a type
    2 |         lwn;
      |         ^~~
a.cc:3:9: error: ‘why’ does not name a type
    3 |         why;
      |         ^~~
```
All these things are in column 9, but would be character 2, because gcc is treating a tab as 8 spaces apparently.

davidcerny: Okay – no biggie either way.

src/core/datatypes/trees/TopologyNode.cpp
```diff
@@ -397,6 +407,49 @@ void TopologyNode::addNodeParameters(std::string const &n, const std::vector<std
...
+ */
+std::string newickEscape(const std::string in_label)
+{
+    bool needs_quoting = false;
+    std::string escaped;
+
+    for (size_t i = 0; i < in_label.size(); ++i) {
+        char c = in_label[i];
+
+        // first check for any illegal characters or underscore
```
davidcerny, line 421: This comment is now out-of-date

davidcerny (approved):

PR #802, opened 2025-07-31
Emergency follow-up to #798

davidcerny:
I got a little too excited in commit 66d9285 about simplifying code and placed the step of drawing a random double outside the `for` loop in which it belonged, causing the loop to draw the same node 10 times instead of drawing (up to) 10 distinct nodes. Please accept this PR, and my apologies.


bredelings (approved):
Oops! :-P

PR #803, opened 2025-07-31
Clean up command line arguments

bredelings:
This PR restructures the command-line arguments to RevBayes.  As we discussed, the new syntax follows Rscript. 
1. if you give no arguments, then rb is interactive.
2. if you give `-e expr` arguments, then rb evaluates the expressions.  These expressions could include `-e 'source("filename")'`.
3. with no `-e expr` arguments, the first positional argument is treated as a filename.

```
Bayesian phylogenetic inference using probabilistic graphical models and an 
interpreted language 

Usage: rb [options]
       rb [options] file [args]
       rb [options] -e expr [-e expr2 ...] [args]

OPTIONS:
  -h,      --help                  Print this help message and exit 
  -v,      --version               Show version and exit 
  -b,      --batch                 Deprecated 
  -j,      --jupyter               Run in jupyter mode 
  -q,      --quiet                 Hide startup message (if no file or -e expr) 
  -i,      --interactive           Force interactive (with file or -e expr) 
  -p,      --echo                  Print commands (with file or -e expr) 
  -c,      --continue              Continue after error (with file or -e expr) 
  -s,      --seed UINT             Random seed 
  -o,      --setOption TEXT ...    Set an option key=value (See ?setOption for 
                                   the list of available keys and their 
                                   associated values) 

Expressions (one or more '-e <expr>') may be used *instead* of 'file'. 
```
For the 2nd and 3rd form, we assume that rb is non-interactive by default, but you can specify `-i` to force it to be interactive.  The `--batch` option can be removed in a follow-up change, but removing it immediately breaks the tutorial tests.  The `--seed` option makes it easy to run a script with a different random seed.

If revbayes is interactive, then:
- the header is printed (can be suppressed with `-q`)
- by default `rb` does _not_ exit on error (continueOnError = TRUE)
- `rb` listens to STDIN for further commands, which means:
   - it is possible to send revbayes commands using a pipe, e.g. `echo "1 + 2" | rb`
   - `q()` is needed to quit

If revbayes is **not** interactive:
- the header is _not_ printed.
- by default user commands are not echoed to the screen (echo = FALSE)
- by default `rb` exits on the first error (error-exit = TRUE)
- by default  `rb` does not listen to STDIN for further commands (interactive = FALSE), which means
   - you cannot send revbayes commands using a pipe
   - `q()` is not needed

These defaults for "script or -e expr" can be overridden with `--echo`, `--continue`, and `--interactive`.

Additionally, there are new settings:
- `echo`
- `continueOnError`

that can be set with `setOption( )`.  By calling `setOption("continueOnError","TRUE")`, a script can avoid exiting on the first error.  This will be useful for test scripts that want to test multiple errors.


src/core/utils/RbSettings.cpp
```diff
@@ -65,6 +66,14 @@ int RbSettings::getLogMCMC( void ) const
     return logMCMC;
 }
 
+std::string bool_to_string(bool b)
+{
+    if (b)
+        return "TRUE";
+    else
+        return "FALSE";
```
davidcerny: Could be a one-liner with the ternary operator

davidcerny (changes requested):
This review probably shouldn't be taken as definitive, since I will need to play with the branch some more, but I'm very impressed by the new interface. I have to admit – I was initially skeptical that this would be a good use of your time, Ben, since I thought the old interface was mostly fine and just needed to be documented better. But the `-e` flag in particular is extremely useful and will make shell scripting with RevBayes much more powerful.

Some change requests and suggestions:

- Not all the changes to `build.yml` are paralleled in `release.yml`. Is this intentional? If we are going to have to bring them in sync for the next release, perhaps we might as well do it right away...
- The fact that `-p` and `-x` have different defaults in the interactive mode vs. when executing a script, and consequently always have to be followed by a boolean (unlike `-i` and `-q`), is very difficult to figure out for a user who doesn't have this PR open in front of them. I would strongly recommend including this information in the `--help` page. In fact, listing the defaults for all four boolean options (`-i`, `-q`, `-p`, `-x`) would be useful.
- Just saying that the `--batch` option is deprecated is not very helpful. I'd suggest either providing a very brief explanation of what it does, or removing it altogether.
- The interface is a little inconsistent in its treatment of unrecognized options. We have at least three scenarios: (1) supplying an invalid flag without a file, (2) supplying it with a file that successfully executes, and (3) supplying it with a file that throws an error. In scenario (1), running `./rb --args 3 5` prints the following nice message:

    ```
    Error: unrecognized option '--args'
    Run with --help for more information.
    ```
    
    In scenario (2), assume the file `test.Rev` just contains the expression `3 + 5`, and we run `./rb test.Rev --args 3 5`. In this case, RevBayes just prints `8` to stdout and ignores the unrecognized flag entirely, which is not terrible either. In scenario (3), assume the file `test.Rev` contains the expression `args[1] + args[2]`, and we again call RevBayes as follows: `./rb test.Rev --args 3 5`. In the old interface, this is actually an alternative way of getting RevBayes to print `8`, but in the new interface, we get the following somewhat mysterious output:
    
    ```
       --args3
    ```

bredelings:
> Not all the changes to build.yml are paralleled in release.yml. Is this intentional? If we are going to have to bring them in sync for the next release, perhaps we might as well do it right away...

Not exactly intentional.  I guess I was going to wait until the changes were mostly done to synchronize them.  Since I'm mostly done, I may as well synchronize them now.

> Just saying that the --batch option is deprecated is not very helpful. I'd suggest either providing a very brief explanation of what it does, or removing it altogether.

The `-b` options will go away completely, but I can't do it in this PR because the tutorial tests all fail if `-b` isn't a valid argument.  So the plan is to merge this PR, then modify the tutorial tests to remove `-b`, then remove the `-b` option altogether.  Given that this is just a temporary placeholder, I think we can solve the poor description by just waiting until the `-b` options is removed.

> The interface is a little inconsistent in its treatment of unrecognized options. We have at least three scenarios: (1) supplying an invalid flag without a file, (2) supplying it with a file that successfully executes, and (3) supplying it with a file that throws an error. In scenario (1), running ./rb --args 3 5 prints the following nice message:

I think I may have a bug here.  What is supposed to happen is that any word after `-e expr` or the script name goes into `args`.  However, when looking into this, it seems like only _unrecognized_ options go into `args`.  Let me see if I can fix this.  The CLI11 library seems much more flexible than boost program_options, but it has been a little challenging figuring out how to get it to do what I want.

> There is also some inconsistency: the defaults for -q are also different between the two use modes, but since the flag doesn't accept a boolean, we can't force it to be FALSE when executing a script. Note that printing the header to stdout during script execution would actually be a really neat way of recording exactly what version of RevBayes was used to run the analysis, potentially improving reproducibility.

I guess I didn't think anyone would want this.  But let me see if I can implement it.

> The fact that -p and -x have different defaults in the interactive mode vs. when executing a script, and consequently always have to be followed by a boolean (unlike -i and -q), is very difficult to figure out for a user who doesn't have this PR open in front of them. I would strongly recommend including this information in the --help page. 

If I understand correctly, you asking for the defaults (and their dependence on interactive mode) to be documented in the --help page?  That makes sense... but let me know if I didn't understand your comment, or didn't fully understand it.

src/core/utils/RbSettings.cpp
bredelings: done.

bredelings:
OK, I think I've fixed the command-line arguments.  Anything after `-e expr` or `scriptname` will be treated as an argument to the script, not an argument to `rb`.  So, `rb -e 'print(args[1])' --args 1 2 3` should print `--args`.  And (unlike before) `rb -e 'print(args[1])' -h 1 2 3` should print `-h`.

About making `-q` take a boolean, I'm still debating with myself.  R only takes `-q`, but doesn't seem to take any boolean options, so that doesn't clarify much.

What seems useful to me is:
- in non-interactive mode, continue after errors.  (making interactive mode quit on the first error seems less helpful)
- in non-interactive mode, echo the command. (making interactive mode NOT echo the command seems less helpful)
- with a script or expr, force interactive mode. (forcing non-interactive mode with no script or expr seems useless)
- in interactive mode, suppress the header. (making non-interactive mode print a header seems less useful.)

If people want to record a version for their scripts, maybe we should add a function `version()` that prints the version?

So then we could have:
- `-i` forces interactive mode for scripts/expressions
- `-q` disables printing a header in interactive mode
- `-c` makes the scripts/expressions continue after the first error.  (Could there be a use for quitting on the first error in interactive mode?  Maybe if we have fake interactivity from piping in commands on STDIN?)
- `-p` enables echo for scripts.  ( Interestingly, we don't actually echo for interactive mode.  We rely on the terminal to print commands that the user types.  If we did echo, commands would be printed twice I think.  Or, if we pipe in commands in interactive mode, then maybe echoing would print the commands once...)

Then we'd *consistently* have no boolean flags.  Although I guess people could still do `-e setOption("interactive","FALSE")`.  Hmm....

Should we really plan for people to do scripts by using interactive mode and piping in scripts?  I kind of think not -- interactive mode should actually be interactive.  Maybe we should just remove `-q`, since if we really want to run something quietly we could just run a script or expression?

Overall, I am still of two minds about the boolean options thing.

davidcerny:
> The `-b` options will go away completely, but I can't do it in this PR because the tutorial tests all fail if `-b` isn't a valid argument. So the plan is to merge this PR, then modify the tutorial tests to remove `-b`, then remove the `-b` option altogether. Given that this is just a temporary placeholder, I think we can solve the poor description by just waiting until the `-b` options is removed.

Makes perfect sense; I agree.

> I think I may have a bug here. What is supposed to happen is that any word after `-e expr` or the script name goes into `args`.

Ah – I didn't understand at first that options have to strictly precede the script or expression to be executed, although now that I'm looking at it, the `--help` page is perfectly clear about that. For what it's worth, I don't think that sending into `args` only those words that can't be parsed as options would necessarily be a bad idea, but I'll defer to your preferences on that.

> If I understand correctly, you asking for the defaults (and their dependence on interactive mode) to be documented in the --help page?

Yes, that's exactly what I was requesting; apologies if I wasn't being clear.

> R only takes `-q`, but doesn't seem to take any boolean options, so that doesn't clarify much.

In my opinion, it's okay to deviate from what R does if it makes sense for our purposes.

> What seems useful to me is:
> 
> - in non-interactive mode, continue after errors. (making interactive mode quit on the first error seems less helpful)
> - in non-interactive mode, echo the command. (making interactive mode NOT echo the command seems less helpful)
> - with a script or expr, force interactive mode. (forcing non-interactive mode with no script or expr seems useless)
> - in interactive mode, suppress the header. (making non-interactive mode print a header seems less useful.)

I agree with all of this except for your very last statement: again, I could see printing a header in non-interactive mode as a neat way of recording the version / commit / branch that the script was executed with. But yes, implementing a separate function to do that would be a good alternative. Monitors already have a `version` argument for that, but it's much more practical to print the version info to stdout than to let it mess up log file headers...

> Interestingly, we don't actually echo for interactive mode.

Yes, I think that option is inherently applicable only to scripts and `-e` expressions.

> Then we'd _consistently_ have no boolean flags.

I like this solution. A less elegant alternative would be to make all four flags (`-i`, `-q`, `-x`, `-p`) boolean, although that would require handling the weird and useless cases of `./rb -i FALSE` and `./rb -p TRUE`. We could perhaps do that by throwing an error – something like "-\<bad option\> can only be used with script or expression".

In any case, I would suggest that the `--help` page should either (1) mention that certain flags only have effect in one mode or the other, if there are no boolean flags; or (2) document the different defaults, if boolean flags do persist.

bredelings:
>  For what it's worth, I don't think that sending into args only those words that can't be parsed as options would necessarily be a bad idea, but I'll defer to your preferences on that.

The idea here is that if you do `rb script --help`, then `script` should see `--help`.  This allows you to start your script with `#! /path/to/rb` and make it executable.  If rb steals arguments from after script, then the script cannot see hose arguments.  So lets keep it the way it is.

I changed the flags to take no arguments, and make it clear that --interactive, --echo, and --continue apply when rb is given a script or `-e expr`, and to make it clear that `-q` applies when `rb` is not given a script or `-e expr`.

In theory I could add a function `version()` that returns the version as a string.  But I feel like that is beyond the scope of this PR, and could be done as a followup.

src/core/help/RbHelpDatabase.cpp
```diff
@@ -4298,48 +4298,60 @@ a # this will be the same as above!)");
...
+
+        DEFAULT: TRUE if no script or -e expr is given, otherwise FALSE.
+
+    echo=<TRUE,FALSE>            Should commands be printed to the screen?
+
+        DEFAULT: TRUE if interactive, otherwise FALSE.
+
+    errorExit=<TRUE,FALSE>       Should we exit on the first error?
+
+        DEFAULT: FALSE if interactive, otherwise otherwise TRUE.)");
```
bjoelle: double otherwise here

src/core/utils/RbSettings.cpp
```diff
@@ -287,11 +323,28 @@ void RbSettings::setOption(const std::string &key, const std::string &v, bool wr
...
+            continue_on_error = *b;
+        else
+            throw RbException()<<"setOption: expected a boolean, but got '"<<value<<"'";
+    }
     else
     {
         std::cout << "Unknown user setting with key '" << key << "'." << std::endl;
     }
 
+    // Maybe we should just write this particular setting instead of all settings?
```
bjoelle, line 310: or just set write = false if what we're changing is echo or continueOnError ? since we are not writing these

src/core/help/RbHelpDatabase.cpp
davidcerny: Huh – this seems to be outdated, but GitHub doesn't mark it as such...

PR #804, opened 2025-07-31
Revamping rerooting

### LINKED ISSUE #738: [BUG] `.reroot()` reassigns branch annotations to the wrong branches
**Describe the bug**
This one is really unfortunate. Rerooting a summary (e.g., MAP or MCC) tree causes the automatically generated branch annotations to become associated with the wrong branches, resulting in things like clades receiving incorrect posterior probabilities.

**To reproduce**
[This zipped directory](https://github.com/user-attachments/files/19875283/tree_traces.zip) contains two tree traces, produced by a nonclock analysis that used taxon "t83" as the outgroup. Since [outgroup specification has no effect on tree summaries](https://github.com/revbayes/revbayes/issues/694), the following code will produce an MCC tree arbitrarily rooted on the first taxon in alphabetical order ("t1"):

```cpp
trace = readTreeTrace("anonymized_nonclock.trees", treetype="non-clock", nruns=2, burnin=0)
mcc_tree = mccTree(trace)
writeNexus(filename="anonymized_nonclock.mcc.tre", mcc_tree)
```

The resulting tree contains branch annotations including posterior probabilities. If we reroot it on "t83" in FigTree, the association between individual branches and their annotations is correctly preserved. In particular, note that the ingroup (= the clade comprising all taxa except "t83" and "t85") has a posterior probability of 1, as expected:

Image
<br><br>

Rerooting in RevGadgets works as well:

```r
library(ggtree)
library(RevGadgets)

MCC <- RevGadgets::readTrees("anonymized_nonclock.mcc.tre", verbose = F)
MCC <- RevGadgets::rerootPhylo(MCC, "t83")
MCC <- MCC[[1]][[1]]

p <- ggtree(MCC) + geom_tiplab(size = 3, offset = 0.01)
p + geom_nodelab(data = p$data[-c(1:(Ntip(MCC) + 1)), ],
                 aes(color = posterior, label = round(posterior, 2)),
                 geom = "label", size = 2, hjust = 0.8, alpha = 0.8,
                 label.padding = unit(0.15, "lines"))
```

Image
<br><br>

However, when we try to reroot directly in RevBayes, after finding the MCC tree but before printing it to a file, the annotations are shifted one node up, so that the ingroup's PP of 1 gets assigned to the root (for which the notion of a posterior probability is obviously meaningless), and the ingroup "inherits" the annotation from the branch immediately below:

```cpp
trace = readTreeTrace("anonymized_nonclock.trees", treetype="non-clock", nruns=2, burnin=0)
mcc_tree = mccTree(trace)
mcc_tree.reroot(clade("t83"), makeBifurcating=TRUE)
writeNexus(filename="anonymized_nonclock_rerooted.mcc.tre", mcc_tree)
```

Image
<br><br>

Note that setting the `makeBifurcating` argument to `TRUE` or `FALSE` makes no difference.

**Expected behavior**
Rerooting should not reassign branch annotations.

**Screenshots**
See above.

**Computer info**
macOS 15.3.2 (Sequoia), ARM64 architecture, RevBayes 1.2.6-preview, current tip of development (9134760).

brpetrucci:
This PR fixes #738, which showed that the `reroot()` method incorrectly reassigns node and branch annotations. 

This was happening because the `reroot` method of the `Tree` class was calling the `setRoot` function, which sets a root and proceeds to reorder the `nodes` object, which made the `node/branch_comments` and `nodes` object not match anymore. To solve this issue, we did the following:

- @bredelings created functions to access the `node_comments` and `branch_comments` member objects from `TopologyNode`. I ended up using a different way to access them, but will leave these methods here because they seem useful to have in general.
- I modified the `reverseParentChild` method, which for all intents and purposes is the actual rerooting function, to also swap node and branch comments between parent and child nodes. This is the bulk of what solved the issue. 
- I then ported the only part of `setRoot` that actually was necessary for `reroot`, _i.e._ the `root = ...` line, to the body of `reroot` itself. So in other words, `reroot` now calls `reverseParentChild` on the parent of the outgroup node, and sets the root to that node.
- Because we removed the call to `setRoot` within `reroot`, we removed one instance of reindexing of nodes. This made some of the rerooting tests fail, so I also updated the non-bifurcating expected output files to reflect that. 
- We are still reindexing in the case when `make_bifurcating = TRUE`, because `makeRootBifurcating` calls `setRoot`. This is not screwing up node and branch annotations. I think that's because we are not changing the root within that function call (and therefore the tree traversal does not change, and the ordering of `nodes` stays the same), but since we *do* change the root before calling `makeRootBifurcating`, I'm a bit skeptical and would appreciate others' thoughts on this.

Left to do:
- @bredelings is of the opinion that we should have a general `reroot` method, that reroots to any node and does just that, and then `reroot_to_outgroup` methods that handle specifically the rerooting to outgroup and make bifurcating tasks. I tend to agree, so I'm planning to do one of the following:
  1. Change the name of all the current `reroot` methods to `reroot_to_outgroup`. Write a new `Tree::reroot` method that simply runs `reverseParentChild` and sets the root, and then have all `reroot_to_outgroup` functions reference that function. This would be the most satisfying solution, but it would require some changes on the Rev language side (and possibly elsewhere), since all the current calls to `reroot` would have to be switched to `reroot_to_outgroup`. Doesn't seem too challenging
  2. Create a `reroot_to_node` method that does the basic steps named above, and then make all current `reroot` methods call `reroot_to_node`. A little less satisfying, but minimizes the changes required elsewhere. 

I'm partial to 1, but we can discuss. Once this idea is more solid I'll make this an actual PR.


davidcerny:
P.S. You probably meant to merge this into development rather than master, @brpetrucci

src/core/datatypes/trees/Tree.cpp
```diff
@@ -2052,10 +2051,24 @@ TopologyNode& Tree::reverseParentChild(TopologyNode &n)
         TopologyNode &p = n.getParent();
         ret = &(reverseParentChild(p));
 
+        // save p values
+        size_t idx_p = p.getIndex();
+        double brlen_p = p.getBranchLength();
+        std::vector<std::string> ncomm_p = p.node_comments;
+        std::vector<std::string> bcomm_p = p.branch_comments;
```
bredelings, line 2058: I think we can consolidate three lines into one if we swap branch comments using:
```
std::swap(p.branch_comments, n.branch_comments);
```

Normally swapping a and b involves:
```
tmp = a;
a = b;
b = tmp;
```

std::swap does all this at once, plus is uses "move construction" to avoid actually copying object contents with expensive objects like std::vector.

src/core/datatypes/trees/Tree.cpp
```diff
@@ -2052,10 +2051,24 @@ TopologyNode& Tree::reverseParentChild(TopologyNode &n)
...
+
         // we need to re-orient the branches/indices so that
         // nodes remain associated with the same parameters
```
bredelings, line 2061: Maybe we should change this to say that we reorient so that _branches_ remain associated with the same parameters: index, branch_length, and comments.

bredelings: Long-term we should probably add a separate index for the node that does not get swapped, so that we can have consistent node indices.  Then we can call the current index the branch index.

brpetrucci: I agree long-term that's a better solution re: branch vs. node indices. And yeah, I can change that comment, it's very confusing anyway

bredelings:
A lot of tests now fail.  I looked at one case, and it was just because the `posterior` annotation is now printed after the `:` as a branch comment.

PR #807, opened 2025-08-01
Remove boost filesystem

bredelings:
We can now use `std::filesystem` in place of `boost::filesystem`, dropping a link-time dependency on one boost library.


davidcerny (approved):
Since we require C++17 anyway, this makes sense. In addition to all the GitHub Actions checks, I can also confirm that this compiles without issues on macOS (ARM64) with cmake and Clang 17.

PR #809, opened 2025-08-04
Fix: Cannot modify clamped node error in Validation analyses

basanta33:
This PR addresses the "Cannot modify the value of a clamped node" error encountered in the latest development branch (Build `rapture-4035-g2d9bdb`).

The issue was identified during a validation analysis and prevented proper functionality. 
This fix resolves the underlying problem, allowing the clamped nodes to be modified if the Simulation Condition is VALIDATION.


bredelings (changes requested):
Hi Basanta.  Adding a special case here seems like a bit of a hack because we retain the "clamped" mark, but the node is (I think?) clamped at a different value at the end of the analysis.

This seems to be very similar to the problem with posterior predictive sampling (#778). The solution there is to copy the model graph, record which nodes are clamped, and then unclamp them.  Would something like that work here?

bredelings:
Probably you should also add a validation test to the regular tests directory in order to ensure that validation tests are not broken again in the future.  It doesn't have to actually run a long time, just to make sure that the run does not crash.

Issue #457, opened 2024-04-23
[BUG] Missing age_95%_HPD for Sampled Ancestors in MCC Tree Output from mccTree Function


minghao-du:
**Describe the bug**
When using the `mccTree` function, the sampled ancestors do not have age_95%_HPD.

**To Reproduce**
When running the script from the tutorial [Combined-Evidence Analysis and the Fossilized Birth-Death Process for Analysis of Extant Taxa and Fossil Specimens](https://revbayes.github.io/tutorials/fbd/fbd_specimen.html), the MCC tree obtained does not include an age_95%_HPD for the sampled ancestor.

**Expected behavior**
Like other fossils, sampled ancestors have an age_95%_HPD.

**Screenshots**
<img width="1144" alt="image" src="https://github.com/revbayes/revbayes/assets/51898298/a325aa49-282b-410d-a2ef-d2170ab74d1b">
<img width="1392" alt="image" src="https://github.com/revbayes/revbayes/assets/51898298/2b3e5c6c-e223-4a25-9d26-5b3533374bf6">

**Computer info**
I am using a Mac, and the version of RevBayes I am using is development (rapture-2545-g7c2b56).

[revbayes_fbd_fbd_specimen.zip](https://github.com/revbayes/revbayes/files/15074161/revbayes_fbd_fbd_specimen.zip)

brpetrucci:
@davidcerny something for us to fix before/in parallel to implementing CCD trees

brpetrucci:
If you continue reading right after the text you boxed in red, you will see that the 95% HPD interval for the node connecting to the _Agricarctos spp_ SA is {6.77354,7.74636}, and the issue here is that that age interval is placed on the node immediately parent to the SA, not the SA itself. If you open the MCC tree on FigTree, for example, you can see the intervals by placing node bars. I'm going to keep this issue open because I do believe we should alter the MCC tree function to include the SA properties on the SA itself, not on the node immediately preceding it, but the information *is* in the tree.

bredelings:
I guess I would say that the tip is fake, and the parent is "the SA itself".

I tried loading bears.mcc.tree and then writing it out again, and it seems to have lost all the node comments.

bredelings:
See #791 -- this means we can't read in `bears.mcc.tre`, but we can recreate it.

davidcerny:
Closed by #810, thanks to @brpetrucci!

Issue #467, opened 2024-05-07
[BUG] clade probabilities with sampled ancestors


hoehna:
**Describe the bug**
The MAP tree function doesn't compute the posterior probabilities of clades correctly when having sampled ancestors. Here we have a fake traces of 5 trees, all the same topology. The only difference is that one tip is set as sampled ancestor 3 out of 5 times. All PP of clades should be 1.0, but aren't.

Two more issues:
1. The first tree is omitted even though the burnin is 0.
2. The node age of the sampled ancestor is set to the non-sampled ancestor state. Why?

**To Reproduce**

trace = readTreeTrace("TryTreeWithSA.trees",burnin=0,treetype="clock")
mapTree(trace,file="map.tre")

[TryTreeWithSA.trees.zip](https://github.com/revbayes/revbayes/files/15235684/TryTreeWithSA.trees.zip)

PR #810, opened 2025-08-04
Adding age HPD to sampled ancestor node/tip

brpetrucci:
This PR attempts to fix issue #457. Turns out the fix is really simple--we just had a check preventing tips from getting the HPD if they were sampled-ancestor tips. I removed those checks, so now any fossil tip will get a 95% age HPD. I suspect I can simplify the check further, but PTSD from removing what looked like a redundant || statement in an if loop and discovering a huge problem months later has made me really cautious, so I just removed the check that prevented SAs from receiving the annotation. 

[Here](https://github.com/user-attachments/files/21585392/bears.mcc.new.tre.txt) is an example of an MCC tree generated with the trace from the issue. Note that the node representing the sampled ancestor and the "tip" coming out of this node have slightly different 95% HPD intervals (e.g. consider the node with index 27, HPD `{14.3451,15.9694}`, representing the SA _Ursavus_primaevus_, which has HPD `{14.2137,15.9694}`). My MCC trees from BEAST2 analyses show the same behavior, so I'm wondering if it has something to do with the fact we consider clades where _Ursavus_primaevus_ is an SA separately from clades where it is (a fact that has led to a lot of disagreement, see #467 and #787). Still, I wonder if this is expected behavior--since the SA is represented by both the node and tip, I assume we'd want them to be calculating the ages HPD from the same sample, be it restricted to the case where that taxon is an SA or not. I'm interested to hear what others think, though.

### DIFF
--- a/src/core/analysis/mcmc/output/TreeSummary.cpp
+++ b/src/core/analysis/mcmc/output/TreeSummary.cpp
@@ -717,7 +717,7 @@ void TreeSummary::annotateTree( Tree &tree, AnnotationReport report, bool verbos
             double saFreq = sampled_ancestor_counts[n->getTaxon()];
 
             // annotate sampled ancestor prob
-            if ( ((n->isTip() && n->isFossil()) || saFreq > 0) && report.sampled_ancestor_probs )
+            if ( (n->isFossil() || saFreq > 0) && report.sampled_ancestor_probs )
             {
                 n->addNodeParameter("sampled_ancestor", saFreq / totalSamples);
             }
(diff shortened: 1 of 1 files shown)


brpetrucci:
It evaded my mind that there would definitely be tests with SA trees and therefore I need to make new expected results files for those. Will get on it tomorrow

bredelings:
I guess I still am not sure about whether the tips should have _different_ age ranges.

However, I suppose that in the case where we are printing the trees with tip nodes, then we may as well print the ranges for them.  If there are more issues (which would not surprise me) we can deal with them later.

brpetrucci:
About the difference in ages, it seems to be related to that discussion we've been having for a while (#787). When the node HPD is calculated, it takes into account the entire set of trees where the clade in question is monophyletic, but when the tip one is, it only considers the set of trees where the clade is monophyletic AND that tip is an SA. I confirmed this is also the case with BEAST2 summary trees. 

I think that's ok, because we should be looking at the tip HPD in any case. Furthermore, since we might decide soon to change how the "clades" are defined in an FBD tree, I think it'd be better to avoid making changes to this specific issue right now. So I'll merge!

PR #816, opened 2025-08-07
Clean up caching in phyloCTMC

bredelings:
Simplify and clean up caching of conditional likelihoods in phyloCTMC.  
- Make a `struct PartialLikelihoods` that contains both the partial likelihoods and the scale factors for a branch / node.
- Make a simplified `IndexCached<T>` object that handles the logic of:
    * invalidating caches at a node
    * sharing the per-node object between current and proposed states
    * touch / restore / keep
- Use it for both partial likelihoods and transition probability matrices.
- Ensure that conditional likelihood pointers are `const` unless we just allocated the conditional likelihood vector (and are currently computing it).
- use `pmatrices` for all transition probability matricies (previous we used `pmatrices` in some places and `transition_prob_matrices` in others)

Benefits:
- readability
- prevent likelihood / pmatrix caching bugs
- make progress towards allowing the dimension of the Q matrix is allowed to change dynamically without crashing the likelihood calculator

This last bit should help with models where the true number of states is unknown.


src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
```diff
@@ -81,6 +81,41 @@ namespace RevBayesCore {
...
+        // With MPI, these might not be ALL the site patterns, but they are all the ones stored in this process.
+        struct Dims
+        {
+            std::size_t num_site_mixtures;
+            std::size_t num_patterns;
+            std::size_t num_states;
+
+            size_t size() const {return num_site_mixtures * num_patterns * num_states;}
+
+            bool operator==(const Dims&) const = default;
```
afmagee, line 115: 13 lines in and I'm already remembering how much C++ I have forgotten, and learning how much C++ 20 I never knew.

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
afmagee: If we're removing `active` we should update this.

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
```diff
@@ -149,6 +184,17 @@ namespace RevBayesCore {
         virtual double                                                      getPInv(void) const;
 
 
+        // functions for accessing mutable caches
+        void                                                                markPartialLikelihoodsDirtyForNode(int node) const;
+        void                                                                markAllPartialLikelihoodsDirty() const;
+        bool                                                                partialLikelihoodsDirtyForNode(int node) const;
+        const PartialLikelihoods&                                           getPartialLikelihoodsForNode(int node) const;
+        PartialLikelihoods&                                                 getCreatePartialLikelihoodsForNode(int node, const PartialLikelihoods::Dims& d) const;
+        PartialLikelihoods&                                                 getModifyPartialLikelihoodsForNode(int node) const;
```
afmagee: I'm probably just forgetting my C++, but how can a function that modifies the object('s partial likelihoods) be const?

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
```diff
@@ -344,18 +387,16 @@ gap_match_clamped( gapmatch )
...
-    
+
     activePmatrixOffset         =  num_nodes * num_site_mixtures;
     pmatNodeOffset              =  num_site_mixtures;
     active_pmatrices            =  std::vector<size_t>(num_nodes, 0);
     pmat_changed_nodes          =  std::vector<bool>(num_nodes, false);
     pmat_dirty_nodes            =  std::vector<bool>(num_nodes, true);
     pmatrices                   =  std::vector<TransitionProbabilityMatrix>(activePmatrixOffset * 2, TransitionProbabilityMatrix(num_chars));
 
+    partialLikelihoods.resize(num_nodes);
```
afmagee: Is this redundant with the `partialLikelihoods(num_nodes)` above?

bjoelle: not completely sure why this needs to be initialized twice ? here and l369

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
```diff
@@ -487,16 +510,16 @@ void RevBayesCore::AbstractPhyloCTMCSiteHomogeneous<charType>::checkInvariants(
...
-	    assert(dirty_nodes[node->getParent().getIndex()]);
+	if (partialLikelihoodsDirtyForNode(node->getIndex()) and not node->isRoot())
+	    assert(partialLikelihoodsDirtyForNode(node->getParent().getIndex()));
     }
 
     // 2. Invariant: if the P-matrix for a node is dirty, then the conditional likelihoods should be dirty
     for(auto node: tree_nodes)
     {
 	int index = node->getIndex();
 	if (pmat_dirty_nodes[index] and not node->isRoot())
```
bjoelle: nothing to do with your changes but I'm wondering if this check has a good reason to exclude the root or if it's a copy-paste from the previous check

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
```diff
@@ -631,6 +650,65 @@ inline bool has_weighted_characters(AbstractHomologousDiscreteCharacterData& dat
...
+template<class charType>
+inline PartialLikelihoods& RevBayesCore::AbstractPhyloCTMCSiteHomogeneous<charType>::getCreatePartialLikelihoodsForNode(int node_index, const PartialLikelihoods::Dims& dims) const
+{
+    assert(partialLikelihoodsDirtyForNode(node_index));
+    partialLikelihoods[node_index] = std::make_shared<PartialLikelihoods>(dims);
+    return *partialLikelihoods[node_index];
+}
+
+template<class charType>
+inline PartialLikelihoods& RevBayesCore::AbstractPhyloCTMCSiteHomogeneous<charType>::getModifyPartialLikelihoodsForNode(int node_index) const
```
bjoelle: would suggest "getNonDirtyPartial..." for that function name

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
```diff
@@ -631,6 +650,65 @@ inline bool has_weighted_characters(AbstractHomologousDiscreteCharacterData& dat
...
+}
+
+template<class charType>
+inline const PartialLikelihoods& RevBayesCore::AbstractPhyloCTMCSiteHomogeneous<charType>::getPartialLikelihoodsForNode(int node_index) const
+{
+    return *this->partialLikelihoods[node_index];
+}
+
+template<class charType>
+inline PartialLikelihoods& RevBayesCore::AbstractPhyloCTMCSiteHomogeneous<charType>::getCreatePartialLikelihoodsForNode(int node_index, const PartialLikelihoods::Dims& dims) const
```
bjoelle: would suggest "getNewPartial...." for that function name

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
```diff
@@ -982,8 +1058,8 @@ double RevBayesCore::AbstractPhyloCTMCSiteHomogeneous<charType>::computeLnProbab
     if ( in_mcmc_mode == false )
     {
         // free the partial likelihoods
-        delete [] partialLikelihoods;
-        partialLikelihoods = NULL;
+        partialLikelihoods.clear();
```
bjoelle: you have the allocate function to do that above

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
```diff
@@ -1683,8 +1753,8 @@ void RevBayesCore::AbstractPhyloCTMCSiteHomogeneous<charType>::executeMethod(con
         if ( delete_partial_likelihoods == true )
         {
             // free the partial likelihoods
-            delete [] partialLikelihoods;
-            partialLikelihoods = NULL;
+            partialLikelihoods.clear();
```
bjoelle: same with the allocate

bjoelle: further down too

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
```diff
@@ -2880,7 +2918,11 @@ template<class charType>
 void RevBayesCore::AbstractPhyloCTMCSiteHomogeneous<charType>::scale( size_t node_index, size_t left, size_t right, size_t middle )
 {
 
-    double* p_node = this->partialLikelihoods + this->activeLikelihood[node_index]*this->activeLikelihoodOffset + node_index*this->nodeOffset;
+    double* p_node = getModifyPartialLikelihoodsForNode(node_index).likelihoods.data();
```
bjoelle: the use of get and getModify seems a bit inconsistent ? in several places like here we use getModify for p_node but then it's not actually modified

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
```diff
@@ -149,6 +184,17 @@ namespace RevBayesCore {
...
 
+        // functions for accessing mutable caches
+        void                                                                markPartialLikelihoodsDirtyForNode(int node) const;
+        void                                                                markAllPartialLikelihoodsDirty() const;
+        bool                                                                partialLikelihoodsDirtyForNode(int node) const;
+        const PartialLikelihoods&                                           getPartialLikelihoodsForNode(int node) const;
+        PartialLikelihoods&                                                 getCreatePartialLikelihoodsForNode(int node, const PartialLikelihoods::Dims& d) const;
+        PartialLikelihoods&                                                 getModifyPartialLikelihoodsForNode(int node) const;
+        void                                                                allocatePartialLikelihoods() const;
+        const double*                                                       getMarginalLikelihoodsForNode(int node) const;
```
bjoelle, line 366: not completely sure if the const and non-const accessors should have the same name - seems like e.g. get and getMutable would be more explicit for people not super familiar with c++, and would make it clearer that the non-mutable is supposed to be the default
you have the same with getCorrection later on

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
afmagee: Agree that the name is a bit confusing, but not sure `getNewPartial` would be good, since, at least to me, that could be confused with an "update" sort of function, which this is not.

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
```diff
@@ -631,6 +650,65 @@ inline bool has_weighted_characters(AbstractHomologousDiscreteCharacterData& dat
...
+}
+
+template<class charType>
+inline const double* RevBayesCore::AbstractPhyloCTMCSiteHomogeneous<charType>::getMarginalLikelihoodsForNode(int node_index) const
+{
+    return this->marginalLikelihoods.data() + node_index*this->nodeOffset;
+}
+
+template<class charType>
+inline double* RevBayesCore::AbstractPhyloCTMCSiteHomogeneous<charType>::getMarginalLikelihoodsForNode(int node_index)
```
afmagee: Why do we need this non-const version and the const version?

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
```diff
@@ -1001,9 +1077,9 @@ void RevBayesCore::AbstractPhyloCTMCSiteHomogeneous<charType>::computeMarginalNo
     this->updateTransitionProbabilities( node_index );
 
     // get the pointers to the partial likelihoods and the marginal likelihoods
-    const double*   p_node                  = this->partialLikelihoods + this->activeLikelihood[node_index]*this->activeLikelihoodOffset + node_index*this->nodeOffset;
-    double*         p_node_marginal         = this->marginalLikelihoods + node_index*this->nodeOffset;
-    const double*   p_parent_node_marginal  = this->marginalLikelihoods + parentnode_index*this->nodeOffset;
+    const double*   p_node                  = getPartialLikelihoodsForNode(node_index).likelihoods.data();
+    double*         p_node_marginal         = getMarginalLikelihoodsForNode(node_index);
+    const double*   p_parent_node_marginal  = getMarginalLikelihoodsForNode(parentnode_index);
```
afmagee, line 1236: Big readability improvement.

src/core/distributions/phylogenetics/substitution/PhyloCTMCSiteHomogeneous.h
afmagee, line 1: Stylistically, `PL_` doesn't quite match the rest of the code, but `pl_` would

src/core/distributions/phylogenetics/substitution/PhyloCTMCClado.h
```diff
@@ -108,16 +108,10 @@ namespace RevBayesCore {
 //        AbstractPhyloCTMCSiteHomogeneous(const TypedDagNode<Tree> *t, size_t nChars, size_t nMix, bool c, size_t nSites, bool amb, bool wd = false, bool internal = false, bool gapmatch = true );
 
 template<class charType>
-RevBayesCore::PhyloCTMCClado<charType>::PhyloCTMCClado(const TypedDagNode<Tree> *t, size_t nChars, bool c, size_t nSites, bool amb, bool internal, bool gapmatch) : AbstractPhyloCTMCSiteHomogeneous<charType>(  t, nChars, 1, c, nSites, amb, false, false, true ),
-
-    cladoPartialLikelihoods(NULL),
-    cladoMarginalLikelihoods(NULL),
```
afmagee, line 114: We can get away without this because the uninitialized std::vector<double> is safely empty, right?

src/core/distributions/phylogenetics/substitution/PhyloCTMCClado.h
```diff
@@ -207,32 +200,26 @@ RevBayesCore::PhyloCTMCClado<charType>::PhyloCTMCClado(const PhyloCTMCClado &n)
     // copy the partial likelihoods if necessary
     if ( this->in_mcmc_mode == true )
     {
-        cladoPartialLikelihoods = new double[2*this->num_nodes*this->num_site_rates*this->num_sites*this->num_chars*this->num_chars];
-        memcpy(cladoPartialLikelihoods, n.cladoPartialLikelihoods, 2*this->num_nodes*this->num_site_rates*this->num_patterns*this->num_chars*this->num_chars*sizeof(double));
+        assert(cladoPartialLikelihoods.size() == 2*this->num_nodes*this->num_site_rates*this->num_sites*this->num_chars*this->num_chars );
```
afmagee: If we don't need to copy these because of how we've refactored the class, then does this assert need to be buried in a "check only if we're in this mode" statement, or can we just check, and the sizes will be determined by the mode?

src/core/distributions/phylogenetics/substitution/PhyloCTMCClado.h
```diff
@@ -207,32 +200,26 @@ RevBayesCore::PhyloCTMCClado<charType>::PhyloCTMCClado(const PhyloCTMCClado &n)
...
     }
     
     // copy the marginal likelihoods if necessary
     if ( this->useMarginalLikelihoods == true )
     {
-        cladoMarginalLikelihoods = new double[this->num_nodes*this->num_site_rates*this->num_sites*this->num_chars*this->num_chars];
-        memcpy(cladoMarginalLikelihoods, n.cladoMarginalLikelihoods, this->num_nodes*this->num_site_rates*this->num_sites*this->num_chars*this->num_chars*sizeof(double));
+        assert(cladoMarginalLikelihoods.size() == this->num_nodes*this->num_site_rates*this->num_sites*this->num_chars*this->num_chars);
```
afmagee: Same comment as above

src/core/distributions/phylogenetics/substitution/PhyloCTMCClado.h
```diff
@@ -282,19 +268,21 @@ void RevBayesCore::PhyloCTMCClado<charType>::computeRootLikelihood( size_t root,
     bool has_sampled_ancestor_child = node.getChild(0).isSampledAncestorTip() || node.getChild(1).isSampledAncestorTip();
     
     // get the pointers to the partial likelihoods of the left and right subtree
-    double* p_node         = this->partialLikelihoods + this->activeLikelihood[root]  * this->activeLikelihoodOffset + root  * this->nodeOffset;
-    const double* p_left   = this->partialLikelihoods + this->activeLikelihood[left]  * this->activeLikelihoodOffset + left  * this->nodeOffset;
-    const double* p_right  = this->partialLikelihoods + this->activeLikelihood[right] * this->activeLikelihoodOffset + right * this->nodeOffset;
+    auto& PL_left  = this->getPartialLikelihoodsForNode(left);
```
afmagee: `PL_` vs `pl_` style comment applies here too

src/core/distributions/phylogenetics/substitution/PhyloCTMCSiteHomogeneousDollo.cpp
afmagee, line 1: Massive readability improvement here thanks to the new `getCorrectionLikelihoodsForNode`

src/core/distributions/phylogenetics/substitution/PhyloCTMCSiteHomogeneousNucleotide.h
```diff
@@ -194,18 +204,26 @@ void RevBayesCore::PhyloCTMCSiteHomogeneousNucleotide<charType>::computeInternal
...
+    auto& PL_left = this->getPartialLikelihoodsForNode(left);
+    auto& PL_right = this->getPartialLikelihoodsForNode(right);
+    const double* p_left   = PL_left.likelihoods.data();
+    const double* p_right  = PL_right.likelihoods.data();
+    assert(PL_left.dims == PL_right.dims);
+
+    double* p_node   = this->getCreatePartialLikelihoodsForNode(node_index, PL_left.dims).likelihoods.data();
+    //    __m128d* p_left   = (__m128d *) this->getPartialLikelihoodsForNode(left);
+    //    __m128d* p_right  = (__m128d *) this->getPartialLikelihoodsForNode(right);
+    //    __m128d* p_node   = (__m128d *) this->getPartialLikelihoodsForNode(node_index);
```
afmagee, line 215: Is this a real comment (in which case, do we need it) or is this some `if defined` shenannigans?

afmagee: (Technically out of scope, admittedly)

bredelings:
Thanks for looking at this!  And thanks for approving.  I am going to go through and address the comments.

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
bredelings: This is classic C++ accessor stuff, where you return a `const thing*` when `this` is `const`, but return a `thing*` when `this` is not `const`.  It basically allows you to mutate the field only if the object containing the field is const.

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
bredelings: Sometimes you have a member function `object.f()` that computes something without modifying `object`, so you mark it `const`.

But now suppose you want to cache the value, so that you compute the function the first time `object.f()` is called, and store the result inside `object`.  Then the next time you call `object.f()` you can reuse the previous result.  You still want the method to be marked `const` since it doesn't "truely" modify `object`.  However, if does modify the cached value, which is stored inside the object, so the compiler complains!  To make the compiler stop complaining, you mark the cached value as `mutable`, which means that you are allowed to modify it even when the object that contains it is `const`.

Here the partial likelihoods are cached values.  You'll see that they are declared as `mutable`.

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
bredelings: Ah, yes.  Removed.

bredelings: Yeah, good point, fixed.

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
bredelings: There is no actual branch above the root node.  So there is no branch length and no P-matrix either.  I've added a comment to clarify this.

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
bredelings: I went with `getMutablePartialLikelihoodsForNode( )`.

Ideally we would not mutate the partials after they are created.  Eventually we can probably reach that ideal state by scaling the partials in the same function that creates them.  Then we can consider eliminating this routine altogether.

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
bredelings: I'm going with `createEmptyPartialLikelihoodsForNode(node)`.  If you have a better idea, let me know.

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
bredelings: fixed

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
bredelings: fixed in multiple places

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
bredelings: Here I think it is modified.  But we do it indirectly, but doing `double* p_site_mixture = p_node + offset;` and then modifying `p_site_mixture[i]`.

bredelings: I did find one other place where it did not need to be mutable.

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
bredelings: I changed the const and non-const versions to have different names.

src/core/distributions/phylogenetics/substitution/PhyloCTMCClado.h
bredelings: Exactly.  The default constructor puts it into a sensible empty state.

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
bredelings: Lol.  The "spaceship operator" `<=>` is even more fun: https://en.cppreference.com/w/cpp/language/default_comparisons.html

src/core/distributions/phylogenetics/substitution/PhyloCTMCSiteHomogeneous.h
bredelings: Done.

src/core/distributions/phylogenetics/substitution/PhyloCTMCClado.h
bredelings: Done.

src/core/distributions/phylogenetics/substitution/PhyloCTMCClado.h
bredelings: Done.

bredelings:
> Surveying the tests, I get the sense that we don't have all of these changes covered? We've got binary and multistate morphological characters in tests, and plenty of nucleotide data. But I'm not sure we have anything that gets into the Dollo or Clado versions? It would be nice for peace of mind if we had tests ensuring all of the changes didn't affect the likelihood, but c'est la vie on such a large scientific codebase.

I think its mostly the Dollo versions that have no coverage.  I'll ask on slack if anyone knows of any Dollo examples.

> Do we know if there are any significant speed changes in the move to the new data structure? Since it's not quite as flat?

I'd be surprised if there were big speed changes, since the slabs of memory for each node are still pretty large themselves, even if they aren't next to each other.  I did a quick benchmark, and it seems that the new code is... consistently about 3% faster on the DNA `primates_and_galeopterus_cytb.nex` data set.

Its possible that the code could take less RAM.  And in fact, on the same simple DNA test, the new code takes 40% less memory.

src/core/distributions/phylogenetics/substitution/PhyloCTMCClado.h
bredelings: fixed

src/core/distributions/phylogenetics/substitution/PhyloCTMCSiteHomogeneousNucleotide.h
bredelings: I think this is more record of something that was tried before.  We could delete it, and let people look in git.  But I think I'll leave it.  I haven't really looked at SSE/AVX instruction coding.

bredelings:
Apparently, I got the benchmarks wrong yesterday - now I see a 5% slowdown on some small DNA data sets.  This seems to be caused by the time taken to malloc and free the buffers.  Apparently this is non-negligible compared the time to compute a small partial likelihood for DNA.  This can probably be fixed by putting the PartialLikelihood objects into a pool, and then reusing them again when we want another one.

After I fix that, if the slowdown is around 1% or less, I plan to merge.

davidcerny:
Would you be willing to share some of those benchmarks here? It would be reassuring to know that the changes made in this PR don't result in a tangible slowdown even with a large number of taxa / alignment length / rate matrix dimensions, etc.

bredelings:
So, I've actually decided to abandon this approach based on reasoning about the benchmarks.  I should have benchmarked this earlier, before I had gotten so far.  The main issue is that this approach requires a malloc and a free for each node.  If an individual partition is short enough, and the number of states is low (i.e. DNA, binary data) then this is too high :cry:   But its fine (sigh) -- I have some pretty good ideas what to do instead.

I think its possible to obtain the safety and readability goals of this PR without any malloc/free overhead.  So thank you for the review @afmagee!  I might not have noticed this if you hadn't asked for benchmarks.

@davidcerny I started constructing some benchmark scripts based on `test_JC_unrooted` with different input files, using both DNA and AA.  These are preliminary though.  I want to add some benchmarking tests for codons (61 states) and some models with a higher number of states as well.  Additionally, I want to run them with and without -march=native, which turns on SSE/AVX support for Intel.  Usually I run each test 5 times without other major software (i.e. firefox, etc) running and take the median user time.

What do you think about adding a benchmarks directory?  It would initially consist of scripts to run, and a README file.  I made some scripts that work like `rb bench-dna file.nex 10000`.  These are not fancy scripts, so if anyone has better benchmarking scripts then we could replace these, or add the other ones.  Additionally, these scripts mostly benchmark the likelihood calculator.

At the moment I'm not considering adding benchmark RESULTS, or at least not standard ones.  The reason is that the exact numbers are going to depend on specific CPUs, so everybody will get different numbers, even if the overall picture is similar.

bredelings:
I have a fix for the speed issue.  It involves dropping the `std::shared_ptr<T>` and moving back toward the original approach of having 2n slots and flipping between slot i and i+n.  I wasn't able to finish it yesterday, so I'll resume next week.

Preliminary benchmarks indicate that it fixes the speed issues, but there's some things I still want to check before merging.

davidcerny:
> What do you think about adding a benchmarks directory? It would initially consist of scripts to run, and a README file. 

That sounds like a great idea!

I might be a bit paranoid, but I'm a little worried that with fundamental changes like this, there could be performance hits that are hard to detect with the relatively small analyses from our CI suite, and only become apparent when scaling up. E.g., trees with >500 tips, alignments with >100,000 sites, analyses with >30 partitions, etc. I don't know to what extent this is a reasonable concern to have (as you point out, your solution based on shared pointers only exhibited a measurable slowdown with _short enough_ partitions), but I guess it can't hurt to have the benchmark suite cover a range of dataset sizes – unlike the CI suite, where large datasets would be impractical.

(An interesting related question is what are the largest datasets that people have actually tried to analyze using RevBayes. I helped conduct a bibliometric study of what researchers actually use RevBayes for, but unfortunately, this wasn't among the information I recorded.)

> At the moment I'm not considering adding benchmark RESULTS, or at least not standard ones. The reason is that the exact numbers are going to depend on specific CPUs, so everybody will get different numbers, even if the overall picture is similar.

Makes sense. To clarify, I was just suggesting that you post the results you've already obtained here as a comment, without any claims to reproducibility.

bredelings:
The thing that was throwing me off was that the times depend on whether `useScaling` is `true` or `false`.

I generally use the median time, because its more stable than an individual run.  If I correctly sorted the rows, then this should be the time in the middle row for each condition.

```
time rb bench-dna.Rev data/primate_and_galeopterus_cytb.nex 100000

        8.471  / development / scaling
        8.492  / development / scaling
        8.522  / development / scaling
        8.527  / development / scaling
        8.553  / development / scaling

        7.752 / refactor-phyloctmc-caching2 / scaling
        7.777 / refactor-phyloctmc-caching2 / scaling
        7.780 / refactor-phyloctmc-caching2 / scaling
        7.815 / refactor-phyloctmc-caching2 / scaling
        7.817 / refactor-phyloctmc-caching2 / scaling

This looks like a 9% speedup.

        2.530 / development / no scaling
        2.531 / development / no scaling
        2.536 / development / no scaling
        2.538 / development / no scaling
        2.554 / development / no scaling

        2.520 /  refactor-phyloctmc-caching2 / no scaling
        2.521 /  refactor-phyloctmc-caching2 / no scaling
        2.525 /  refactor-phyloctmc-caching2 / no scaling
        2.534 /  refactor-phyloctmc-caching2 / no scaling
        2.538 /  refactor-phyloctmc-caching2 / no scaling

This looks like no change.
        

time rb bench-aa.Rev EF-Tu-12.fasta 10000

        7.569 / development / scaling
        7.570 / development / scaling
        7.574 / development / scaling
        7.579 / development / scaling
        7.585 / development / scaling

        7.509 / refactor-phyloctmc-caching2 / scaling
        7.534 / refactor-phyloctmc-caching2 / scaling
        7.537 / refactor-phyloctmc-caching2 / scaling
        7.540 / refactor-phyloctmc-caching2 / scaling
        7.540 / refactor-phyloctmc-caching2 / scaling

This looks like a 0.4% speedup.
                
        6.848 / development / no scaling
        6.872 / development / no scaling
        6.875 / development / no scaling
        6.893 / development / no scaling
        6.932 / development / no scaling
        
        6.791 / refactor-phyloctmc-caching2 / no scaling
        6.811 / refactor-phyloctmc-caching2 / no scaling
        6.841 / refactor-phyloctmc-caching2 / no scaling
        6.845 / refactor-phyloctmc-caching2 / no scaling
        6.875 / refactor-phyloctmc-caching2 / no scaling
        
This looks like no change.
```

bredelings:
I added a benchmarking script in the directory `benchmarking`.  I changed it so that you run it like this:

```
time rb bench-unrooted.Rev ../tests/data/primates_and_galeopterus_cytb.nex DNA 100000
time rb bench-unrooted.Rev ../tests/data/EF-Tu-12.fasta AA 10000
```

More benchmarks would be good.  That would include data sets with more taxa, more sites, and more partitions, and also different model setups.

I have been wishing for an `argparse`-like facility for revbayes.  Then would could write something like `--generations=10000` or `-g 10000`.

davidcerny:
Hmm... the latest changes are failing on the [`mvFossilTimeSlideUniform_multiple_runs.Rev`](https://github.com/revbayes/revbayes/blob/master/tests/test_FBD/scripts/mvFossilTimeSlideUniform_multiple_runs.Rev) test script, and it seems to be a consistent failure, rather than a stochastic one that we run into every once in a while.

I don't think this is a reliable indicator there's something wrong with the PR, though. The test was added in PR #895, which (seemingly) solved issue #887 by making sure `mvFossilTimeSlideUniform` recalculates node indices every time the proposal is performed. In theory, that shouldn't have been necessary – it was a brute-force solution to a problem whose actual causes were unclear. Moreover, the problem in question – that different runs of the same MCMC exhibit very different behavior – has also popped up elsewhere (see issue #903).

I'm now wondering if I can reproduce #903 using this branch. Will report back.

bredelings:
The difference for the failing test looks something like this:

```
--- output_expected/mvFossilTimeSlideUniform_multiple_runs.errout    2026-04-03 00:17:44.000000000 -0400
+++ output/mvFossilTimeSlideUniform_multiple_runs.errout    2026-04-03 01:16:28.000000000 -0400
@@ -2633,7 +2633,7 @@
    
    
    Drawing new initial states ... 
-   Could not compute lnProb for node 'unord_ctmc[1]': lnProb = -inf
+   Could not compute lnProb for node 'unord_ctmc[1]': lnProb = -nan
    Standard character matrix with 43 taxa and 212 charac
    
    
@@ -4236,7 +4236,7 @@
    
    
    Drawing new initial states ... 
-   Could not compute lnProb for node 'unord_ctmc[1]': lnProb = nan
+   Could not compute lnProb for node 'unord_ctmc[1]': lnProb = -nan
    Standard character matrix with 43 taxa and 212 charac
```

I hope to look at this on Wednesday...

bredelings:
Oh, another interesting thing: I tried recompiling with `-march=x86-64-v3 -mtune=native` on Intel linux.  In addition to being 10% faster, it changed some of the test results :grimacing: Mostly the same ones that change between Linux/Mac, but maybe not exactly the same ones.

bredelings:
So what's happening with the bug is that the probability of the observed characters in one column is zero.  This leads to a log-likelihood of -Inf, which is what our test output expects.  However, at some point the probability that the column is variable also becomes zero.  We then have `0/0` which is NaN.

If the probability of the data and of the condition are both zero, then NaN seems like the right answer.

The questions there are:
* should the condition have a small positive probability?
* why didn't this trigger before?

bredelings:
I found a case where we were failing to initialize the scaling factor when the probability of a site was zero.  This was a pre-existing bug -- it doesn't appear to have had much effect since the probability was already zero.  But fixing it makes the tests pass.

However, while looking for this bug, I found that in PhyloCTMCSiteHomogeneousConditional.h (which is used for `coding="variable"`  and `coding="informative"`, we had a check that set the probability to NaN if a probability was more than 1.  In practice, this probability was sometimes 1.0000000000000001 (1+1e-16) because of round-off error.  Instead of returning NaN, we should just cap the probability at 1, and complain if the error is too large.

This changed the results of test_MC3 and test_FBD/scripts/mvFossilTimeSlideUniform_multiple_runs.Rev, so I updated them.  **I wonder if this will make the output on Mac and Linux the same now?**

davidcerny:
Certainly sounds like it's worth giving it a try – let me pull, recompile, and see what results I get on my Macbook!

davidcerny:
> **I wonder if this will make the output on Mac and Linux the same now?**

It doesn't, unfortunately. The results for [`Test_heat_tuning_and_checkpointing.Rev`](https://github.com/revbayes/revbayes/blob/refactor-phyloctmc-caching2/tests/test_MC3/scripts/Test_heat_tuning_and_checkpointing.Rev) I'm getting on macOS with the current tip of this branch (commit 29654b4) are not identical to the updated expected output that you pushed with commit b27b0a5, with subtle differences in parameter values, likelihoods, and numbers of attempted moves.

For the [`mvFossilTimeSlideUniform_multiple_runs.Rev`](https://github.com/revbayes/revbayes/blob/refactor-phyloctmc-caching2/tests/test_FBD/scripts/mvFossilTimeSlideUniform_multiple_runs.Rev) script, I'm still getting `-inf` log likelihoods on macOS, rather than the `-nan` ones we're getting on Linux.

bredelings:
BTW, I'm pretty happy with this now.  It solves a few issues and makes the code substantially more maintainable.  I have a list of other things lined up that depend upon this and will further improve readability and safety, and also make it faster.

In its current form this code so  scary that almost nobody feels comfortable modifying it.   It has a lot of interlocking parts, and is written in C nearly at the level of assembly language.  My hope is that in addition to speeding up the code, it will become safer and something that more people can safely modify.

I'm considering merging this PR tomorrow -- but I could possibly wait a few days if someone wants me to.

davidcerny:
Just to follow up on my earlier comment, issue #903 can still be reproduced using this branch, confirming that it is unrelated to the failing test (instead being related to the cloning / copying / initialization of moves or sampler objects, most likely).

That's certainly not an objection to merging this PR as-is, though. It would be nice if the refactoring of likelihood caching solved seemingly unrelated issues such as test discrepancies between platforms or chains behaving weirdly, but just because it didn't do all that doesn't mean it's not worth doing on its own merits.

I assume the benchmarks no longer show any difference between this branch and development?

bredelings:
The benchmarks show no difference when scaling is off.  They show an improvement when scaling is on.  I'm not completely sure why.  Based on the magnitude (medium for DNA, low for AA), it looks like the old code might have been doing scaling twice in some circumstances.

davidcerny:
Perfect!

PR #821, opened 2025-08-14
Do not print messages about trivial tip age adjustments

### LINKED ISSUE #820: [BUG] Warning about adjusting taxon ages without any adjustments happening
**Describe the bug**
When setting a starting value via `.setValue()` on an FBD tree, I get for all taxa the message 
`Age of taxon Y was above the specified maximum and has been adjusted.`  or the equivalent message with `below`. This happens although taxon ages match those in the tree. Checking the taxon ages afterwards, I found that their ages were actually not adjusted.

This might be due to a strict comparison in Tree.cpp or numeric imprecision when importing the ages from .csv

**To Reproduce**

[example_data.zip](https://github.com/user-attachments/files/21754628/example_data.zip)

```
seed(42)
taxa <- readTaxonData("fossils_1_nchar100_cont_rho1.csv")
init_tree = readTrees("tree_1_nchar100_cont_rho1.nex")[1]
fbd_tree ~ dnFBDP(origin=2.0, lambda=1.0, mu=0.3, psi=5.0, rho=1.0, taxa=taxa)
fbd_tree.taxa()[1] # taxon t9_2
fbd_tree.taxa()[1].getAge() # 1.099357

fbd_tree.setValue(init_tree)

fbd_tree.taxa()[24] # taxon order changed, select t9_2
fbd_tree.taxa()[24].getAge() # 1.099357 age was not adjusted
```

**Expected behavior**
No message is displayed if taxon ages are not adjusted.

**Computer info**
Running on WSL2 (Ubuntu 24.04.2 LTS), RevBayes version (1.3.0) Build from development (rapture-4174-g376e80) on Thu Jul 31 16:15:21 CEST 2025

davidcerny:
Often, it happens that the ages read in from a taxon file and the ages of the corresponding tips in a tree should be exactly identical, but are not recognized as such by RevBayes. We have already introduced several solutions to this problem: an optional precision argument for the tip age checks that tree distributions run on their `initialTree` argument (#455, #481), and automatic tip age adjustment used for things like `.setValue()` or `.initializeFromCheckpoint()` (#517). In the latter case, `Tree::checkTaxonAges()` prints a message that an adjustment has been performed.

In general, it seems worthwhile informing the user that we are actually setting the tip ages to something other than the specified values, but as pointed out in #820, it's perhaps more confusing than helpful to do so when the adjustment is very small – say, smaller than the tolerance for comparing doubles (`userSettings().getTolerance()`; 1e-10 by default). That said, it's still important to perform the adjustment in such situations, even if we do it quietly: failing to do so would cause moves to be rejected, potentially with confusing warnings of their own (and that's still the best-case scenario).

This PR closes #820. It also adds a test for #516, which was solved by making extremely small adjustments to the tip ages obtained from a checkpoint file.


src/core/datatypes/trees/Tree.cpp
```diff
@@ -950,19 +951,40 @@ void RevBayesCore::Tree::checkTaxonAges(bool forceAdjust)
...
+            double diff = node->getAge() - taxon.getMaxAge();
+            if (forceAdjust)
+            {
                 node->setAge(taxon.getMaxAge());
-                RBOUT("Age of taxon " + taxon.getName() + " was above the specified maximum and has been adjusted.");
-            } else throw RbException() << "Age of taxon " << taxon.getName() << " is above the specified maximum.";
+                
+                // only notify the user if we are making a non-trivial adjustment
+                if ( diff > RbSettings::userSettings().getTolerance() )
+                    RBOUT("Age of taxon " + taxon.getName() + " was above the specified maximum and has been adjusted.");
```
bredelings, line 982: Possibly we should say adjusted from old_age -> new_age?

davidcerny: Good idea – done.

davidcerny:
I've ended up changing `checkTaxonAges()` so that it is quiet unless the adjustment is large enough for the "before" and "after" ages to be different when printed to six significant digits. For what it's worth, I've never seen that happen.

Planning to merge this unless there are any objections @bredelings.

src/core/datatypes/trees/Tree.cpp
bredelings: In theory you could say how much the age was adjusted by, instead of adjusted from x -> y.  But you could also just merge it now.

PR #832, opened 2025-09-04
Conditioning options for the 'coding' argument in dnPhyloCTMC

basantakhakurel:
Adding the options to coding argument in dnPhyloCTMC. 
Previously, the help for dnPhyloCTMC appeared like this:
```
                    coding : 
                             Type:       String, <any>, value
                             Default:    all
```
This PR adds the options available for 'coding':
```
                    coding : Should any type of conditioning be applied?
                             Type:       String, <any>, value
                             Default:    all
                             Options:    all|noabsencesites|nopresencesites|informative|variable|nosingletonpresence|nosingletonabsence|nosingletons
```

### DIFF
--- a/src/revlanguage/distributions/phylogenetics/character/Dist_phyloCTMC.cpp
+++ b/src/revlanguage/distributions/phylogenetics/character/Dist_phyloCTMC.cpp
@@ -548,7 +548,7 @@ const MemberRules& Dist_phyloCTMC::getParameterRules(void) const
         dist_member_rules.push_back( new ArgumentRule( "treatAmbiguousAsGap", RlBoolean::getClassTypeSpec(), "Should we treat ambiguous characters as gaps/missing?", ArgumentRule::BY_VALUE, ArgumentRule::ANY, new RlBoolean( false ) ) );
 
         std::vector<std::string> codingOptions = {"all", "noabsencesites", "nopresencesites", "informative", "variable", "nosingletonpresence", "nosingletonabsence", "nosingletons"};
-        dist_member_rules.push_back(new OptionRule("coding", new RlString("all"), codingOptions, "Should any type of conditioning be applied?"));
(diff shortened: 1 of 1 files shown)


src/revlanguage/distributions/phylogenetics/character/Dist_phyloCTMC.cpp
```diff
@@ -547,7 +547,8 @@ const MemberRules& Dist_phyloCTMC::getParameterRules(void) const
 
         dist_member_rules.push_back( new ArgumentRule( "treatAmbiguousAsGap", RlBoolean::getClassTypeSpec(), "Should we treat ambiguous characters as gaps/missing?", ArgumentRule::BY_VALUE, ArgumentRule::ANY, new RlBoolean( false ) ) );
 
-        dist_member_rules.push_back( new ArgumentRule("coding", RlString::getClassTypeSpec(), "", ArgumentRule::BY_VALUE, ArgumentRule::ANY, new RlString("all") ) );
+        std::vector<std::string> codingOptions = {"all", "noabsencesites", "nopresencesites", "informative", "variable", "nosingletonpresence", "nosingletonabsence", "nosingletons"};
+        dist_member_rules.push_back(new OptionRule("coding", new RlString("all"), codingOptions, "Should any type of conditioning be applied?"));
```
davidcerny: This is super nitpicky, but strictly speaking "all" doesn't really work as an answer to: "Should any type of conditioning be applied?" Maybe the description could instead say something like "What character patterns should we condition on?" or "What character patterns have been sampled?"

davidcerny (changes requested):

src/revlanguage/distributions/phylogenetics/character/Dist_phyloCTMC.cpp
basantakhakurel: That's more clear way to put it. I'll change that.

davidcerny (approved):

PR #833, opened 2025-09-08
Options for calculating clade probabilities with sampled ancestors

davidcerny:
We have discussed multiple times (see #467, #787) whether an (A, B) clade in which A is a sampled ancestor of B should be considered different from an (A, B) clade in which both taxa descend from an unnamed internal node – a question that comes up when summarizing the posterior distributions of serially sampled time trees. Currently, we do indeed regard them as different, and there are some good reasons for this, but there are also good reasons for treating them as the same clade – see the linked discussions.

This PR simply equips the relevant functions (`annotateTree`, `consensusTree`, `mapTree`, and `mccTree`, as well as the `.cladeProbability()` method of class `TraceTree`) with an additional argument called `differentiateSAs` which enables switching between both behaviors. When `FALSE` (default), both clades are considered to be the same; when `TRUE`, the current behavior is restored and both clades are treated as distinct.


brpetrucci:
For others reading this PR, I'll add some context to the code since the way `Split` and `splitFrequency` are implemented makes things a little confusing (and is the reason David implemented things the way he did). 

The middle argument in `Split` is a taxon set representation of the MRCA of the clade in question, so either an empty taxon set (when the MRCA is a speciation node) or a one-taxon taxon set (when the MRCA is a sampled ancestor). We're not sure why it is a taxon set since it does not seem like it will ever be more than one taxon, but that's a problem for another time. So with that knowledge, let's see what must happen for each possible value of `differentiate_SAs`. Assume we're calculating the clade probability for a clade (A, B, C, D), with A as a sampled ancestor, in each case.

When `differentiate_SAs` is true, the if-statement within the for-loop will be hit exactly once, when taxon i is A, and the middle argument of `Split` in this case will be A. This call of `splitFrequency` will be equivalent to the function before David modified it, which makes sense. The if-statement outside the loop will not be hit (since `any_member_is_ancestor` is true), and the return will be as expected. As a side note, if the MRCA of the clade in question were a speciation node, this final if-statement would be hit, and once again we have the same return as before David altered the function. 

When `differentiate_SAs` is false, the if-statement within the for-loop will be hit once for every taxon, so we can add to `prob` the `splitFrequency` of splits where each possible taxon is a sampled ancestor (and therefore set as the middle argument of `Split`). Then the final if-statement will add to `prob` the `splitFrequency` of the case not differentiating SAs (since the middle argument is empty), and we'll have considered all possible cases where (A, B, C, D) is a monphyletic clade when not differentiating SAs, including the cases where B, C and D are the MRCAs as SAs of the rest.

brpetrucci:
Now as to my actual comment regarding the PR, and I'm interested in hearing others' opinions (in particular @bredelings) about it. 

I think `differentiate_SAs` should be true by default. This is not only because it'd make the retro-compatibility with how the summary tree functions are being used in tutorials and previously written code trivial, but because the true case is conceptually how every other software considers SAs when calculating posterior probabilities. I understand we have some disagreements within the community on the biological validity of that idea, and I think adding this argument is a good compromise, but I'd still like to have the default reflect the current consensus among researchers working with SA trees.

ixchelgzlzr:
David, I really like this solution of adding an argument that allows to decide how to treat sampled ancestors in multiple functions. Thank you so much for implementing! I'll find some time tonight to test these functions on my outputs.

Regarding what should be the default for the `differentiate_SAs` argument, I do not have strong opinions. I truly think that it all depends on your question. All I will say is that given the number of times this topic has independently come up for users (I am particularly thinking about how it can give some counterintuitive empirical results in the context of summarizing trees), it would be super beneficial to have the functions well documented and provide an example in one of the tutorials that use SAs?

bredelings:
It seems that the Split and Clade data types in RevBayes here modify the traditional definition of what a Split or Clade is by adding an optional MRCA taxon -- but only if the tree is rooted. So, whereas 
* the traditional clade would be a monophyletic set of taxa $S$, and two clades are the same clade if they have the same taxa
* the modified definition is that a clade is a monophyletic set of taxa $S$ _plus_ an MRCA, and two clades are the same if they contain the same taxa **AND** have the same MRCA (i.e. the two MRCAs must be either both missing, or both the same taxon).

So, we really have two things: a Clade and a CladeWithPossibleMRCA.  The problem is partly that they are going by the same name, despite being different things.  If a clade is a set of taxon $S$, then a CladeWithPossibleMRCA is either:
- $(S,  \emptyset)$ -- if there is _no_ MRCA
- $(S, \\{ m \\})$ -- if there _is_ an MRCA.

The other part of the problem is that Clade probabilities help you answer a specific question: 
> Q1: What is the probability that this set of taxa $S$ form a monophyletic group?

In contrast, CladeWithPossibleMRCA probabilities help you answer a different specific question.  Or perhaps "questions", since the question depends on the form of the CladeWithPossibleMRCA.

Thus, if the CladeWithPossibleMRCA has the form $(S,\emptyset)$, then we have:
> Q2a: What is the probability that the set of taxa $S$ forms a monophyletic group **AND** the MRCA of that group is not a sampled taxon?

In contrast, if the CladeWithPossibleMRCA has the form $(S,\\{m\\})$ then we have:
> Q2b: What is the probability that the set of taxa $S$ forms a monophyletic group **AND** the MRCA of that group is the sampled taxon $m$?

I guess this explains the code that @davidcerny wrote, since I think:
```math
\Pr ( S \text{ is monophyletic} ) = \sum_m \Pr( S \text{ is monophyletic} \wedge MRCA(S)=\{m\}) + \Pr(S\text{ is monophyletic} \wedge MRCA(S)=\emptyset)
```

The key point of confusion with biologists is that "clade" is well-known to mean a monophyletic group.  So when we report "clade probabilities" that are not the probability of being a monophyletic group, then that doesn't satisfy what people mean by "clade probability". The probability may be useful, but it is not the probability of a monophyletic group!  I would argue that we should call them clade-with-possible-mcra-probabilities :joy: .

bredelings:
I'm going to approve the PR as-is.  I think that defaulting to clade probabilities (and not clade-with-possible-mrca-probabilities) meets the expectation of most biologists -- and also fits the definition of "clade" as a monophyletic group.  Choosing this default may cause some pain, if we need to update some tutorials.  But it will also remove pain, as we will no longer necessarily need to explain that a "clade probability" actually isn't a "clade probability".

We may also want to add some documentation that define swhat the clade-with-possible-mrca-probability actually means, since it is not super-obvious.

Are there any RevBayes tutorials that depend on the clade-with-possible-mrca-probability?

bredelings (approved):

bredelings:
I notice that `collectTreeSample( )` ignores sampled ancestors at out-degree-1 nodes.  It only handles sampled-ancestor tips that are not internal nodes.  I have a small patch to fix this.  Should I push it here, or should that be a separate PR?

davidcerny:
Unsurprisingly, I'm with @bredelings. I agree it is important for us to follow "the current consensus among researchers working with SA trees" wherever possible, but I don't think that researchers working with SA trees get to redefine a term as old and basic as "clade" – and as Ben points out, when `differentiateSAs=true`, the values we are getting are _not_ clade probabilities, however useful or interesting they might be.

I would also point out that "researchers working with SA trees" include not just developers but also users, and in this repo alone, we've seen that multiple users consider the current "clade probabilities" to be either flat out wrong when they first encounter them, or not what they were looking for after the whole concept of "clade-with-possible-MRCA probabilities" is explained to them. So I'm not sure much consensus there really is.

> We may also want to add some documentation that define swhat the clade-with-possible-mrca-probability actually means, since it is not super-obvious.

Yes, this PR should probably be followed by one that adds documentation for (at least) the `mapTree()` and `mccTree()` functions. Any volunteers? 😛 

> Are there any RevBayes tutorials that depend on the clade-with-possible-mrca-probability?

Not that I can tell, although maybe I'm missing something @brpetrucci?

> I notice that `collectTreeSample( )` ignores sampled ancestors at out-degree-1 nodes. It only handles sampled-ancestor tips that are not internal nodes. I have a small patch to fix this. Should I push it here, or should that be a separate PR?

Either works as far as I'm concerned!

bredelings:
Hmm... maybe we should wait to merge this.  I was thinking about these probabilities mostly in terms of annotations, but if they also affect how we generate majority consensus, MAP and MCC trees, then we'd better check that the resulting MAP and MCC trees still make sense.  I think the CladeWithPossibleMRCA was partly invented in order to make MCC trees work in a simple way.  So if we use the normal definition of Clade, then we need to ensure that sampled ancestors are handled sensibly by these summary trees.

For majority consensus trees its pretty simple.  Basically we can't ignore leaf branches anymore, but otherwise everything else is the same.  So, a sampled ancestor gets its own leaf branch if the sampled-ancestor probability is < 50%, and otherwise it does not (and is a sampled ancestor).

For MAP trees, it should also be pretty simple.  You just count the number of times the topology occurs.

For MCC trees, when we take the product of the clade probabilities for a given tree, we multiply by $1 - Pr(SA)$ for each tip, and by $Pr(SA)$ for each taxon that is not a tip.

Technically the majority consensus and MCC part is probably worthy of a paper... but that would take time :roll_eyes: :grimacing:

brpetrucci:
> For MCC trees, when we take the product of the clade probabilities for a given tree, we multiply by $1 − Pr(SA)$ for each tip, and by $Pr(SA)$ for each taxon that is not a tip.

@bredelings could you clarify whether you mean $Pr(SA)$ for the set of trees where the taxon in question is in the clade in question, or in general? Either way I am wondering whether this makes sense. 

Let's say we have a clade $C$ that appears in a set of trees in our posterior, say $p$ in frequency. Say that clade has taxon $A$ as a sampled-ancestor MRCA with frequency $q$ (considering only the set of trees where $C$ is monophyletic), and a speciation node as MRCA with frequency $1-q$. Let $r$ be the total frequency of $A$ appearing as a sampled ancestor throughout the entire tree sample (i.e. $r = q + t$, if $t$ is the frequency with which $A$ appears as an SA MRCA of other clades). When not differentiating SAs, we would have the annotated clade probability of $C$ as $p$. Now the question is what do we multiply $p$ by when calculating the contribution of $C$ to the score of a tree with $C$ monophyletic. So there's two ways to interpret what Ben mentioned:

1. The score should be $pq$ if A is an SA, and $p(1-q)$ if not. Correct me if I'm wrong, but I think that would be equivalent to the current algorithm, where we consider the clade probability of $C$ taking into account whether its MRCA is a sampled ancestor or not. So I don't see why we would need to change the code, since the only difference would be in the interpretation of the probabilities, and therefore only the outward facing (i.e. tree annotation) side of things really matters.
2. The score should be $pr=pq+pt$ if A is an SA, and $p(1-r)=p(1-q)-pt$ if not. In this case it seems to me that all we're changing from the previous implementation is we're giving more weight to whether $A$ appears as an SA in non-$C$ clades (the $pt$ term). I don't see how that's desirable--why would we want to select trees with $A$ as an SA at a different rate based on when $A$ is an SA with a different set of sampled descendants? If the only difference between this and the previous implementation is that we're considering $A$'s status as an SA in other clades when calculating clade probabilities, I don't see how that's an improvement.

So either way I think that the clade probability calculations for MCC trees shouldn't change. For MAP trees, I'd argue that the topology IS different when $A$ is an SA vs. when it isn't, so I also don't see why we'd change the implementation there.

bredelings:
@brpetrucci Now that we have two ways to compute "clade" probabilities, it is possible to get two different answers for what the MAP tree and MCC tree are, depending on which "clade" probabilities are used.  If we use `differentiateSAs = TRUE`, then we will get the old result, which is fine.  

The question is what happens when we use the _existing_ code for MAP and MCC trees, but pass in _actual_ clade probabilities, not clade-with-possible-mrca probabilities.  For MCC trees, this strategy would simply ignore sampled ancestors, since it is assuming that they are included in the clade probabilities -- but if `differentiateSAs = FALSE`, then they are not. 

Does that help?

> For MAP trees, I'd argue that the topology IS different when A is an SA vs. when it isn't, so I also don't see why we'd change the implementation there.

I'm not quite sure what you mean here.  I think everyone agrees that the topology is different when A is a sampled ancestor or not.

> So there's two ways to interpret what Ben mentioned:

I can't quite tell if you mean $q=Pr(C \wedge MRCA(C)=\\{A\\})$ or $q=Pr(MRCA(C)=\\{A\\}|C)$?

What I was proposing is that when computing MCC trees with `differentiateSAs=FALSE`, we could do $Pr(C) \cdot Pr(A\text{ is a sampled ancestor})$ to score trees where A is a sampled ancestor, and $Pr(C) \cdot Pr(A\text{ is not a sampled ancestor})$ for trees when A is not a sampled ancestor.  Is that equal to your option (1) or (2)?

Note that if `differentiateSAs=FALSE`, we don't have access to $Pr(C \wedge MRCA(C) = \\{A\\}).$  That probability is equal to $Pr(MRCA(C) = \\{A\\}|C) \cdot PR(C)$. The MCC tree ignores conditional _clade_ probabilities, but it looks like with `differentiateSAs=TRUE` it accounts for conditional _sampled-ancestor_ probabilities.

P.S. Have you see this? https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1012789

Conditional clade probabilities are an improvement, but if you really want them then you should use a summary like this instead of the MCC tree, which is kind of a hack.

brpetrucci:
> I can't quite tell if you mean...

I meant $q = P(MRCA(C) = {A} | C)$, so the frequency of $A$'s presence as an SA at the root of $C$ considering the set of trees where $C$ is present.

> Is that equal to your option (1) or (2)?

That is what I meant by option 2. I see how that would require a change to the code due to what you mentioned regarding clade probabilities vs. clade-with-MRCA probabilities. But I'm not sure I see how that would be an improvement. The current code, using clade-with-MRCA probabilities, seems equivalent to me to my option 1, where we're considering the clade's frequency times the frequency of that clade's MRCA as an SA (by only considering clade-with-MRCA probabilities). So the difference between options 1 and 2 seem to me to be that option 2 puts more weight, when scoring a clade probability, on the frequency that that clade's MRCA (if it is an SA) appears as an SA in other clades in the sample besides this one. I don't see why that's better than only considering the frequency with which A appears as an MRCA in **this** clade. That seems more important to me when scoring trees than considering what happens in other clades

> P.S. Have you see this? https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1012789

Yes! My original idea for getting a budding tree summary going for my chapter 2 was to implement this on RevBayes. I think it's the future. But they don't even have a working implementation for it on trees with SAs (I'm in contact with the beast2 team to try and help make that happen), so it seems to me like this won't be an option on the time-frame I need for my chapter. Hence my insistence on getting MCC trees to work with SAs, and then with budding trees.

bredelings:
@brpetrucci OK, I think we agree that if we compute an MCC tree score using clade-with-MRCA probabilities, then it will likely be somewhat closer (but still not very close) to the probability of the topology.

> I don't see why that's better than only considering the frequency with which A appears as an MRCA in this clade. 

I never said it was better :-)  But to be clear, by "better", I think you mean "leads to a probability that is closer to the MAP tree probability when you multiply all the probabilities together assuming independence".

> Yes! My original idea for getting a budding tree summary going for my chapter 2 was to implement this on RevBayes. I think it's the future. But they don't even have a working implementation for it on trees with SAs (I'm in contact with the beast2 team to try and help make that happen), so it seems to me like this won't be an option on the time-frame I need for my chapter. Hence my insistence on getting MCC trees to work with SAs, and then with budding trees.

Hmm... maybe we should talk about this over zoom.  I will send you a slack message.

brpetrucci:
Talked to Ben and realized I had misunderstood the argument being had here. There's two things to consider

1. If we're not differentiating SAs in the calculation of clade probabilities, we won't consider SAs **at all** when scoring clades for the MCC tree calculations. That clearly isn't right, so we need to differentiate somewhere the clade probabilities we're using to calculate the MCC tree, and the ones we're reporting (which, in this case, would be pure clade probabilities). This situation hadn't clicked for me, and I agree we need to fix that.
2. The argument I was actually getting at is **how** we're including the SA probabilities in the clade probabilities. I was arguing that the current way (which considers only the conditional SA probabilities, only for the sample of trees where the given clade is monophyletic) was better. But I think I see the value in instead considering the full sample's frequency of a given taxon as an SA, because that leads to more accurate clade probabilities in the case where we don't want to differentiate SAs. I'll paste the example I wrote for Ben that convinced myself below:

Say clade {A, B, C, D} appears in 60% of our trees. In half of those, A is the MRCA as an SA, and in the other half the MRCA is a speciation node. In 30% of our trees, a clade {A, E, F, G} is present, and in all of those A is the MRCA as an SA. Let's say no other trees have A as an SA (so P(A is an SA) = 0.6)

If we consider the current calculations of clade probabilities in RevBayes and BEAST2, that would make $P({A,B,C,D}\ with\ A\ MRCA) = 0.3$, $P({A,B,C,D}\ with\ speciation\ node\ MRCA) = 0.3,$ and $P({A, E, F, G}\ with\ A\ MRCA) = 0.3$. 

If we instead consider total "pure" clade probabilities, and multiply them by the total frequency of A as an SA, that would become $P({A, B, C, D}\ with\ A\ MRCA) = 0.6\times0.6 = 0.36$, $P({A, B, C, D}\ with\ sp.\ node\ MRCA) = 0.6\times0.4 = 0.24$, and $P({A, E, F, G}\ with\ A\ MRCA) = 0.3\times0.6 = 0.18$.

So here, the second option gives a higher "score" to the more frequent clade, when A is an SA, because A is more frequently an SA than not and that clade is better represented in the sample. If instead A was not an SA in the {A, E, F, G} clade, we'd conversely give a higher score to the more frequent clade when A is not an SA. This seems more in line with the philosophy of not differentiating SAs. So I agree that'd be a desired fix. Maybe the fix would be to create separate methods for the clade probability calculations used to calculate the MCC tree and the ones used to annotate it?

All that said, I still think the default should be `differentiate_SAs = true` for backwards compatibility.

brpetrucci:
Made the necessary modifications to handle SAs when `differentiate_SAs`/`differentiate_MRCAs` is false. When that's the case, 

1. The `mccTree` function now multiplies the scores of a tree by the probability of each of its taxa being an SA (if it is an SA in the given tree) or not (if it isn't). 
2. The `mrTree` function now uses `cladeProbability` instead of `splitFrequency` to find the frequency of a given clade, so as to consider the `differentiate_SAs` argument, and sets the SA status of each taxon in the tree based on whether the taxon is an SA in `>= (100*cutoff)`% of trees in the sample.

In other words, when `differentiate_SAs` is true, we are considering the probability that a given taxon is an SA independently from the clade probabilities of the clades in which that taxon is present, both to score MCC trees and to build MR trees. Ben argued, and I've come to agree, that this generally produces tree summaries that are more representative of the tree sample. More thorough testing is needed to ensure that's the case, but giving users the option to do so is a positive.

That said, I changed the default of `differentiate_MRCAs` to `true`, because I think it is generally a bad idea to introduce an argument that completely changes the behavior of a user-facing function, and not make it backwards-compatible. I'm assuming this will still lead to some test failures, however, and will correct them as they come up.

Since I ended up contributing to this PR as well I don't feel comfortable reviewing it, so I'd love to hear others' thoughts on these changes and what else we'd like to change before merging this.

bredelings:
> That said, I changed the default of differentiate_MRCAs to true, because I think it is generally a bad idea to introduce an argument that completely changes the behavior of a user-facing function, and not make it backwards-compatible. I'm assuming this will still lead to some test failures, however, and will correct them as they come up.

This isn't a new argument.  David and I are not going to agree with you on this.  Please change it back.

hoehna (approved):
Looks good from my side.

brpetrucci:
After discussing in the monthly meeting, we have decided to change `differentiateMRCAs` to handle specifically how to *build* the summary trees, and report annotations both for the clade probability and the "configuration" probability--i.e. clade + MRCA. This will of course require us to change all tests that use tree summary functions on samples with SAs to have new expected outputs with both types of PPs. 

We also need to discuss what the names should be. We always output clade probabilities as `posterior`. Maybe that could be the name of the comment relating the actual "clade" probability (i.e. taking into account only species membership in a group), and the configuration probability can be reported as `posterior_with_mrca` or something of the sort. I suppose as long as we update the documentation it shouldn't be too bad. 

In other words, here's the TODO:

0. Fix segfault with `mrTree` when `differentiate_SAs = true`.
1. Change `posterior` comment in annotation functions to always consider `cladeProbability` with `differentiate_SAs = false`.
2. Add another comment for the configuration probability that uses `cladeProbability` with `differentiate_SAs = true`.
3. Ensure we're always using `differentiate_SAs` to determine how to actually build the summary trees.
4. Change test outputs of existing tests, and add tests with `differentiateMRCAs = false`.
5. Change documentation to reflect new annotation convention.

bredelings:
So the suggested fix is to print two attributes, perhaps PP_CLADE and PP_CONFIGURATION.  It then doesn't matter to @bjoelle which method is used to construct the MCC tree, though it might matter to @brpetrucci.

Are there other suggestions for the second name?  Configuration seems like it does a good job conveying what Bruno and Joelle liked about them.  I've been calling them "clades-with-mrca", which is a little more specific.  I'm curious if there's a good name that is specific because technically you could have other combinations of features (such as nodes, see below) that could be considered a "configuration".

Other questions: does this matter mostly for the MCC tree, or also for a majority-consensus tree with sampled ancestors?

Lastly, I think when we eventually switch to conditional clade (or split) trees, we won't need these configuration probabilities.  We will need clade (or split) probabilities and node probabilities, where a node equals the combination of all branches connected to a node plus any sampled ancestors at that node.  Presumably we might still want to print PP_CONFIGURATION?  And maybe also PP_NODE?

davidcerny:
> So the suggested fix is to print two attributes, perhaps PP_CLADE and PP_CONFIGURATION. It then doesn't matter to @bjoelle which method is used to construct the MCC tree, though it might matter to @brpetrucci.

I thought we all agreed that was an acceptable compromise but then @brpetrucci pointed out that's not really a workable solution, since the choice of clades vs. "configurations" determines not just which attributes we should print but also how to construct the tree they should be printed onto. Or did we find a solution to that?

> I've been calling them "clades-with-mrca", which is a little more specific.

I agree, I just find it a bit of a mouthful. (Also, should the plural be "clades-with-MRCA", "clade-with-MRCAs", or "clades-with-MRCAs"?). :) I'm pretty sure there is no pre-existing name for this concept, so no matter what we end up calling it, we'll have to provide an explanation of our terminology in the documentation. But yes, something succint and descriptive would be great.

PR #837, opened 2025-09-17
Update NCL

### LINKED ISSUE #768: [Feature Request] RevBayes contains a modified, old NCL instead of using upstream NCL
RevBayes contains an old copy of NCL (the Nexus Class Library) instead of using an upstream version.

This means that when bugs are fixed in upstream NCL, we do not benefit from them, and have to rediscover and re-fix them.  It also means that if we discover bugs and fix bugs, they do not propagate to the upstream NCL.  

If we could re-synchronize our local copy with upstream we could get whatever bug fixes they have applied since our copy diverged.

Probably we cannot simply replace our copy of NCL with the upstream version, due to local modifications.  Ideally we could:
- itemize our local modifications
- submit local bugfixes upstream
- determine if we want to keep some local changes that aren't accepted upstream
- switch to a newer version of NCL and re-apply local changes.

davidcerny:
This PR brings our version of the Nexus Class Library (NCL) up to date with the [upstream repo](https://github.com/mtholder/ncl) in a way suggested by @bredelings in #768: I switched to the current version of the library (commit [b7f4394](https://github.com/mtholder/ncl/commit/b7f4394591146f830ee117b93a9678712f6ffa22) from December 2025), itemized our local changes, and did my best to re-apply them.

Because the history of the repo was flattened by the 2019 "rapture", it was impossible to cherry-pick any commits older than that. Instead, I had to create new, equivalent commits. I made sure to (1) do this in the same chronological order in which the original commits were added, (2) point to the corresponding original commit in the commit message of each new commit, and (3) credit the authors of the original commits by making them co-authors on the new commits. In most cases, I did not preserve the original commit messages, because these usually made no reference to the changes made specifically to NCL.

## Comments on re-applied changes

Some of our local changes affect the functioning of NCL, but in ways that are specific to RevBayes and that we might not necessarily want to push upstream:

- [1cc6d90](https://github.com/revbayes/revbayes.archive/commit/1cc6d90f4cc43501e2240c0c9a2337661f5a31a1) (old) / bed953f (new): allow `_` and `-` in taxon labels. We could achieve the same effect by setting the bit flags `preserveUnderscores` and `hyphenNotPunctuation` that NCL specifically provides for this purpose, but this actually seems cleaner – it only requires editing one file as opposed to three.
- [189b94c](https://github.com/revbayes/revbayes.archive/commit/189b94cf07cc1af864d2df0e927f6ad5191571a3) (old) / fec4d71 (new): disable code for determining quoting requirements. This (1) allows unquoted taxon names to contain symbols such as `/`, but more importantly, it (2) ensures that names containing underscores are not quoted. The latter is needed for tests to pass: setting `preserveUnderscores` alone prevents the replacement of underscores by whitespace, but still causes the resulting names to appear in quotes.
- d7d78a5 (newly added): don't emit status messages. This re-applies a local change so ancient that there is no old commit to point to.

## Local changes that have not been not re-applied

Purely cosmetic changes:

- [cfe1e36](https://github.com/revbayes/revbayes.archive/commit/cfe1e36adbc422a3cdeb42275dd3daf0b9ff275d): adds a few blank lines to `nxsmultiformat.cpp`; seems pointless. My guess is the intention was to add some extra code there, but that has never happened.
- [b5c7f8a](https://github.com/revbayes/revbayes.archive/commit/b5c7f8ad82b34e5a1fef02605570ed3d337a0b3b): changes to function and variable names
- [c61385a](https://github.com/revbayes/revbayes.archive/commit/c61385aa97cde5cb4ebb1e30e821a9d061e404d1): changes to capitalization
- [177df69](https://github.com/revbayes/revbayes.archive/commit/177df6974f2b9815a42f8cd6a79fac9ddeeb2b88): camelCase -> snake_case
- [31b82b8](https://github.com/revbayes/revbayes.archive/commit/31b82b8ec55c6d818df0c8cc9091c15ca2699f3b): camelCase -> snake_case, again
- [fc67823](https://github.com/revbayes/revbayes.archive/commit/fc678233695cb3130fb23af3a7f871c3ae062869): whitespace insertions to bring NCL in line with our (very loose) style guidelines (e.g., `if (x)` rather than `if(x)`, etc.).

Questionable changes:

- [af4913b](https://github.com/revbayes/revbayes.archive/commit/af4913b27a47b41c784d7a232d1d5b250546ea7b): this may have been accidental; it likely stemmed from the confusion between `REALMATRIX` (a way of specifying a parsimony step matrix, sometimes found in the `ASSUMPTIONS` block of Nexus files produced by MacClade) and `MatrixReal` (a Rev type). In any case, we don't actually parse the `ASSUMPTIONS` block, so the change could not have been functional.

Redundant changes:

- [64d8e6d](https://github.com/revbayes/revbayes.archive/commit/64d8e6d78d2181e096c2f243b796f0b8c99e6619): make an implicit typecast explicit. I submitted this fix upstream where it got merged (see [#26](https://github.com/mtholder/ncl/pull/26)), so there is nothing left to re-apply.
- [08a107b](https://github.com/revbayes/revbayes.archive/commit/08a107b7ff152cb6f44ad687e734be7cb9d82c91): allow parsing Phylip files containing data of type `STANDARD`. I submitted this fix upstream where it got merged (see [#27](https://github.com/mtholder/ncl/pull/27)), so there is nothing left to re-apply.
- [ef40189](https://github.com/revbayes/revbayes.archive/commit/ef4018938650ec40d07f8828b4dc02933d817c9f): allow parsing files specifying `DATATYPE=BINARY` (BEAST's equivalent of the `RESTRICTION` type). I submitted this fix upstream where it got merged (see [#26](https://github.com/mtholder/ncl/pull/26)), so there is nothing left to re-apply.
- [33015c0](https://github.com/revbayes/revbayes.archive/commit/33015c03b66b718b42425f8fa26e2c5857efc7aa): an exception was commented out to circumvent a compiler issue, but a proper fix (consisting of a typecast) has been committed upstream in the meantime (see [22cb1cd](https://github.com/mtholder/ncl/commit/22cb1cdbe5ad6747628dc7fa87b16bbe18cc6086), [f517830](https://github.com/mtholder/ncl/commit/f5178302abddc24b8d8cb17cb149a9b646f579b3)).
- [652f80b](https://github.com/revbayes/revbayes.archive/commit/652f80be51f3541a2ecf33c8408c093d7836b20e): don't catch exceptions by value. I submitted this fix upstream where it got merged (see [#26](https://github.com/mtholder/ncl/pull/26)), so there is nothing left to re-apply.
- [093456c](https://github.com/revbayes/revbayes.archive/commit/093456c95eca1b641e8d98cd7ea58582a07fbb37): @bredelings's attempted fix of what GCC flags as "misleading indentation". Redundant with respect to a slightly different patch that I submitted upstream (see [#28](https://github.com/mtholder/ncl/pull/28)).
- [c223b0d](https://github.com/revbayes/revbayes.archive/commit/c223b0d50fe935296766114005c9e7033335dd34): copy `taxSets` properly. @bredelings later submitted this fix upstream where it got merged (see [#19](https://github.com/mtholder/ncl/pull/19)), so there is nothing left to re-apply.
- d2a1429: replacement of `sprintf` by `snprintf`. @bredelings later submitted this fix upstream where it got merged (see [#25](https://github.com/mtholder/ncl/pull/25)), so there is nothing left to re-apply.
- b90e3d2: my attempted fix for [#731](https://github.com/revbayes/revbayes/issues/731). Redundant with respect to a slightly different patch that had already been committed upstream (see [5e2cd01](https://github.com/mtholder/ncl/commit/5e2cd01bc70c299cbefe5f3ca1c40308813f99e1)).
- e60386e: a partial update rendered redundant by this comprehensive update

It also looks like several changes were made to the old copy of NCL even before it was first pushed to GitHub – these were mostly instances of indiscriminate find-and-replace, such as `boolean` -> `RlBoolean` and (inexplicably) `signals` -> `MemberObject`.


bredelings:
Wow!  Thanks for doing this!  Some initial thoughts:

* If we can drop edit to NCL to make it conform to our style requirements, then it should be easier to import changes from upstream NCL in the future.

* Ideally we would get our NCL changes upstream, and then we could remove the local changes when we re-import NCL from upstream.

* Some of our local changes seem...um... bad.  Like disabling the NCL quoting rules.  So, we could try and drop these local changes, or we could keep them for now if removing them means that tests fail.  But ideally we could drop them and tell people that they need to quote their taxon names if they want to include weird characters in them.  

* Manually reapplying individual edits from before the rapture means that at least we have a specific patch that we can point to and possibly revert.  Its unfortunate that the rapture lost the ancient history...

bredelings:
It looks like NCL has a few problems with C++20.  I wonder if the upstream NCL fixes these.

(Hmm... it seems like it doesn't?  So I guess this isn't really related to the current PR, so maybe I should move this comment to its own issue)

The issue seems to be that `nxscharactersblock.h` does 

``` C++
class NxsDiscreteDatatypeMapper;

class NxsCharactersBlock
{
...
		typedef std::pair<NxsDiscreteDatatypeMapper, NxsUnsignedSet> DatatypeMapperAndIndexSet;
		typedef std::vector<DatatypeMapperAndIndexSet> VecDatatypeMapperAndIndexSet;
...
};

class NxsDiscreteDatatypeMapper
{
...
		NxsDiscreteDatatypeMapper(NxsCharactersBlock::DataTypesEnum datatypeE, bool hasGaps);
...
};
```

The error is because the definition for `NxsDiscreteDatatypeMapper` is not known when `std::pair<NxsDiscreteDatatypeMapper, NxsUnsignedSet>` is first used.  This was allowed in C++17 in some circumstances, but is not longer allowed in C++20.

We could simply move the definition of NxsDiscreteDatatypeMapper above the definition of NxsCharactersBlock, except that NxsDiscreteDatatypeMapper also references NxsCharactersBlock through NxsCharactersBlock::DataTypesEnum, creating a reference loop.

davidcerny:
> - If we can drop edit to NCL to make it conform to our style requirements, then it should be easier to import changes from upstream NCL in the future.

I agree this is a good idea – the only downside is that the diff becomes a little unwieldy. That's an acceptable trade-off, though, so I have dropped the commits that re-applied purely cosmetic changes. (I have also changed the PR description to account for this, since I think it's important to keep a record of what was done and why.)

> - Ideally we would get our NCL changes upstream, and then we could remove the local changes when we re-import NCL from upstream.

In principle I agree with this too, but as I said on Slack, I'm a little worried about how realistic it is. The last time a PR got merged upstream was in October 2023.

> - Some of our local changes seem...um... bad. Like disabling the NCL quoting rules. So, we could try and drop these local changes, or we could keep them for now if removing them means that tests fail. But ideally we could drop them and tell people that they need to quote their taxon names if they want to include weird characters in them.

Yes, dropping that one change unfortunately does make a huge number of tests fail. It's not so much about including weird characters in unquoted taxon names – we do have [one test](https://github.com/revbayes/revbayes/blob/master/tests/test_coalescent/scripts/mcmc_heterochronous_BSP.Rev) that involves reading in a FASTA file with unquoted taxon names containing a forward slash (`/`), but we could obviously do what you suggest and make it pass by quoting those names instead. The real problem is that without the change, all names containing underscores are automatically quoted, which is the case even if we prevent the automatic replacement of underscores by whitespace. This then leads to multiple issues:

- taxon name discrepancies in tests that involve matching a matrix to a tree
- taxon name discrepancies in tests that involve matching a matrix to a taxon file (e.g., one that specifies tip ages)
- taxon name discrepancies in tests that involve specifying an outgroup by name
- mismatches with expected output even in those tests that do actually run

So unfortunately, I don't think we can afford to drop the change.

One somewhat dubious change we could drop is [1cc6d90](https://github.com/revbayes/revbayes.archive/commit/1cc6d90f4cc43501e2240c0c9a2337661f5a31a1), which allows taxon names to contain underscores (`_`) and hyphens (`-`). NCL provides the bit flags `preserveUnderscores` and `hyphenNotPunctuation` that have the same effect, so we could just set those instead. But that would requite modifying 3 NCL files in 5 places, as opposed to a single edit. So in terms of making it easier to import upstream changes in the future, this would actually seem counterproductive.

davidcerny:
FWIW, there are some compiler warnings that we may or may not want to address before merging this:

```
../src/libs/ncl/nxscxxdiscretematrix.cpp: In function ‘void NxsTransposeCompressedMatrix(const std::vector<NxsCharacterPattern>&, ScopedTwoDMatrix<signed char>&, std::vector<unsigned int>*, std::vector<double>*)’:
../src/libs/ncl/nxscxxdiscretematrix.cpp:260:5: warning: this ‘if’ clause does not guard... [-Wmisleading-indentation]
  260 |     if (patternWeights)
      |     ^~
../src/libs/ncl/nxscxxdiscretematrix.cpp:262:9: note: ...this statement, but the latter is misleadingly indented as if it were guarded by the ‘if’
  262 |         for (unsigned p = 0; p < npatterns; ++p)
      |         ^~~
```

and

```
../src/libs/ncl/nxsmultiformat.cpp: In member function ‘void MultiFormatReader::readPhylipData(FileToCharBuffer&, const NxsDiscreteDatatypeMapper&, std::__cxx11::list<std::__cxx11::basic_string<char> >&, std::__cxx11::list<std::vector<int> >&, unsigned int, unsigned int, bool)’:
../src/libs/ncl/nxsmultiformat.cpp:529:34: warning: ‘currentTaxon’ may be used uninitialized [-Wmaybe-uninitialized]
  529 |                 if (currentTaxon + 1 != n_taxa)
      |                     ~~~~~~~~~~~~~^~~
../src/libs/ncl/nxsmultiformat.cpp:455:18: note: ‘currentTaxon’ was declared here
  455 |         unsigned currentTaxon;
      |                  ^~~~~~~~~~~~
../src/libs/ncl/nxsmultiformat.cpp: In member function ‘void MultiFormatReader::readInterleavedPhylipData(FileToCharBuffer&, const NxsDiscreteDatatypeMapper&, std::__cxx11::list<std::__cxx11::basic_string<char> >&, std::__cxx11::list<std::vector<int> >&, unsigned int, unsigned int, bool)’:
../src/libs/ncl/nxsmultiformat.cpp:666:34: warning: ‘currentTaxon’ may be used uninitialized [-Wmaybe-uninitialized]
  666 |                 if (currentTaxon + 1 != n_taxa)
      |                     ~~~~~~~~~~~~~^~~
../src/libs/ncl/nxsmultiformat.cpp:564:18: note: ‘currentTaxon’ was declared here
  564 |         unsigned currentTaxon;
      |                  ^~~~~~~~~~~~
```

and also

```
../src/libs/ncl/nxstreesblock.cpp: In static member function ‘static void NxsTreesBlock::ProcessTokenStreamIntoTree(NxsToken&, NxsFullTreeDescription&, NxsLabelToIndicesMapper*, std::map<std::__cxx11::basic_string<char>, unsigned int>&, bool, NxsReader*, bool, bool, bool, bool, bool, bool)’:
../src/libs/ncl/nxstreesblock.cpp:1714:9: warning: this ‘if’ clause does not guard... [-Wmisleading-indentation]
 1714 |         if (allowUnquotedSpaces)
      |         ^~
../src/libs/ncl/nxstreesblock.cpp:1717:17: note: ...this statement, but the latter is misleadingly indented as if it were guarded by the ‘if’
 1717 |                 token.GetNextToken();
      |                 ^~~~~
```

bredelings (approved):

PR #838, opened 2025-09-18
Allow handling partial ambiguities involving a gap state

### LINKED ISSUE #784: [BUG] Segmentation fault when loading data with ambiguous inapplicable character
**Describe the bug**
RevBayes crashes when parsing a matrix that contains the state `(- 1)`

**To Reproduce**
Run `readDiscreteCharacterData("ambig-inapp.txt")`.

The file [ambig-inapp.txt](https://github.com/user-attachments/files/20768474/ambig-inapp.txt) contains:

```nexus
#NEXUS
BEGIN DATA;
  DIMENSIONS NTAX=2 NCHAR=2;
	FORMAT DATATYPE=STANDARD GAP=- MISSING=? SYMBOLS="01";
	MATRIX
    A 11
    B (0 1)(- 1)
  ;
END;
```

**Expected behavior**
The file should open.  If the combination (- 0) cannot be handled correctly by RevBayes, then it could be treated as `?`, perhaps with a warning?

I suspect that Martin Brazeau's recent activity in the Nexus Class Library may pertain to this issue – so we may be able to pull a fix from upstream.

**Computer info**
Latest RevBayes build, Win 10.

**Additional context**
In case `(- 1)` leaves you scratching your head: this can arise where a certain state can be excluded.  For example, consider coding a fossil taxa for the two characters:

1. Specialized feeding organ: (0) absent, (1) present
2. Feeding organ composition: (-) Inapplicable: organ not present; (1) Made of soft tissue with no chance of fossilization; (2) Biomineralized tissue with strong fossilization potential

If a specialized feeding organ is not observed, but other organisms made of Biomineralized tissue with strong fossilization potential are present, the two scenarios are (i) the organ was absent, and should be scored `0-`; (ii) it was present, but didn't fossilize, so should be scored `0 (- 1)`

I suspect that most models implemented by RevBayes would treat the `(- 1)` state as equivalent to `(? 1)`, and by extension, `?` – but such combinations are likely to occur in well-designed palaeontological matrices that are intended to be analysed with inapplicable-aware algorithms.

davidcerny:
This PR fixes #784 the way @ms609 suggested: a partial ambiguity that involves a gap / inapplicable state, such as `(0 -)` or `{1 -}`, gets converted to `?`, and a warning is emitted indicating that this is what happened.

This mechanism is only implemented for the `Standard` data type, since I'm pretty sure there can't possibly be a molecular alignment where there is an ambiguity as to whether a site is a gap or occupied by a particular residue. If I'm wrong about this, then a more general solution might be in order.

A more satisfying solution would be to treat ambiguously inapplicable matrix cells as such, but that would require massive changes to the way we represent and handle `Standard` data. Currently, `is_gap` is a simple boolean – a matrix cell cannot be a gap just in part.


ms609 (approved):
This looks great to me – this is a sensible placeholder approach until it's justified to implement a more sophisticated special treatment.  I'm afraid I won't have time to scrutinise the code or to test the code locally over the next couple of weeks.

bredelings:
Hmm.. this is fine, I guess.  The other option would be to throw an exception and say that people need to fix the data file.  But I think a warning is sufficient.  Presumably this will still crash if we say {- A} for a DNA file?

davidcerny:
I was pleasantly surprised to discover that for a DNA file, we already throw a nice and informative exception on encountering {- A}:

```
   Error:	Cannot add state 'A' to a DNA character with value '-'!
```

Based on what @ms609 wrote in #784, though, I do think it makes sense to handle this situation differently for standard vs. DNA sequence data (using a warning for the former and an exception for the latter). {0 -} is something a paleontologist might conceivably want to score a fossil as having, in some very specific circumstances.

PR #839, opened 2025-09-22
updated help file for mvFNPR

mcranium:
I added a help file for mvFNPR

### DIFF
--- a/help/md/mvFNPR.md
+++ b/help/md/mvFNPR.md
@@ -1,9 +1,16 @@
 ## name
 mvFNPR
 ## title
+Fixed Nodeheight Prune and Regraft Proposal move.
 ## description
+A subtree is pruned and re-attached without changing the node heights.
 ## details
 ## authors
 ## see_also
+mvSPR
 ## example
 ## references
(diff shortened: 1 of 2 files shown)


mcranium:
Not sure why this does not proceed. The only potential root that comes to my mind is that I used one "special" character: ö for the author name Höhna in the references.

davidcerny:
Thank you for submitting this, @mcranium! Some brief feedback:

- The words "move" and "proposal" are near-synonyms (in the context of RevBayes, a move's main job is to perform or revert a proposal), so it's redundant for the title to use both.
- People who have experience using other phylogenetic software like BEAST might be more familiar with the term "[Wilson-Balding](https://academic.oup.com/genetics/article-abstract/150/1/499/6034402) move", so it may be worth explaining the relationship between the two (namely, that FNPR is a subset of Wilson-Balding that often exhibits higher acceptance rates).
- It would be nice for the help file to contain a small self-contained example: you can take a look at the `mvSPR` help page for inspiration.

> Not sure why this does not proceed. The only potential root that comes to my mind is that I used one "special" character: ö for the author name Höhna in the references.

That's not it, I think. One of the GitHub Actions was failing stochastically, which is annoying but does occasionally happen – I've solved that just by re-running it. The one which is still failing now, and which prevents the PR from being merged, is just a Slack notification. I'm pretty sure this is happening because you're trying to merge from an external repo.

hoehna (approved):
Thanks for this.

Issue #840, opened 2025-09-23
[BUG] Error in bootstrap estimation of marginal likelihood estimate error


ms609:
**Describe the bug**
`Error: NaN produced in rgeom` when estimating path/stepping stone sampler `stdError` with `bootstrap = TRUE`

**To Reproduce**
Check out https://github.com/neo-trans/104_by_kv

```revbayes
ppFile = <path to>/104_by_kv/104_by_kv.pp 

ss = steppingStoneSampler(
  file = ppFile,
  powerColumnName = "power",
  likelihoodColumnName = "likelihood"
)
ps = pathSampler(
  file = ppFile,
  powerColumnName = "power",
  likelihoodColumnName = "likelihood"
)
# Will warn to recommend bootstrapping
write(ss.marginal(), ps.marginal(), ss.stdError(), ps.stdError())

ss.stdError(bootstrap = TRUE)
#    Error:       NaN produced in rgeom

ps.stdError(bootstrap = TRUE)
#    Error:       NaN produced in rgeom
```

**Expected behavior**
Values to be computed, or informative error to be produced

**Computer info**
Windows 11, RevBayes 1.3.1 release

ms609:
Closed by #842 – thank you!

PR #842, opened 2025-09-25
Fix block bootstrap with small values of n

bredelings:
Issue #840 was caused by trying to run a block bootstrap with n*prop=0.9, which was calling `RbStatistics::Geometric::rv( )` with `p=1.1111111` (not NaN).


davidcerny (approved):
Thank you!

ms609:
Built and installed; I can confirm that this fixes #840 for me, so merging.

PR #851, opened 2025-09-30
Make sure `?<x>` and `help("<x>")` are equivalent for all `<x>`

### LINKED ISSUE #850: [BUG] `help("args")` not recognized
**Describe the bug**
`help("args")` fails, but `?args` succeeds

**To Reproduce**

Image

**Expected behavior**
`help("args")` should serve as an alias for `?args`, as is the case for e.g. `help("dnNormal")`

**Computer info**
Win 10, RB devel

davidcerny:
Just a quick follow-up to #830, intended to solve #850.


bredelings:
Can you quickly explain what this patch is doing?  Why is ` hs.isHelpAvailableForQuery(baseQuery)` failing?

davidcerny:
Sure: `args` is neither a type nor a function, so our usual help infrastructure contained in `RbHelpSystem` doesn't apply to it. To get around this, I introduced a more general help-rendering method into `RbHelpRenderer`, along with a check in `Parser` that tries to look for an entry in `RbHelpDatabase` if it doesn't find it in `RbHelpSystem`. This made `?args` work, but it turns out that `help("<x>")` is handled elsewhere, so this additional check needs to be inserted into `Func_help` as well.

bredelings:
Hmm... I have a vague memory that a whole bunch of things were lumped under "type" that maybe weren't types.

In any case, it still seems odd that `hs.isHelpAvailableForQuery(baseQuery)` answers 'false'.  Couldn't we make this answer `true` and make `getHelp( )` work instead having two parallel systems?

davidcerny:
I don't find it too strange. The only thing that `RbHelpSystem::isHelpAvailableForQuery` does is:

https://github.com/revbayes/revbayes/blob/cbe72b0372cf2c869d951293b383181bacdb9e99/src/core/help/RbHelpSystem.cpp#L341

So it makes sense it would return `false` for a `baseQuery` that is neither.

I guess we could define a new class on par with `RbHelpType` and `RbHelpFunction` (maybe `RbHelpArbitrary`?) and teach `RbHelpSystem` to look for it and to parse it. I'm not sure that would be cleaner or more concise, though...

hoehna (approved):
This seems to be a peculiar bug, but solves the issue.

Issue #855, opened 2025-10-03
[BUG] Function calls can't be used as vector index, within []


Boussau:
Hello,

**Describe the bug**
If a function call is used as an index of a vector, within [], revbayes complains

**To Reproduce**
> abs(-3)
   3
> test[abs(-3)] <- 3
   Error:	Index 0 not allowed in 'test[0]'.  The first element should be 'test[1]'
> test[3] <- 3 

**Expected behavior**
I'd expect function calls to work within []

**Computer info**
The problem was encountered with rebayes 1.3.0, both on Ubuntu and Windows.

Thanks!

bredelings:
This seems to result from the fact that `abs(-3)` is `RealPos`, but the code for `test[...]` assumes it is `IntegerPos`.

In `revlanguage/parser/SyntaxIndexOperation.cpp`:
>   191     int idx = (int)static_cast<Integer&>( indexVar->getRevObject() ).getValue();                                                                  │

The value for `indexVar` is actually correct -- it is 3 -- but somehow this line returns a zero.

The same thing happens with `test[1.5]`:
```
> test[1.5] = 5
   Error:	Index 0 not allowed in 'test[0]'.  The first element should be 'test[1]'
```

Issues:
- we should probably have an integer version of abs( )
- we should give a sensible error message if the subscript is not an integer

bredelings:
Fixed by #856

PR #856, opened 2025-10-03
Fix `Real` subscript to array

bredelings:
This PR fixes issue #855.  The issue is that `abs(-3)` returns RealPos instead of Natural, and also subscripting by Reals doesn't work -- it does a bad typecast and ends up treating the real subscript as zero.  Additionally, it seems that errors in interactive mode were causing the interpreter to quit?

Each commit/patch pretty much does what it says in the title.

### DIFF
--- a/src/revlanguage/functions/math/Func_absInt.cpp
+++ b/src/revlanguage/functions/math/Func_absInt.cpp
@@ -1,0 +1,117 @@
+#include <iosfwd>
+#include <string>
+#include <vector>
+
+#include "AbsoluteValueFunction.h"
+#include "Func_absInt.h"
+#include "Integer.h"
+#include "Natural.h"
+#include "RlDeterministicNode.h"
+#include "TypedDagNode.h"
+#include "Argument.h"
+#include "ArgumentRule.h"
+#include "ArgumentRules.h"
+#include "DeterministicNode.h"
+#include "DynamicNode.h"
+#include "RbHelpReference.h"
+#include "RevPtr.h"
+#include "RevVariable.h"
+#include "RlFunction.h"
+#include "RlTypedFunction.h"
+#include "StringUtilities.h"
+#include "TypeSpec.h"
+#include "TypedFunction.h"
+#include "GenericFunction.h"
+
+using namespace RevLanguage;
+
+/** default constructor */
+Func_absInt::Func_absInt( void ) : TypedFunction<Natural>( )
+{
+    
+}
+
+
+/**
+ * The clone function is a convenience function to create proper copies of inherited objected.
+ * E.g. a.clone() will create a clone of the correct type even if 'a' is of derived type 'b'.
+ *
+ * \return A new copy of the process.
+ */
+Func_absInt* Func_absInt::clone( void ) const
+{
+    
+    return new Func_absInt( *this );
+}
+
+// Create a non-polymorphic function.
+std::int64_t* absInt(std::int64_t x)
+{
+    return new std::int64_t(std::abs(x));
+}
+
+RevBayesCore::TypedFunction<std::int64_t>* Func_absInt::createFunction( void ) const
+{
+    auto* arg = static_cast<const Integer &>( this->args[0].getVariable()->getRevObject() ).getDagNode();
+
+    return RevBayesCore::generic_function_ptr< std::int64_t >( absInt, arg );
+}
+
+
+/* Get argument rules */
+const ArgumentRules& Func_absInt::getArgumentRules( void ) const
+{
+    
+    static ArgumentRules argumentRules = ArgumentRules();
+    static bool          rules_set = false;
+    
+    if ( !rules_set )
+    {
+        
+        argumentRules.push_back( new ArgumentRule( "x", Integer::getClassTypeSpec(), "A (possibly negative) number.", ArgumentRule::BY_CONSTANT_REFERENCE, ArgumentRule::ANY ) );
+        
+        rules_set = true;
+    }
+    
+    return argumentRules;
+}
+
+
+const std::string& Func_absInt::getClassType(void)
+{
+    
+    static std::string rev_type = "Func_absInt";
+    
+    return rev_type; 
+}
+
+
+/* Get class type spec describing type of object */
+const TypeSpec& Func_absInt::getClassTypeSpec(void)
+{
+    
+    static TypeSpec rev_type_spec = TypeSpec( getClassType(), new TypeSpec( Function::getClassTypeSpec() ) );
+    
+	return rev_type_spec; 
+}
+
+
+/**
+ * Get the primary Rev name for this function.
+ */
+std::string Func_absInt::getFunctionName( void ) const
+{
+    // create a name variable that is the same for all instance of this class
+    std::string f_name = "abs";
+    
+    return f_name;
+}
+
+
+const TypeSpec& Func_absInt::getTypeSpec( void ) const
+{
+    
+    static TypeSpec type_spec = getClassTypeSpec();
+    
+    return type_spec;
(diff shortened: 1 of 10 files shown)


src/revlanguage/functions/math/Func_absInt.cpp
```diff
@@ -0,0 +1,119 @@
...
+{
+    return new std::int64_t(std::abs(x));
+}
+
+RevBayesCore::TypedFunction<std::int64_t>* Func_absInt::createFunction( void ) const
+{
+    
+    auto* arg = static_cast<const Integer &>( this->args[0].getVariable()->getRevObject() ).getDagNode();
+
+    // Select the version of sqrt that takes a double.
```
davidcerny: This comment should be replaced

src/revlanguage/functions/math/Func_absVectorInt.cpp
```diff
@@ -0,0 +1,131 @@
...
+    return abs_x;
+}
+
+
+RevBayesCore::TypedFunction<RevBayesCore::RbVector<std::int64_t> >* Func_absVectorInt::createFunction( void ) const
+{
+    
+    auto* arg = static_cast<const ModelVector<Integer> &>( this->args[0].getVariable()->getRevObject() ).getDagNode();
+
+    // Select the version of sqrt that takes a double.
```
davidcerny: Another comment to replace

tests/test_basics/scripts/indexing.Rev
```diff
@@ -0,0 +1,20 @@
...
+
+print("> x[-1] = -1")
+x[-1] = -1
+print("> x[0] = 0")
+x[0] = 0        
+print("> x[1] = 1")
+x[1] = 1
+print("> x[2] = -2")
+x[2] = -2
+print(x)
```
davidcerny, line 16: Maybe add `print("> print(x)")` before this line, just for consistency's sake

src/revlanguage/functions/math/Func_absInt.cpp
```diff
@@ -0,0 +1,119 @@
...
+#include "RlTypedFunction.h"
+#include "StringUtilities.h"
+#include "TypeSpec.h"
+#include "TypedFunction.h"
+#include "GenericFunction.h"
+
+using namespace RevLanguage;
+
+/** default constructor */
+Func_absInt::Func_absInt( void ) : TypedFunction<Natural>( )
```
davidcerny, line 29: Not a big deal by any means, but this line triggers a compiler warning using Clang 17:

```
/Users/David/Grive/Slater_Lab/RevBayes_fork/revbayes/src/revlanguage/functions/math/Func_absInt.cpp:29:14: note: in instantiation of member function 'RevLanguage::TypedFunction<RevLanguage::Natural>::execute' requested here
   29 | Func_absInt::Func_absInt( void ) : TypedFunction<Natural>( )
      |              ^
/Users/David/Grive/Slater_Lab/RevBayes_fork/revbayes/src/core/dag/DagNode.h:72:69: note: hidden overloaded virtual function 'RevBayesCore::DagNode::executeMethod' declared here: type mismatch at 2nd parameter ('const std::vector<const DagNode *> &' vs 'const std::vector<Argument> &')
   72 |         void                                                        executeMethod(const std::string &n, const std::vector<const DagNode*> &args, double &rv) const; //!< Map the member methods to internal function calls
      |                                                                     ^
```

src/revlanguage/functions/math/Func_absVectorInt.cpp
```diff
@@ -0,0 +1,131 @@
...
+#include "StringUtilities.h"
+#include "TypeSpec.h"
+#include "TypedFunction.h"
+#include "UserFunctionNode.h"
+#include "GenericFunction.h"
+
+using namespace RevLanguage;
+
+/** default constructor */
+Func_absVectorInt::Func_absVectorInt( void ) : TypedFunction< ModelVector<Natural> >( )
```
davidcerny, line 36: Another line triggering compiler warnings

src/revlanguage/functions/math/Func_absVectorInt.h
```diff
@@ -0,0 +1,48 @@
...
+     * The RevLanguage wrapper of the absolute value function.
+     *
+     * The RevLanguage wrapper of the abs function connects
+     * the variables/parameters of the function and creates the internal AbsoluteValueFunction object.
+     * Please read the AbsoluteValueFunction.h for more info.
+     *
+     *
+     * @copyright Copyright 2009-
+     * @author The RevBayes Development Core Team (Sebastian Hoehna)
+     * @since 2014-07-27, version 1.0
```
davidcerny: Technically, this is inaccurate insofar as it concerns this particular file rather than RevBayes as a whole... Again, not a big deal, but I can imagine situations where knowing who added a particular function might be useful.

davidcerny (changes requested):
This looks mostly good to me, with a nice test to boot. (I suppose this is one case where it would be nice to have an `--echo` command-line option, so that the commands don't have to be printed to the standard output manually one by one.)

Playing around with the code on this branch made me realize how... interestingly... our vectors actually behave. E.g.:

```
> x[abs(-3)] = 3
> x[3]
   3
> x
Error:	Cannot create vector variable with name 'x' because element with name 'x[1]' is NULL.
```

But:

```
> y[3] = 3
> y[2] = 2
> y[1] = 1
> y
   [ 1, 2, 3 ]
```

It'd be good to have these nuances documented somewhere, but that's outside the scope of this PR.

src/revlanguage/functions/math/Func_absVectorInt.h
bredelings: Fixed by removing that block.

src/revlanguage/functions/math/Func_absInt.cpp
bredelings: This looks like a problem in DagNode<T>::executeMethod( ) that comes from code in TypedFunction<T>::execute( ).

It is not great, but perhaps too far afield to solve here.

src/revlanguage/functions/math/Func_absVectorInt.cpp
bredelings: Same as above, I presume.

tests/test_basics/scripts/indexing.Rev
bredelings: OK.

src/revlanguage/functions/math/Func_absVectorInt.cpp
bredelings: Fixed.

src/revlanguage/functions/math/Func_absInt.cpp
bredelings: Fixed.

bredelings:
Regarding the "interesting" behavior, it seems like its just saying that if you define `x[n]` then you can't create print `x` until you also create `x[1]...x[n-1]`, which kind of makes sense.

This reminds me that I have been thinking that setting the individual values of the vector sequentially is kind of problematic.  Instead of writing
```
for(i in 1:n)
{
   x[i] = .....i......
}
```
it seems more logical to define all the elements at once by doing something like:
```
x = [sqrt(i) for i in 1:n]
```
In Haskell you define arrays by giving bounds and a function of the index that defines all the elements.  You then don't need to set the first element, second element, third element, etc.

src/revlanguage/functions/math/Func_absInt.cpp
davidcerny: Fair enough!

davidcerny (approved):

PR #857, opened 2025-10-03
Update README.md for meson project

bredelings:
Update the meson build instructions.  For example, we don't use git submodules any more.


projects/meson/README.md
```diff
@@ -1,6 +1,6 @@
 # Building with meson
 
-[Meson](https://mesonbuild.com/) ([git](https://github.com/mesonbuild/meson)) is a new build system that aims to replace autotools and fills the same role as CMake.
+[Meson](https://mesonbuild.com/) ([git](https://github.com/mesonbuild/meson)) is build system that fills the same roll as CMake.
```
basantakhakurel: role?

projects/meson/README.md
```diff
@@ -9,131 +9,110 @@ Meson makes it easy to cross-compile to Windows, Macintosh, and ARM systems from
...
     If you are not using the system version of boost, first [tell meson where to find BOOST](#Telling-meson-where-to-find-BOOST).
-   
-    ``` sh
-    projects/meson/generate.sh
-    meson build --prefix=$HOME/Applications/revbayes
-    ninja -C build install
-    ```
+
+        projects/meson/generate.sh
+        meson build --prefix=$HOME/Applications/revbayes
```
basantakhakurel: meson complains if you use `meson build` instead of `meson setup build`. 
```
WARNING: Running the setup command as `meson [options]` instead of `meson setup [options]` is ambiguous and deprecated.
```
Maybe better to have it `meson setup build`?

basantakhakurel (changes requested):
Hi Ben,
This looks good. 
There is a typo and also see the comment about deprecated usage.

projects/meson/README.md
davidcerny: Also, it's okay to remove the adjective ("new") but not the indefinite article ("a").

projects/meson/README.md
```diff
@@ -9,131 +9,110 @@ Meson makes it easy to cross-compile to Windows, Macintosh, and ARM systems from
...
-``` sh
-dnf install gcc-c++ meson boost-devel
-```
+
+    dnf install gcc-c++ meson boost-devel
+
 
 ### Other systems or not root - use python3
 
 1. You need pip to install meson. Either pip or pip3 is fine, as long as its using python 3.7 or later.
```
davidcerny: While we're at it, this should read "as long as it is" rather than "as long as its".

bredelings: I changed `its` to `it's`.  They are pronounced identically...

projects/meson/README.md
bredelings: Good point, that is the new style, we should use a verb (`setup`) as the first argument everywhere.

davidcerny (approved):

basantakhakurel (approved):

PR #859, opened 2025-10-03
Fix majority consensus crash with sampled ancestors

bredelings:
The majority consensus code starts with a star tree, and iterates over supported clades.  For each supported clade it marks any mrca taxa as sampled ancestors, and then divides the children to impose the split described by the clade bitmask.

The problem here is that sometime tree already contains the split, and the code bailed early _before_ marking mrca taxa as sampled ancestors.  Such tips then correspond to the wrong clade.  In this case, the wrong clade had zero sampled times.  We then looked at an element of an empty vector and got a SEGFAULT.

The fix is to move the code which bails on splits that we already have _after_ the code to mark sampled-ancestor tips.

(I just realized this patch might be hard to review because it changes a lot of whitespace and other cleanups.  Let me know if you want me to tweak it for easier review.)


src/core/analysis/mcmc/output/TreeSummary.cpp
```diff
@@ -925,71 +925,67 @@ Tree* TreeSummary::mrTree(AnnotationReport report, double cutoff, bool verbose)
...
-                    continue;
-                }
-                else
-                {
-                    for (size_t i = 0; i < mrca.size(); i++)
-                    {
-                        mrca[i]->setSampledAncestor(true);
-                    }
-                }
+                // Add the child to the mrca if its a tip and its taxon is in clade.second
```
davidcerny: The first "its" here should read "it's" or "it is"

src/core/analysis/mcmc/output/TreeSummary.cpp
```diff
@@ -925,71 +925,67 @@ Tree* TreeSummary::mrTree(AnnotationReport report, double cutoff, bool verbose)
...
+                if ( child->isTip() && std::find(clade.second.begin(), clade.second.end(), child->getTaxon() ) != clade.second.end() )
+                    mrca.push_back(child);
             }
 
-            nIndex++;   //increment node index
-            TopologyNode* intNode = new TopologyNode(nIndex); //Topology node constructor, with proper node index
-
-            // move the children to a new internal node
-            for (size_t i = 0; i < children.size(); i++)
+            // if we couldn't find all the the mrca, then this clade is not compatible
```
davidcerny: "all the the mrca" -> "all the mrca" (or "all the mrcas")

davidcerny (approved):
This makes perfect sense – thank you for tracking down the bug and clearly explaining the changes you made!

> I just realized this patch might be hard to review because it changes a lot of whitespace and other cleanups.

Yes, the diff is maybe a little more complicated than it needs to be because of changes that basically amount to style preferences (range-based loops, no braces, etc.), but I think that's okay.

src/core/analysis/mcmc/output/TreeSummary.cpp
bredelings: Done.

src/core/analysis/mcmc/output/TreeSummary.cpp
bredelings: Fixed.

Issue #858, opened 2025-10-03
[BUG] Skyline FBD analyses produce compiler-dependent output


davidcerny:
**Describe the bug**
Fossilized Birth-Death (FBD) analyses in general are wonky in RevBayes, since they produce different output on different platforms: it's why we [special-case](https://github.com/revbayes/revbayes/blob/cbe72b0372cf2c869d951293b383181bacdb9e99/.github/workflows/build.yml#L312) our test runner to ignore the FBD test output produced on macOS or Windows. However, it turns out that skyline FBD analyses – in which one or more of the rate parameters can vary through time in a piecewise-constant manner – are even more wonky, since their output differs between different compilers on the same platform.

Until recently, all of our FBD tests only specified constant-rate analyses, but in PR #845, a skyline FBD test was added. [GitHub Actions](https://github.com/revbayes/revbayes/actions/runs/18109603397/job/51553905219) show that this test produces consistent output under the following OS/compiler combos:

- Ubuntu 22.04, GCC 10
- Ubuntu 24.04, GCC 11 (no MPI)
- Ubuntu 22.04, GCC 12
- Ubuntu 24.04, GCC 13
- Ubuntu 24.04, GCC 15 (incl. with `buildtype: debug-optimized`)
- Ubuntu 24.04, Clang 15

The first few lines of the log file here looks as follows:

```
Iteration	Posterior	Likelihood	Prior	branch_rates_morpho	extinction_rate	origin_time	psi[1]	psi[2]	psi[3]	psi[4]	psi[5]	speciation_rate
0	-7477.243	-7240.708	-236.5342	0.04483802	0.0352606	3.150825	8.13918	6.77304	12.8996	6.34757	0.762531	0.07298359
5	-3153.135	-3085.028	-68.1064	0.1031445	0.7223114	2.051456	1.39008	0.867115	1.91905	0.307876	6.90674	0.9265235
10	-3148.35	-3083.157	-65.19248	0.1072652	0.7644451	2.033234	6.55731	0.134521	4.06573	1.05351	6.6125	1.13495
```

But under Ubuntu 24.04, GCC 14, we instead get the following:

```
Iteration	Posterior	Likelihood	Prior	branch_rates_morpho	extinction_rate	origin_time	psi[1]	psi[2]	psi[3]	psi[4]	psi[5]	speciation_rate
[...]

PR #861, opened 2025-10-04
Fixes for using the BDSTP with global timelines

bredelings:
Fixes for using the BDSTP with global timelines


davidcerny (approved):
Wow – this proved to be an amazingly simple fix for what I expected to be a really complicated issue. Thank you so much, Ben!

davidcerny:
(I could try adding a test before we merge, but because of #858, I think that's unlikely to work.)

davidcerny:
Yep, just checked, and it indeed doesn't work. Merging this as-is.

PR #867, opened 2025-10-10
Fix dnPhyloCTMC help page formatting

ms609:
This corrects the halp page syntax to fix some formatting issues - see
https://revbayes.github.io/documentation/dnPhyloCTMC.html

Other fixes would be useful - the markdown does not seem to be rendering as expected, and I'm not sure how the 'example' section should be typeset so as to be displayed as code.


davidcerny:
To get the example section to render properly, don't use backticks (` ``` `) but simply indent the code by 4 spaces. You can use (for example) the [`fnJC` help file](https://github.com/revbayes/revbayes/blob/master/help/md/fnJC.md?plain=1) as a template.

ms609:
> Examples indented.

help/md/dnPhyloCTMC.md
```diff
@@ -1,10 +1,12 @@
...
+dnPhyloCTMC
 
 ## title
-The parameters of a phylogenetic model – a tree topology with branch lengths, a substitution model that describes how observations evolve over the tree, etc. – collectively form a distribution called the _phylogenetic continuous-time Markov chain_.
+Distribution of a phylogenetic continuous-time Markov chain
 
 ## description
+The parameters of a phylogenetic model – a tree topology with branch lengths, a substitution model that describes how observations evolve over the tree, etc. – collectively form a distribution called the _phylogenetic continuous-time Markov chain_.
+
 dnPhyloCTMC gives the probability distribution of tip labels on a phylogenetic tree given an phylogenetic continuous-time Markov chain model.
```
davidcerny: While we're at it, we might want to change line 10 as well – I find it a little confusing. `dnPhyloCTMC` gives the probability distribution of _character state vectors at the tips_ of a phylogenetic tree. I guess you could argue that tips are labeled by their character state vectors, but normally the phrase "tip label" would refer to the taxon name. Also, there is a typo: "an phylogenetic" -> "a phylogenetic".

ms609: I agree – Updated per your suggested wording

davidcerny (approved):

PR #868, opened 2025-10-10
Include character simulation in dnPhyloCTMC documentation

ms609:
I'm working through how to simulate characters using a `dnPhyloCTMC` object.  I've updated the examples accordingly.

What I'm still missing is a means of getting these simulated characters out of RevBayes; `write(sim)` just provides the summary, and `write(sim.show())` fails with an error as `sim.show()` returns `NULL`.  I can't see any other `methods()` that look relevant.


davidcerny:
Can your problem be solved by something like `writeFasta(filename="my_alignment.fasta", data=sim)`, or am I misunderstanding what you're trying to do?

ms609:
Looks good, thanks; I've added `writeNexus` to the examples.

I wonder whether, akin to `x.methods()`, it would be useful to have a way of finding functions that take an argument of class `typeof(x)` – to make it easier for a user to locate relevant functions?  I didn't manage to find `writeFasta` this morning.

davidcerny (approved):
If you'd like to take a stab at implementing that, I certainly wouldn't object!

In the meantime, this looks good – merging.

PR #871, opened 2025-10-15
Miscellaneous documentation formatting fixes

### LINKED ISSUE #848: Missing documentation for srMinESS
It would be useful if the behaviour of `srMinESS(burninMethod = "SEM")` was documented on its manual page.

### LINKED ISSUE #849: [BUG] Encoding issue with Windows documentation
**Describe the bug**
Non-ASCII characters are mis-rendered on Windows (launching RB from the cmd command prompt)

**To Reproduce**
`help("srMinESS")`

**Screenshots**

Image

**Computer info**
Windows 11, RevBayes development

davidcerny:
This PR makes several changes to the documentation:

(1) Replace all non-ASCII characters by their closest ASCII equivalents. This mostly involved converting en dashes to hyphens and omitting accents (ö, Č, Á) from authors' names. It should prevent the rendering problems seen in Windows terminal emulators that do not support UTF-8 encoding (#849).

(2) Fix several other problems that caused the help files to be rendered incorrectly on the website, including "See also" entries that couldn't be parsed and code blocks delimited using triple backticks rather than indentation.

(3) As requested (see #848), I added a detailed description of how the `burninMethod` argument works in the convergence rules (`srGelmanRubin`, `srGeweke`, `srMinESS`, `srStationarity`).

(4) As I was going through the help files affected by these changes, I also made several other minor improvements, such as fixing typos and inconsistent spacing or capitalization, filling in missing citation info, unifying the citation format, and removing redundant content (e.g., re-iterating the parameter descriptions that are already printed out automatically). Almost all of these were purely cosmetic, with a few exceptions:

  - The help page for `dnDuplicationLoss` was just a duplicate of `dnMultiSpeciesCoalescent`, so I deleted it entirely;
  - The description of how the `max` argument works in `dnMultiSpeciesCoalescentUniformPrior` actually pertained to how the `Ne` argument works in `dnMultiSpeciesCoalescent`;
  - The example for `srGeweke` erroneously stated that the rule stops an MCMC run when the test statistic becomes significant at the specified level. It's the other way around; the run is terminated when the test statistic _ceases_ to be significant.
  
(5) In the README file, I changed the link to our user forum from the old Google Group to the new GitHub Discussions page.


EmiliaJarochowska:
> (5) In the README file, I changed the link to our user forum from the old Google Group to the new GitHub Discussions page.

That group is still linked on RB's main page

davidcerny:
I assume you mean https://revbayes.github.io by that? I can submit another PR in the website repo to change that as well.

EmiliaJarochowska:
> I assume you mean https://revbayes.github.io/ by that?

yes. I only learned from this PR that the old forum is obsolete(?). I filed https://github.com/revbayes/revbayes.github.io/issues/97

help/md/dnCDBDP.md
```diff
@@ -33,7 +33,7 @@ inferred tree (using `.clamp()`), and to a matrix recording which state is
 observed at a given tip (using `.clampCharData()`).
 
 2. Cladogenetic State-dependent Speciation Extinction (ClaSSE) (Goldberg &
-   Igić 2012)
+   Igic 2012)
```
bredelings: Are you changing `ć` to `c` to make it ASCII?

help/md/dnDuplicationLoss.md
ms609, line 1: Looks like a blank page is being recreated here – if this duplicates `dnMultiSpeciesCoalescent` should it redirect the user there?

help/md/mvBetaSimplex.md
```diff
@@ -3,14 +3,11 @@ mvBetaSimplex
 ## title
 Beta Simplex move
 ## description
-The Beta Simplex move selects one element of the a vector and proposes a new value for it drawn from a Beta distribution.
+The Beta Simplex move selects one element of the a vector and proposes a new value for it drawn from a Beta distribution. A usage example can be found at https://revbayes.github.io/tutorials/chromo/#root
```
ms609, line 6: 👍

help/md/dnCDBDP.md
davidcerny: Regrettably, yes. See (1) in the PR description.

bjoelle: I'm OK with minimizing the amount of non-ASCII in the main parts of the documentation, but I don't think we should do the same for people's names
I'm sure @davidcerny will agree that it gets pretty annoying when people with broken computer systems insist that your name is the problem

help/md/srGelmanRubin.md
```diff
@@ -1,52 +1,65 @@
...
+be calculated when two or more independent runs are performed, and its value
+tends to unity (1) as the runs converge.
+
+The number of samples to be removed as burnin before calculating the test
+statistic is determined using the `burninMethod`. Different burnin lengths are
+tested, increasing from 0 to 50% (for `ESS`) or 100% (for `SEM`) of the length
+of the trace in increments of 10 samples. If the `ESS` option is chosen
+(default), effective sample sizes (ESS) are calculated for all monitored
+parameters after removing the number of samples corresponding to each candidate
+burnin length. The best burnin length for a given parameter is the one that
```
ms609, line 21: I'm intrigued by how this balance works. 

In the case where an 80% burnin gives the lowest PSRF (but a satisfactory ESS), but a 20% burnin gives the highest ESS (and a PSRF above the user-specified threshold), it sounds like the PSRF stopping rule would not be triggered.

It would be helpful to provide an example that articulates how a user could stop a run as soon as it is possible to determine a burnin such that ESS > 100 and PSRF < 1.02.

help/md/srMinESS.md
```diff
@@ -1,55 +1,68 @@
 ## name
 srMinESS
 ## title
-Estimated sample size stopping rule
-
+Effective sample size stopping rule
 ## description
-Allow an MCMC run to terminate once the specified criterion has been met.
-
+Terminates an MCMC run when the effective sample sizes (ESS) of all parameters
```
ms609, line 6: 👍

help/md/srMinESS.md
```diff
@@ -1,55 +1,68 @@
...
+The MCMC analysis will terminate once all parameters in every run meet the ESS
+threshold. As such, performing additional runs will not decrease the number
+of generations required to meet the ESS threshold, even though it may increase
 the number of indepedent samples in the final, pooled posterior sample.
 
-## authors
-ESS explanation adapted from Luiza Fabreti and Sebastian Höhna's [tutorial](https://revbayes.github.io/tutorials/convergence/)
+The number of samples to be removed as burnin before calculating the test
+statistic is determined using the `burninMethod`. Different burnin lengths are
+tested, increasing from 0 to 50% (for `ESS`) or 100% (for `SEM`) of the length
```
ms609, line 21: It would be useful to clarify whether the analysis terminates when _all_ stopping rules are met, or when _any_ stopping rule is met.

If the former, is it possible that one SR is met with one burnin%, and another SR with a different burnin%, but there is no burnin% for which both rules are satisfied?

ms609 (commented):
Great - thanks for working this all through in such detail!

help/md/dnCDBDP.md
bredelings: Sorry I didn't see that in the description at first.

I don't know if we should do that.  It seems that for Windows 11 the default terminal for cmd.exe has changed to use "Windows Terminal", which _does_ support UTF-8.  And on Windows 10, Windows Terminal is installable.

https://learn.microsoft.com/en-us/windows/terminal/

UTF-8 is pretty standard these days, so pretty much everything that doesn't support it yet will support it soon.

It seems like part of the problem is that the original issue report didn't involve much investigation.

davidcerny: That's fair. I'll revert it. (And yes, having two accents in my last name can be a recipe for a world of pain...)

ms609: Here's a screen grab if I run rb.exe 1.3.1 from Win 11 using "Windows Terminal"
image

bredelings: > I'm sure @davidcerny will agree that it gets pretty annoying when people with broken computer systems insist that your name is the problem

Oh no. :roll_eyes:

help/md/dnDuplicationLoss.md
davidcerny: That seems to be a misunderstanding; sorry for the confusion. It's not that `dnDuplicationLoss` (the distribution) is a duplicate of `dnMultiSpeciesCoalescent`; it's that `dnDuplicationLoss.md` (the help file) is a duplicate of `dnMultiSpeciesCoalescent.md`. I.e., the help file was actually describing a different distribution than it should. Since I didn't feel qualified to write a new help file for `dnDuplicationLoss` from scratch (having never used it myself), I just made it blank.

help/md/dnCDBDP.md
bredelings: @ms609 Have you tried:
- `chcp 65001` 
- Installing a unicode font such as "Cascadia Mono"?
- How does the text look in RevStudio?

To be clear, my "investigation" here involved asking chatgpt what the problem was, so it might be insufficient.

@bjoelle You use windows, right?  Do you have a preferred terminal?

However, at this point I don't think we should remove UTF-8 just because there are some problems on windows.  I'm fairly sure its possible to work around them, and I guess the question (in my mind) is how much effort is required to do the workaround once we find a good one.

bredelings: As a side note, UTF-8 technically allows us to put emoji in the documentation, since emoji are made up of unicode code points :joy: :roll_eyes:. I have checked that its possible to make git branch names with emoji in them.

help/md/srGelmanRubin.md
davidcerny: There is no PSRF-based burnin estimation method, so this is a moot point. If a 20% burnin maximizes the ESS of the remaining sample, then that's the fraction that will be excluded before the ESS and PSRF values (both of them) are calculated, and the analysis will be stopped once both thresholds have been satisfied.

> It would be helpful to provide an example that articulates how a user could stop a run as soon as it is possible to determine a burnin such that ESS > 100 and PSRF < 1.02.

This is not currently possible, so no such example can be provided.

help/md/srMinESS.md
davidcerny: > It would be useful to clarify whether the analysis terminates when _all_ stopping rules are met, or when _any_ stopping rule is met.

Yes, which is why this information is provided in the help page for `mcmc`. In fact, [it was you](https://github.com/revbayes/revbayes/commit/7a250282fa9ce37877bf2b9177c3bc3a6185bc67) who put it there. 😛

> If the former, is it possible that one SR is met with one burnin%, and another SR with a different burnin%, but there is no burnin% for which both rules are satisfied?

No, that's not possible unless you deliberately specify different `burninMethod` values for different convergence rules (e.g., `ESS` for `srGelmanRubin` but `SEM` for `srMinESS`).

I think you are confused by the fact that the variable a given stopping rule calculates to determine the burnin fraction is generally not the same as the variable it calculates for the sample that remains once the burnin has been discarded. There is a single exception to this – `srMinESS(..., burninMethod="ESS")` – but the Gelman-Rubin rule, for example, cannot pick the burnin fraction that minimizes the Gelman-Rubin statistic for the remaining sample. It can only pick the fraction that maximizes the ESS or minimizes the SEM, because those are the only two options that the `burninMethod` argument accepts.

I thought the current help pages were clear about the choices you have when it comes to determining burnin, but let me know how I can rephrase them to avoid this confusion.

help/md/dnCDBDP.md
ms609: @bredelings My goal in opening the issue was to make RevBayes as friendly as possible to inexperienced Windows users; these workarounds sound more complex than we probably want to detail in any "instructions for new users" document, and indeed my skim of stack overflow suggests that `chcp 65001` can have negative side-effects.

As none of us have spotted a straightforward workaround (i.e. this is not an issue in how the doc pages are being built internally), it might be that the efficient solution is to avoid UTF-8 where it doesn't matter (e.g. dashes), and to work around it where a malformed UTF8 character may change the meaning (e.g. the summand symbol)?

davidcerny: I agree with @ms609 – I don't think it's reasonable to expect Windows users to mess around with their command line settings just to properly display an en dash.

bjoelle: @bredelings the terminal I use the most is the one that comes with the git install for windows - I believe that it has UTF8 support, but don't quote me on that

bredelings: On the other hand, I'm not sure its fair to remove functionality for linux and mac users because windows is broken.  

And it doesn't seem like we have actually established that yet.  Installing a font doesn't seem like an extreme hardship for a bug reporter?  Nor does running RevStudio to see if it has the same problem?  Also, are you confusing "changing the system locale to UTF-8" with just running "chcp 65001"?  Because it seems that the latter affects only the terminal that it is running in.

I do agree that some things might be too complex to recommend for end users.  But at this point, I'm just trying to understand what the problem is.  At some point I may find time to install one of the free windows VMs and investigate, but I don't have time today.  And it doesn't seem like we should remove the functionality before investigating.

bredelings (approved):

help/md/srMinESS.md
ms609: > information is provided in the help page for mcmc

I thought I remembered covering this before! I couldn't see it in the srXXX pages.  Perhaps a cross-reference would be useful.

> is it possible that one SR is met with one burnin%, and another SR with a different burnin%, but there is no burnin% for which both rules are satisfied?

OK, so I think this _is_ possible, but it wouldn't trigger the stopping rule. I had read "Different burnin lengths are tested" to mean that the test statistic is calculated for different burnin lengths, but in fact I think I now understand that the ESS (or SEM) is calculated for different burnin lengths; then the test statistic is calculated once, for the single burnin length that maximizes ESS.

Consequently I do not think that it is currently possible to stipulate "I want the analysis to terminate as soon as there is a subset of the posterior distribution that has a satisfactory test statistic, and an ESS of at least X".  This may be the case some time before the sample with the _maximal_ ESS passes the test statistic, meaning that the run continues beyond the minimum time necessary.

Have I understood this now, or am I still in a muddle?

davidcerny: > Perhaps a cross-reference would be useful.

I've added `mcmc` to the "See also" section of the help pages for individual rules. Other than that, I don't think the information on how different stopping rules interact belongs in the help page of any / every particular rule.

> the ESS (or SEM) is calculated for different burnin lengths; then the test statistic is calculated once, for the single burnin length that maximizes ESS.

That's exactly right. If you look in the code, you'll see that burnin determination is handled by a dedicated class (`BurninEstimatorContinuous`) that's separate from the stopping rules, and when the latter need to perform this step, they just call that class's member function `estimateBurnin()`.

> Consequently I do not think that it is currently possible to stipulate "I want the analysis to terminate as soon as there is a subset of the posterior distribution that has a satisfactory test statistic, and an ESS of at least X".

Yes, in theory it's possible to find a scenario where our code doesn't guarantee this outcome. Let's say we want PSRF < 1.01 and ESS > 500. Perhaps excluding a burnin of 1000 generations would give us an ESS of 550 (acceptable) and a PSRF of 1.009 (acceptable). Excluding a burnin of 900 generations would give us an ESS of 600 (better than before) and a PSRF of 1.011 (worse than before and not acceptable). So we will keep the analysis running.

For this to actually occur in practice, though, ESS and PSRF would need to be anticorrelated. That's difficult to imagine given that (according to [this preprint](https://arxiv.org/pdf/1812.09384), which @bredelings posted in Slack some time ago)

$$
\text{PSRF} \approx \sqrt{1 + \frac{m}{\text{ESS}}},
$$

where $m$ is the number of runs. So I don't think this is worth getting worried about.

Issue #870, opened 2025-10-14
[BUG] Validation with SBC: true values in simplex format are not printed out


prilau:
**Describe the bug**
When running a validation script, if a stochastic variable has a Dirichlet prior, the output of the true values in each replicate is a pointer, instead of a vector of values.

**To Reproduce**
inspect pi.out after running Validation_F81_uniform.Rev

**Computer info**
MacOS, development branch

davidcerny:
Fixed by #876 – thanks @kopperud!

PR #876, opened 2025-10-23
Fix for printing Simplex in validation

kopperud:
Suggested fix for #870


davidcerny (commented):
Great work, thanks so much again! My only comment is that it would be nice to add an integration test to make sure this problem doesn't resurface in the future.

The current integration test for validation analysis ([`Test_validation.Rev`](https://github.com/revbayes/revbayes/blob/master/tests/test_validation/scripts/Test_validation.Rev)) is based on the [`Validation_mk.Rev`](https://github.com/revbayes/revbayes/blob/master/validation/scripts/Validation_mk.Rev) script. If we were to base it on[`Validation_F81_uniform.Rev`](https://github.com/revbayes/revbayes/blob/master/validation/scripts/Validation_F81_uniform.Rev) instead, that would be one way to test that we can write out simplex variables to a file.

To convert an existing validation script into an integration test, only the following changes should be needed: decrease `NUM_VALIDATION_REPLICATES` (e.g., from 1000 to 10), decrease `NUM_MCMC_ITERATIONS` (e.g., from 10000 to 1000), and change the `directory` argument to `"output"`.

src/core/datatypes/math/Simplex.cpp
```diff
@@ -175,3 +175,17 @@ std::ostream& RevBayesCore::operator<<(std::ostream& o, const Simplex& x)
...
 
 }
+
+
+void Simplex::writeToFile(const path &dir, const std::string &fn) const
+{
+    path filename = dir / (fn + ".out");
+    create_directories(dir);
+
+    std::ofstream o( filename.string() );
```
bredelings: I think maybe we no longer need the `.string()` part, now that we are on C++17.

davidcerny (approved):

prilau (approved):

PR #880, opened 2025-10-31
Avoided in-place conversion between integer and real

### LINKED ISSUE #879: [BUG]Integer variable reused after type conversion produces corrupted value
**Describe the bug**
When an integer constant variable is reused during type conversion, it can become corrupted if it is later interpreted as a real type.

**To Reproduce**
```
> a<-1
> b:=a + 1
> c := a+2.0
> a
   1
> b
   4607182418800017409
> c
   3
```

**Additional context**
Maybe it is not a general using, but it does not warn if someone just fotget the 'point zero'.

curiosusJR:
suggest change can close #879

### DIFF
--- a/src/revlanguage/functions/argumentrules/ArgumentRule.cpp
+++ b/src/revlanguage/functions/argumentrules/ArgumentRule.cpp
@@ -6,6 +6,8 @@
 #include "Argument.h"
 #include "ArgumentRule.h"
 #include "RbException.h"
+#include "Integer.h"
+#include "Real.h"
 #include "RlFunction.h"
 #include "TypeSpec.h"
 #include "Workspace.h"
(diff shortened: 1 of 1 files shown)


curiosusJR:
test_coalescent/scripts/mcmc_isochronous_EBSP.Rev always seems to fail.

davidcerny:
@curiosusJR I wonder if closing this may have been premature? It looks like there was a point at which the tests passed: https://github.com/revbayes/revbayes/actions/runs/19032885827

curiosusJR:
> @curiosusJR I wonder if closing this may have been premature? It looks like there was a point at which the tests passed: https://github.com/revbayes/revbayes/actions/runs/19032885827 

@davidcerny Oh thank you for looking at this. At that time I passed the test but did not fix the issue. The discussions in the issue seems point to more complex considerations like performance, which I didn’t considered. I don’t know if the expect output of mcmc_isochronous_EBSP is effected due to this issue. When I add a warning in type replace, many test scripts print out the warning, including mcmc_isochronous_EBSP.Rev. However, let me try more and see if it could work.

curiosusJR:
It seems the `test_coalescent/scripts/mcmc_isochronous_EBSP.Rev` is the only failed test. I tried to add a warning when a conversion occurs between integer and real number and here is the output of this test: 
[warning.txt](https://github.com/user-attachments/files/23329560/warning.txt)

I tried to add 'point zero' manually on some values which looks like caused the warnings and the new test script is: 
[mcmc_isochronous_EBSP.txt](https://github.com/user-attachments/files/23329580/mcmc_isochronous_EBSP.txt)

I run the new test script with current version of revbayes(1.3.1) and get a different output compared with the 'output_expected' but same with the changed version of revbayes runs on the original test script, which implys adding 'point zero' on one or some of these values will change output because an unknown integer corruption has been fixed.

If that's the case, it seems like the expect of output maybe should be changed and then the test maybe pass.

PR #882, opened 2025-11-06
new functionality to write phylip formatted files

hoehna:
Write our alignments into phylip.


davidcerny (approved):
I've checked, and the function works for morphological ("Standard") data as well. Approving!

basantakhakurel (approved):

PR #883, opened 2025-11-06
Dev bds shift condition

kopperud:
This branch is a small fix for our (Sebastian and me) investigation into assessing whether rate shift events on extinct and/or unsampled lineages should be taken into account (in the birth-death-shift model).

Before the fix, the likelihoods were incorrect. We validate the fix by comparing the branch probability (D) on a single branch, using i) RevBayes and ii) RPesto ( https://github.com/kopperud/RPesto/blob/main/src/rust/src/branch_probability.rs#L56-L64 )

Figures attached (branch probability D before and after fix)

# Before fix (incorrect D)
rb_rpesto_compare_old

# After fix (correct D)
rb_rpesto_compare_new


davidcerny (approved):
Looks good to me.

PR #886, opened 2025-11-07
Remove null terminators from `.methods()` output

### LINKED ISSUE #885: [BUG] Embedded null strings in `.methods()` output
**Describe the bug**
RevBayes output contains `\0` null strings, which cannot be processed in R (using package "processx")

**To Reproduce**

```rb
g ~ dnGeom(1/2)
g.methods()

! embedded nul in string: 'MixtureLikelihoods = MixtureLikelihoods\0()\r\n'
```

**Expected behavior**
It's not clear that the `\0` is desirable; could this be replaced with a space?

**Screenshots**
If applicable, feel free to provide screenshots to help explain your problem.

**Computer info**
Windows, RB 1.4.0 (devel)

**Additional context**
The context is that I'm working on some R functions that allow me to launch and interact with a RevBayes session (which I intend to eventually become a standalone R package).  This works by reading and parsing output lines via "processx", but R does not support null strings, and processx uses c to convert the RB output into R format - there's no obvious workaround to read raw bytes from the terminal.

davidcerny:
Should close #885.

### DIFF
--- a/src/revlanguage/workspace/FunctionTable.cpp
+++ b/src/revlanguage/workspace/FunctionTable.cpp
@@ -689,34 +689,36 @@ void FunctionTable::printValue(std::ostream& o, bool env) const
         i->second->printValue( s, true );
         
         // nice printing with arguments aligned
-        size_t offset = i->first.size() + 3;
-        std::string temp = "";
-        std::string repl = "";
-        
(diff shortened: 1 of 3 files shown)


ms609 (approved):
Looks promising, thanks for the quick fix!

PR #892, opened 2025-11-11
Update macOS runner images

davidcerny:
The macOS 13 runner we are currently using is [deprecated](https://github.com/actions/runner-images/issues/13046), and will be fully phased out in less than a month. This PR replaces the corresponding image label in our workflows (both "build" and "release") with `macos-15-intel`, which is the last macOS runner for the x86_64 architecture and which will continue to be supported [until August 2027](https://github.com/actions/runner-images/issues/13045). Since macOS 15 (Sequoia) doesn't support Xcode 15 without workarounds, I also bumped up the corresponding Xcode version from 15 to 16.

For the ARM64 build, no upgrade is currently necessary, but since it's probably a good idea to keep it up-to-date anyway, I changed the image label from `macos-14` to `macos-latest`, and specified Xcode 26 instead of Xcode 16.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,8 +21,8 @@ jobs:
                ubuntu-gcc-15,
                ubuntu-gcc-15-debugoptimized,
                ubuntu-clang-15,
-               macos-xcode-15-intel,
-               macos-xcode-16-arm,
+               macos-xcode-16-intel,
+               macos-xcode-26-arm,
                windows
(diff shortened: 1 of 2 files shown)


davidcerny:
If this PR is approved, it will have to be merged by someone with maintainer status, because it doesn't pass the checks it is intended to replace. @bredelings?

bredelings (approved):

Issue #896, opened 2025-11-17
[BUG] Get Error when calling dnPhyloBrownian()


curiosusJR:
**Describe the bug**
`dnPhyloBrownian` always return error. eg:
```
tree <- readTrees(text="((A:1,B:1):1,C:2);")[1]
traits ~ dnPhyloBrownian(tree=tree, sigma=0.5)
traits
```

Get return: 
```
> tree <- readTrees(text="((A:1,B:1):1,C:2);")[1]
> traits ~ dnPhyloBrownian(tree=tree, sigma=0.5)
   Error:	Vector index out of range. You tried to access index '4' for a vector of size '0'.
```

PR #897, opened 2025-11-17
fix dnPhyloBrownian

curiosusJR:
A suggestion solution for #896


davidcerny (approved):
After consulting with @hoehna, this looks good – thank you!

PR #900, opened 2025-11-20
Record vectors of stochastic variables when `stochasticOnly = TRUE`

### LINKED ISSUE #899: [BUG] `mnModel(..., stochasticOnly = TRUE)` ignores vectors
**Describe the bug**
I found this while preparing the integration test added in PR #895 – if you specify `stochasticOnly=TRUE` in an `mnModel` monitor, this will remove not just any deterministic variables from the resulting trace files, but also those stochastic variables that are vectors.

**To reproduce**
Take the [`bears_FBD_ex3.Rev`](https://github.com/revbayes/revbayes/blob/master/tests/test_FBD/scripts/FBD_dating_ex3.Rev) integration test and add `, stochasticOnly=TRUE` to line 167. Before making this change, the script produces a [log file](https://github.com/revbayes/revbayes/blob/master/tests/test_FBD/output_expected/bears_FBD_ex3.log) with the following header, consisting of 56 columns:

```
Iteration	Posterior	Likelihood	Prior	age_ursinae	branch_rates[1]	branch_rates[2]	branch_rates[3]	branch_rates[4]	branch_rates[5]	branch_rates[6]	branch_rates[7]	branch_rates[8]	branch_rates[9]	branch_rates[10]	branch_rates[11]	branch_rates[12]	branch_rates[13]	branch_rates[14]	branch_rates[15]	branch_rates[16]	branch_rates[17]	branch_rates[18]	branch_rates[19]	branch_rates[20]	branch_rates[21]	branch_rates[22]	branch_rates[23]	branch_rates[24]	branch_rates[25]	branch_rates[26]	branch_rates[27]	branch_rates[28]	branch_rates[29]	branch_rates[30]	branch_rates[31]	branch_rates[32]	branch_rates[33]	branch_rates[34]	branch_rates[35]	branch_rates[36]	branch_rates[37]	branch_rates[38]	branch_rates[39]	branch_rates[40]	branch_rates[41]	branch_rates[42]	branch_rates_mean	diversification	extant_mrca	extinction_rate	num_samp_anc	origin_time	psi	speciation_rate	turnover
```

Re-running the script after making the change will instead produce a log file with just 9 columns:

```
Iteration	Posterior	Likelihood	Prior	branch_rates_mean	extinction_rate	origin_time	psi	speciation_rate
```

The fact that we no longer see variables like `diversification`, `num_samp_anc`, or `turnover` is good, because these are deterministic. However, the 42 `branch_rates` were stochastic variables, so we should still be recording them in the output file.

**Expected behavior**
Vectors of stochastic variables should not be suppressed by `stochasticOnly=TRUE`.

**Computer info**
macOS 26.0.1 (Tahoe), ARM64 architecture, RevBayes 1.4.0-preview, commit db0bf01.

### LINKED ISSUE #911: [BUG] Missing header when restarting an analysis without a pre-existing log file
**Describe the bug**
When we restart an analysis from a checkpoint, we almost always want to append to a pre-existing log file. It therefore makes sense not to print the monitor header (containing the iteration, posterior, etc. columns, as well as the names of the monitored parameters) before we start logging again – the header is presumably already there at the beginning of the file, and we don't want to have another one in the middle.

However, for testing or debugging purposes, it may be desirable to "restart" an analysis for which we have a checkpoint file but not the previous log files. For example, this might be the case when:

1. It is the ability to resume the analysis that is of interest, rather than what came before it;
2. A problem occurs at a late point in the analysis. Usually, we provide the random seed to let others reproduce it, but that may require them to re-run the whole analysis (potentially for many hours) on their own. Providing a checkpoint file instead might be a convenient alternative.

In fact, we have a test that exemplifies these two situations: [`mvFossilTimeSlideUniform_checkpointing.Rev`](https://github.com/revbayes/revbayes/blob/development/tests/test_FBD/scripts/mvFossilTimeSlideUniform_checkpointing.Rev). However, as can be seen from its expected output, the log files generated by the analysis that was initialized from the provided checkpoint are missing a header.

https://github.com/revbayes/revbayes/blob/04a1b09b5460206d86b6ec307cab0bce422226ad/tests/test_FBD/output_expected/mvFossilTimeSlideUniform_checkpointing.log#L1-L5

**To reproduce**
Re-run the `mvFossilTimeSlideUniform_checkpointing.Rev` test from the development branch.

**Expected behavior**
Ideally, we'd want to keep the current behavior (i.e., not printing the monitor header) when the log files to which the monitors are supposed to write already exist, but suppress it (and print the header) when they do not.

**Computer info**
macOS 26.1 (Tahoe), ARM64 architecture, RevBayes 1.4.0-preview, current tip of development (04a1b09).

davidcerny:
Fixes #899.


bjoelle (commented):
I don't understand why this works as a fix - the `!the_node->isElementVariable()` check is also present when `stochastic_nodes_only` is false, so it should not be the cause of the different behaviour between `stochasticOnly` options

davidcerny:
Well, it quickly turned out that it doesn't really "work" – it does make the issue go away, but at the cost of introducing a new one. I'll need to think about the logic of that conditional more carefully.

It's weird that stochastic vector elements are getting captured at all (even when `stochasticOnly = FALSE`), since they don't satisfy either of the two conditions!

davidcerny:
So it turns out that when `ModelMonitor::resetDagNodes()` iterates over the model graph, it encounters each vector twice: first as a single node, and then as a set of nodes corresponding to individual vector elements. Since we don't want to include each vector element twice, we keep the former and suppress the latter by requiring `!the_node->isElementVariable()`. However, that's where issue #899 comes from: since a vector of stochastic nodes is itself a deterministic node, we can't extract these nodes to record their values.

This branch goes about it the opposite way: it suppresses the vector-node and iterates over the nodes corresponding to its elements. This change causes multiple test failures, but it seems to me that in most (though not all) cases, it's the new behavior that's correct and the current behavior that's wrong. Here is a detailed description of the differences:

`test_basics/ignore.Rev`

- The `y[1][1]`, `y[1][2]`, `z[1][1][1]`, `z[1][1][2]`, `z[1][2][1]`, `z[1][2][2]` nodes are included in the new output but not the expected output. That's a bug on this branch – I need to do a better job of identifying which nodes are clamped.

`test_Coal_skyline/mcmc_Coal_skyline_fixed.Rev`

- The current (i.e., expected) output only shows 6 significant digits for vector elements, whereas the new output shows 7, which is the default `outputPrecision`. It makes sense for the same precision to be applied to both scalars and vector elements, so I'd argue the new output is correct here.

`test_coalescent/mcmc_heterochronous_GMRF_maptreebased.Rev`

- Again, 7 sig. figs instead of 6. The script produces 3 log files; this accounts for discrepancies in all 3.

`test_coalescent/mcmc_isochronous_BSP.Rev`

- 7 sig. figs instead of 6. The script produces 3 log files; this accounts for discrepancies in all 3.

`test_coalescent/mcmc_isochronous_EBSP.Rev`

- The new output doesn't include the `indicator[1]` column. However, the script shows that while the remaining elements of the `indicator` vector are deterministic variables, `indicator[1]` is constant. So the new behavior is correct. The script produces 2 log files; this accounts for discrepancies in both.

`test_coalescent/mcmc_isochronous_GMRF_maptreebased.Rev`

- 7 sig. figs instead of 6. The script produces 3 log files; this accounts for discrepancies in all 3.

`test_coalescent/mcmc_isochronous_GMRF_treebased.Rev`

- 7 sig. figs instead of 6. The script produces 3 log files; this accounts for discrepancies in all 3.

`test_coalescent/mcmc_isochronous_GMRF.Rev`

- 7 sig. figs instead of 6. The script produces 3 log files; this accounts for discrepancies in all 3.

`test_coalescent/mcmc_isochronous_HSMRF.Rev`

- 7 sig. figs instead of 6. The script produces 3 log files; this accounts for discrepancies in all 3.

`test_coalescent/mcmc_isochronous_piecewise_3const.Rev`

- 7 sig. figs instead of 6. The script produces 3 log files; this accounts for discrepancies in all 3.

`test_coalescent/mcmc_isochronous_piecewise_3lin.Rev`

- 7 sig. figs instead of 6. The script produces 3 log files; this accounts for discrepancies in all 3.

`test_coalescent/mcmc_isochronous_piecewise_6diff.Rev`

- 7 sig. figs instead of 6. The script produces 3 log files; this accounts for discrepancies in all 3.

`test_coalescent/mcmc_isochronous_Skygrid.Rev`

- 7 sig. figs instead of 6. The script produces 3 log files; this accounts for discrepancies in all 3.

`test_coalescent/mcmc_isochronous_Skyline.Rev`

- 7 sig. figs instead of 6. The script produces 3 log files; this accounts for discrepancies in all 3.

`test_coalescent/mcmc_isochronous_Skyride.Rev`

- 7 sig. figs instead of 6. The script produces 3 log files; this accounts for discrepancies in all 3.

`test_conversions/UniformInteger.Rev`

- The `y[1]`, ..., `y[100]` nodes are included in the expected output but not the new output. This seems correct to me – these nodes are all clamped, so recording their values is redundant.

`test_EBD/mcmc_EBD.Rev`

- 7 sig. figs instead of 6.

`test_EBD/mcmc_EBD_empirical.Rev`

- 7 sig. figs instead of 6.

`test_HSMRF_BDP/mcmc_HSMRF_BDP.Rev`

- 7 sig. figs instead of 6.

`test_large_model_graph/mcmc_large_normal_model.Rev`

- 7 sig. figs instead of 6.

`test_mixtures/mcmc_mixture.rev`

- The `pi[1][1]`, `pi[1][2]`, ..., `pi[2][4]` nodes are included in the new output but not the expected output. The new output is clearly correct here: the individual elements of `pi` are stochastic nodes in the form of vectors drawn from `dnDirichlet()`. It looks like the current system can handle vectors of i.i.d. stochastic variables (that is, unless we set `stochasticOnly = TRUE`), but not vectors which themselves represent a single draw from a distribution.

`test_mixtures/mcmc_mixture2.rev`

- As above.

`test_Mk/mcmc_Mk.rev`

- 7 sig. figs instead of 6.

`test_Mk/mcmc_Mki.rev`

- 7 sig. figs instead of 6.

`test_Mk/mcmc_Mkv.rev`

- 7 sig. figs instead of 6.

`test_negCladeConst/negativeCladeConstraint.Rev`

- 7 sig. figs instead of 6.

`test_Partition/mcmc_Partition_codon.Rev`

- The `er[1][1]`, `er[1][2]`, ..., `er[4][6]` and `gamma_rates[1][1]`, `gamma_rates[1][2]`, ..., `gamma_rates[4][4]` nodes are included in the new output but not the expected output. This represents another case of vectors that have been created "wholesale" (rather than element by element) being invisible to the current system. The elements of `er` again represent draws from `dnDirichlet()` (i.e., each one is a stochastic node), and the elements of `gamma_rates` were obtained using `fnDiscretizeGamma()` (i.e., each one is a deterministic node). The script produces 2 log files; this accounts for discrepancies in both.

`test_Partition/mcmc_Partition_gene.Rev` 

- The `er[1][1]`, `er[1][2]`, ..., `er[2][6]` and `gamma_rates[1][1]`, `gamma_rates[1][2]`, ..., `gamma_rates[2][4]` nodes are included in the new output but not the expected output. The reason for this is the same as above. The script produces 2 log files; this accounts for discrepancies in both.

`test_Partition/mcmc_Partition_uniform.Rev`

- The `er[1]`, ..., `er[6]` and `gamma_rates[1]`, ..., `gamma_rates[4]` nodes are included in the new output but not the expected output. The reason for this is the same as above. The script produces 2 log files; this accounts for discrepancies in both.

`test_skygrid/mcmc_skygrid.Rev`

- 7 sig. figs instead of 6.

`test_steppingstone/ML_partition_uniform.Rev`

- 7 sig. figs instead of 6. The script produces 10 log files; this accounts for discrepancies in all 10.

`test_steppingstone/ML_partition_uniform_one_stone_at_a_time.Rev`

- 7 sig. figs instead of 6. The script produces 10 log files; this accounts for discrepancies in all 10.

`test_stoch_map/mcmc_corr.Rev`

- This is a very interesting case. There are 24 nodes (called `rates[1][1]`, `rates[1][2]`, ..., `rates[4][4]`) that are included in the new output but not the expected output, but only 16 have unique names – the remaining 8 are duplicates. The reason for this becomes clear once we look at the original script, where we first create a vector of 16 constant nodes, and then replace 8 of its elements by deterministic nodes. This is a bug on this branch. We should be recording just those 8 deterministic nodes.

`test_UCLD_noncentered/mcmc_UCLD_noncentered.rev`

- 7 sig. figs instead of 6.

I'm going to go ahead and change the expected output of those tests for which I believe the new results to be the correct ones.

davidcerny:
There are a few more failing tests that were either recently added to development, or whose failures were previously masked by the fact that we disabled them for macOS:

`test_BDSTP/BDSTP.Rev`

- The `F[1]`, ..., `F[14]` nodes are included in the expected output but not the new output. This is correct – these nodes are all clamped, so recording their values is redundant. (In fact, they don't even represent any real, interpretable quantities – they are just an artefact of the tortured way in which we currently handle tip age uncertainty.)

`test_coalescent/mcmc_heterochronous_BSP.Rev`

- 7 sig. figs instead of 6 in the `pop_size` columns. The script produces 3 log files; this accounts for discrepancies in all 3.

`test_coalescent/mcmc_heterochronous_GMRF_treebased.Rev`

- 7 sig. figs instead of 6 in the `delta_log_population_size` columns. The script produces 3 log files; this accounts for discrepancies in all 3.

`test_FBD/FBD_dating_ex3.Rev`

- 7 sig. figs instead of 6.

`test_FBD/FBD_one_parameter_skyline.Rev`

- 7 sig. figs instead of 6 in the `psi` columns.

`test_FBD/FBDP_fossil_tip_age_uncertainty.Rev`

- The `F[1]`, ..., `F[14]` nodes are included in the expected output but not the new output. See above.

`test_FBD/FBDP.Rev`

- The `F[1]`, ..., `F[14]` nodes are included in the expected output but not the new output. See above.

`test_FBD/mvFossilTimeSlideUniform_checkpointing.Rev`

- The exact discrepancy here was obscured by issue #911, but when the latter is solved, it's clear the difference is the same as in other BDSTP / FBD tests (see above): the `F[1]`, ..., `F[14]` nodes are included in the expected output but not the new output.

`test_FBD/mvFossilTimeSlideUniform_multiple_runs.Rev`

- The `F[1]`, ..., `F[43]` nodes are included in the expected output but not the new output. See above. The script produces 2 log files; this accounts for discrepancies in both.

`test_MkError/mcmc_MkError.rev`

- 7 sig. figs instead of 6 in the `br_lens` columns.

`test_Partition/mcmc_Partition_codon.Rev`

- Two more differences became apparent when examining the JSON output files. First, instead of recording the branch length vector as follows:

  ```json
  "bl":[0.04668089246107779,0.08518248961710673,0.24508170576439664,...],
  ```
  
  we record it as:
  
  ```json
  "bl[1]":0.04668089246107779,"bl[2]":0.08518248961710673,"bl[3]":0.24508170576439664,...
  ```
  
  This is different, but IMO neither obviously better nor worse.
  
  Second, we also record the nested `er` and `gamma_rates` vectors in the JSON files generated as part of the new output but not of the expected output. This difference was already pointed out for the standard (i.e., tab-separated rather than JSON) output in the comment above. The new output is clearly correct here, since these are stochastic nodes that our current monitoring system ignores for no good reason. Note how these nested vectors are written to JSON:
  
  ```json
  "gamma_rates[1]":[0.8959733745739022,0.9709424352126861,1.0250762799985929,1.1080079102148188],"gamma_rates[2]":[0.8977148201454996,0.9714712971734394,1.0246822805217213,1.1061316021593397],...
  ```
  
  The script produces 2 JSON files; this accounts for discrepancies in both.

`test_proposed_bad_index/model_average_primates_cytb.Rev`

- 7 sig. figs instead of 6.

`test_proposed_bad_index/model_average_primates_cytb_debug.Rev`

- 7 sig. figs instead of 6.

davidcerny:
This is ready for review, I think. Two final notes:

(1) Performance-wise, the new solution is mostly indistinguishable from the old, but there are exceptions. The [large model graph test](https://github.com/revbayes/revbayes/blob/master/tests/test_large_model_graph/scripts/mcmc_large_normal_model.Rev), which uses `mnModel` to sample a vector of 15,000 stochastic variables 100 times (1 sample every 10 MCMC generations, for a total of 1,000 generations), currently takes about 23 seconds to run on my machine, but on this branch the runtime goes up to about 53 seconds.

(2) When I stated in a previous comment that this branch

> suppresses the vector-node and iterates over the nodes corresponding to its elements

this wasn't completely accurate. An interesting edge case is provided by the [MC3 heat tuning and checkpointing test](https://github.com/revbayes/revbayes/blob/master/tests/test_MC3/scripts/Test_heat_tuning_and_checkpointing.Rev), which contains the following Rev code:

https://github.com/revbayes/revbayes/blob/cbe72b0372cf2c869d951293b383181bacdb9e99/tests/test_MC3/scripts/Test_heat_tuning_and_checkpointing.Rev#L56-L58

Now, if we had _just_ `root_freqs` and no `rtfq1`, the simplistic solution would work just fine. Here, `root_freqs` is a single stochastic node, rather than a vector of two such nodes:

Screenshot 2025-12-02 at 23 33 24
</br></br>

The notion of suppressing a vector in favor of its elements therefore doesn't apply. The outcome is still as desired – we get two columns in the resulting log file, called `root_freqs[1]` and `root_freqs[2]` – but it was the `Simplex` object itself that handled its own "unfolding" into two elements. It wasn't unfolded by a walk across the model graph, in which `root_freqs` appears as an atom.

If we defined deterministic nodes for _both_ elements of `root_freqs` – i.e., `rtfq1 := root_freqs[1]; rtfq2 := root_freqs[2]` – we wouldn't have an issue either. In doing so, we have quietly created two deterministic nodes called `root_freqs[1]` and `root_freqs[2]`:

Screenshot 2025-12-02 at 23 43 38
</br></br>

We correctly suppress `root_freqs` in favor of these two, and the outcome is once again what we'd expect: we have four columns in the resulting log file, labeled `root_freqs[1]`, `root_freqs[2]`, `rtfq1`, and `rtfq2`.

It's only the code from the original test script that poses an issue:

Screenshot 2025-12-02 at 23 50 40
</br></br>

We have implicitly created a deterministic node called `root_freqs[1]`, which causes us to suppress `root_freqs` – but then we are left with no information about `root_freqs[2]`, and no such column will therefore appear in the output file.

This PR solves the problem by first checking if the number of element-nodes of a given vector-node corresponds to that vector's size. If it does, then it suppress the vector in favor of the elements, as stated in my previous comment. If it doesn't, it instead suppresses the elements in favor of the vector, as we currently do on development.

Is this edge case weird and artificial? Yes. Both before and after this PR we would still record `root_freqs[1]` even if we didn't define a deterministic node for it, so the extra node adds absolutely nothing – just a duplicated column in the log file. But it's probably better to handle edge cases correctly even if they are weird and artificial.

src/core/monitors/ModelMonitor.cpp
```diff
@@ -56,35 +74,101 @@ void ModelMonitor::resetDagNodes( void )
...
+                    // remove base name so vector itself can be monitored
+                    vector_base_names.erase(base_name);
                 }
-                
             }
+        }
         
+        // Step 3: add variables, including vector elements but not the vectors themselves (unless we made an exception
+        // in the previous step)
+        for (DagNode* the_node : n)
```
bjoelle, line 145: I'm wondering if part of the performance hit is because we are running this loop several times
it seems to me we could do this only once, building the list of nodes that should be added as we go and then just adding those
also it seems that the clamped & hidden check could be done way earlier since we don't care about these nodes

src/core/monitors/Monitor.cpp
```diff
@@ -363,7 +363,7 @@ void Monitor::setMcmc(Mcmc *m)
 /**
  * Sort the nodes by name so that the order is guaranteed of replicated runs.
  */
-void Monitor::sortNodesByName( void )
+void Monitor::sortNodesByName( bool natural )
```
bjoelle, line 368: could be good to mention in a comment what is a "natural" sort as opposed to the default ?

bjoelle (changes requested):
overall this looks good to me
I think the algo can be reorganized to minimize the performance hit however, see detailed comments

src/core/monitors/ModelMonitor.cpp
davidcerny: @bjoelle In this case, I don't think we can merge the loops. We do need two passes: the first one to collect all vector-nodes and element-nodes in the graph, and the second one to add those we have determined should be added. We can't add nodes as we collect them, because we first need to run Step 2 to decide if we should suppress the vector-node in favor of its element-nodes, or _vice versa_.

In any case, this doesn't really matter. I've actually experimented with lots of ways to make the new code more performant: move some of the checks upfront and add early exits/`continue`s for them, like you suggested; use unordered sets and unordered maps for faster lookup; add non-vector and non-element nodes in Step 1 rather than Step 3; or even cache node names instead of repeatedly calling `the_node->getName()`. None of it made any difference at all, because `ModelMonitor::resetDagNodes()` is not where the bottleneck is. It only gets called twice per analysis, and takes only a few milliseconds each time. `VariableMonitor::monitorVariables()` has a much bigger effect, but still doesn't seem to be the main culprit.

src/core/monitors/Monitor.cpp
davidcerny: done

src/core/monitors/ModelMonitor.cpp
bjoelle: ok, fair enough

bjoelle (approved):

PR #901, opened 2025-11-21
Observation Error Model

basantakhakurel:
This PR implements a morphological observation error model.
It introduces a "soft" likelihood initialization at the tips, parameterized by an error probability (`observation_error_probability`). This allows the model to distinguish between the True State and the Observed Score.

Upon using the `using_observation_error` flag, the tip likeihoods are intialized as follows.

For an observed state $o$ and a potential true state $j$:

$$
P(X_{obs}=o | X_{true}=j) = \begin{cases} (1-\epsilon) + \epsilon\pi_o & \text{if } j = o \\ 
\epsilon\pi_o & \text{if } j \neq o 
\end{cases}
$$

An integration test `test_MkError` has also been added.


davidcerny (changes requested):
I definitely intend to take a more detailed look at this later on, but for now I'd just suggest adding a sentence or two to the [`dnPhyloCTMC` help page](https://github.com/revbayes/revbayes/blob/master/help/md/dnPhyloCTMC.md) about how and why one might want to use the new arguments. The argument descriptions added to `Dist_phyloCTMC.cpp` are rather terse, so they don't really give the full picture on their own.

basantakhakurel:
Changes added.

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
```diff
@@ -3162,6 +3193,69 @@ void RevBayesCore::AbstractPhyloCTMCSiteHomogeneous<charType>::setMcmcMode(bool
...
+    // set the value
+    using_observation_error = r != NULL;
+    observation_error_probability = r;
+
+    // add the new parameter
+    this->addParameter( observation_error_probability );
+
+    if ( observation_error_frequencies == NULL )
+    {
+        Simplex* tmp_val = new Simplex( this->num_chars );
```
davidcerny, line 3216: Does this mean you need as many frequencies as there are characters in the alignment?

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
```diff
@@ -3162,6 +3193,69 @@ void RevBayesCore::AbstractPhyloCTMCSiteHomogeneous<charType>::setMcmcMode(bool
...
+
+    // remove the old parameter first
+    if ( observation_error_frequencies != NULL )
+    {
+        this->removeParameter( observation_error_frequencies );
+        observation_error_frequencies = NULL;
+    }
+
+    // set the value
+    observation_error_frequencies = r;
```
davidcerny, line 3256: Do you need a check on whether the simplex has the right number of elements?

src/core/distributions/phylogenetics/substitution/PhyloCTMCSiteHomogeneous.h
```diff
@@ -464,10 +495,31 @@ void RevBayesCore::PhyloCTMCSiteHomogeneous<charType>::computeTipLikelihood(cons
...
+                                else
+                                {
+                                    tmp += tp_begin[c1*this->num_chars+c2] * (obs_error_prob*obs_error_freqs[org_val]);
+                                }
+                            }
+                            p_site_mixture[c1] = tmp;
+                        }
                         // store the likelihood
-                        p_site_mixture[c1] = tp_begin[c1*this->num_chars+org_val];
+                        // p_site_mixture[c1] = tp_begin[c1*this->num_chars+org_val];
```
davidcerny: Why do you need to keep this as a comment, instead of just removing it entirely?

src/core/distributions/phylogenetics/substitution/PhyloCTMCSiteHomogeneous.h
```diff
@@ -464,10 +495,31 @@ void RevBayesCore::PhyloCTMCSiteHomogeneous<charType>::computeTipLikelihood(cons
...
                     else // no ambiguous characters in use
                     {
-                        std::uint64_t org_val = char_node[site];
+                      std::uint64_t org_val = char_node[site];
 
+                      if (this->using_observation_error == false)
+                      {
+                        // store the likelihood
+                        p_site_mixture[c1] = tp_begin[c1 * this->num_chars + org_val];
+                      }
```
davidcerny, line 504: Messed up indentation

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
```diff
@@ -3040,6 +3052,25 @@ void RevBayesCore::AbstractPhyloCTMCSiteHomogeneous<charType>::simulate( const T
...
+                        // there was an error
+                        const Simplex& error_probs = this->observation_error_frequencies->getValue();
+                        double u3 = rng->uniform01();
+                        size_t new_state = 0;
+                        while ( error_probs[new_state] < u3 )
+                        {
+                            u3 -= error_probs[new_state];
+                            ++new_state;
+                        }
+                        c.setStateByIndex(new_state);
```
davidcerny, line 3071: So here we are picking the true state given the observed state?

help/md/dnPhyloCTMC.md
```diff
@@ -13,42 +13,47 @@ a substitution model that describes how observations evolve over the tree, etc.
...
+distribution to a `AbstractHomologousDiscreteCharacterData` object) is computed
 using Felsenstein's pruning algorithm, with partial likelihoods stored for each
 branch of the tree. It is automatically outputted in the `Likelihood` column of
 the `mnFile()` and `mnScreen()` monitors (which can be suppressed with
 `likelihood = FALSE`).
+Optionally, an observation error probability can be specified
+to account for scoring ambiguity (e.g., in morphological data), which treats observed
+states as probabilistic classifications rather than absolute ground truth.
+This can be specified using `observationErrorProbability =` and `observationErrorFrequencies =`
+arguments for the observation error probabilitiy and the frequencies respectively.
```
davidcerny: My problem with this explanation is that it doesn't really tell me what these variables _mean_. Is it the case that `observationErrorProbability` = $\epsilon$ and `observationErrorFrequencies` = $\pi_o$ from the equation in your PR description?

Naïvely, I would assume that `observationErrorProbability` is just the probability that the observed state differs from the true state, but your test script says it's the probability the character was scored randomly, and those do not mean the same thing. Even if the character _was_ scored randomly, there is a nonzero probability that the scored / observed state is equal to the true state, which you give in your equation. I'd suggest considering whether something along the lines of `randomScoringProbability` wouldn't be a less misleading name for that argument.

Also, what are the `observationErrorFrequencies`? The phrase "observational error frequencies of the states" is not very clear. It almost suggests that different states can have different probabilities of being mistaken for another state, which is almost certainly not what you were going for. (Such probabilities wouldn't have to sum up to 1, for example.) So are they the unobservable true state frequencies? But those would presumably be either fixed by the substitution model (e.g., JC69) or estimated, no? Or are they the observed state frequencies? That's what the notation $\pi_o$ would suggest, but in that case, it's still not clear to me why the user would need to specify them – can't they just be trivially calculated from the alignment?

In any case, one basic piece of information I'd expect the help page to give me is the expected length of the `observationErrorFrequencies` simplex. Equal to the number of character states, presumably?

davidcerny (changes requested):

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
basantakhakurel: The size of the observation frequencies must be matching the state space. You can either have frequencies for each character or separated by state space.

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
basantakhakurel: Great point. I have added a check.

src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
basantakhakurel: I believe here would be the opposite as it is simulation part. 
We're taking the true state (which just finished evolving) and potentially corrupting it to create the observed state.

davidcerny: Ah, makes sense!

help/md/dnPhyloCTMC.md
basantakhakurel: Thanks for the feedback. 
You're right about the distinction between the probability of the process (random scoring) and the probability of an outcome (state mismatch). I've made following clarifications:
- `observationErrorProbability` is the probability of the uninformative process occuring 
- `observationErrorFrequencies` represents the distribution used for those uninformative scores.
I think the idea was to also allow for users to have the frequencies (possibly calculated by scoring multiple specimens or scoring the same specimen by multiple people) so it is not calculated from the data itself. 
- The simplex length must strictly match the number of character states ($k$).

davidcerny (approved):
Excellent, this is much clearer – thank you!

PR #693, opened 2025-03-14
DRAFT Stochmap tp dirty merge

bredelings:
Preparing to merge this soon.

PR #904, opened 2025-11-27
state-dependent Ornstein-Uhlenbeck model

prilau:
- new distribution: dnOrnsteinUhlenbeckStateDependent
  modeling adaptation of a univariate continuous trait to different discrete states
- new test for dnOUSD

- new distribution: dnBrownianProcessStateDependent
  modeling evolution of a univariate continuous trait to different discrete states; collapsed model of dnOUSD (when alpha = 0 for all states)

- new datatype: CharacterHistory
- new RevLanguage utility function: readCharacterHistory
  read simmap into CharacterHistory

- modified distribution: dnPhyloCTMCDASiteIID
  - root branch will not be imputed if root branch is absent in the phylogeny
  - fixed likelihood calculation of the character history when root branch is absent
  - if root branch is absent and argument rootFrequencies is NULL, compute analytically the stationary frequencies at root
  - if root branch is absent and argument rootFrequencies is provided, use the provided distribution as root frequencies (allows root frequencies to be fixed or estimated)

- modified move: mvCharacterHistory
  - added root rejection sampler. Same as node rejection sampler, but only samples the root
  - added tip rejection sampler. Used when some/all discrete states at the tips are unobserved


.github/workflows/build.yml
```diff
@@ -6,28 +6,29 @@ on:
       - "*"
   pull_request:
     branches:
-    - "*"
```
bredelings, line 9: I think this branch should not be  changing build.yml from development?

src/core/datatypes/phylogenetics/characterhistory/BranchHistory.h
```diff
@@ -48,9 +48,9 @@ class CharacterEvent;
         void                                                            print(const TopologyNode* nd=NULL) const;
 
         // setters
-        void                                                            setParentCharacters(const std::set<CharacterEvent*>& s);
+//        void                                                            setParentCharacters(const std::set<CharacterEvent*>& s);
```
bredelings: Should we just delete this?

src/core/datatypes/phylogenetics/characterhistory/BranchHistory.h
```diff
@@ -48,9 +48,9 @@ class CharacterEvent;
...
         void                                                            setParentCharacters(const std::vector<CharacterEvent*>& s);
-        void                                                            setChildCharacters(const std::set<CharacterEvent*>& s);
+//        void                                                            setChildCharacters(const std::set<CharacterEvent*>& s);
```
bredelings: Can we just delete this?

src/core/datatypes/phylogenetics/characterhistory/BranchHistory.cpp
```diff
@@ -289,15 +301,15 @@ void BranchHistory::setChildCharacters(const std::vector<CharacterEvent*>& s)
...
-void BranchHistory::setChildCharacters(const std::set<CharacterEvent*>& s)
-{
-    std::set<CharacterEvent*>::iterator it;
-    for (it = s.begin(); it != s.end(); it++)
-    {
-        child_characters[ (*it)->getSiteIndex() ] = *it;//new CharacterEvent(**it);
-    }
-
-}
+//void BranchHistory::setChildCharacters(const std::set<CharacterEvent*>& s)
```
bredelings: Should we just delete this?

src/core/datatypes/phylogenetics/characterhistory/CharacterEvent.cpp
```diff
@@ -5,23 +5,20 @@
 using namespace RevBayesCore;
 
 
-CharacterEvent::CharacterEvent(void)
+CharacterEvent::CharacterEvent(void) :
+    site_index(0),
```
bredelings: Can you move this to the header, like `int site_index = 0;`?
And the same with age, event_type, and missing.

src/core/datatypes/phylogenetics/characterhistory/CharacterEventDiscrete.cpp
```diff
@@ -13,13 +13,8 @@ CharacterEventDiscrete::CharacterEventDiscrete(void) : CharacterEvent()
...
 CharacterEventDiscrete::CharacterEventDiscrete(size_t i, size_t s, double a, size_t t) : CharacterEvent(i,a,t),
-    state(s)
-{
-
-}
-
-CharacterEventDiscrete::CharacterEventDiscrete(const CharacterEventDiscrete& c) : CharacterEvent(c),
-    state(c.state)
+    state(s),
+    ambiguous_state()
```
bredelings: I don't think you need to specifically call the default constructor.

src/core/datatypes/phylogenetics/ratematrix/RateGenerator.cpp
```diff
@@ -62,7 +62,8 @@ double RateGenerator::getSumOfRates(std::vector<CharacterEvent*> from, const std
     double sum = 0.0;
     for (size_t i = 0; i < num_states; ++i)
     {
-        sum += -getRate(i, i, age, rate) * counts[i];
+        sum += -getRate(i, i, age, rate) * counts[i]; // This needs checking!!!
```
bredelings: Does this still need checking?

bredelings (dismissed):
Hi, can you please edit the description of the PR to give an overview of the code  changes that you are proposing?

src/core/moves/characterhistory/RootRejectionSampleProposal.h
```diff
@@ -0,0 +1,954 @@
+#ifndef RootRejectionSampleProposal_H
+#define RootRejectionSampleProposal_H
```
davidcerny, line 2: I may be missing something, but it looks like this is a newly added move that has no corresponding `revlanguage` file or entry in the move register – i.e., no Rev interface, and therefore no way to actually use it.

davidcerny (changes requested):
It looks like on the user-facing / Rev side of things, this PR adds 5 new elements (and there should arguably be a 6th one for the newly added root rejection-sampling move; see above). Specifically, we have:

3 new distributions:

- `dnPhyloBrownianProcessStateDependent`
- `dnPhyloCharacterEvent`
- `dnPhyloOrnsteinUhlenbeckStateDependent`

1 new function:

- `readCharacterHistory()`

and 1 new type:

- `CharacterHistory`

I would like to see these documented with dedicated help files.

For the distribution help pages, you can draw inspiration from [`dnCDBDP.md`](https://github.com/revbayes/revbayes/blob/development/help/md/dnCDBDP.md), [`dnCoalescent.md`](https://github.com/revbayes/revbayes/blob/development/help/md/dnCoalescent.md), or [`dnPhyloCTMC.md`](https://github.com/revbayes/revbayes/blob/development/help/md/dnPhyloCTMC.md). For an I/O function help page, you can use [`readTreeTrace.md`](https://github.com/revbayes/revbayes/blob/development/help/md/readTreeTrace.md) as a template. And [`Trace.md`](https://github.com/revbayes/revbayes/blob/development/help/md/Trace.md) or [`TraceTree.md`](https://github.com/revbayes/revbayes/blob/development/help/md/TraceTree.md) can give you a sense of what a help page might look like for a type.

src/core/moves/characterhistory/RootRejectionSampleProposal.h
prilau: the root rejection sampler is part of `mvCharacterHistory`, and can be called with `mvCharacterHistory(..., graph="root",   proposal="rejection")` in Rev

davidcerny: Ah, I see – thank you for the clarification!

In that case, it would be good to expand the help page for `mvCharacterHistory` to clarify what these arguments actually do and what effect all the different options have. Especially because the move is missing even the basic descriptions for these two arguments (i.e., those stored in [`Move_CharacterHistory::getParameterRules()`](https://github.com/revbayes/revbayes/blob/cbe72b0372cf2c869d951293b383181bacdb9e99/src/revlanguage/moves/Move_CharacterHistory.cpp#L314)).

davidcerny:
It looks like this PR includes 4 files that are commented out in their entirety:

- [src/core/distributions/phylogenetics/characterhistory/PhyloCharacterEventDistribution.cpp](https://github.com/revbayes/revbayes/pull/904#diff-fad1ed1ec222ef1cb29007bc7aa4d96d98b33147a383f68d43d02724e72140c6)
- [src/core/distributions/phylogenetics/characterhistory/PhyloCharacterEventDistribution.h](https://github.com/revbayes/revbayes/pull/904#diff-62f41688aafab069f0b1cf78ed42c2d74240c041d155d12ee2c18c43acb41bb2)
- [src/revlanguage/distributions/phylogenetics/Dist_PhyloCharacterEvent.cpp](https://github.com/revbayes/revbayes/pull/904#diff-6a5ea96ca1b86a43e4a02133592a6b2777e844a4e6d4fe93a594d448f99c7269)
- [src/revlanguage/distributions/phylogenetics/Dist_PhyloCharacterEvent.h](https://github.com/revbayes/revbayes/pull/904#diff-dac7b8787733db6c8fdc8dc5a3a5d8835bfca70c423bbd42dab351e56ea6da8a)

If these represent work in progress, it would be better to put them on a separate branch and delete them from this one.

src/core/moves/characterhistory/RootRejectionSampleProposal.h
prilau: I edited the help file, but only for the use case related to trait evolution models (dnPhyloOUSD, dnPhyloBMSD, and the MuSSCRat model). Specifically, the moves for cladogenic events are used in biogeographic reconstruction (and perhaps other stuff that I'm not aware of), but are not applicable in trait evolution models. The uniformization sampler at internal nodes and along branches, as far as I know, is deprecated. (I am hesistant to remove it since I was not involved in its development)

help/md/dnPhyloBrownianProcessStateDependent.md
```diff
@@ -0,0 +1,53 @@
+## name
+dnPhyloBrownianProcessStateDependent (dnPhyloBMSD)
```
davidcerny: This file needs to be renamed to `dnPhyloBrownianProcessStateDependent.md`. Currently, its contents don't properly combine with the help page that's autogenerated for `dnPhyloBMSD` from `getParameterRules()`.

help/md/dnPhyloBrownianStateDependent.md
```diff
@@ -0,0 +1,53 @@
+## name
+dnPhyloBrownianProcessStateDependent (dnPhyloBMSD)
+## title
+Phylogenetic state-dependent Brownian process
+## description
+Univariate Brownian process over an augmented phylogeny (i.e., discrete character history)
+## details
+The phylogenetic state-dependent Brownian process is collapsed from the phylogenetic state-dependent Ornstein-Uhlenbeck process, in which `alpha` is always 0. The probability is computed using a pruing algorithm. Specifically, `dnPhyloBMSD` uses an augmented tree structure, i.e., the character history with different character states (also referred to as "regimes"; Hansen 1997). Shifts in the character state is discrete and does not occur concurrently with speciation events, and can be represented by nodes of degree 2. As such, each branch in the augmented tree structure takes exactly one character state. Under this process, branches of the same character state share the same OU parameters (sigma). These character states can correspond to the states of an observed discrete character.
```
davidcerny: Instead of saying that branches governed by the same state "share the same OU parameters", we can maybe just say they share the same rate...

help/md/dnPhyloBrownianStateDependent.md
```diff
@@ -0,0 +1,53 @@
...
+    Q <- fnJC(num_disc_states)
+    X ~ dnPhyloCTMCDASiteIID(tree, Q, branchRates=1, type="Standard")
+    
+    # set state-dependent OU parameters
+    for (i in 1:num_states){
+      sigma2[i] ~ dnLognormal(1, 0.587405)
+    }
+    
+    # basic use of the function
+    Y ~ dnPhyloOUSD(char_hist, sigma=sigma2^0.5)
```
davidcerny: We should use `dnPhyloBMSD` here, not `dnPhyloOUSD`

help/md/dnPhyloCTMCDASiteIID.md
```diff
@@ -1,9 +1,42 @@
 ## name
 dnPhyloCTMCDASiteIID
 ## title
+CTMC Distribution with sampled history and independent and identically distributed sites
 ## description
+The probability distribution of the character state vectors at the tips of a phylogenetic tree, given a phylogenetic continuous-time Markov chain model where each site evolves independently.
 ## details
+The parameters of a phylogenetic model -- a tree topology with branch lengths, a substitution model that describes how observations evolve over the tree, etc. -- collectively form a distribution called the _phylogenetic continuous-time
+Markov chain_. In this distribution, the branch history is sampled (through data augmentation), instead of integrated out (as in `dnPhyloCTMCDASiteIID`). An instance of character history can therefore be retrieved from the distribution (using `.characterHistories`) The probability of observed character state vectors (specified via clamping the
```
davidcerny: Here you are contrasting `dnPhyloCTMCDASiteIID` with... itself? Did you mean to write "(as in `dnPhyloCTMC`)"?

help/md/mvCharacterHistory.md
```diff
@@ -1,9 +1,34 @@
...
 ## see_also
 ## example
+The following example shows the set-up of character history moves when the CTMC distribution is used together with a trait evolution model (e.g, `dnPhyloOUSD`).
+
+X ~ dnPhyloCTMCDASiteIID(tree, Q, branchRates=1, type="Standard", nSites=1)
+
+moves.append( mvCharacterHistory(ctmc=X, qmap_site=Q, graph="root",   proposal="rejection", weight=10.0) )
+moves.append( mvCharacterHistory(ctmc=X, qmap_site=Q, graph="node",   proposal="rejection", weight=10.0) )
+moves.append( mvCharacterHistory(ctmc=X, qmap_site=Q, graph="branch", proposal="rejection", weight=10.0) )
+moves.append( mvCharacterHistory(ctmc=X, qmap_site=Q, graph="tip",    proposal="rejection", weight=10.0) )
```
davidcerny: For this section to render properly on the website, the introductory sentence ("The following example...") should be a comment, and the whole thing should be indented by four spaces.

help/md/dnPhyloCTMCDASiteIID.md
```diff
@@ -1,9 +1,42 @@
 ## name
 dnPhyloCTMCDASiteIID
 ## title
+CTMC Distribution with sampled history and independent and identically distributed sites
 ## description
+The probability distribution of the character state vectors at the tips of a phylogenetic tree, given a phylogenetic continuous-time Markov chain model where each site evolves independently.
```
davidcerny: I wonder if it's worth explaining somewhere the rather tortured name of this distribution... if I understand correctly, "CTMC" = continuous-time Markov chain, "DA" = data augmentation, "IID" = "independent and identically distributed.

davidcerny (changes requested):
I'm super impressed by the help files you added! They are very detailed and do a great job of explaining the new functionality, in both practical and theoretical terms. I have a few minor comments, but the files are very nearly there, I think.

src/core/datatypes/phylogenetics/characterhistory/BranchHistory.h
```diff
@@ -60,12 +58,11 @@ class CharacterEvent;
...
         void                                                            updateHistory(const std::multiset<CharacterEvent*,CharacterEventCompare>& updateSet);
 
-        
+
 
     protected:
         //BranchHistory(void);
         BranchHistory(size_t nc, size_t idx);
-        BranchHistory(size_t nc, size_t idx, std::set<int> sc);
         BranchHistory(const BranchHistory& m);
```
davidcerny, line 66: I notice that you keep

```
BranchHistory(const BranchHistory& m);
```

as a protected member here (and the corresponding overloaded `=` operator), but you remove

```
BranchHistoryContinuous(const BranchHistoryContinuous& m);
```

from `BranchHistoryContinuous.h`

and

```
BranchHistoryDiscrete(const BranchHistoryDiscrete& m);
```

from `BranchHistoryDiscrete.h`... is this intentional?

src/core/datatypes/trees/TopologyNode.cpp
```diff
@@ -526,6 +528,71 @@ std::string TopologyNode::buildNewickString( bool simmap = false, bool round = t
...
+        const std::vector<CharacterEvent*>& parent_chars = bh.getParentCharacters();
+        o << parent_chars[0]->getStateStr() << "," << (total_branch_length - used_branch_length) << "}";
+
+    }
+
+    // 5. Write ";" if we're done with the tree.
+    if ( isRoot() )
+    {
+        // FIXME - move to caller?
+        o << ";";
```
davidcerny: Still in need of fixing?

src/core/distributions/phylogenetics/characterhistory/TreeHistoryCtmc.h
```diff
@@ -514,27 +533,28 @@ template<class charType>
...
     // NOTE: we may remove this is if we want to require users to supply a root branch length
     // if the root node is not assigned a branch length, assign a branch length
-    if (root_branch_length == 0.0) {
-        root_branch_length = tau->getValue().getRoot().getAge() * rootBranchLengthMultiplier;
-    }
-    
+    // edit (07.17.2025): root branch imputation disabled
+    // if (root_branch_length == 0.0) {
+    //     root_branch_length = tau->getValue().getRoot().getAge() * rootBranchLengthMultiplier;
+    // }
```
davidcerny: Why was it disabled? Do we now require users to supply a root branch length, as stated above? If so, do we provide a clear error message if a tree with no root branch is supplied?

src/core/distributions/phylogenetics/continuousCharacter/PhyloBrownianProcessStateDependent.cpp
```diff
@@ -0,0 +1,864 @@
...
+    // we need to reset the contrasts
+    resetValue();
+}
+
+
+/**
+ * Destructor. Because we added ourselves as a reference to tau when we added a listener to its
+ * TreeChangeEventHandler, we need to remove ourselves as a reference and possibly delete tau
+ * when we die. All other parameters are handled by others.
+ */
```
davidcerny, line 70: Is this description still accurate, given that the body of the function is commented out?

src/core/distributions/phylogenetics/continuousCharacter/PhyloBrownianProcessStateDependent.cpp
```diff
@@ -0,0 +1,864 @@
...
+    size_t site_index = 0;
+    for (size_t i = 0; i < this->num_sites; ++i)
+    {
+        while ( this->value->isCharacterExcluded(site_index) )
+        {
+            ++site_index;
+            if ( site_index >= this->value->getNumberOfCharacters()  )
+            {
+                throw RbException( "The character matrix cannot set to this variable because it does not have enough included characters." );
+            }
```
davidcerny, line 449: I'm not sure I understand what this error message is saying

src/core/distributions/phylogenetics/continuousCharacter/PhyloBrownianProcessStateDependent.cpp
```diff
@@ -0,0 +1,864 @@
...
+    
+    // remove the old parameter first
+    this->removeParameter( homogeneous_sigma );
+    this->removeParameter( state_dependent_sigma );
+    homogeneous_sigma       = NULL;
+    state_dependent_sigma   = NULL;
+    
+    
+    // set the value
+    state_dependent_sigma = s;
```
davidcerny, line 626: Do you need to check if `s` is of the right size?

src/core/distributions/phylogenetics/continuousCharacter/PhyloBrownianProcessStateDependent.h
```diff
@@ -0,0 +1,91 @@
...
+    class PhyloBrownianProcessStateDependent : public TypedDistribution< ContinuousCharacterData > {
+        
+    public:
+        // Note, we need the size of the alignment in the constructor to correctly simulate an initial state
+        PhyloBrownianProcessStateDependent(const TypedDagNode<CharacterHistoryDiscrete> *bh, size_t n_sites );
+        virtual                                                            ~PhyloBrownianProcessStateDependent(void);                                                              //!< Virtual destructor
+        
+        // public member functions
+        // pure virtual
+        virtual PhyloBrownianProcessStateDependent*                         clone(void) const;                                                                      //!< Create an independent clone
```
davidcerny: My C++ is not the best, but wouldn't you need a `= 0` here to make it a _pure_ virtual function?

src/core/distributions/phylogenetics/continuousCharacter/PhyloOrnsteinUhlenbeckStateDependent.cpp
```diff
@@ -0,0 +1,1099 @@
...
+    // we need to reset the contrasts
+    resetValue();
+}
+
+
+/**
+ * Destructor. Because we added ourselves as a reference to tau when we added a listener to its
+ * TreeChangeEventHandler, we need to remove ourselves as a reference and possibly delete tau
+ * when we die. All other parameters are handled by others.
+ */
```
davidcerny, line 75: Same thing as with the BM counterpart. It looks like this description fits the way the destructor functioned at one point, but not the way it functions now.

src/core/distributions/phylogenetics/continuousCharacter/PhyloBrownianProcessStateDependent.cpp
```diff
@@ -0,0 +1,864 @@
...
+        const TopologyNode &child = *(*it);
+        
+        // get the branch length for this child
+        double branch_length = child.getBranchLength();
+        
+        // get the branch specific rate
+        double branch_time = computeBranchTime( child.getIndex(), branch_length );
+        
+        //@TODO: Need to change the simulation to actually use the states from the character history
+
```
davidcerny: So what does the simulation currently do? Is it worth documenting this behavior somewhere, at least as a code comment?

src/core/distributions/phylogenetics/continuousCharacter/PhyloOrnsteinUhlenbeckStateDependent.cpp
```diff
@@ -0,0 +1,1099 @@
...
+
+
+            // Do the merging rule
+            // calculate and store the node variance
+            double var_node = (var_left*var_right) / (var_left+var_right);
+            this->variances[this->active_likelihood[node_index]][node_index] = var_node;
+
+
+            // this loop is broken currently, does not work for multiple characters
+            for (int i=0; i<this->num_sites; i++)
```
davidcerny: Why not? Is it because it doesn't properly handle missing data, unlike the corresponding loop in `PhyloBrownianProcessStateDependent::recursiveComputeLnProbability()`?

To be honest, I'm not sure I understand what "multiple characters" / `num_sites` refers to here. The continuous traits, or the discrete characters that control their evolution? Is the distribution supposed to be able to handle more than one of each? Are there any differences in this regard between `dnPhyloBMSD` and `dnPhyloOUSD`?

src/core/distributions/phylogenetics/continuousCharacter/PhyloOrnsteinUhlenbeckStateDependent.cpp
```diff
@@ -0,0 +1,1099 @@
...
+
+    // remove the old parameter first
+    this->removeParameter( homogeneous_alpha );
+    this->removeParameter( state_dependent_alpha );
+    homogeneous_alpha       = NULL;
+    state_dependent_alpha   = NULL;
+
+
+    // set the value
+    state_dependent_alpha   = a;
```
davidcerny, line 603: As with BM: do you need to check that `a` has the right number of elements?

src/core/distributions/phylogenetics/continuousCharacter/PhyloOrnsteinUhlenbeckStateDependent.cpp
```diff
@@ -0,0 +1,1099 @@
...
+
+    // remove the old parameter first
+    this->removeParameter( homogeneous_sigma );
+    this->removeParameter( state_dependent_sigma );
+    homogeneous_sigma       = NULL;
+    state_dependent_sigma   = NULL;
+
+
+    // set the value
+    state_dependent_sigma = s;
```
davidcerny, line 697: Also here

src/core/distributions/phylogenetics/continuousCharacter/PhyloOrnsteinUhlenbeckStateDependent.cpp
```diff
@@ -0,0 +1,1099 @@
...
+
+    // remove the old parameter first
+    this->removeParameter( homogeneous_theta );
+    this->removeParameter( state_dependent_theta );
+    homogeneous_theta        = NULL;
+    state_dependent_theta    = NULL;
+
+
+    // set the value
+    state_dependent_theta = t;
```
davidcerny, line 752: And here

src/core/distributions/phylogenetics/continuousCharacter/PhyloOrnsteinUhlenbeckStateDependent.cpp
```diff
@@ -0,0 +1,1099 @@
...
+            double y = RbStatistics::Normal::rv(theta, sqrt(stationary_variance), *rng);
+            chars[i] = y;
+        }
+        else if (root_treatment == PARAMETER)
+        {
+            chars[i] = computeRootValue();
+        }
+        else
+        {
+            throw RbException( "Cannot simulate under a state-dependent Ornstein-Uhlenbeck process because the root treatment is not set correctly." );
```
davidcerny: Is it worth reminding the user in the error message what the valid options are?

src/core/distributions/phylogenetics/continuousCharacter/PhyloOrnsteinUhlenbeckStateDependent.h
```diff
@@ -0,0 +1,104 @@
...
+
+    public:
+        enum                                                                ROOT_TREATMENT { OPTIMUM, EQUILIBRIUM, PARAMETER };
+        // Note, we need the size of the alignment in the constructor to correctly simulate an initial state
+        PhyloOrnsteinUhlenbeckStateDependent(const TypedDagNode<CharacterHistoryDiscrete> *bh, size_t n_sites, ROOT_TREATMENT rt);
+        virtual                                                            ~PhyloOrnsteinUhlenbeckStateDependent(void);                                                              //!< Virtual destructor
+
+        // public member functions
+        // pure virtual
+        virtual PhyloOrnsteinUhlenbeckStateDependent*                       clone(void) const;                                                                      //!< Create an independent clone
```
davidcerny: As with BM

src/core/io/NewickConverter.cpp
```diff
@@ -404,32 +512,235 @@ TopologyNode* NewickConverter::createNode(const std::string &n, std::vector<Topo
...
+//            {
+//                ss.ignore();
+//            }
+//        }
+//    }
+//    else
+//    {
+//        nodes.push_back( node );
+//        brlens.push_back( 0.0 );
+//    }
```
davidcerny: Can we just delete this? What is the current treatment of branch lengths in the SIMMAP format? Do we allow them but ignore them?

src/core/monitors/StochasticCharacterMappingMonitor.h
```diff
@@ -177,24 +175,24 @@ void StochasticCharacterMappingMonitor<characterType>::monitorVariables(std::uin
...
+        // sse_process->drawStochasticCharacterMap( character_histories );
         bool set_amb_char_data = false; // this is the default value for the arg
         sse_process->drawStochasticCharacterMap( character_histories, set_amb_char_data, use_simmap_default );
     }
     else
     {
-        
-        
-    	glhbdsp_process->drawStochasticCharacterMap( character_histories, use_simmap_default );
+    	// glhbdsp_process->drawStochasticCharacterMap( character_histories );
```
davidcerny: These calls were removed in PR #693 and replaced by the current ones. I don't think there's any need to put them back as comments. This file should ideally stay untouched by the PR.

src/core/moves/characterhistory/RootRejectionSampleProposal.h
```diff
@@ -0,0 +1,977 @@
...
+namespace RevBayesCore {
+
+    /**
+     * The scaling operator.
+     *
+     * A scaling proposal draws a random uniform number u ~ unif (-0.5,0.5)
+     * and scales the current vale by a scaling factor
+     * sf = exp( lambda * u )
+     * where lambda is the tuning parameter of the Proposal to influence the size of the proposals.
+     *
```
davidcerny: Copied over; should be removed

src/core/moves/characterhistory/RootRejectionSampleProposal.h
```diff
@@ -0,0 +1,977 @@
...
+}
+
+/**
+ * Perform the Proposal.
+ *
+ * A scaling Proposal draws a random uniform number u ~ unif (-0.5,0.5)
+ * and scales the current vale by a scaling factor
+ * sf = exp( lambda * u )
+ * where lambda is the tuning parameter of the Proposal to influence the size of the proposals.
+ *
```
davidcerny: See above

src/core/moves/characterhistory/TipRejectionSampleProposal.h
```diff
@@ -1,414 +1,591 @@
...
+}
+
+/**
+ * Perform the Proposal.
+ *
+ * A scaling Proposal draws a random uniform number u ~ unif (-0.5,0.5)
+ * and scales the current vale by a scaling factor
+ * sf = exp( lambda * u )
+ * where lambda is the tuning parameter of the Proposal to influence the size of the proposals.
+ *
```
davidcerny: See above

src/revlanguage/distributions/phylogenetics/branchrate/Dist_PhyloOrnsteinUhlenbeckStateDependent.cpp
```diff
@@ -0,0 +1,309 @@
...
+    RevBayesCore::PhyloOrnsteinUhlenbeckStateDependent::ROOT_TREATMENT rtr;
+    if (rt == "optimum"){
+        rtr = RevBayesCore::PhyloOrnsteinUhlenbeckStateDependent::ROOT_TREATMENT::OPTIMUM;
+    }else if (rt == "equilibrium"){
+        rtr = RevBayesCore::PhyloOrnsteinUhlenbeckStateDependent::ROOT_TREATMENT::EQUILIBRIUM;
+    }else if (rt == "parameter"){
+        rtr = RevBayesCore::PhyloOrnsteinUhlenbeckStateDependent::ROOT_TREATMENT::PARAMETER;
+   }else{
+        throw RbException("argument rootTreatment must be one of \"optimum\", \"equilibrium\" or \"parameter\"");
+    }
```
davidcerny: If we perform this check here, then we perhaps don't need it in `PhyloOrnsteinUhlenbeckStateDependent.cpp`...

src/revlanguage/distributions/phylogenetics/branchrate/Dist_PhyloOrnsteinUhlenbeckStateDependent.cpp
```diff
@@ -0,0 +1,309 @@
...
+        rv = static_cast<const Real &>( root_value->getRevObject() ).getDagNode();
+        dist->setRootValue( rv );
+    }
+
+    // @TODO: Need some way to check if rootTreatment = "parameter" is specified,
+    // then the user must also specify the random variable for the ancestral value
+    // and if not, then throw an error
+
+    // @TODO: Likewise, don't allow specifying "optimum" or "equilibrium" and at
+    // the same time supplying a parameter for the ancestral value.
```
davidcerny: It sounds like these cases shouldn't be too hard to handle. Worth adding?

src/revlanguage/distributions/phylogenetics/character/Dist_phyloCTMCDASiteIID.h
```diff
@@ -38,8 +38,8 @@ class TypeSpec;
         Dist_phyloCTMCDASiteIID*                        clone(void) const;                                                                      //!< Clone the object
         static const std::string&                       getClassType(void);                                                                     //!< Get Rev type
         static const TypeSpec&                          getClassTypeSpec(void);                                                                 //!< Get class type spec
-        virtual MethodTable                             getDistributionMethods( void ) const;                                                                       //!< Get the member methods
         std::string                                     getDistributionFunctionName(void) const;                                                //!< Get the Rev-name for this distribution.
+        virtual MethodTable                             getDistributionMethods( void ) const;                                                                       //!< Get the member methods
```
davidcerny: You could also align the comment while alphabetizing :P

src/revlanguage/moves/Move_CharacterHistory.cpp
```diff
@@ -339,6 +582,8 @@ const MemberRules& RevLanguage::Move_CharacterHistory::getParameterRules(void) c
         optionsGraph.push_back( "branch" );
         optionsGraph.push_back( "cladogenetic" );
         optionsGraph.push_back( "cladogenetic2" );
+        optionsGraph.push_back( "tip" );
+        optionsGraph.push_back( "root" );
```
davidcerny, line 577: GitHub doesn't let me properly highlight lines in this file, but there are a few changes I'd like to see:

- Lines 571--578 were clearly just copied over here by someone reminding themselves of how `OptionRule` works. Let's delete them.
- The `graph` and `proposal` arguments lack argument descriptions: see the empty strings (`""`) on lines 587, 593.
- No need to duplicate the "Inherit weight..." comment, so let's delete lines 595--596.

tests/test_OUsd/scripts/mcmc_OUsd_stochasticHistory.Rev
```diff
@@ -0,0 +1,129 @@
+################################################################################
+#
+# RevBayes Validation Test: Brownian-Motion on a fixed tree
+#
```
davidcerny, line 4: This is an integration test, not a validation test, and it tests sdOU rather than BM

tests/test_OUsd/scripts/mcmc_OUsd_stochasticHistory.Rev
```diff
@@ -0,0 +1,129 @@
...
+# authors: Priscilla Lau
+#
+################################################################################
+
+seed(12345)
+
+#######################
+# Reading in the Data #
+#######################
```
davidcerny, line 16: Is it worth also testing the `readCharacterHistory()` function, either in this script or in a separate one?

davidcerny (changes requested):
Apologies for submitting my review in several rounds, @prilau! I also took the liberty of pushing a commit fixing a few typos in the help files.

Aside from the individual comments below, I also noticed that the following files:

- [src/core/analysis/mcmc/output/JointAncestralStateTrace.cpp](https://github.com/revbayes/revbayes/pull/904/files#diff-9abd3a9d87eb3a6e597ae6d86c9534322f06642a1c96214f6707881a5e54377b)
- [src/core/datatypes/phylogenetics/ratematrix/RateGenerator.cpp](https://github.com/revbayes/revbayes/pull/904/files#diff-593f3a31a8fab37faa2f58c2c990fa1bc5a20c7525c8590119431652ce403265)
- [src/core/moves/MetropolisHastingsMove.cpp](https://github.com/revbayes/revbayes/pull/904/files#diff-c1d9bc2f3a20ab7e005202b2f5762aa6283a58423d35af26a2c7016c20f8b3b8)
- [src/revlanguage/moves/Move_CharacterHistory.h](https://github.com/revbayes/revbayes/pull/904/files#diff-4827ee8bff0abd76e11318d5286158ee73c0f5efd4067bed68d7f06a6011b65a)

only differ from their respective versions on development by whitespace. The git diff would probably be cleaner if this PR left them untouched.

src/core/distributions/phylogenetics/characterhistory/TreeHistoryCtmc.h
prilau: The root branch imputation was used when the user wanted to calculate the stationary frequencies at the root _numerically_, but did not/forgot to supply a root branch themselves.
In this implementation, (if rootFrequencies is null,) the numerical calculation will only be triggered if a root branch is supplied by the user; otherwise, the stationary frequencies will be computed _analytically_.
Furthermore, if the user provides root frequencies, then stationarity at the root will not be assumed, and the likelihood calculation will be based on the provided root frequencies.

src/core/distributions/phylogenetics/continuousCharacter/PhyloOrnsteinUhlenbeckStateDependent.cpp
prilau: The code can handle one discrete character (with multiple character states) only. `num_sites` probably comes from the molecular world where there are numerous sites (character) per sequence.
The loop is 'broken' in a sense that (1) it exists even though the current implementation is intended to use for one character (`num_sites`=1) only, and (2) supplying multiple discrete characters may lead to unexpected behavior.

src/core/distributions/phylogenetics/continuousCharacter/PhyloBrownianProcessStateDependent.h
prilau: @hoehna

src/core/io/NewickConverter.cpp
prilau: @hoehna

src/core/distributions/phylogenetics/continuousCharacter/PhyloOrnsteinUhlenbeckStateDependent.cpp
```diff
@@ -0,0 +1,1099 @@
...
+    return sum_partial_probs;
+}
+
+
+void PhyloOrnsteinUhlenbeckStateDependent::touchSpecialization( const DagNode* affecter, bool touchAll )
+{
+    const Tree& tau = character_histories->getValue().getTree();
+
+    // if the topology wasn't the culprit for the touch, then we just flag everything as dirty
+    if ( affecter == this->state_dependent_sigma && false )
```
davidcerny: If you include `&& false` in the condition, the body of the conditional will never be executed. Presumably that's not what you want?

src/core/distributions/phylogenetics/characterhistory/TreeHistoryCtmc.h
davidcerny: Thank you! This explanation is super useful, and makes perfect sense. It would be great to have this behavior documented somewhere – if not in a help file, then at least as a comment somewhere in the code.

src/core/distributions/phylogenetics/continuousCharacter/PhyloOrnsteinUhlenbeckStateDependent.cpp
davidcerny: Hmm... okay. If the code can only handle one discrete character, then it makes no sense to have a `num_sites` member variable, and it would be better to get rid of it entirely. This is also a really important point that needs to be documented in the help file.

I've experimented with eliminating all traces of `num_sites`, and managed to get it to compile and run. The integration test produces results that are very similar overall, but nevertheless different in detail, so clearly treating the single character as a vector of length 1 vs. a scalar does make a difference. Would you like me to push that change, @prilau?

src/core/distributions/phylogenetics/continuousCharacter/PhyloBrownianProcessStateDependent.h
prilau: this should not (and cannot) be a pure virtual function (as per Sebastian)

davidcerny: Let's remove the misleading comment, then

src/core/distributions/phylogenetics/continuousCharacter/PhyloOrnsteinUhlenbeckStateDependent.cpp
```diff
@@ -656,6 +679,13 @@ void PhyloOrnsteinUhlenbeckStateDependent::setSigma(const TypedDagNode<RbVector<
     state_dependent_sigma   = NULL;
 
 
+    size_t number_states = character_histories->getValue().getNumberStates();
+    // make sure that the state-space is correct
+    if ( s->getName().size() != number_states )
+    {
+    //    throw RbException() << "The number of states (" << number_states << ") in the character history doesn't match the number of sigma parameters (" << s->getValue().size() << ")";
```
davidcerny: Why are the exceptions in `::setAlpha()`, `::setSigma()`, and `::setTheta()` commented out? Wouldn't these be useful to the user?

src/core/datatypes/phylogenetics/characterhistory/BranchHistory.h
prilau: Yes this is as intended

src/core/distributions/phylogenetics/characterhistory/TreeHistoryCtmc.h
```diff
@@ -755,14 +768,14 @@ void RevBayesCore::TreeHistoryCtmc<charType>::setTipProbs(const HomologousCharac
 
     throw RbException("Bogus implementation for setTipProb (Sebastian)");
```
davidcerny, line 774: This looks concerning, but since it's pre-existing code, maybe this is not the time to fix it...

src/core/distributions/phylogenetics/continuousCharacter/PhyloBrownianProcessStateDependent.cpp
```diff
@@ -0,0 +1,865 @@
...
+    this->removeParameter( state_dependent_sigma );
+    homogeneous_sigma       = NULL;
+    state_dependent_sigma   = NULL;
+
+    size_t number_states = character_histories->getValue().getNumberOfStates();
+    // make sure that the state-space is correct
+    if ( s->getName().size() != number_states )
+    {
+        // throw RbException() << "The number of states (" << number_states << ") in the character history doesn't match the number of sigma parameters (" << s->getValue().size() << ")";
+    }
```
davidcerny: It looks to me like this is not quite ready? The last communication I remember us having about this was from December 15:

> the latest commit of my branch is a work-in-progress. I pushed it so that Sebastian can see it and maybe fix it during the week. (To be more specific, the checking number of states thing works with stochastic character history, but not with fixed history
> 
> That’s why the code is currently commented out

prilau: fixed

src/core/distributions/phylogenetics/continuousCharacter/PhyloOrnsteinUhlenbeckStateDependent.cpp
davidcerny: Could you please comment one more time on what solution you ended up adopting, @prilau? The code still has `num_sites` all over the place, but the help file for `dnPhyloCTMCDASiteIID` does say that "if this distribution is used together with `dnPhyloOUSD` or `dnPhyloBMSD`, specify `nSites=1`." Is it still the case that

> the current implementation is intended to use for one character (`num_sites`=1) only

?

prilau: `dnPhyloCTMCDASiteIID` should take `nSites=1` (i.e., 1 discrete character) when used with `dnPhyloOUSD` and `dnPhyloBMSD`, but that is not related to the code here.

Instead of deleting `num_sites` in `dnPhyloOUSD`, I fixed the previously broken loop such that `nSites` can now take any positive integers. Specifically, if one wanna model evolution of multiple continuous traits that follows the same state-dependent OU process, one should use `nSites>1`. I would say this scenario is possible but not too likely. The main purpose of this loop is to provide a framework for future extension to multivariate OU models.

davidcerny: Thanks so much for the clarification!

davidcerny (approved):

PR #905, opened 2025-11-28
Fix multi-sample BM and OU processes

HaoqingDu:
Fix issue with missing values in multi-sample OU and Add one test for the fixed function


src/core/distributions/phylogenetics/continuousCharacter/PhyloOrnsteinUhlenbeckPruning.cpp
```diff
@@ -282,52 +277,125 @@ void PhyloOrnsteinUhlenbeckPruning::recursiveComputeLnProbability( const Topolog
...
+
+                }
+                else if ( use_missing_data == true && missing_data[left_index][i] == false && missing_data[right_index][i] == true )
+                {
+                    missing_data[node_index][i] = false;
+                    
+                    p_node[i]  = p_left[i] + p_right[i];
+                    mu_node[i] = mu_left[i];
+                    
+                    this->variances_per_site[this->active_likelihood[node_index]][node_index][i] = var_left;
```
kopperud: If the data belonging to one of the subtrees (left or right) is missing, then the propagating variables (mu, var, log_nf or p_node) must still be updated to account for the branch duration, and this is missing here

src/core/distributions/phylogenetics/continuousCharacter/PhyloMultiSampleOrnsteinUhlenbeckProcessREML.cpp
```diff
@@ -419,52 +461,127 @@ void PhyloMultiSampleOrnsteinUhlenbeckProcessREML::recursiveComputeLnProbability
...
+
+                }
+                else if ( use_missing_data == true && missing_data[left_index][char_index] == false && missing_data[right_index][char_index] == true )
+                {
+                    missing_data[node_index][char_index] = false;
+                    
+                    p_node[char_index]  = p_left[char_index] + p_right[char_index];
+                    mu_node[char_index] = mu_left[char_index];
+                    
+                    this->variances_per_site[this->active_likelihood[node_index]][node_index][char_index] = var_left;
```
kopperud: I think also here, p_node is not being updated correctly

kopperud:
For this PR I suggest to:

Verify that the missing data calculation is correct. This can for example be done by adding another Drosophila dataset file where the missing data are pruned (missing taxa are also pruned from the tree). Then, you make a Rev script (lets say with some fixed OU model parameters) and test whether the likelihoods from the 5-taxon data and the 6-taxon data sets are equal. In the test data, you have (A,(B,(C,(D,(E,F))))) where there is fully missing data for D, but I also suggest to test the code a dataset where the clade (E,F) has fully missing data, to cover the part of the code where one subtree (and not just a tip) is completely missing.

Other suggestions:
* Add a validation test for the Brownian motion (REML), and the Brownian motion (non-REML)
* Add a test for the VCV version of the multi-sample OU process

Long-term (not in this PR) I suggest to
* Rename PhyloMultiSampleOrnsteinUhlenbeckREML to PhyloMultiSampleOrnsteinUhlenbeckPruning (because it is not a REML algorithm)
* Try and re-factor the recursive likelihood function to only cover **this branch** and not (**left branch**, **right branch**), because there is a lot of code duplication, which makes the code base difficult to understand and maintain

HaoqingDu:
Thanks for the comments from Bjorn. I changed the pull request to draft to have more tests on the feature.

bredelings:
Hi @kopperud, do you think this is ready to merge now?  (I hear you are defending soon, so no rush)

kopperud:
I have added tests for the BM multisample distribution (REML), and rewritten the likelihood function such that the tests pass (non-pruned and pruned datasets give equal likelihoods). But the simpler, species-level distributions remain to be tested

kopperud:
I have added two more tests for the following distributions
* `dnPhyloBrownianREML`
* `dnPhyloOrnsteinUhlenbeckPruning`
which tests whether the likelihoods of two datasets:
1. a six-species phylogeny, with 3 missing data and 3 observations
2. a three-species phylogeny with 3 observations
give the same likelihood score given the same parameter values. The tests pass on my local machine.

kopperud:
I think that the likelihoods for the species-level distributions were correct as they were. However I think the likelihoods for the multisample BM (REML) process were incorrect (even if there is no missing data), so I tried my best to change it. Here is my sketched approach for figuring out/validating the pruning-REML algorithm for a dataset with two species (A, B) and four samples (A1, A2, B1, B2). Capital letters are random variables, and lower-case letters are realizations

image

I represent the "standard" probability score for this graph as 

$$
P(A1 = a1, A2 = a2, B1 = b1, B2 = b2 | R = r, \sigma^2_\text{BM}, \sigma^2_\text{SP})
$$
or more compactly
$$
P(a1,a2,b1,b2|r,\sigma^2_\text{BM}, \sigma^2_\text{SP}).
$$

This can be calculated as follows

$$
P(a1,a2,b1,b2|r,\sigma^2_\text{BM}, \sigma^2_\text{SP}) =  \int_{-\infty}^\infty \int_{-\infty}^\infty P(b|r) P(a|r) P(a1|a) P(a2|a) P(b1|b) P(b2|b) da db.
$$

The main idea with the "REML" approach is to also integrate out the mean parameter of the BM process (in this case the $r$ parameter)

$$
P(a1,a2,b1,b2| \sigma^2_\text{BM}, \sigma^2_\text{SP}) =  \int_{-\infty}^\infty \int_{-\infty}^\infty \int_{-\infty}^\infty P(b|r) P(a|r) P(a1|a) P(a2|a) P(b1|b) P(b2|b) da db dr.
$$

I made a script where this three-dimensonal integral is calculated for a specific example (apologies for the Julia code, I was not able to find an R function to evaluate this type of integral)
```julia
using Integrals
using Distributions

# set up some parameter values
# mu is not necessary since we will integrate it out (the REML approach)
sigma_sq = 0.1
delta_t =  0.5
within_sp_var = 0.05

# the observations are chosen specifically such that Var([A1, A2]) = 0.05
A = 0.1
B = 0.2

A1 = A - 0.1581139
A2 = A + 0.1581139

B1 = B - 0.1581139
B2 = B + 0.1581139


# Treat A1, A2 and B1, B2 as observations: integrate out A and B and the mean
function p_specimen(x::Vector{Float64}, parameters) 
  p_A = pdf(Normal(x[1], sqrt(sigma_sq * delta_t)), x[2])
  p_A1 = pdf(Normal(x[2], sqrt(within_sp_var)), A1)
  p_A2 = pdf(Normal(x[2], sqrt(within_sp_var)), A2)

  p_B = pdf(Normal(x[1], sqrt(sigma_sq * delta_t)), x[3])
  p_B1 = pdf(Normal(x[3], sqrt(within_sp_var)), B1)
  p_B2 = pdf(Normal(x[3], sqrt(within_sp_var)), B2)
  res = p_A * p_A1 * p_A2 * p_B * p_B1 * p_B2
  return(res)
end

p_multispecies_naive = solve(IntegralProblem(p_specimen, ([-10, -10, -10], [10, 10, 10])), HCubatureJL()).u
p_multispecies_naive 
```
and the result turns out to be `P = 0.5833289181194762`.

Here is the same probability calculation but using a pruning algorithm
```julia

# Treat A1, A2 and B1, B2 as observations: integrate out A and B and the mean
# this time use a pruning algorithm
mu_A1 = A1
v_A1 = 0.0 + within_sp_var

mu_A2 = A2
v_A2 = 0.0 + within_sp_var

contrastA = mu_A1 - mu_A2
pContrastA = pdf(Normal(0.0, sqrt(v_A1 + v_A2)), contrastA)

mu_B1 = B1
v_B1 = 0.0 + within_sp_var

mu_B2 = B2
v_B2 = 0.0 + within_sp_var

contrastB = mu_B1 - mu_B2
pContrastB = pdf(Normal(0.0, sqrt(v_B1 + v_B2)), contrastA)

mu_A = (mu_A1 * v_A2 + mu_A2 * v_A1) / (v_A1 + v_A2)
v_A = (v_A1 * v_A2) / (v_A1 + v_A2) + delta_t * sigma_sq

mu_B = (mu_B1 * v_B2 + mu_B2 * v_B1) / (v_B1 + v_B2)
v_B = (v_B1 * v_B2) / (v_B1 + v_B2) + delta_t * sigma_sq

pRootContrast = pdf(Normal(0.0, sqrt(v_A + v_B)), mu_A - mu_B)

p_multispecies_pruning = pContrastA * pContrastB * pRootContrast
p_multispecies_pruning
```
and the result is the probability `P = 0.5833289223547985` which is not identical but similar enough.

My changes in the `dnPhyloBrownianMultiSampleREML` distribution are based on this line of thinking

davidcerny:
@kopperud So interestingly, I think the fact we're getting different output on macOS and Ubuntu reflects deeper issues with `dnPhyloOrnsteinUhlenbeckPruning`. The problem appears the second time we apply `mvSlide` to `optimum`: there is a small floating-point error associated with the likelihood evaluation:

macOS:

```
generation = 0    proposal = 2/6    Sliding(optimum)
    BEFORE:   optimum:  -2.9957322736
    BEFORE:   traits_pruned:  -57.51754963
    BEFORE:   traits:  -57.517550711

    PROPOSED: optimum:  -2.9957322736
    PROPOSED: traits_pruned:  -57.51754963
    PROPOSED: traits:  -57.517550711

    FINAL:    optimum:  -2.9957322736
    FINAL:    traits_pruned:  -57.51754963
    FINAL:    traits:  -57.517550711

    log(posterior_ratio) = -1.4210854715e-14  log(likelihood_ratio) = -1.4210854715e-14   log(prior_ratio) = 0
    log(acceptance_ratio) = -1.4210854715e-14  log(hastings_ratio) = 0
  The move was ACCEPTED.
```

Ubuntu:

```
generation = 0    proposal = 2/6    Sliding(optimum)
    BEFORE:   optimum:  -2.9957322736
    BEFORE:   traits_pruned:  -57.51754963
    BEFORE:   traits:  -57.517550711

    PROPOSED: optimum:  -2.9957322736
    PROPOSED: traits_pruned:  -57.51754963
    PROPOSED: traits:  -57.517550711

    FINAL:    optimum:  -2.9957322736
    FINAL:    traits_pruned:  -57.51754963
    FINAL:    traits:  -57.517550711

    log(posterior_ratio) = 2.842170943e-14  log(likelihood_ratio) = 2.842170943e-14   log(prior_ratio) = 0
    log(acceptance_ratio) = 2.842170943e-14  log(hastings_ratio) = 0
  The move was ACCEPTED.
```

Clearly, the `log(posterior_ratio)` should be exactly 0, because the likelihoods of `traits` and `traits_pruned` remain unchanged... but why do they remain unchanged if we have changed the `optimum`? That seems like a bug.

(Stranger still, there does eventually come a point when `Sliding(optimum)` results in `traits` and `traits_pruned` having different likelihoods between the `BEFORE` and `PROPOSED` states: it happens the 16th time the move is applied on macOS, and the 19th time it is applied on Ubuntu.)

I asked Cursor if there could be a problem with the way the distribution caches likelihoods, and got the following answer:

> **Cause in the distribution**
>
> In `PhyloOrnsteinUhlenbeckPruning::touchSpecialization`, when `touchAll` is true (e.g. when theta/optimum is the toucher), we only flip the `active_likelihood` buffer when `changed_nodes[index] == false`:
>
> ```
> if ( changed_nodes[index] == false )
> {
>     active_likelihood[index] = (active_likelihood[index] == 0 ? 1 : 0);
>     changed_nodes[index] = true;
> }
> ```
>
> Right after construction we call `resetValue()`, which sets all `changed_nodes[index] = true`. So for the first several touches (until the first `keep()`), every node has `changed_nodes[index] == true` and we never flip. We keep writing into the same buffer (buffer 0) and never switch to the alternate buffer. That can leave the likelihood tied to the old buffer and make the first several proposed values look like they don’t change the likelihood until a `keep()` finally resets `changed_nodes` to false and flips start happening.

I'm not sure if this makes sense or if it's just an LLM hallucination, though. The fixes that Cursor proposed didn't really solve the issue, so its diagnosis could easily be wrong too.

davidcerny:
Hmm... after a lot more experimentation, Cursor concluded:

> So the key takeaway: the move and DAG wiring are fine; the OU likelihood, for this test’s starting configuration, is genuinely (numerically) insensitive to theta.
>
> Your test uses:
> - `alpha ~ dnExponential(1.0)` → can easily be quite small early on.
> - Realistic branch lengths and fairly large $\sigma$
> - A lot of missing data.
>
> That combination can put you in a regime where, for the current alpha and tree, the sensitivity of the log‑likelihood to theta is so tiny that changing theta by ~0.18 moves the log‑likelihood by less than ~1e-12, i.e. below double‑precision noise and your printed precision.

Again, I'm not sure whether any of this is true, but if it is, we could try adding some sort of noise floor to our Metropolis-Hastings machinery, to ensure that sufficiently small log-likelihood differences are treated as equal to 0.

bredelings:
Interesting.  I'm not entirely sure if there is a problem with the current state, but ideally we would rule it out by switching away from the current situation (where we store 2N values plus info about whether `[0][i]` or `[1][i]` is the "active" one) to a situation where we store a N current values and N previous values, and simply revert to the old values if the move is rejected.  This way its impossible to get the active flag wrong.  Additionally, instead of storing a dirty flag, we store an `optional<T>` so that it is impossible to access a dirty value.  If `T` is large, we use a `std::shared_ptr<T>` instead of `optional<T>` so that a single value can be shared between the previous and current state.

We do this for conditional likelihoods here:

https://github.com/revbayes/revbayes/pull/816

I read through this PR, reading the individual patch titles, plus parts of each patch, and I think its mostly understandable.

For the current case you could define something like:
```
struct CachedNodeInfo
{
   std::vector<double> means;
   double variance;
   std::vector<double> partial_likelihoods;
};

std::vector<std::shared_ptr<CachedNodeInfo>> curr_node_info;
std::optional<std::vector<std::shared_ptr<CachedNodeInfo>>> prev_node_info;
```


Anyway, if someone wants to tackle this I could provide suggestions.  I think the key thing is to make incremental changes introducing the new framework, and add asserts checking that the new/old answers are the same (although you might find that the old answers are wrong).  Then at the end of the patch series you remove the old infrastructure.

bredelings:
@davidcerny There were indeed some bugs in the active flags for PhyloOrnsteinUhlenbeckPruning.  I fixed them by storing the active flags before the first touch and then comparing the current flags with them.

So now I wonder if this also fixes the difference between Mac and Linux, or not?

> Clearly, the log(posterior_ratio) should be exactly 0, because the likelihoods of traits and traits_pruned remain unchanged... 

One thing you'll note is that the BEFORE/PROPOSED/AFTER values are not printed to 17 digits.  This hides some possibly differences (say) the 14th digit.  So relative errors on the size of 1e-14 won't show up.  Normally I think this is what we want, but if you are trying to track down platform differences, you could locally change the precision to 17 digits.

> but why do they remain unchanged if we have changed the optimum? That seems like a bug.

I haven't investigated this.  You can put a breakpoint on touchSpecialization and/or computeLnProbability to double-check if the likelihood SHOULD be changing when it isn't.

davidcerny:
Thanks so much for looking into this, @bredelings! I'll recompile on my local (macOS) machine, regenerate the expected output, and push it to GitHub. That'll allow us to see if the inter-platform differences persist...

bredelings:
The new expected output was unchanged on my linux box, and the new expected output fails.  So I presume that Mac and Linux are still different :-(

I wonder if the library implementations for some mathematical functions like `exp` differ in the least significant bit, or something like that.

davidcerny:
Ah, that's unfortunate. In that case, though, I wonder if we should just special-case the workflow scripts and call it a day, especially now that you've made sure the distribution handles likelihood caching correctly (so we know that's not the source of the issue). It's not a very elegant solution, but we already do it for a number of other tests, so doing it for one more is not that egregious.

src/core/distributions/phylogenetics/continuousCharacter/PhyloBrownianProcessMultiSampleREML.cpp
```diff
@@ -472,7 +536,56 @@ void PhyloBrownianProcessMultiSampleREML::resetValue( void )
...
+                missing_data[species_index][site] = false;
+            }
+        }
+    }
+    use_missing_data = false;
     std::vector<TopologyNode*> nodes = this->tau->getValue().getNodes();
+    for (size_t site = 0; site < this->num_sites; ++site)
+    {
+        
+        for (std::vector<TopologyNode*>::iterator it = nodes.begin(); it != nodes.end(); ++it)
```
bredelings: Perhaps change this to a range-based for loop like:
```
for(auto node: nodes)
{
   if (node->isType())
...
```

davidcerny (approved):

PR #907, opened 2025-11-29
Fix the geometric distribution

### LINKED ISSUE #906: [BUG] Geometric distribution `rgeom()` and `dnGeometric()` returns incorrect mean
**Describe the bug**
The mean of the geometric distribution returned by rgeom() is incorrect in (at least) the development and master branches, returning approximately 171 instead of the expected value of ~99 given p=0.01.

**To Reproduce**
```
a <- rgeom(10000, 0.01)
mean(a)
```

```
val ~ dnGeometric(0.01)

for ( i in 1:10000 ) {
   values[i] <- val
   val.redraw()
}
mean(values)
```

**Expected behavior**
For a geometric distribution with parameter p, the mean should be 1/p or (1-p)/p depending on the parameterization:

Trials until first success: mean = 1/p
Failures before first success: mean = (1-p)/p   <-- this seems be used in RevBayes

Therefore, in the case of p = 0.01, the mean should be ~99. However:

`development (rapture-4406-g876a98)`
```
> a <- rgeom(10000, 0.01)
> mean(a)
   170.6947
```

`master (rapture-4205-gcbe72b)`
```
>  a <- rgeom(10000, 0.01)
> mean(a)
   171.8158
```

For comparison, this one seems to be correct
`dev_PoMo_SNP (rapture-3258-g41b2a2)`
```
> a <- rgeom(10000, 0.01)
> mean(a)
   98.9082
```

**Computer info**
macOS 26.1 (Tahoe)

davidcerny:
Cherry-picks @hoehna's fix to the geometric distribution from `dev_PoMo_SNP`, and adds an integration test.


wzhu0 (approved):

davidcerny:
As a heads-up, one case where these changes make a real-world difference is the calculation of standard errors for marginal likelihood estimates. In particular, it might explain why in my [attempts to compare the two](https://github.com/revbayes/revbayes/pull/779#issuecomment-3037285442), RevBayes seemed to consistently yield slightly higher and/or more narrowly distributed standard errors than the `mcmc3r` R package. (The differences were so small that they could be stochastic, so I thought nothing of it at the time.) @ms609 might find this of interest.

bredelings:
I wonder, though if we shouldn't use something more like this:

```
unsigned geometric(double p) {
    assert(0 <= p and p <= 1);
    return std::geometric_distribution<>(p)(standard_rng);
}
```

To some extent, this problem came from rolling our own code, which turned out to be half-baked, and not very well tested until now.  The code in std::geometric_distribution<> or boost::geometric_distribution<> should be much more thoroughly tested -- by other people.

davidcerny:
Huh, yes, that would indeed make a lot more sense! Does `std::geometric_distribution` produce output of type `unsigned`? Because `RbStatistics::Geometric::rv()` is supposed to yield a `std::int64_t`...

bredelings:
I think we use `std::geometric_distribution<std::int64_t>` in that case.  I am trying to remember how this works.

One annoying fact is that on Mac we compile with clang and use libc++, whereas on Linux we compile with gcc and use libstdc++.  That means that the exact numbers from the random number generator might be different if I remember correctly.  Our solution was to use `boost::geometric_distribution<std::int64_t>` to ensure the same behavior on both architectures.  It is annoying, as it would be nicer to using `std` instead of `boost`.  But that is what we are doing now, if I remember correctly.

src/core/math/distributions/DistributionGeometric.cpp
```diff
@@ -184,9 +177,11 @@ std::int64_t RbStatistics::Geometric::quantile(double q, double p)
...
 {
-    if (!RbMath::isFinite(p) || p <= 0 || p > 1) 
-        throw RbException()<<"Sampling geometric random variable: invalid success probability p = "<<p;
+    if (!RbMath::isFinite(p) || p <= 0 || p > 1)
+    {
+        throw RbException() << "Sampling geometric random variable: invalid success probability p = " << p;
+    }
     
-    return RbStatistics::Poisson::rv(exp(rng.uniform01()) * ((1 - p) / p),rng)+1;
+    return RbStatistics::Poisson::rv(rng.exponential() * ((1 - p) / p), rng);
```
bredelings: Should we be using a Poisson here at all??

bredelings (changes requested):
Using a Poisson to generate a geometric seems weird.

davidcerny:
Many thanks for the review, @bredelings! I've refactored the `rv()` function to use `boost::geometric_distribution()` like you suggested. Hopefully this guarantees platform-independent output.

The `propose_bad_index` and ML standard error tests are going to fail one more time; I'll replace the expected output with whatever the Linux-based GitHub Actions give me.

bredelings:
I looked at what you implemented, and for some reason the mean didn't seem right.  I changed it to use `boost::random::geometric_distribution<std::int64_t>` instead of using quantiles.  This seems less like rolling our own, and the numbers seem correct.  It required refactoring RandomNumberGenerator to expose the underlying generator.

This way basically imitates what we would do with `std::geometric_distribution`, except that we use `boost::geometric_distribution` to get the same result on different architectures.  Apparently that is not a strict guarantee though, as even the `log` function can different by at least 1 bit on different architectures.

@wzhu0 Can you check that this produces output that you expect?

bredelings (approved):

PR #908, opened 2025-11-29
Fixing `mvRateAgeProposal` for sampled ancestors

hoehna:
This is a cherry picked merged from code by Mike May


davidcerny:
What issue does this fix exactly? Can you provide a test for it?

davidcerny (changes requested):
Hmm, okay, I understand now what this does. It's supposed to bring `mvRateAgeProposal` (not `mvRateAgeBetaShift`, contrary to the original title of the PR) in line with other topology moves – specifically, `mvNodeRateTimeSlideUniform`, `mvNodeTimeScale`, `mvNodeTimeSlideBeta`, `mvRateAgeBetaShift`, and `mvSpeciesNodeTimeSlideUniform` – that pick a random internal node that is not the root or the parent of a sampled ancestor "tip".

However, we implemented a better solution to this problem in PRs #798 and #802, which has the advantage of not getting stuck in an infinite loop on "linear" trees in which all internal nodes are sampled ancestors. So we should replace lines 110–117 by the following:

```
// pick a random node which is not the root, a tip, or the parent of a sampled ancestor
TopologyNode* node = tau.pickRandomInternalNode(rng);
if (node == NULL)
{
    if (logMCMC >=1 or debugMCMC >=1)
    {
        std::cerr << "mvRateAgeProposal has no effect; the tree only contains the root, tips, and sampled ancestors." << std::endl;
    }
}
```

mikeryanmay:
Fixed another proposal mvRateAgeSubtreeProposal, and updated the FBD_dating_ex3.Rev tests also use these proposals. I updated the expected output for those tests as well, but per @davidcerny there may still be some platform-specific errors with these tests.

davidcerny (approved):
Thanks a lot, @mikeryanmay!

PR #909, opened 2025-11-29
fixing an issue with mpi and MC3

hoehna:
Some additional safety checks for MC3 when using with MPI


src/core/utils/MpiUtilities.cpp
```diff
@@ -45,8 +45,20 @@ void RevBayesCore::MpiUtilities::DebugMsg(const std::string& s) {
...
-    MPI_Barrier();
+    std::cout.flush();
+
+    MPI_Barrier(MPI_COMM_WORLD);
+
+    using namespace std::this_thread;     // sleep_for, sleep_until
+    using namespace std::chrono_literals; // ns, us, ms, s, h, etc.
+    using std::chrono::system_clock;
+
+    sleep_for(1s);
```
bredelings, line 56: What is this doing?

davidcerny (commented):
When I build the MPI version and use it to run the [`Test_heat_tuning_and_checkpointing.Rev`](https://github.com/revbayes/revbayes/blob/master/tests/test_MC3/scripts/Test_heat_tuning_and_checkpointing.Rev) script on 8 cores, the resulting output is the same between this branch and development. What problem is this PR trying to solve? When do we expect it to arise?

src/core/utils/MpiUtilities.cpp
hoehna: Giving me time to blink ;)

hoehna:
> When I build the MPI version and use it to run the [`Test_heat_tuning_and_checkpointing.Rev`](https://github.com/revbayes/revbayes/blob/master/tests/test_MC3/scripts/Test_heat_tuning_and_checkpointing.Rev) script on 8 cores, the resulting output is the same between this branch and development. What problem is this PR trying to solve? When do we expect it to arise?

This issue happens when you use #CPUs > (#chains * #reps), which means that the additional CPUs are used for parallel likelihood computation.

davidcerny:
I tried changing the heat tuning and checkpointing test to `nruns = 2, nchains = 3`, then I ran it on 12 cores (so with 2 cores per chain). The changes on this branch made no difference to the resulting likelihoods. I then tried to change the number of cores to 8, i.e., something other than an integer multiple of (`nchains` * `nruns`). This made no difference, either.

I also see that commit c21aa62 removes the 1-second sleep from lines 82 and 87 but not 56 and 61. Is this intentional?

davidcerny:
As a further test, I took the `mcmc_dNdS_branch.Rev` and `mcmc_dNdS_sites.Rev` scripts from the (never finished?) [codon model tutorial](https://revbayes.github.io/tutorials/selection/), uncommented the lines calling `mcmcmc()` (and commented out the lines calling `mcmc()` instead), specified 2 runs of 3 chains each, fixed the random seed, and ran the analyses on 12 cores. Again, the results obtained before and after making the change on this branch were completely identical.

I also tried running these experiments on the departmental cluster (i.e., using Linux and GCC instead of macOS and Clang), and there was no difference, either.

I suggest closing this PR – it looks like whatever problem this was originally intended to fix has been fixed in the meantime in some other way.

PR #910, opened 2025-11-29
Mcmc tree moves

hoehna:
This was some ongoing work to explore and add new moves on tree topologies and branch-lengths (only for unrooted trees). In detail, these new moves are:

- mvGibbsPruneAndRegraft: SPR move that tries all reattachment points and proposes proportional to likelihood.
- mvGibbsSubtreeSwap: Subtree Swap move that tries all swaps and proposes proportional to likelihood.
- mvWeightedBranchLengthScale: A branch-length move that tries to approximate the likelihood function over the fraction of left / right branch while keeping the sum the same. The move proposes from this distribution instead of random slides.
- mvWeightedPruneAndRegraft: SPR move that tries all reattachment points and integrates the attachment location using numerical approximation (thus getting a marginal likelihood). It proposes proportional to the marginal likelihood.
- mvWeightedSubtreeSwap: Subtree Swap move that tries all swaps and integrates the branch length of swap nodes using numerical approximation (thus getting a marginal likelihood). It proposes proportional to the marginal likelihood.


davidcerny:
This fails a single existing test – [`FBD_one_parameter_skyline.Rev`](https://github.com/revbayes/revbayes/blob/master/tests/test_FBD/scripts/FBD_one_parameter_skyline.Rev), because that's the only one to use `mvGPR`, whose behavior this PR modifies. We can probably just replace its expected output.

Other than that, it's just the usual – we need some tests and help files. 🙂

src/core/moves/proposal/tree/BiasedNearestNeighborInterchangeProposal.cpp
```diff
@@ -0,0 +1,436 @@
...
+    }
+    
+    
+    if ( attempts > attempts_before_using )
+    {
+        std::map<RbBitSet, TopologyNode*> current_clades = tau.getBitsetToNodeMap();
+        size_t num_clades = current_clades.size();
+        
+        throw RbException("Need to finish this implements (Sebastian).");
+
```
davidcerny: Seems like this needs addressing. I assume this is why the proposal has no Rev interface, and no entry in the register?

src/core/moves/proposal/tree/WeightedBranchLengthScaleProposal.cpp
```diff
@@ -0,0 +1,380 @@
...
+                    sol = sol2;
+                }
+            }
+            
+            //
+            proposed_branch_fraction = x1 + sol;
+            if ( proposed_branch_fraction < x1 || proposed_branch_fraction > x2 )
+            {
+                throw RbException("Wrong proposal");
+            }
```
davidcerny, line 241: A more descriptive error message would be good

davidcerny:
Hmm... the Gibbs versions of existing topology moves (`mvGibbsSPR`, `mvGibbsSubtreeSwap`) seem to work fine, as does `mvWeightedBranchLengthScale`. However, the weighted versions (`mvWeigthedSPR`, `mvWeightedSubtreeSwap`) cause a segfault when I try to use them.

I tried reverting the branch all the way back to commit 39fde58, and encountered the same problem. Therefore, as far as I can tell, these two moves never worked. We'll need to make an executive decision on whether we should try to fix them, or just remove them from the PR.

So far, I haven't been able to fix them on my own. I did find out that the segfault was due to the `sum_of_weights` variable becoming `nan`, which is in turn caused by the fact that when we first call the `computeMarginal()` function from within the loop that iterates over all "second nodes":

https://github.com/revbayes/revbayes/blob/2b138c2225fcd7c95f715b0e709d57e934503c31/src/core/moves/proposal/tree/WeightedSubtreePruneAndRegraftProposal.cpp#L242-L250

the log likelihood of the `tree` is `-inf`:

https://github.com/revbayes/revbayes/blob/2b138c2225fcd7c95f715b0e709d57e934503c31/src/core/moves/proposal/tree/WeightedSubtreePruneAndRegraftProposal.cpp#L112

Unfortunately, that's as far as I was able to get – I never figured out why that tree likelihood goes to 0.

PR #913, opened 2025-12-05
separate implementation of the birth-death-shift model

kopperud:
## New distribution: `dnFastBirthDeathShift`

Currently, the existing implementation `dnCDBDP` is a general and flexible state-dependent birth-death process. Depending on how the Rev-script is set up, the `dnCDBDP` can accommodate various analyses, including tests for trait-dependent diversification (with observed tip states), trait-agnostic diversification (i.e., the birth-death-shift model or LSBDS as in Höhna et al. 2026, in press), or other analyses such as those with cladogenetic events.

This flexibility is good, but there are also a few drawbacks:
* The large and complex code makes it difficult to change behaviour and add new features
* If the goal is to run a LSBDS analysis or similar, then the person who writes the Rev script needs to discretize the base distributions themselves, which can be tedious/verbose/error-prone
* The use of the Rev-side rate matrix object (the RateGenerator) has an unfortunate performance disadvantage: in particular it looks like the `getRate(...)` method introduces branches, and in turn the compiler is not able to produce efficient code. 
* In the ODE solver, it is assumed that in the vector product $Av$ that the matrix $A$ is always dense and the elements can take any value. Thus a general matrix-vector product implementation needs to be used

The goal of this PR is to implement a separate distribution `dnFastBirthDeathShift` that is simpler than dnCDBDP. There are some notable features:
* It does not allow for clamping character data
* Only log-normal base distribution is supported (less flexible)
* It does the base distribution discretization inside the `c++` code, making the Rev scripts simpler
* The run time of the (LSBDS and similar) analyses are quite a bit faster, by about 60-70 times:
    * The `RateGenerator` object is replaced by simple `double`s, allowing the compiler to better optimize the code
    * In the ODE solver, the matrix $A$ is sparse, and there are only one or two different numbers on the off-diagonal. Thus, the matrix-vector product $Av$ can be more efficiently with a special implementation, as opposed to a general matrix-vector product.
    * The if-statements in the hot loops of the ODE solver are removed, and replaced with additional calculations where necessary. I believe this makes it easier for the compiler to take advantage of CPU vector instructions

If anyone has comments I would appreciate it!


help/md/dnFastBirthDeathShift.md
```diff
@@ -0,0 +1,86 @@
...
+The possible speciation rates (lambda) and extinction rates (mu) are
+assumed to be distributed according to discretized base distributions.
+Currently, the log-normal base distribution is hard coded. The log-normal
+base distributions have location parameters (speciationScale, extinctionScale) 
+and spread parameters (speciationSD, extinctionSD), which can be fixed or
+estimated using prior distributions.
+
+In comparison with dnCDBDP, the dnFastBirthDeathShift distribution is 
+simplified and more optimized, allowing it to be significantly faster than 
+the implementation in dnCDBDP (some tests indicate 60-70x performance).
```
davidcerny: This is super nitpicky, but you could perhaps highlight argument and distribution names using backticks (` `` `).

src/core/distributions/phylogenetics/tree/birthdeath/BDS_ODE.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+            dxdt[i + 2*num_categories] += -no_event_rate * safe_x[i + 2*num_categories] + 2 * lambda[i] * safe_x[i] * safe_x[i + 2*num_categories];
+        }
+    }
+
+    if (forward){
+        for (size_t i = 0; i < (num_categories*2); i++){
+            dxdt[i] = -dxdt[i];
+        }
+    }
```
davidcerny: These two conditionals share the same condition, so you could perhaps simplify slightly by merging them...

src/core/distributions/phylogenetics/tree/birthdeath/FastBirthDeathShiftProcess.cpp
```diff
@@ -0,0 +1,2688 @@
...
+        rv = num_speciation_shift_events;
+    }
+    else if ( name == "numberExtinctionShiftEvents" )
+    {
+        rv = num_extinction_shift_events;
+    }
+    else
+    {
+        throw RbException() << "The state dependent birth-death process does not have a member method called '" << name << "'.";
+    }
```
davidcerny, line 1095: "state dependent birth-death process" -> "fast birth-death-shift process", maybe?

src/core/distributions/phylogenetics/tree/birthdeath/FastBirthDeathShiftProcess.cpp
```diff
@@ -0,0 +1,2688 @@
...
+        rv = delta_extinction; 
+    }
+    else if ( name == "getTimeInStates" )
+    {
+        rv = time_in_states;        
+    }
+    else
+    {
+        throw RbException() << "The state dependent birth-death process does not have a member method called '" << name << "'.";
+    }
```
davidcerny, line 1126: See also here

src/core/distributions/phylogenetics/tree/birthdeath/FastBirthDeathShiftProcess.cpp
```diff
@@ -0,0 +1,2688 @@
...
+            success = simulateTree(attempts);
+        }
+
+        if (success == true)
+        {
+            return;
+        }
+        ++attempts;
+    }
+    throw RbException("After 100000 attempts a character-dependent birth death tree could not be simulated. Try changing minNumLineages or maxNumLineages.");
```
davidcerny: Again, the exception should refer to FastBDS rather than CDBDP, and note that you actually set the number of attempts to 10,000 rather than 100,000.

src/core/distributions/phylogenetics/tree/birthdeath/FastBirthDeathShiftProcess.cpp
```diff
@@ -0,0 +1,2688 @@
...
+
+/**
+ * Set the current value.
+ */
+void FastBirthDeathShiftProcess::setValue(Tree *v, bool f )
+{
+    if (v->isBinary() == false)
+    {
+        throw RbException("The character-dependent birth death process is only implemented for binary trees.");
+    }
```
davidcerny, line 1451: Also change the process name here

src/core/help/RbHelpDatabase.cpp
```diff
@@ -5061,7 +5061,7 @@ for specifying how to separate values in the RevObject (default is "").)");
 # define RevObject to write
     x <- matrix([[1, 1],[1, 1]])
 # write to CSV file
-    write(x, "/path/to/file/filenmae.csv", false, ","))");
+    write(x, "/path/to/file/filename.csv", false, ","))");
```
davidcerny: While we are at it, might be also worth fixing the other typos in `write.md`: "This function write" -> "This function writes", "seperator" -> "separator".

src/revlanguage/distributions/phylogenetics/tree/Dist_FastBirthDeathShiftProcess.cpp
```diff
@@ -0,0 +1,379 @@
...
+
+        std::vector<std::string> slabels;
+        slabels.push_back("speciationScale");
+        std::vector<TypeSpec> speciationTypes;
+        speciationTypes.push_back( RealPos::getClassTypeSpec() );
+        memberRules.push_back( new ArgumentRule( slabels     , speciationTypes ,                          "The scale (in the log-normal distribution) of the speciation rate.", ArgumentRule::BY_CONSTANT_REFERENCE   , ArgumentRule::ANY ) );
+
+        std::vector<std::string> elabels;
+        elabels.push_back("extinctionScale");
+        memberRules.push_back( new ArgumentRule( elabels     , RealPos::getClassTypeSpec() , "The scale (in the log-normal distribution) of the speciation rate.", ArgumentRule::BY_CONSTANT_REFERENCE   , ArgumentRule::ANY ) );
```
davidcerny: extinction, not speciation

src/revlanguage/distributions/phylogenetics/tree/Dist_FastBirthDeathShiftProcess.cpp
```diff
@@ -0,0 +1,379 @@
...
+
+        memberRules.push_back( new ArgumentRule( "speciationSD", RealPos::getClassTypeSpec(), "The spread (sigma-parameter in the log-normal distribution) of the speciation rate.", ArgumentRule::BY_CONSTANT_REFERENCE, ArgumentRule::ANY, new RealPos(0.587)));
+        memberRules.push_back( new ArgumentRule( "extinctionSD", RealPos::getClassTypeSpec(), "The spread (sigma-parameter in the log-normal distribution) of the extinction rate.", ArgumentRule::BY_CONSTANT_REFERENCE, ArgumentRule::ANY, new RealPos(0.587)));
+
+        memberRules.push_back( new ArgumentRule( "alpha", RealPos::getClassTypeSpec()        , "the rate of shifts in speciation rate", ArgumentRule::BY_CONSTANT_REFERENCE   , ArgumentRule::ANY, NULL ) );
+        memberRules.push_back( new ArgumentRule( "beta", RealPos::getClassTypeSpec()        , "the rate of shifts in extinction rate", ArgumentRule::BY_CONSTANT_REFERENCE   , ArgumentRule::ANY, NULL ) );
```
davidcerny: Maybe capitalize the word "the" for consistency's sake...

src/revlanguage/monitors/Mntr_StochasticBranchRate.cpp
```diff
@@ -69,6 +70,19 @@ void Mntr_StochasticBranchRate::constructInternalObject( void )
...
+        RevBayesCore::StochasticNode<RevBayesCore::Tree>* bds_sn  = static_cast<RevBayesCore::StochasticNode<RevBayesCore::Tree>* >( bds_tdn );
+
+        RevBayesCore::FastBirthDeathShiftProcess *bds_process = NULL;
+        bds_process = dynamic_cast<RevBayesCore::FastBirthDeathShiftProcess*>( &bds_sn->getDistribution() );
+        bds_process->setSampleCharacterHistory( true );
+
+        m = new RevBayesCore::StochasticBranchRateMonitor( bds_sn, (std::uint64_t)print_gen, file_name, sep );
+        m->setAppend( app );
+        m->setPrintVersion( wv );
+    }
```
davidcerny: Since we are adding a third valid option in addition to CDBDP and GLHBDSP, the exception thrown in the `else` branch should be updated accordingly.

src/revlanguage/monitors/Mntr_StochasticBranchRate.cpp
```diff
@@ -139,6 +153,7 @@ const MemberRules& Mntr_StochasticBranchRate::getParameterRules(void) const
     {
         monitor_rules.push_back( new ArgumentRule("cdbdp"          , TimeTree::getClassTypeSpec(),  "The character dependent birth-death process to monitor.",                      ArgumentRule::BY_REFERENCE, ArgumentRule::ANY, NULL ) );
         monitor_rules.push_back( new ArgumentRule("glhbdsp"        , TimeTree::getClassTypeSpec(),  "The lineage-heterogeneous birth-death process to monitor.",                    ArgumentRule::BY_REFERENCE, ArgumentRule::ANY, NULL ) );
+        monitor_rules.push_back( new ArgumentRule("bds"            , TimeTree::getClassTypeSpec(),  "The birth-death-shift process to monitor.",                                    ArgumentRule::BY_REFERENCE, ArgumentRule::ANY, NULL ) );
```
davidcerny: The fact that we have separate arguments for alternative processes seems strange to me. Wouldn't it be better to have a single argument called something like `process`, overloaded so that it accepts CDBDP, GLHBDSP, and FastBDS?

tests/test_FastBDS/scripts/mcmc_BDS.Rev
```diff
@@ -0,0 +1,121 @@
+################################################################################
+#
+# RevBayes test: Bayesian inference of diversification rates under a
+#                relaxed birth-death-shift model
+#
+#
+# authors: Bjorn Kopperud & Sebastian Hoehna
+#
+################################################################################
```
davidcerny, line 9: Very nice test! Just two things:

- The [test runner](https://github.com/revbayes/revbayes/blob/development/tests/run_integration_tests.sh) wants the comparison directory to be called `output_expected` rather than `expected_output`;
- This directory should also include an `.errout` file, obtained by running the test and redirecting the stdout and stderr to a file with the same base name as the script:

```
../projects/cmake/build/rb ./test_FastBDS/scripts/mcmc_BDS.Rev &> ./test_FastBDS/output_expected/mcmc_BDS.errout
```

src/revlanguage/distributions/phylogenetics/tree/Dist_FastBirthDeathShiftProcess.cpp
```diff
@@ -0,0 +1,379 @@
...
+ * Get the alternative Rev names (aliases) for the constructor function.
+ *
+ * \return Rev aliases of constructor function.
+ */
+std::vector<std::string> Dist_FastBirthDeathShiftProcess::getDistributionFunctionAliases( void ) const
+{
+    // create alternative constructor function names variable that is the same for all instance of this class
+    std::vector<std::string> a_names;
+    a_names.push_back( "dist_alias_here" );
+
```
davidcerny, line 204: Don't forget to change this placeholder! Personally, I'd suggest at least "FastBDS" as a valid alias.

davidcerny (changes requested):
The math here is above my pay grade, especially when it comes to the ODE solver, but overall, this PR looks really nice, clean, and well-documented. I only have a few suggestions relating mostly to the interface.

kopperud:
Thanks for all the comments, David. I will try to fix these next week

davidcerny:
Many thanks for the changes, @kopperud! Just a few notes about the test:

- ~~`mnStochasticBranchRate` doesn't seem to be working as intended. The script says it should record every generation (`printgen=1`), but the only thing it prints is the header.~~
- I noticed that you tune the moves during the MCMC run in your script. As @bredelings recently pointed out on Slack, this makes it (at the very least) questionable whether the analysis converges to the posterior.
- The test seems to produce the same output for every platform/OS/compiler combo except for Intel Macs. Well, that's a first... usually, when we get platform-specific output, it's macOS, Windows, or both deviating from Linux. For what it's worth, the divergence first happens in the 3rd proposal of generation 8. Here is what it looks like on Intel Macs:

    ```
    generation = 8    proposal = 3/8    Scaling(speciation_shift_rate)
    BEFORE:   speciation_shift_rate:  4.162180712
    BEFORE:   timetree:  -1569.2183985

    PROPOSED: speciation_shift_rate:  3.8697271022
    PROPOSED: timetree:  -1569.0821148

    FINAL:    speciation_shift_rate:  3.8697271022
    FINAL:    timetree:  -1569.0821148

    log(posterior_ratio) = -0.15616987823  log(likelihood_ratio) = 0.13628373154   log(prior_ratio) = -0.29245360977
    log(acceptance_ratio) = 0.046287405629  log(hastings_ratio) = 0.20245728386
  The move was ACCEPTED.
  ```
  
  And here is the output for every other platform:
  
  ```
  generation = 8    proposal = 3/8    Scaling(speciation_shift_rate)
    BEFORE:   speciation_shift_rate:  4.162180712
    BEFORE:   timetree:  -1569.2183985

    PROPOSED: speciation_shift_rate:  4.1465044079
    PROPOSED: timetree:  -1569.207332

    FINAL:    speciation_shift_rate:  4.1465044079
    FINAL:    timetree:  -1569.207332

    log(posterior_ratio) = -0.0046098103579  log(likelihood_ratio) = 0.011066493737   log(prior_ratio) = -0.015676304095
    log(acceptance_ratio) = 0.0087824646086  log(hastings_ratio) = 0.013392274966
  The move was ACCEPTED.
  ```
  
  There are two possible ways forward. We could try to investigate where the difference comes from, or just special-case the [test runner](https://github.com/revbayes/revbayes/blob/master/.github/workflows/build.yml) to ignore it.

davidcerny:
Out of curiosity, I asked Claude to help me diagnose the source of the discrepancy. After some experimentation, this was as far as I got:

> we’ve now ruled out “same parameters → different ODE behavior” and instead confirmed **“parameters themselves diverge across platforms,”** which is exactly what you’d expect if:
> - The RNG stream is identical but **borderline Metropolis–Hastings decisions** (accept vs reject) flip on Intel due to extended-precision / libm differences, or
> - Some earlier part of the model (e.g. lognormal/normal quantile, log-likelihood, or prior) computes slightly different doubles on Intel.
>
> Two concrete next directions you can choose between:
> - **Numerical / build-level fix targeted at macOS+Intel**
>
>    Add strict FP flags for Intel macOS builds (e.g. to force SSE and disable extended precision) in your build system, so it behaves more like Linux/ARM:
>     - In CMake (pseudo-code): under `if(APPLE AND CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64")` add
`-mfpmath=sse -msse2 -ffloat-store -ffp-contract=off -fno-fast-math`.
>     - In Meson: similarly add those to `cxx_args` when `host_machine.system() == 'darwin' and host_machine.cpu_family() == 'x86_64'`.
>
>    This is the most realistic way to make Intel macOS numerically behave like the others without invasive code changes.
> - **Test-level adjustment instead of bit‑for‑bit equality**
>
>    Accept that MCMC state will diverge across platforms and adjust the test to compare:
>    - Summary statistics (e.g. mean log-likelihood, posterior mean of a parameter) within a tolerance, or
>    - A deterministic quantity computed _without_ MCMC (e.g. a single log-likelihood evaluation at fixed parameters) as the cross-platform invariant.

I tried changing `projects/meson/build.sh` in accordance with the first of these suggestions, but it didn't fix the problem, unfortunately.

bredelings:
Hi @kopperud, can you please extend the cover letter for this PR to say why it makes sense to have multiple implementations of the same process?  Please also give an overview of what the PR does so that it is easier to review.

src/revlanguage/monitors/Mntr_StochasticBranchRate.cpp
kopperud: I spoke with David and we are in agreement that these arguments are confusing and overall not necessary. But we thought that the issue could be ignored for now and we could open a separate PR to make the monitor simpler to use

davidcerny: Actually, I found a way to do what I had suggested, so I took the liberty of making the change and including it as part of this PR...

davidcerny (approved):

bredelings:
Hi Bjorn,

Thanks for writing the cover letter.  It is definitely easier to understand about what you are doing now.

In general, duplicating code and modifying it is a bad design pattern.  It is often easier to write code this way, but much hard to maintain it.  For example, if any changes to the BDS code are needed, we now need to fix them in 2 places -- will it even be possible to keep then in sync?  This creates 1 extra copy, but in the future we might end up with even more copies.  For example, if someone wanted to use a distribution other than the log-normal, would we duplicate it _again_?  

Or, as Cory Doctorow says: [Code is a liability, not as asset](https://doctorow.medium.com/https-pluralistic-net-2026-01-06-1000x-liability-graceful-failure-modes-d69f384af9e4).

So, I think the conversation we need to have first is what would be needed to put the functionality from this new distribution into the existing dnCDBDP distribution.  If that ends up being too painful, we would then talk about how to share as much code as possible between the 2 implementations.

bredelings (changes requested):
I think we need to avoid so much code duplication.

bredelings:
Here are some preliminary thoughts that are based more on the cover letter than the code.  (I will have more time to look at this when I get back from Egypt.  I should be back around Feb 16, so I can look more that week.)

* I agree that accessing individual rate elements through `rm->getRate(from_state, to_state, age, 1.0);` could be really slow.  When the rate is independent of the age, this is excessively general.  Also, doing a virtual method call is going to be really slow.

* However, if the RateGenerator is actually a RateMatrix, then it should be possible to dynamic_cast it an AbstractRateMatrix, where accessing individual elements should be fast.

* If the main ideas here involve faster computation of $Av$ when $A$ is sparse, then one thing we could to is to provide a virtual method that computes $Av$.  Then we could make a derived class for rate generators that are (i) constant in time and (ii) sparse.  Then a virtual method to compute $Av$ could be very fast.

I still need to look at what you are doing with the integrator.  In theory, if the new integrator is better, we might want to use that in both classes.

Thoughts?

bredelings:
Also, we are talking about making Eigen always available in RevBayes.  Eigen can generate really good matrix multiplication code.  For dense matrices, it does both vectorization and tiling.  The tiling is important for large matrices that do not fit in the cache.

Eigen also has classes for banded and sparse matrices.

kopperud:
> Thanks for writing the cover letter. It is definitely easier to understand about what you are doing now.
> 
> In general, duplicating code and modifying it is a bad design pattern. It is often easier to write code this way, but much hard to maintain it. For example, if any changes to the BDS code are needed, we now need to fix them in 2 places -- will it even be possible to keep then in sync? This creates 1 extra copy, but in the future we might end up with even more copies. For example, if someone wanted to use a distribution other than the log-normal, would we duplicate it _again_?

I agree that it will make the code base more difficult to maintain, that is unfortunate. But I thought this was relatively standard practice (see other distributions on birth-death models or continuous-trait models)

> So, I think the conversation we need to have first is what would be needed to put the functionality from this new distribution into the existing dnCDBDP distribution. If that ends up being too painful, we would then talk about how to share as much code as possible between the 2 implementations.

I think many of the performance improvements could also be done for the dnCDBDP. But the interfaces (the distribution arguments) are incompatible, for example I don't want to make a Q matrix in my Rev script, as I find it is not pedagogical/not necessary for interpreting this particular model. Refactoring the distributions to share more code would be very difficult for me to do, though, as I dont fully understand how it works

kopperud:
>     * However, if the RateGenerator is actually a RateMatrix, then it should be possible to dynamic_cast it an AbstractRateMatrix, where accessing individual elements should be fast.
>
>     * If the main ideas here involve faster computation of Av when A is sparse, then one thing we could to is to provide a virtual method that computes Av. Then we could make a derived class for rate generators that are (i) constant in time and (ii) sparse.  Then a virtual method to compute Av could be very fast.
>

I completely agree here, but I am not experienced enough with c++ to know how to implement this (and without making a mess)

> I still need to look at what you are doing with the integrator. In theory, if the new integrator is better, we might want to use that in both classes.

If by integrator you mean the euler/RK4/dopri etc. algorithms, then there is no new integrator here. The only difference is the gradient function (calculating du/dt) which can be done more efficiently in this special case

kopperud:
> I think we need to avoid so much code duplication.

If you think it is too much to add then I understand if it can't be merged (or not at this time). I am not relying on this for any particular project, I just thought that I put quite a lot of work into it, and that it would be a shame to throw it away. But thanks for all the comments!

bredelings:
Hi Bjorne,

Yes, it would be a shame if this could not be merged.  Perhaps I could help you out on the stuff that seems challenging.

I see what you mean about wanting the interface to be different.  Its possible you could add a new interface but keep the StateDependentSpeciationExtinctionProcess backend.  Then you could add special cases in the StateDependentSpeciationExtinctionProcess backend that make it faster for the special case you are interested in.  However, I'm not sure why the new interface means that character data cannot be clamped -- why is that?

Is there a reason why you chose dnFastBirthDeathShift instead of (say) dnLineageSpecificBirthDeathShift?

I can help with C++ stuff.  I need to actually look at the C++ code you wrote in order to make a specific recommendation, but in general RevBayes would benefit from being able to handle banded and sparse matrices, so it seems worth the effort.  I need to wait until after I get back from Egypt though.

PR #916, opened 2025-12-11
Merge pull request #824 from revbayes/development

bredelings:
Move master  forward to development

This exists to make master an ancestor of development.

PR #919, opened 2025-12-17
Brief follow-up to v1.3.2 release

davidcerny:
As pointed out by @bredelings on Slack, much of the current RevBayes infrastructure wasn't built by currently active developers, and when it comes to actions we don't perform that often (prepare a new release, keep the documentation on the website in sync with the documentation available from within RevBayes, etc.), we end up having to rediscover how things work every time, sometimes by trial and error. This PR is meant to create a permanent record of some of these "discoveries".

Comments welcome!

### DIFF
--- a/doc/release-workflow.md
+++ b/doc/release-workflow.md
@@ -34,20 +34,19 @@ Neither the [master](https://github.com/revbayes/revbayes/tree/master) branch no
 
 4. Merge development into master
     * On doing this, GitHub creates a merge commit, giving master 1 unique change that is not on development, and making this whole dance necessary next time
+    
+5. Generate a YAML-formatted help file
 
-5. Update website from within the [revbayes.github.io](https://github.com/revbayes/revbayes.github.io) repo
+    1. Re-build RevBayes from scratch. This automatically generates a `rb-help2yml` executable outside the `build` directory.
+    2. Run the executable to generate a `help.yml` file from the contents of [`help/md`](https://github.com/revbayes/revbayes/tree/master/help/md)
+        * `./rb-help2yml`
 
-    1. Checkout the source branch
-        * `git checkout source`
-        * `git pull`
-    2. Create a release branch off of source
-        * `git branch release-vX.Y.X`
-        * `git checkout release-vX.Y.Z`
-    3. Update download version by modifying `_config.yml`
-    4. Re-build RevBayes from scratch. This automatically generates a `help.yml` file from the contents of [`help/md`](https://github.com/revbayes/revbayes/tree/master/help/md).
-    5. Copy `help.yml` to the `_data` directory
-    6. Commit the changes
+6. Update website from within the [revbayes.github.io](https://github.com/revbayes/revbayes.github.io) repo
+
+    1. Update download version by modifying `_config.yml`
+    2. Copy `help.yml` to the `_data` directory
+    3. Commit the changes
         * `git add _config.yml`
         * `git add _data/help.yml`
         * `git commit -m "Update RB version to <X.Y.Z>"`
-    7. Open a pull request to merge the release branch into source
+
--- a/doc/release-workflow.md
+++ b/doc/release-workflow.md
@@ -1,27 +1,53 @@
 # Release workflow
 
-0. Update NEWS.md file
+Neither the [master](https://github.com/revbayes/revbayes/tree/master) branch nor the [development](https://github.com/revbayes/revbayes/tree/development) branch is a direct ancestor of the other: each has at least one unique commit. Therefore we cannot merge either one into the other using a PR, because the PR will always say that the branch to be merged is out-of-date with respect to the base branch. What we can do is create our own commit that is an ancestor of both, instead of relying on GitHub PRs to merge things for us. But that has to be on a non-protected branch, since protected branches can only be modified by PRs.
 
-1. Build packages
+1. Create a (non-protected) release branch
 
     1. Checkout the development branch
         * `git checkout development`
         * `git pull`
+    2. Create the release branch off of development
+        * `git branch release-vX.Y.Z`
+        * `git checkout release-vX.Y.Z`
+    3. Update `NEWS.md` file
+    4. Make sure the release workflow (`release.yml`) is in sync with the build workflow (`build.yml`)
+
+2. Prepare the release
+
+    1. Merge master into the release branch using git
+        * `git merge master`
     2. Create a tag for the release
-        * `git tag vD.D.D`
-        * tag names
-            * v2.0 for a major release
+        * `git tag vX.Y.Z`
+        * tag names: no strict semantic versioning; use v2.0 for a major release
     3. Push the tag to github
         * `git push --tags`
         * This will create a draft release.
-        * GitHub actions will upload packages for Linux, Linux/Singularity, Mac, Windows
+        * GitHub actions will upload packages for Linux, Linux/Singularity, macOS, Windows
     4. Wait for packages to build.
-    5. Edit draft-release page on github
+    5. Edit the draft release page on GitHub
         * See https://github.com/revbayes/revbayes/releases
-        * Add some basic release notes.
-        * Make the draft release public.
+        * Add release notes from `NEWS.md`
+        * Make the draft release public
+        
+3. Open a pull request to merge the release branch into development, making development a strict descendant of master
 
-2. Update website
+4. Merge development into master
+    * On doing this, GitHub creates a merge commit, giving master 1 unique change that is not on development, and making this whole dance necessary next time
 
-    1. Update version number on download page.
-    2. Update links on download page.
+5. Update website from within the [revbayes.github.io](https://github.com/revbayes/revbayes.github.io) repo
+
+    1. Checkout the source branch
+        * `git checkout source`
+        * `git pull`
+    2. Create a release branch off of source
+        * `git branch release-vX.Y.X`
+        * `git checkout release-vX.Y.Z`
+    3. Update download version by modifying `_config.yml`
+    4. Re-build RevBayes from scratch. This automatically generates a `help.yml` file from the contents of [`help/md`](https://github.com/revbayes/revbayes/tree/master/help/md).
+    5. Copy `help.yml` to the `_data` directory
+    6. Commit the changes
+        * `git add _config.yml`
+        * `git add _data/help.yml`
+        * `git commit -m "Update RB version to <X.Y.Z>"`
+    7. Open a pull request to merge the release branch into source
--- a/NEWS.md
+++ b/NEWS.md
@@ -53,13 +53,15 @@
 
 ## Documentation improvements
   * Expand documentation for `mvSlice` (#614).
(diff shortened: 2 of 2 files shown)


bjoelle:
I don't understand why we have to do this work-around with respect to the `master` and `development` branch - if I remember correctly, you just need to merge `master` into `development`, either just after or just before release, and then `development` can be merged into `master` normally

bjoelle:
also, I just checked and you can absolutely do a PR of a branch that has diverged, it just means you will have to resolve any conflict manually

davidcerny:
@bjoelle The sequence you suggest doesn't seem to be possible, unfortunately, for reasons already stated in the release workflow doc. @bredelings can correct me if I got this wrong, but it looks like you can merge master into development as long as it's strictly ahead of the latter (#823 is a successful example), but not once the two branches have diverged (see #916 for a failed attempt to do just that).

> also, I just checked and you can absolutely do a PR of a branch that has diverged, it just means you will have to resolve any conflict manually

Yes, I don't think that was ever in question? We have a whole bunch of currently open PRs doing just that. The problem is specific to merging one protected branch into another.

bjoelle:
looks like there is a setting we can turn off that will remove the "out-of-date" issue
what happens if you approve #921 ?

bjoelle:
but even without changing anything it sounds like updating development with the new master before doing anything else to development would also work ?

davidcerny:
I've approved, so let's see. But I'd expect #921 to succeed, for the same reason #823 succeeded: right now, master is strictly ahead of development, instead of having diverged from it.

> but even without changing anything it sounds like updating development with the new master before doing anything else to development would also work ?

Yes, maybe that's the real solution – to not let master and development diverge in the first place. I wonder what @bredelings thinks.

davidcerny:
Also, what's the setting in question? That sounds like a potentially useful thing to jot down, too!

bjoelle:
so it's in settings > branches > Require branches to be up to date before merging
I don't remember this so it was likely on by default - we could turn it off for good or just keep the option in mind if we run into the issue again

ms609:
Just a thought: most of these steps look like they could be incorporated into a GitHub action script, with a `workflow_dispatch:` trigger in which a version number is specified as an argument.  If the release notes are to be taken verbatim from `NEWS.md`, then from my quickish look the only human input required is to specify the version number of the next release?

Perhaps a cautious approach would be a two-step workflow: the first step preparing a draft release, with release notes, in its own branch (thus allowing for any validation and edits at that point); the second merging the release into master and marking it as the current release?

davidcerny:
I might have gotten this wrong, and the step of extracting release notes from `NEWS.md` might already be automated. I've never handled this part of the release process myself, so I'm not 100% sure.

bredelings:
So, "not letting master and development diverge in the first place" isn't any simpler than what's being proposed.  The problem is that if we merge development into master using a PR, github creates a merge commit instead of doing a fast-forward merge.  So as soon as we merge development into master, they have diverged!  This is ridiculous of course, but it seems that github won't do fast-forward merges.

So we have to do a synchronization dance.  We can do it right after a new release, or we can do it right before the next release (as proposed here).  Perhaps we are atypical in having two protected branches instead of just one?

As far as documenting things, I think that writing down notes is good.  However, its probably important that these are labeled as "notes" or "discoveries", and that they contain explanations about WHY we do things, so that later people can decide when/whether to modify them.  They wouldn't be "laws". If we call them a "tutorial" then it implies that you can just type the command and it should work.  But just because something worked in 2025 doesn't mean it will work in 2027.  I suppose we could also add a date stating when something worked, continually updating that would be annoying.

bredelings:
That said, turning off the "Require branches to be up-to-date before merging" _seems_ safe.  I'm trying to think of any problems this would cause...

So it seems that our tests run on both the branch and the proposed merge: that's why we always run the tests twice -- once for the branch, and once for the PR that actually tests the proposed merge.

In theory, if we require "up-to-date" branches, then any changes due to the merge would show up in the "changed files".  But currently we approve PRs, and you can then do the merge afterwards without getting a re-approval.  

So it seems like disabling this strict up-to-date requirement should be fine.  We could always change it back afterward if there was a problem.

doc/release-workflow.md
```diff
@@ -1,27 +1,53 @@
...
+5. Update website from within the [revbayes.github.io](https://github.com/revbayes/revbayes.github.io) repo
+
+    1. Checkout the source branch
+        * `git checkout source`
+        * `git pull`
+    2. Create a release branch off of source
+        * `git branch release-vX.Y.X`
+        * `git checkout release-vX.Y.Z`
+    3. Update download version by modifying `_config.yml`
+    4. Re-build RevBayes from scratch. This automatically generates a `help.yml` file from the contents of [`help/md`](https://github.com/revbayes/revbayes/tree/master/help/md).
```
bredelings: Is this automatic?  I thought you had to run the rb-help2yml executable to produce the help.yml file.  It looks like ./build.sh specifies -DHELP=ON, so it builds rb-help2yml, but I don't see where it runs it.

doc/release-workflow.md
```diff
@@ -1,27 +1,53 @@
...
+3. Open a pull request to merge the release branch into development, making development a strict descendant of master
 
-2. Update website
+4. Merge development into master
+    * On doing this, GitHub creates a merge commit, giving master 1 unique change that is not on development, and making this whole dance necessary next time
 
-    1. Update version number on download page.
-    2. Update links on download page.
+5. Update website from within the [revbayes.github.io](https://github.com/revbayes/revbayes.github.io) repo
+
```
bredelings, line 43: Maybe the instructions are too detailed here? The main "discovery" seems to be that we need to update `_config.yml` and update `_data/help.yml`, plus the info about how to generate `help.yml`.  How to update the website seems like it would fit better in `README.md`.  At the very least, we could point to the website `README.md`, which would make this seem less like a magic incantation.

bredelings:
> Just a thought: most of these steps look like they could be incorporated into a GitHub action script, with a `workflow_dispatch:` trigger in which a version number is specified as an argument. If the release notes are to be taken verbatim from `NEWS.md`, then from my quickish look the only human input required is to specify the version number of the next release?
> 
> Perhaps a cautious approach would be a two-step workflow: the first step preparing a draft release, with release notes, in its own branch (thus allowing for any validation and edits at that point); the second merging the release into master and marking it as the current release?

@ms609 It might be possible to make the `release.yml` workflow specify the release notes for the draft release it creates.  In theory we could make the `release.yml` workflow update the version number and the `help.yml` in the website; but unfortunately you can only modify the website `source` branch using a PR to ensure that the tutorials don't break.  The master branch is (currently at least) unprotected.

We should note that, in reality, what usually happens is that I push a tag, it creates the release, but something is wrong with it.  I then delete the tag locally, google how to delete a remote tag, make some git comments, retag, and push a new tag.  Rinse, lather, repeat.  I think for this last release, I only had the push the tag twice, which is pretty low.

I like the cautious approach that automatic actions should only create draft releases.

davidcerny:
Apologies for dropping the ball on this for so long, there's been a lot of interesting points raised in this discussion! I'll try to address them one by one:

@bredelings:

> So, "not letting master and development diverge in the first place" isn't any simpler than what's being proposed. The problem is that if we merge development into master using a PR, github creates a merge commit instead of doing a fast-forward merge. So as soon as we merge development into master, they have diverged! This is ridiculous of course, but it seems that github won't do fast-forward merges.
>
> So we have to do a synchronization dance. We can do it right after a new release, or we can do it right before the next release (as proposed here).

I think @bjoelle's point was that if we do things in the right order (i.e., merging `master` into `development` immediately after `master` has been updated, before making other changes to `development`), we can get rid of the requirement to introduce a third, non-protected branch that is an ancestor of both `master` and `development`. In that sense, doing the synchronization dance right after a new release _is_ simpler than what's being proposed here, as it involves one fewer branches (but a stricter timing requirement).

We discussed this argument in Baton Rouge, drawing git branching diagrams on a whiteboard and not really reaching any conclusion. But based on the fact that #921 (a proposed merge of `master` into an unchanged `development`) was capable of being merged, I think @bjoelle was right about this.

Of course, this might all be beside the point if, as @bjoelle suggests, we instead decide to play around with the

> setting we can turn off that will remove the "out-of-date" issue

we're currently running into. I'm a bit wary of disabling this setting, if only because it has been our default for so long. It should indeed be safe in most situations, but there is a tiny chance it could cause issues that our current setup is guaranteed to avoid: we would no longer subject the merged state to CI testing, so on the (admittedly slim) chance that the three-way merge broke something, we would not detect it before the breakage had already made it to the base branch.

> As far as documenting things, I think that writing down notes is good. However, its probably important that these are labeled as "notes" or "discoveries", and that they contain explanations about WHY we do things, so that later people can decide when/whether to modify them. They wouldn't be "laws". If we call them a "tutorial" then it implies that you can just type the command and it should work.

I agree. I did try to include explanations of why we do what we do in the current version of the notes (e.g., why it's necessary to create a separate release branch, though see above). Note that I only used the (potentially misleading) word "tutorial" in the cover letter for the PR, rather than in the actual document I updated. That file is currently called `release-workflow.md`, which sounds relatively neutral, so I didn't feel the need to rename it.

> But just because something worked in 2025 doesn't mean it will work in 2027. I suppose we could also add a date stating when something worked, continually updating that would be annoying.

Hmm... I don't hate the idea of updating the workflow notes continually to keep them in sync with our actual workflow :stuck_out_tongue: As for adding a date, that could be done, although it should also be possible to extract that info using blame.

@ms609:

> Perhaps a cautious approach would be a two-step workflow: the first step preparing a draft release, with release notes, in its own branch (thus allowing for any validation and edits at that point); the second merging the release into master and marking it as the current release?

Perhaps I'm misunderstanding your suggestion, but as far as I can tell, this is exactly what we are already doing (as described in the notes added by this very PR).

@bredelings

> It might be possible to make the `release.yml` workflow specify the release notes for the draft release it creates.

Interesting! I guess that would involve parsing the `NEWS.md` file, then? Is it worth attempting to do that as part of this PR?

> We should note that, in reality, what usually happens is that I push a tag, it creates the release, but something is wrong with it. I then delete the tag locally, google how to delete a remote tag, make some git comments, retag, and push a new tag. Rinse, lather, repeat. I think for this last release, I only had the push the tag twice, which is pretty low.

Do you remember what went wrong? That process sounds like quite a headache, so if there is any info we could add to the workflow doc to minimize it going forward, that would probably be worth doing.

doc/release-workflow.md
davidcerny: You're 100% correct, I got this wrong. The extra executable _is_ automatically generated, but it still needs to be run. Fixed now.

doc/release-workflow.md
davidcerny: I agree, those instructions were overly detailed and out-of-place. I tried to make them more concise, although I'm not sure how to strike the right balance. The current version of the notes says to commit particular files, for example, but doesn't say _where_ (on which branch) to commit them...

bredelings (approved):

PR #924, opened 2026-01-02
Fix CI on mac where boost 1.90 is broken

bredelings:
This PR unbreaks the tests on Mac by forcing boost 1.89 even though homebrew has boost 1.90


davidcerny (approved):
Seems like the only viable solution until the next boost release. Approving.

PR #925, opened 2026-01-08
Update build.sh for setting boost location with modern cmake

bredelings:
So apparently build.sh wasn't setting the boost location correctly for modern cmake and modern boost.  We should be setting -DCMAKE_PREFIX_PATH.  The old -DBOOST_ROOT has no effect any more.

Mostly we didn't need to set it, but currently homebrew is pulling in boost 1.90, which is broken.  So try to get this working again.

I also removed some other stuff that doesn't work with modern cmake and modern boost, such as BOOST_INCLUDEDIR and BOOST_LIBRARYDIR

### DIFF
--- a/projects/cmake/build.sh
+++ b/projects/cmake/build.sh
@@ -34,10 +34,7 @@ debug="false"
 mpi="false"
 cmd="false"
 boost_root=""
-boost_lib=""
-boost_include=""
 boost_verbose=""
-boost_debug=""
 static_boost="false"
 j=4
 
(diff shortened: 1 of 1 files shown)


projects/cmake/build.sh
```diff
@@ -53,22 +50,19 @@ while echo $1 | grep ^- > /dev/null; do
 -ninja          <true|false>    : set to true to build with ninja instead of make
 -mpi            <true|false>    : set to true if you want to build the MPI version. Defaults to false.
 -cmd            <true|false>    : set to true if you want to build RevStudio with GTK2+. Defaults to false.
--boost_root     string          : specify directory containing Boost headers and libraries (e.g. `/usr/`). Defaults to unset.
--boost_lib      string          : specify directory containing Boost libraries. (e.g. `/usr/lib`). Defaults to unset.
--boost_include  string          : specify directory containing Boost libraries. (e.g. `/usr/include`). Defaults to unset.
--boost_verbose  <true|false>    : log some info about finding Boost
--boost_debug    <true|false>    : log MORE info about finding Boost
--static_boost	<true|false>    : link using static Boost libraries. Defaults to false.
+-boost-root     string          : specify directory containing Boost headers and libraries (e.g. `/usr/`). Defaults to unset.
```
davidcerny: Is the change from `-boost_root` to `-boost-root` intentional? It's incompatible with the example given on line 64.

davidcerny (changes requested):
This looks good, thanks! If the change from `-boost_root` to `-boost-root` is in fact intentional, we'll have to update [this tutorial](https://revbayes.github.io/compile-osx) (and possibly some others?) accordingly.

projects/cmake/build.sh
bredelings: I made it so that both work. We could change the help text back to using underscore if that simplifies things.

bredelings: I changed the help text back so we can deal with this later.

davidcerny (approved):

PR #926, opened 2026-01-09
Switch from c++17 -> c++20

bredelings:
Switch from c++17 -> c++20


bredelings:
Actually maybe we should do this before c++23

PR #932, opened 2026-01-19
Treat unrooted trees as such when generating summaries

### LINKED ISSUE #929: [BUG] Map tree calculations are effected by the order of strings.
**Describe the bug**
`mapTree()` appears to choose the most frequent exact Newick string (including child order) rather than the most frequent topology. This can yield a MAP tree that is not the most frequent topology when the same topology appears with different child orders in the posterior trace.

**To Reproduce**
A example tree trace file with 7 generations:
```
Iteration	Replicate_ID	Posterior	Likelihood	Prior	topology
0	0	-100.0	-90.0	-10.0	((A:1.0,B:1.0):1.0,(C:1.0,D:1.0):1.0);
1	0	-100.0	-90.0	-10.0	((C:1.0,D:1.0):1.0,(A:1.0,B:1.0):1.0);
2	0	-100.0	-90.0	-10.0	((A:1.0,B:1.0):1.0,(C:1.0,D:1.0):1.0);
3	0	-100.0	-90.0	-10.0	((C:1.0,D:1.0):1.0,(A:1.0,B:1.0):1.0);
4	0	-100.0	-90.0	-10.0	((A:1.0,C:1.0):1.0,(B:1.0,D:1.0):1.0);
5	0	-100.0	-90.0	-10.0	((A:1.0,C:1.0):1.0,(B:1.0,D:1.0):1.0);
6	0	-100.0	-90.0	-10.0	((A:1.0,C:1.0):1.0,(B:1.0,D:1.0):1.0);
```
The first 4 iterations show the same topology `{A,B}|{C,D}` but different string order (twice as `((A,B),(C,D))` and twice as `((C,D),(A,B))` ) and other 3 iterations show the topology like `{A,C}|{B,D}`. 

Run the command: 
```
rb -e 'treetrace = readTreeTrace("posterior.trees", treetype="non-clock", burnin=0);  mapTree(treetrace, "map_tre.nex")'
```
and I got the topology of `{B,D}|{A,C}` in `map_tre.nex`, which is the most frequent exact Newick string with count 3.

**Expected behavior**
MAP would be equivalent to {A,B}|{C,D} because this topology occurs 3 times. 

**Computer info**
MacOS Tahoe 26.1 M1
RevBayes version (1.3.2) Build from tags/v1.3.2 (3d84ac) on Thu Dec 11 13:36:01 UTC 2025

davidcerny:
This PR fixes #929. As it turns out, the problem actually comes from the fact that even when we read in unrooted trees as such (i.e., using `treetype="non-clock"`, and with the `unroot_nonclock` argument kept at its default value of `TRUE`), we do not explicitly _mark_ them as unrooted. So when we ask whether the first tree in the trace is rooted or not

https://github.com/revbayes/revbayes/blob/db9aca3168a4c71199b28d6b2738917c1de79da7/src/core/analysis/mcmc/output/TreeSummary.cpp#L1932

we find that it is. However, because unrooting did actually take place, the original `((A,B),(C,D))` Newick string got turned into `((C,D),A,B)`, while the equivalent string `((C,D),(A,B))` turned into `((A,B),C,D)`. Since we treat these as rooted trees with a trifurcation at the root, we conclude that they are different, and yield the wrong MAP tree.

Why do we unroot trees but then refuse to mark them as unrooted? @bredelings made this change in commit f9b9bf2 (part of PR #282) in response to an edge case I had found. I was getting a segfault when I tried to summarize a trace containing trees with outdegree-1 nodes (sampled ancestors) that I was deliberately treating as unrooted (`treetype="non-clock"`). Personally, I don't consider safeguarding against this edge case all that important, because non-clock trees with knuckles have no biological interpretation I can think of. However, I changed the `annotateTree()` function to throw an exception instead of causing a segfault, which in turn allowed me to safely revert Ben's commit.


src/core/analysis/mcmc/output/TreeSummary.cpp
```diff
@@ -200,6 +200,13 @@ void TreeSummary::annotateTree( Tree &tree, AnnotationReport report, bool verbos
             node_ages = tree_clade_ages[newick][split];
         }
 
+        // Verify that we have age data for this node
+        if ( node_ages.empty() )
+        {
+            std::string node_desc = n->isTip() ? ("tip '" + n->getName() + "'") : ("internal node with clade " + clade.toString());
+            throw RbException() << "No age samples found for " << node_desc << ". This may indicate a mismatch between the tree to be annotated and the tree trace.";
```
bredelings: This has also sometimes indicated a bug, where we incorrectly computed the clade object used to look up the node ages, and then can't find any.  I don't know if we need to change the message, but I thought I'd mention that.

bredelings (approved):
Putting back the code that was commented out seems like a nice fix.  The new code seems to make sense, so I haven't looked into why the code was originally crashing.

src/core/analysis/mcmc/output/TreeSummary.cpp
davidcerny: Good idea; I have removed that part of the message. Even if it were true, that information would be unlikely to help most users, I think.

davidcerny:
Just for anyone who might be interested in looking into this further, the code was originally crashing on this trace:

[test_trace.txt](https://github.com/user-attachments/files/24741037/test_trace.txt)

I tracked down the segfault to the fact that the `summarize()` function initializes the taxon bitset `b` with `tree.getNumberOfTips()`, and then passes it to the `collectTreeSample()` function that uses it as the argument for the `.getTaxa()` method. If sampled ancestors are present, the total number of taxa will exceed the number of tips, leading to invalid access problems.

So I fixed this by changing how the bitset is initialized in `summarize()`, and then immediately ran into _another_ segfault, caused by the fact that the `annotateTree()` function was trying to access an empty `node_ages` vector. I solved this by introducing the corresponding check, and throwing an exception when the check fails.

To demonstrate that the crash no longer occurs, I originally included the trace above as yet another test, showing that we now fail gracefully in this edge case (i.e., with an exception rather than a segfault). But our test runner doesn't like exceptions either, so people will have to either take my word for it, or run RB on the trace above on their own. :stuck_out_tongue:

PR #933, opened 2026-01-19
Options for calculating clade posterior probabilities 2: electric boogaloo

brpetrucci:
Second try at implementing choices for calculating clade posterior probabilities in trees with sampled ancestors. 

@bredelings started us out by renaming the `Split` and `Clade` objects to `SplitWithMRCA` and `CladeWithMRCA` to reflect their current usage (where they include information on whether the split/clade in question has an SA as an MRCA), and implementing `Split` and `Clade` objects that do not include an MRCA. This required some extra member function implementations for `Split` to access the bitset within. This then let us split the `counts`, `frequency` etc. functions for each of these objects, so we can use the `differentiate_SAs` (`differentiate_MRCAs` on the Rev language side) option to choose which function/option to use in each case. 

Ben implemented changes in `annotateTree` to handle the split (pun intended) in functionality, taking into account that we now would like to report both clade posterior probability and clade-with-mrca posterior probability for each clade (annotated as `posterior` and `posterior_mrca` respectively). 

I then implemented changes in `mccTree` and `mrTree`:
- `mccTree`: if `differentiate_SAs` is true, we calculate tree scores the same way we did before. If it is false, we add to the score the probability of each taxon being an SA (if it is an SA in the tree in question) or not (if it isn't). In this way we can still consider SA probabilities in our tree scores, while considering only taxon membership in clade probabilities.
- `mrTree`: if `differentiate_SAs` is true, we again build the tree just the same as before. If it is false, we consider clade (as opposed to clade-with-mrca) probabilities when deciding to add a clade to the tree, and then set each taxon to an SA if it is an SA in more than `cutoff` frequency in our sample.
- For `mapTree`, we did not modify anything but adding a `differentiate_SAs` argument, since differentiating SAs or not should not affect the construction of the tree (though we have an ongoing discussion about that, if @davidcerny wants to elaborate on it). We delegate the calculation of node ages to `annotateTree` (see below for TO-DO list), so we must take the argument here to pass it to that function.

Things look good (though I'd take any feedback about how to improve the current implementation), but there's still some work to do:
1. Add machinery in `annotateTree` to calculate and set node age based on `differentiate_SAs` (since the argument will change which subset of the tree sample we consider when calculating node agres). We also would like to add node parameters for both node age calculations (with and without differentiating SAs).
2. Create toy posterior samples to test the functions and ensure `mapTree`, `mccTree` and `mrTree` are working as intended.
3. Create new expected output files for the tests that are failing.


src/core/analysis/mcmc/output/TreeSummary.cpp
```diff
@@ -213,27 +250,41 @@ void TreeSummary::annotateTree( Tree &tree, AnnotationReport report, bool verbos
...
+                    age_mrca = (node_with_mrca_ages[idx_mrca-1] + node_with_mrca_ages[idx_mrca]) / 2;
                 }
 
             }
 
             // finally, we set the age/length
             if ( clock )
             {
-                n->setAge( age );
+                // What do to here??
```
bredelings: Hello from Athens! We need to select between `age` and `age_mrca` here.  I'll look at this more when I get back from Egypt.

bredelings (changes requested):

bredelings:
Thanks for working on this!

There seems to be a SEGFAULT in `summarizeTreeTrace/scripts/mr-consensus-sa.Rev`.

bredelings:
TODO from Bruno:
*    Add machinery to use differentiate_SAs to figure out which node ages to set on annotateTree, and add both (or just the other?) node ages as node parameters
*    Create fake posterior samples to test and ensure both options are working as intended for mcc, mr, and map trees
*    Create new expected output files for tests that are failing

davidcerny:
Just a quick update regarding the to-dos:

> Add machinery to use differentiate_SAs to figure out which node ages to set on annotateTree, and add both (or just the other?) node ages as node parameters

IIRC, Ben established that while we can do the former ("use `differentiate_SAs` to figure out which node ages to set on `annotateTree`"), it's not possible to do the latter, since we might end up with no samples to calculate mean ages from. This is different from annotating the tree with both clade probabilities and clade-with-MRCA probabilities: in this case, if we find no samples containing a given branch, we can simply set its PP to 0.

That said, I wonder if the situation is a little more complicated, or perhaps if I'm misunderstanding the argument. It seems to me that the two cases are not really symmetric. If we are building a summary tree using clade-with-MRCAs, we should always be able to find corresponding clade samples. For example, if a given clade-with-MRCA is present in exactly 20 samples, the clade of the same content will also be present in _at least_ 20 samples. But I suspect there's something I'm not seeing here.

> Create fake posterior samples to test and ensure both options are working as intended for mcc, mr, and map trees

Since the issue this PR is intended to solve has resurfaced a number of times (#467, #787), we can maybe solicit test cases from the users who have been affected by it...

> Create new expected output files for tests that are failing

This has now been done, although we'll have to regenerate the expected output once more to account for the changes made in #1006.

I'm wondering if it'd make sense to add fixing issue #943 to the to-do list – I can confirm it affects this branch as well. It's not strictly related to the split / split-with-MRCA distinction, but it _is_ another instance of the broader problems of handling sampled ancestors in MAP trees, and guarding against the possibility of finding no samples satisfying a given criterion.

PR #935, opened 2026-01-19
Use temp checkpoint file and atomically rename in mcmc. 

### LINKED ISSUE #930: [BUG] checkpoint file overwriting may cause lost
**Describe the bug**
RevBayes currently writes checkpoint files by overwriting existing files directly. If an unexpected exit occurs during writing (e.g. job preemption or node failure), the previous checkpoint can be lost or replaced by a partially written file.

Would it be possible to improve robustness by writing to a temporary file and atomically renaming it on success (or keeping a backup of the previous checkpoint)?

curiosusJR:
This change makes RevBayes checkpointing more robust against unexpected exits by writing to hidden temporary files and then performing an atomic replace. The main `Mcmc::checkpoint()` now writes to `.<checkpoint>.tmp` (and the associated _mcmc and _moves files) before replacing the original, so a partial write can’t corrupt the prior checkpoint. On Unix this uses `std::filesystem::rename` (atomic on the same filesystem). On Windows it uses `MoveFileExW` with `MOVEFILE_REPLACE_EXISTING | MOVEFILE_WRITE_THROUGH` to preserve atomic replace semantics.

This may close #930.

### DIFF
--- a/src/core/analysis/mcmc/Mcmc.cpp
+++ b/src/core/analysis/mcmc/Mcmc.cpp
@@ -31,6 +31,7 @@
 #include "RbVector.h"
 #include "RbVectorImpl.h"
 #include "RbSettings.h" // for logMCMC setting
+#include "RlUserInterface.h"
 #include "StringUtilities.h"
 
 #ifdef RB_MPI
@@ -264,6 +265,14 @@ void Mcmc::checkpoint( void ) const
         
         // clean up
         out_stream.close();
+        const bool ok = out_stream.good();
+        if ( !ok )
+        {
+            RBOUT( "Warning: failed to write checkpoint file \"" + checkpoint_file_name.string() + "\"; keeping existing file." );
+            std::error_code ec;
+            std::filesystem::remove(tmp_checkpoint_file_name, ec);
+        }
+        else
 #ifdef _WIN32
         if ( MoveFileExW(tmp_checkpoint_file_name.wstring().c_str(), checkpoint_file_name.wstring().c_str(), MOVEFILE_REPLACE_EXISTING | MOVEFILE_WRITE_THROUGH) == 0 )
         {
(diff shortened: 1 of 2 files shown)


davidcerny:
I haven't been able to verify that this works as intended. I changed the [`Test_heat_tuning_and_checkpointing.Rev`](https://github.com/revbayes/revbayes/blob/master/tests/test_MC3/scripts/Test_heat_tuning_and_checkpointing.Rev) script to print checkpoint files every 100 rather than 1000 generations, and used the following bash script to monitor the directory into which they were being written:

```sh
while true
do
    ls -a
    sleep 0.1
done
```

I never saw the temporary back-up files the script was supposed to generate. I tried using one of my previous scripts that employs checkpointing (with regular `mcmc` as opposed to `mcmcmc`), and the results were the same – I could not verify any temporary files were being written.

Of course, it is possible the entire action of writing a temp file and performing an atomic replacement was completed in less than 0.1 seconds, i.e., between two successive checks of the directory performed by my bash script. But if that's the case, then the problem which this PR intends to solve is very unlikely to arise in the first place – you'd need to be _really_ unlucky for an unexpected exit to just so happen during that extremely short period of time which it takes to overwrite the previous checkpoint file.

My final remark is that if this solution in fact (1) works and (2) is worthwhile implementing, it should be extended to `mcmcmc` as well. MC^3 delegates most of its checkpointing to `Mcmc::checkpoint()`, but it also produces its own `_heat` checkpoint files that might benefit from an analogous change.

curiosusJR:
It doesn't seems to be a quota limit issue that I thought, at least not in the APFS filesystem in my test. Different file systems seem to handle file writability in different ways.... 

Anyway, indeed there is not much content need to be writen in checkpoint files so the visialbe window of empty files is quite short but it still exists and depends on OS, filesystem, disk type, scheduling, etc. 

I made a tiny test for catching the empty files and another for proving the temp file exist. 
[test.tar.gz](https://github.com/user-attachments/files/24780292/test.tar.gz)

`empty_files.sh` is trying to capture empty files with 
```
while true; do
  sleep 0.1
  find output -type f -empty
done
```
in total 20s. I tested this on 3 device many times and got 5~40 output lines (it is unstable on MacOS). 
There is a comment line in the script could change it into modified version `rb` and then the output lines become almost zero, only 1 or 2 temp files may be outputed. 

If there is no temp file outputed from previous test, following test can prove temp files are used. 

Since `std::ofstream()` needs write permission and `std::filesystem::rename()` doesn't, I make those checkpoint files unwritable and if they are edited directly, the content would not be changed, and if they are renamed from temp files, their permission should be changed. These could be observed by running `temp_exist.sh`. (Also switch rb version by editing the comment line)

I do make a big `nruns` value and super frequently checkpoint for a higher possibility of empty checkpoint file capture and everytime I catched successfully. It could be worse on mpi version because of higher IO pressure, i think.

davidcerny (approved):
Many thanks for these scripts, @curiosusJR! These are cleverly designed, and successfully demonstrate that the PR does what it's supposed to. Approving!

PR #937, opened 2026-01-23
Fix `rb-mpi-help2yml`

### LINKED ISSUE #898: [BUG] The `rb-mpi-help2yml` executable doesn't work
**Describe the bug**
The [cmake build script](https://github.com/revbayes/revbayes/blame/master/projects/cmake/build.sh) automatically generates not only the main RevBayes executable (called `rb`, or `rb-mpi` for the MPI version), but also a secondary executable that translates the Markdown help files stored in `help/md` to YAML and updates `RbHelpDatabase.cpp`. (The [meson build script](https://github.com/revbayes/revbayes/blob/master/projects/meson/build.sh) still requires setting a flag to achieve the same outcome, which may have been an oversight in #659.) This secondary executable is called `rb-help2yml`, or `rb-mpi-help2yml` for the MPI version.

While the former works fine and does what it's supposed to, the latter immediately crashes with the following error message:

```
*** The MPI_Comm_rank() function was called before MPI_INIT was invoked.
*** This is disallowed by the MPI standard.
*** Your MPI job will now abort.
[MacBook-Pro-3.local:42590] Local abort before MPI_INIT completed completed successfully, but am not able to aggregate error messages, and not able to guarantee that all other processes were killed!
--------------------------------------------------------------------------
Primary job  terminated normally, but 1 process returned
a non-zero exit code. Per user-direction, the job has been aborted.
--------------------------------------------------------------------------
--------------------------------------------------------------------------
mpirun detected that one or more processes exited with non-zero status, thus causing
the job to be terminated. The first process to do so was:

  Process name: [[27593,1],0]
  Exit code:    1
--------------------------------------------------------------------------
```

**To reproduce**
```
git clone https://github.com/revbayes/revbayes.git
cd revbayes/projects/cmake
./build.sh -mpi true
mpirun -np 1 ./rb-mpi-help2yml
```

**Expected behavior**
`rb-mpi-help2yml` should run and produce the same results as `rb-help2yml`.

**Computer info**
macOS 26.0.1 Tahoe (ARM64 architecture), cmake 4.0.2, OpenMPI 4.1.6, RevBayes 1.4.0-preview, commit db0bf01.

davidcerny:
Fixes #898.

### DIFF
--- a/src/help2yml/main.cpp
+++ b/src/help2yml/main.cpp
@@ -11,10 +11,21 @@
 #include "YAMLHelpRenderer.h"
 #include "Workspace.h"
 
+#ifdef RB_MPI
+#include <mpi.h>
+#endif
+
 using namespace RevBayesCore;
 
 int main(int argc, const char * argv[])
 {
+#ifdef RB_MPI
+    // We have const char* argv[], but MPI_Init() requires non-const argv
+    int mpi_argc = argc;
+    char** mpi_argv = const_cast<char**>(argv);
+    MPI_Init(&mpi_argc, &mpi_argv);
+#endif
+
     path file = "help.yml";
 
     if( argc > 1 )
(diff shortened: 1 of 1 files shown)


basantakhakurel (approved):
Can confirm it fixed the issue.

PR #938, opened 2026-01-26
Correct reasoning about return types for numeric vector functions

### LINKED ISSUE #793: [BUG] max returns type Real even if all entries of vector are RealPos
**Describe the bug**
When selecting a maximum taxon age via looping over individual taxa, the resulting variable is of type `Real` instead of the expected `RealPos`, although all entries of the vector are `Real`. This leads to downstream issues, where I define exponential priors as the rootAge for an FBD analysis. Because the offset (maximum fossil age) is `Real`, the prior is Real<stochastic> and throws an error (expects RealPos<any>)

**To Reproduce**
[fossils_cont_rho1_1.csv](https://github.com/user-attachments/files/21381403/fossils_cont_rho1_1.csv)
```
# see attached file
taxa <- readTaxonData("fossils_cont_rho1_1.tsv")
for (i in 1:taxa.size()){
    ages[i] = taxa[i].getMaxAge()
    type(ages[i]) # always RealPos
}
type(ages) # is Real

[fossils_cont_rho1_1.csv](https://github.com/user-attachments/files/21381355/fossils_cont_rho1_1.csv)

max_age <- max(ages)
type(max_age) # is Real
```

**Expected behavior**
max should return RealPos if its value is positive

**Computer info**
Running on WSL2 (Ubuntu 24.04.2 LTS), revBayes 1.3.0

**Additional context**
Potentially related: [#585 ](https://github.com/revbayes/revbayes/issues/585)

davidcerny:
This PR fixes #793. It does so in a way suggested by @bredelings, i.e., by converting the relevant functions (`max`, `min`, `sum`, `median`) into templates analogous to [`Func_round.h`](https://github.com/revbayes/revbayes/blob/master/src/revlanguage/functions/math/Func_round.h), with the difference that these operate on vectors rather than atomic values.

For `min` and `max`, this was further complicated by the fact that the functions are supposed to work not just for vectors of arbitrary numeric types, but also for real-valued matrices, so there is some overloading going on in addition to templating. I refactored `sum` to work in the same way, which (1) greatly simplified the backend (allowing me to remove a number of files such as `Func_sumNatural.[h|cpp]`, `Func_sumPositive.[h|cpp]`, etc.) and (2) extended the function to matrices. `median` was refactored as well in a slightly different way, since the matching of input and output types is a little more sophisticated.

This passes all existing tests, and I've added a new test as well showcasing how the refactored functions work on different numeric types. However, I know that things having to do with types and type matching have been a huge pain in the past, so I would be grateful if someone could take a look and make sure all this makes sense to them.


src/revlanguage/functions/math/Func_median.h
```diff
@@ -11,31 +10,174 @@ namespace RevLanguage {
...
+
+namespace RevBayesCore {
+    // Compute median as double first to preserve fractional results, then convert to ReturnType
+    template <typename ReturnType, typename InputType>
+    ReturnType* median_value(const RbVector<InputType>& v)
+    {
+        double m = 0.0;
+        
+        // get a copy of the original vector
+        RbVector<InputType> w = v;
```
bjoelle: have you checked that this is a true copy ? I'm always worried when it comes to vectors

src/core/functions/builtin/SumFunction.h
```diff
@@ -5,33 +5,33 @@
...
 template <class valueType> class RbVector;
 template <class valueType> class TypedDagNode;
     
     /**
-     * @brief Function for computation of the sum of a vector of  real numbers.
-     *
-     * This class is the function that computes the sum of a vector of numbers.
-     * The numbers are passed in as a DAG node whose value type is a std::vector<double>.
-     *
+     * @brief Declaration of the deterministic variable for minimum.
```
bjoelle: typo minimum -> sum

src/revlanguage/functions/math/Func_max.h
```diff
@@ -1,40 +1,150 @@
...
+#include "Integer.h"
+#include "IntegerPos.h"
+#include "ModelVector.h"
+#include "Natural.h"
+#include "RealPos.h"
+#include "RlDeterministicNode.h"
+#include "TypedDagNode.h"
+
+
+namespace RevBayesCore {
```
bjoelle, line 45: it's not super clear to me why we have to redefine the max calculation for the vector version, while the matrix version is simply handed over to the Core function ?
same for the others actually - if there is a reason it may be good to document it for the future

src/revlanguage/functions/math/Func_median.h
davidcerny: Great point. I changed that line to explicitly call the copy constructor, just for clarity's sake, and I convinced myself that it performs a deep copy rather than a shallow copy. Bear with me:

- The `RbVector` copy constructor calls the `RbVectorImpl` copy constructor
- The latter iterates over the elements of the original vector and uses `this->push_back` (where `this` is a `std::vector<valueType>`) to add them one by one to the new vector
- `median()` is allowed to work on the following derived types: `Integer`, `IntegerPos`, `Natural`, `Probability`, `Real`, `RealPos`. We can reduce these just to `Integer` and `Real`: `IntegerPos` inherits from `Natural` and `Natural` inherits from `Integer`, while `Probability` inherits from `RealPos` and `RealPos` inherits from `Real`
- `Integer` inherits from `ModelObject<std::int64_t>`, so `Integer::valueType` is just `std::int64_t`, and `Real` inherits from `ModelObject<double>`, so `Real::valueType` is just `double`
- For primitive types like `std::int64_t` and `double`, `std::vector<>::push_back()` performs a deep copy

Let me know if this reasoning doesn't sound right.

I should also emphasize that the line in question was adapted from our current median function:

https://github.com/revbayes/revbayes/blob/db9aca3168a4c71199b28d6b2738917c1de79da7/src/core/functions/builtin/MedianFunction.cpp#L55

So the new code shouldn't mess up anything that the old code got right.

src/core/functions/builtin/SumFunction.h
davidcerny: Fixed

src/revlanguage/functions/math/Func_max.h
davidcerny: I knew how to make a function template to operate on vectors with elements of different types, and I knew how to create a function that operates on both `Real`-valued matrices and `Real`-valued vectors, but I couldn't figure out how to make a single template handle all of those inputs. 

Since our current `max` function already handles both `Real`-valued matrices and `Real`-valued vectors by setting a `matrix` argument and passing it to the Core function, which then decides what it should do depending on that argument's value, it seemed to me that the path of least resistance was to keep this system in place and only add a function template to deal with vectors whose elements are of a type other than `Real`.

I tried to document this with these lines:

https://github.com/revbayes/revbayes/blob/30351aa541cfdfb8807c609d251f5bc7a86d3885/src/revlanguage/functions/math/Func_max.h#L11

https://github.com/revbayes/revbayes/blob/30351aa541cfdfb8807c609d251f5bc7a86d3885/src/revlanguage/functions/math/Func_maxMatrix.h#L12-L13

But maybe that's not clear enough.

Let me know what the best solution is – document these decisions in more detail, remove the Core function, etc. Or if you can think of a way to unify the `ModelVector` and `MatrixReal` versions of the function, please let me know as well!

bjoelle: so I think part of my confusion was that there was only one version for the median function, but of course median for a matrix doesn't necessarily make sense duh
I can't think of a way to unify both aside from unrolling the matrix and passing it to the vector function, and that seems to complicate things unnecessarily
I do think it would be clearer to have the same structure for both versions, i.e. either both entirely in the Rev or both with a Core function, and maybe this way we could also allow for matrices of other types in case we ever need that
or we can leave it like that if you don't have more time to spend on this, fine by me :)

davidcerny: Thank you; I sort of ended up going for something halfway between your two recommendations. :) I removed the Core functions for `max`, `min`, and `sum`. These are now defined entirely in Rev-level header files, and therefore follow the same structure as `round` and `median`. Hopefully this makes for clearer and more consistent code. It also allows us to be terser, as we now only need two files (e.g., `Func_max.h` and `Func_maxMatrix.h`) in place of four (`MaxFunction.cpp`, `MaxFunction.h` in the Core and `Func_max.cpp`, `Func_max.h` in Rev).

Another advantage is that the "matrix" versions of the `max`, `min`, and `sum` functions really only handle matrices now. There is no need to decide if we are dealing with a vector or a matrix anymore, like the original Core functions did, since all vectors (including vectors of `Real` elements) are instead handled by the templated, non-matrix counterparts of these functions.

However, I didn't try to turn the matrix versions into templates, allowing for proper reasoning about the return type when the functions are called on matrices of types other than `MatrixReal`. Right now, we only have one such matrix type (`MatrixRealPos`) and its use is extremely limited (I think only the weird quadrature discretization functions use it?), so it doesn't seem to be super urgent. Maybe it can be left for a future PR...

bjoelle (approved):

PR #939, opened 2026-01-28
PseudoDataLikelihood

bredelings:
### The problem(s)

1. RevBayes currently adds uniform bounds on fossils in a somewhat hacky and unreadable way.  For example, in `tutorials/fbd/scripts/model_FBDP.Rev` we have:

```
    F[i] ~ dnUniform(t[i] - b_i, t[i] - a_i)
    F[i].clamp( 0 )
```
By enforcing $`0 \in (t_i-b_i, t_i-a_i)`$, we end up enforcing that $`t_i \in (a_i,b_i)`$.  It might seem simpler to simply write
```
   t[i] ~ dnUniform(a[i],b[i])
```
but that isn't possible because $`t_i`$ already has a prior -- we can't add a second prior!  So we have to write really ugly model code.

2. Constraints like `t[i] < b[i]` or `t[i] > a[i]` cannot be enforced as distributions, because the corresponding distributions cannot be normalized.

3. In some cases `a[i]` and `b[i]` could be variable themselves.  For example, if a fossil is between two date-able layers, but the age of the layers is unknown.  In such a case, using `dnUniform( )` incorrectly yields a higher probability if `a[i]` and `b[i]` are closer together.  This is because `dnUniform( )` is normalized so that the integral on `t` is 1, which is incorrect here.

### The solution -- pseudo-data

What's really happening here is that we have some unspecified fossil data `F[i]`.  If we knew the true distribution `Dist` of the fossil data, and the observed value `f` of the fossil data, then we could write:

```
   F[i] ~ dnDist(t[i], a[i], b[i])
   F[i].clamp(f)
```
This adds the likelihood term $`\Pr_{Dist}(F_i=f|t_i, a_i, b_i)`$ to the probability.  In our hypothetical example, this term would be 1 if $`t_i \in (a_i, b_i)`$ and 0 otherwise.  

However, sometimes we don't know $`f`$ or `Dist`, but we do know $`\Pr(F_i | t_i, a_i, b_i)`$ as a function of $`t_i`$, $`a_i`$, $`b_i`$.  This isn't a prior on $`t_i`$ because $`t_i`$ is on the _right_ side of the conditional, but it does give information about $`t_i`$!

We propose to express this kind of thing by writing:
```
   F[i] ~ dnPseudo(pdBetween(t[i], a[i], b[i]))                  # there's some fossil data that gives information about t[i]
   F[i].clamp( pseudoObservation() )                               # this fossil data imposes a likelihood function on t[i] that is 1 between a[i] and b[i] and 0 otherwise
```

Interestingly, `a[i]` and `b[i]` can be stochastic or deterministic nodes.  So, for example, we can enforce that $`t_i \in (a_i, b_i)`$ even when `a[i]` and `b[i]` are changing! 

### Not distributions
Unlike distributions, pseudodata objects are _not_ normalized to sum to 1.  For example, `pdAbove(t, a)` which returns 1 if $`t > a`$ and 0 otherwise.  Additionally, `pdNormal(x,m,s)` has a maximum of 1.  `pdLogNormal(x,m,s)` has a peak at m, whereas dnLogNormal(m,s) has a peak density _below_ m.

### Functionality

* `dnPseudo( PseudoDataLikelihood )` is used to specify likelihoods as a function of `parameter`
* `pdAbove(x,a)`: 1 for x >=  a, otherwise 0.  Has an optional exponential decay for a soft bound.
* `pdBelow(x,b)`: 1 for x <= b, otherwise 0.  Has an optional exponential decay for a soft bound.
* `pdBetween(x,a,b)`: 1 for x in [a,b], 0 otherwise. Has an optional exponential decay for a soft bound.
* `pdRequire( <boolean expression> )` requires that a boolean statement be true.  By default, the false condition has a likelihood of zero, but by specifying `weight=w` you can set the odds ratio between true and false to `exp(w)`
* `pdNormal(x,m,s)`: Likelihood $`e^{-(x-m)^2/2s^2}`$.
* `pdExponential(x,lambda,shift)`: Likelihood 0 if x < shift, otherwise $`e^{-\lambda (x-shift)}`$
* `pdLogNormal(x,m,s,shift)`: Likelihood 0 if x < shift, otherwise $`e^{-(\log(x-shift)-m)^2/2s^2}`$
* `pdConstant(x)`: always 1.
* `pd1 && pd2`: multiplies the likelihoods.  For example, `pdAbove(x,a) && pdBelow(x,b)` == `pdBeween(x,a,b)`.
* `pd`: print the log of the PseudoDataLikelihood `pd`

It will be possible _eventually_ to require that a tree contains a clade by writing:
```
fossil ~ dnPseudo( pdContainsClade(tree, clade))
fossil.clamp( pseudoObservation() )
```
but this will be implemented later.  It may require the LogDensity branch to be completed in order to solve the constraint.  This PR lays the groundwork though.


bredelings:
This still needs documentation for new functions besides dnPseudo, but is otherwise very close to ready.  So if there are problems with the approach it would be good to get feedback now.

I've made some modifications to the tutorials on the `pseudodata` branch.  

In addition to the added tests, I've checked that the modified tutorials run.

acapomorphic:
Hi Ben,
Could you clarify what `pseudoObservation()` represents here?

```
   F[i] ~ dnPseudo(pdBetween(t[i], a[i], b[i]))                  # there's some fossil data that gives information about t[i]
   F[i].clamp( pseudoObservation() )                               # this fossil data imposes a likelihood function on t[i] that is 1 between a[i] and b[i] and 0 otherwise
```

bredelings:
@acapomorphic 

> Could you clarify what pseudoObservation() represents here?
>
>   F[i] ~ dnPseudo(pdBetween(t[i], a[i], b[i]))                  # there's some fossil data that gives information about t[i]
 >  F[i].clamp( pseudoObservation() )                               # this fossil data imposes a likelihood function on t[i] that is 1 between a[i] and b[i] and 0 otherwise

It represents whatever you observe when you discover a fossil. It could include the stratigraphic location of the fossil, etc.

To be more detailed, the code block introduces a likelihood term $\Pr_D(F_i=f_i|t_i, a_i, b_i)$.  
When we represent the fossil as pseudodata, both the distribution $D$ and the observed value $f_i$ are unspecified:
- dnPseudo takes the place of the unspecified distribution $D$
- pseudoObservation() takes the place of the unspecified observed value $f_i$

Does that answer your question?

help/md/pdAbove.md
```diff
@@ -0,0 +1,47 @@
+## name
+pdAbove
+
+## title
+PseudoDataLikelihood Above
+
+## description
+Pseudodata that x is above lower
+
```
davidcerny, line 9: This is very nitpicky, but maybe write "above a lower bound", or use back ticks around "lower" to make it clear you are referring to a specific argument.

help/md/pdBelow.md
```diff
@@ -0,0 +1,46 @@
+## name
+pdBelow
+
+## title
+PseudoDataLikelihood Below
+
+## description
+Pseudodata that x is below upper
+
```
davidcerny, line 9: See my comment on `pdAbove`

help/md/pdBetween.md
```diff
@@ -0,0 +1,41 @@
+## name
+pdBetween
+
+## title
+PseudoDataLikelihood Between
+
+## description
+PseudoData that x is between lower and upper
+
```
davidcerny, line 9: See my comment on `pdAbove`

help/md/pdConstant.md
```diff
@@ -0,0 +1,25 @@
...
+
+## title
+Constant PseudoDataLikelihood
+
+## description
+Pseudodata that gives no information.
+
+## details
+This represents a pseudodata likelihood that is always 1.
+
```
davidcerny, line 12: This section could definitely benefit from a brief statement on when / why the user might want to use this form of pseudodata. It cannot possibly make any difference to the inference, right? So why would we choose to include it in the first place?

help/md/pseudoObservation.md
```diff
@@ -0,0 +1,41 @@
...
+write:
+
+    F ~ dnPseudo( pdAbove(t, age_min, age_max) )
+    F.clamp( pseudoObservation() )
+
+Here pseudoObservation() takes the place of the unspecified
+value f.  This is similar to how dnPseudo takes the place
+of the unspecified distribution for F.
+
+See the help for dnPseudo for more explanation.
```
davidcerny: This is the one part of the PR that I still find conceptually unclear. While the basic idea makes sense to me, I don't understand why `pseudoObservation()` should be a function. What is it a function _of_?

Aside from the conceptual difficulties, interface-wise this also seems to be incongruent with how the rest of RevBayes works. We do have a few other functions that don't take arguments (`getwd()`, `listOptions()`, `VectorMonitors()`, `VectorMoves()`), but they seem to be all related to the workspace rather than the model.

It'd make more sense to me to think of the pseudo-observation as an object, and of `pseudoObservation()` as a function for constructing that object. That's the relationship between the monitor vector and the `VectorMonitors()` function, after all, and in the back end, we do have an object to go with the function. But that's not how the help file puts it – it says (in somewhat awkward language) that it is the function itself that represents the observed value of a draw from `dnPseudo`. That doesn't sound right.

Another conceptual difficulty has to do with the question of whether pseudo-observations differ between different analyses. In `PseudoObservation.h`, we define a comparison between different objects of that type to always return `false`, which implies that every instance is unique. (At the Rev level, in contrast, such a comparison cannot be carried out at all.) But the object has no member variables that could be set to different values between different instances thereof, so I wonder if it would be more appropriate to treat the pseudo-observation as a constant. But then again, we are not really big on constants in RevBayes, either. Hmm... this is hard.

In many ways, I actually prefer your old suggestion from Slack:

```
f ~ dnPseudo(t[i])
f.clamp( pdBetween(a[i], b[i]) )
```

to the current interface:

```
f ~ dnPseudo( pdBetween(t[i], a[i], b[i]) )
f.clamp( pseudoObservation() )
```

src/revlanguage/functions/pseudodata/Func_PseudoDataAnd.cpp
```diff
@@ -0,0 +1,100 @@
...
+/* Get argument rules */
+const ArgumentRules& Func_PseudoDataLikelihoodAnd::getArgumentRules( void ) const
+{
+    static ArgumentRules argumentRules = ArgumentRules();
+    static bool          rules_set = false;
+
+    if ( !rules_set )
+    {
+        argumentRules.push_back( new ArgumentRule( "left", PseudoDataLikelihood::getClassTypeSpec(), "The left argument.", ArgumentRule::BY_CONSTANT_REFERENCE, ArgumentRule::ANY ) );
+        argumentRules.push_back( new ArgumentRule( "right", PseudoDataLikelihood::getClassTypeSpec(), "The right argument.", ArgumentRule::BY_CONSTANT_REFERENCE, ArgumentRule::ANY ) );
```
davidcerny, line 57: These argument descriptions are not very... descriptive. Maybe "left / right pseudodata object" or just "condition 1" and "condition 2" would work better?

src/revlanguage/functions/pseudodata/Func_PseudoDataLogNormal.cpp
```diff
@@ -0,0 +1,144 @@
...
+        L->log() = -std::numeric_limits<double>::infinity();
+        return L;
+    }
+    x = log(x);
+
+    // Normal
+    double z = (x-mu)/sigma;
+    L->log() = -z*z/2;
+
+    // NOTE that we do NOT include the 1/(x*sigma) term that is include in log-normal DENSITIES.
```
davidcerny, line 44: "is included", not "is include"

src/revlanguage/functions/pseudodata/Func_PseudoDataNormal.cpp
```diff
@@ -0,0 +1,124 @@
...
+#include "TypeSpec.h"
+
+using std::vector;
+using std::unique_ptr;
+
+namespace Core = RevBayesCore;
+
+// NOTE: This is intentionally NOT normalized to sum to 1 with integrating over x.
+//       Normalizing is wrong -- it is not a distribution on x, but on the pseudo-observation.
+//       We need to avoid making the likelihood at x=mu increase as with sigma decreases,
```
davidcerny, line 26: There seems to be an extra "with" in this sentence

src/revlanguage/functions/pseudodata/Func_PseudoObservation.h
```diff
@@ -0,0 +1,52 @@
...
+#include "TypedDagNode.h"
+#include "TypedFunction.h"
+
+namespace RevLanguage {
+class ArgumentRules;
+class TypeSpec;
+
+    /**
+     * The RevLanguage wrapper of Interval pseudo-data.
+     *
```
davidcerny, line 24: Doesn't seem right

tests/test_pseudo/scripts/and.Rev
```diff
@@ -0,0 +1,16 @@
+seed(1)
+
+t ~ dnNormal(0,10)
+
+a ~ dnNormal(-1,1)
+b ~ dnExponential(3) + a
+
+fossil ~ dnPseudo( pdAbove(t, a) && pdBelow(t, b) )
```
davidcerny, line 8: It would be neat to combine this with one of the interval tests and include some code for comparing the resulting traces, just to verify that the two approaches do indeed yield identical posteriors...

davidcerny (commented):
Overall, this PR extends RevBayes functionality in a very clean and well-documented way, which is great. When it comes to the specific use case of fossil calibrations, which helped motivate it, it perhaps doesn't improve things as much as I hoped in terms of flexibility and readability – we still rely on the difficult-to-interpret auxiliary variables `F[i]`, and I see no obvious way of making one fossil's age a deterministic function of another's – but it still represents a step forward, and I'm excited for the kinds of things this framework might enable when combined with the functionality that's being developed on the [LogDensity](https://github.com/revbayes/revbayes/tree/LogDensity) branch.

As detailed below, the only major change I think is still worth considering at this point is getting rid of `pseudoObservation()`.

I would perhaps also appreciate a brief description of the differences between this implementation of pseudodata and the one underlying the now closed PR #745. I see that `pdRequire()` has become a function rather than a distribution, in line with the other pseudodata likelihoods, and that it gained a `weight` argument that allows it to subsume the formerly separate `pdWeight()` function. That definitely looks like a simpler, cleaner solution. My guess is that `pdAdd`, `pdSub1`, and `pdSub2` were replaced by the `shift` argument of `pdExponential` and `pdLogNormal`, which also makes sense, as those are (currently) the only two cases where the notion of shifting the base likelihood function actually applies. But it's not clear to me why we lost the ability to combine pseudodata likelihoods using `or` as well as `and`, and what was gained by introducing the always-redundant `pdConstant`.

davidcerny:
By the way, I think a whole bunch of `.errout` files is missing from `test_pseudo/output_expected`.

tests/test_pseudo/scripts/interval2.Rev
```diff
@@ -1,4 +1,4 @@
-seed(1)
+seed(152)
```
davidcerny, line 2: Isn't the intermittent failure under some random seeds a sign of a deeper problem?

help/md/pseudoObservation.md
bredelings: > In many ways, I actually prefer your old suggestion from Slack:

I kind of like the old interface too.  However, it turned out to cause problems computing the right likelihoods.  First, RevBayes has no infrastructure for updating the value of clamped objects when their parameters change.  Second, the parameters `t[i]`, `a[i]`, and `b[i]` are really parameters of the unspecified distribution.  So in the end putting the parameters into the distribution was necessary to get the pseudolikelihoods to update when `x` or `a[i]` or `b[i]` changed.

> It'd make more sense to me to think of the pseudo-observation as an object, and of pseudoObservation() as a function for constructing that object.

Yeah, that's correct.  When I write `pseudoObservation()` I'm not referring to the function itself, but to the result of calling the function.

It sounds like the documentation of pseudoObservation is confusing.  But how about the documentation of dnPseudo?  If the documentation for dnPseudo is just as confusing, then we have one problem.  But if the documentation for dnPseudo is clear, and the the documentation for pseudoObservation is confusing, then I think we have a different problem.

bredelings: I do still wonder if we should replace "pseudo" with "unspecified".  As in 

    F ~ dnUnspecified( pdBetween(t[i], a[i], b[i]) )
    f.clamp( unspecifiedObservation() )

If I have to keep saying over and over in the documentation that dnPseudo represents an unspecified distribution, and pseudoObservation() represents an unspecified observation, then maybe this indicates that "Pseudo" isn't really conveying helpful information.

Along those same likes, we could rename `pdBetween(...)` to `lkBetween(...)` or `plkBetween(...)` or `plBetween(...)`.

bredelings: > Another conceptual difficulty has to do with the question of whether pseudo-observations differ between different analyses. In PseudoObservation.h, we define a comparison between different objects of that type to always return false, which implies that every instance is unique. (At the Rev level, in contrast, such a comparison cannot be carried out at all.) But the object has no member variables that could be set to different values between different instances thereof, so I wonder if it would be more appropriate to treat the pseudo-observation as a constant. But then again, we are not really big on constants in RevBayes, either. Hmm... this is hard.

I would be somewhat happier to simply not define these operators at all, but if I don't define them then some other code won't compile.

I don't think this matters really, but I changed the code to be more like a standard "singleton" type, where there is a single value that inhabits the type and equality always returns true.  I used some fancy new C++23 stuff so that (a) we don't need to define `operator !=` and we can define all comparison operators in terms of the new "spaceship operator" `operator <=>`.

help/md/pdConstant.md
bredelings: I don't think of it that way.  This is one of the possible shapes of the likelihood function -- completely flat.  Its kind of like the number "1", which has no effect when you multiply by it.  But nobody has ever proposed eliminating it for that reason.

bredelings:
> I would perhaps also appreciate a brief description of the differences between this implementation of pseudodata and the one underlying the now closed PR https://github.com/revbayes/revbayes/pull/745. I see that pdRequire() has become a function rather than a distribution, in line with the other pseudodata likelihoods, and that it gained a weight argument that allows it to subsume the formerly separate pdWeight() function. That definitely looks like a simpler, cleaner solution. My guess is that pdAdd, pdSub1, and pdSub2 were replaced by the shift argument of pdExponential and pdLogNormal, which also makes sense, as those are (currently) the only two cases where the notion of shifting the base likelihood function actually applies. But it's not clear to me why we lost the ability to combine pseudodata likelihoods using or as well as and, and what was gained by introducing the always-redundant pdConstant.

The main difference is that the old implementation had `dnPseudo( t, pdBetween(a,b) )` whereas the new implementation has `dnPseudo( pdBetween(t,a,b) )`.  The old implementation had a problem that `pdBetween(a,b)` had the type "pseudo-data on Real" but a function `pdContainsClade(C)` would be of a different type "pseudo-data on trees".  This in turn caused a lot of other problems like needing to have different functions pdConstantOnReal(), pdConstantOnTrees(), pdConstantOnFOO(), ...  etc.  This PR switches to have a single new type PseudoDataLikelihood does _not_ depend on the data type we are applying the function to.  In addition, there's no mathematical reason to split out the "t" from `pdBetween(t,a,b)`.

However, with this new argument structure, its no longer clear how to use pdAdd, pdSub1 and pdSub2 to shift the first argument of the pseudodatalikelihood functions like pdBetween.  We can write `x |> pdBetween(a,b)` in revbayes, so we could possibly do something like `x |> shift(3) |> pdBetween(a,b)`.  However, what I'd really like to do is have something like `x |> pdLog() |> pdNormal(mu,sigma)` return `-Inf` if `x<0`, and this doesn't work.  In some languages we could put `x` last  so that `pdAbove(a,b)` would be a function of `x`.  Then we could do `x |> pdLog(pdAbove(a,b))`.  But RevBayes isn't that advanced.

I think the new `pdRequire( )` is simpler in a number of ways, as you note.  Now `dnPseudo( )` is the _only_ way of introducing pseudolikelihoods.  This is really independent of the other changes.

Previously, we had `pdWeightReal(...)`.  However, this actually violated the normalization constraint of pseudodatalikelihoods, which is that the maximum is always 1.  The "or" functionality allowed approximating mixture distributions, but didn't really do it correctly.  A correct mixture distribution might look something like this:

    F ~ dnPseudo( pdMixture([0.9,  0.09,  0.01], [pd1,  pd2,  pd3]) )
    F.clamp( pseudoObservation() )

However, I'm not sure we really need this so, I left it out for now.  We can always add it if/when we have an actual use case.

mikeryanmay:
The current approaches for fossil calibrations and tip dates are a real headache. Easily the worst part of learning/teaching divergence-time estimation in RevBayes (which is hard enough as it is), and I suspect many people either give up, or worse, make serious mistakes without realizing it. I 100% support anything that improves the user experience on this front. I have not gotten into the gory detail of the code yet, so here I am just going to give my thoughts about user experience and coherence.

This PR seems like a step in the right direction. It's especially nice that this is abstract and extensible to things like clade constraints. The code for writing tip-age intervals is certainly more intuitive! I have to admit I'm having a hard time imagining what node calibration code looks like in this framework. Can someone provide an example? (I've also tried sketching a graphical model and failed, but to be fair I think that's just a general node calibration problem.)

The real rough edge here (in my opinion) is `x.clamp(pseudoObservation())`. I think users will find this confusing. In all other cases, users are providing concrete objects with specific values to clamp, but here they are calling a function they don't understand with no arguments and they will wonder why they have to do it at all. This seems like another opportunity for users to check out, or instructors to wave their hands and say "don't think about this", which somewhat goes against the RevBayes philosophy of "understand your code". On our side, I understand the utility of using clamp this way, I just wonder if there are alternatives? What would be the consequences of clamping `PseudoDist` objects by default? It appears they are always constructed with a `PseudoDataLikelihood` so a likelihood could always be computed.

Statistically, I don't think that stratigraphic age uncertainty and node calibrations are really the same sort of thing, but this new framework treats them so. To me, a node calibration is saying: given the age of this node, t_i, what is the probability of a fossil at time f_i? The fossil age f_i is an observation so I'm happy to call this probability a likelihood---so for example the probability does not integrate to 1 over all t_i. Importantly, f_i in this case is not ambiguous, uncertain, or incomplete (or at least traditionally we treat it as known without error). This way of thinking is solidly instatiated by the PR (I think).

By contrast, tip age intervals/distributions reflect incomplete information. If we know the fossil s sampled between two dated strata, a_i and b_i, the "likelihood" should integrate over values of its age, f_i. This is analogous to nucleotide ambiguity: if all we know is that a particular nucleotide is a purine, the likelihood averages over whether that nucleotide is A or G. My view is that current approaches to tip dating constitute data augmentation: we augment our state space with the exact age of the tip so we can easily compute a probability, but as a result what we're really computing is just a fraction of the actual likelihood. (Of course, we could incorporate fossil age uncertainty into node calibrations, though I'm not aware of anyone having done so.)

So, I'm not totally convinced that these two different types of things (arbitrary likelihood functions vs. incomplete observations) can abstracted as the same, but I'm open to being convinced otherwise. I think in practice treating them the same will often be safe, but there are cases where the distinction is important. I'm thinking specifically of marginal likelihood estimation with power posteriors, where what goes into the "likelihood" bin matters. However, we already have implementations of data augmentation that will break power posteriors, and the current tip-age-interval implementation suffers from this too, so this PR is not really introducing a new problem. Perhaps I'm just complaining that this PR doesn't fix something it was not intended to fix!

mikeryanmay:
I suppose a node calibration would look something like this? From:
```
obs_age_ursinae ~ age_ursinae - dnExponential(1.0)
obs_age_ursinae.clamp(1.84)
```

To:
```
obs_age_ursinae ~ dnPseudo(pdExponential(age_ursinae, lambda = 1, shift = 1.84))
obs_age_ursinae.clamp(pseudoObservation())
```

bredelings:
> The real rough edge here (in my opinion) is x.clamp(pseudoObservation()). I think users will find this confusing. In all other cases, users are providing concrete objects with specific values to clamp, but here they are calling a function they don't understand with no arguments and they will wonder why they have to do it at all. This seems like another opportunity for users to check out, or instructors to wave their hands and say "don't think about this", which somewhat goes against the RevBayes philosophy of "understand your code".

If instructors don't explain it, then I suppose students won't learn anything.  But I actually think of this as a pedagogical tool, where people that are using pseudodata are forced to think "what does this mean?"  The explanation that I am giving is that even though people often think of this as a constraint, what they are actually doing is taking a real observation and making up a likelihood.

So this is an opportunity for instructors to ask students -- when you observe a fossil, what do you actually observe?  When you summarize this information down to "the age is between a and b", what are you throwing away?  How subjective is this summarization?  What would it look like if you used all the information? 

> On our side, I understand the utility of using clamp this way, I just wonder if there are alternatives? What would be the consequences of clamping PseudoDist objects by default? It appears they are always constructed with a PseudoDataLikelihood so a likelihood could always be computed.

I don't think about this in terms of utility -- I think this is conceptually accurate.  It raises the question "what is the REAL observation that you are refusing to write?"

Additionally, the answer to "why do I have to do this?" is that clamping a node is the way that we specify observation, so if we don't clamp the variable, we are saying that it isn't observed.  The real question is -- if we are observing something, why aren't we specifying the observed value?  And why aren't we specifying what distribution that value comes from?

Though of course if the instructor doesn't explain anything and says "just write this magic line and it will work" then the students won't learn anything.  And maybe this is too hard for students to learn, but I'm not sure that it has to be.  Part of the issue is that people have been trained for years not to think about fossils as observations, and to suppress all the details, so it will be initially hard for both instructors and students to think "what do we observe when we observe a fossil?"

Thoughts?

One possibility is that in the future we could add new syntax for this.  Instead of writing
```
x ~ dnPseudo( pdAbove(y,a,b) )
x.clamp( pseudoObservation() )
```
then could just write something like
```
!~ pdAbove(y,a,b)
```
and say that it is shorthand for the above, but I'm not completely sure how to implement this.

bredelings:
> I suppose a node calibration would look something like this? From:
> 
> obs_age_ursinae ~ age_ursinae - dnExponential(1.0)
> obs_age_ursinae.clamp(1.84)
> 
> To:
> 
> obs_age_ursinae ~ dnPseudo(pdExponential(age_ursinae, lambda = 1, shift = 1.84))
> obs_age_ursinae.clamp(pseudoObservation())
> 

Yeah, that's right.  That arguably shows why we should clamp _something_.  And this at least avoids the question of asking what it means to say that we observed `1.84`.

The additional goal is to explain that we DID observe something and we are REFUSING TO SAY WHAT IT IS.

So we could write
```
obs_age_ursinae.clamp(ItsASecret())
```
or
```
obs_age_ursinae.clamp(IRefuseToTellYouWhatISaw())
```

I also am thinking we could possible write the following to communicate that:
```
obs_age_ursinae ~ dnUnspecified(pdExponential(age_ursinae, lambda = 1, shift = 1.84))
obs_age_ursinae.clamp(unspecifiedObservation())
```

bredelings:
> By contrast, tip age intervals/distributions reflect incomplete information. If we know the fossil s sampled between two dated strata, a_i and b_i, the "likelihood" should integrate over values of its age, f_i. This is analogous to nucleotide ambiguity: if all we know is that a particular nucleotide is a purine, the likelihood averages over whether that nucleotide is A or G.  Because we are augmenting the MCMC state with the node age, we also do not do a direct sum.

I think this is a good analogy -- an `R` is definitely an observation.  Normally we sum over A,T,G,C and the likelihood is 1 for A/G and 0 for T/C.  However, in some cases, we augment the MCMC state with either (i) the letter at each internal node or (ii) the letter at every point in time.  In those cases, because we augment the MCMC state, we do not sum.  Likewise, if we observe a fossil between two datable layers with ages a and b, then this observation has probability zero if the fossil is younger than a or older than b.

> My view is that current approaches to tip dating constitute data augmentation: we augment our state space with the exact age of the tip so we can easily compute a probability, but as a result what we're really computing is just a fraction of the actual likelihood. (Of course, we could incorporate fossil age uncertainty into node calibrations, though I'm not aware of anyone having done so.)

I think for node ages, we have to sum them out regardless of whether they are for node calibrations or stratigraphic ages.  But I also think in both cases we can sum them out using MCMC sampling.  Say we have data $Y$ and parameter $\Theta$, and we augment with latent variable $L$, and $P(Y,L|\Theta) = P(Y|L) P(L|\Theta)$.  Then, clearly $P(Y|L)$ isn't the marginal likelihood.  But $P(Y|\Theta)$ wasn't the marginal likelihood either.  So I wouldn't say this is a "fraction of the actual likelihood".  If I'm understanding how path sampling works, then we are able to integrate out these latent variables, so including them doesn't change the marginal likelihood.

I could be making a mistake here, but it also looks like you can do $q_\beta(\theta) = p(y|l)^\beta \cdot p(l|\theta) \cdot p(\theta)$ and have $\Pr(Y=y) = \int_0^1 E [\log p_\beta(y|L)] d\beta$.  But you can (I think) _also_ do $q_\beta(\theta) = p(y|l)^\beta \cdot p(l|\theta)^\beta \cdot p(\theta)$ and have $\Pr(Y=y) = \int_0^1 E [\log p_\beta(y|L)+\log p_\beta(L|\theta)] d\beta$.

bredelings:
1. Currently we have dnPseudo / pseudoObservation().  I had a chat with some LLMs and came up with the following alternatives that all seem better:
* dnSuppressed(...) / suppressedObservation()
* dnUnspecified(...) / unspecifiedObservation()
* dnVirtual(...) / virtualObservation()
* dnImplicit(...) / implicitObservation()

2. I also updated clamp so that it takes a default value `pseudoObservation()` if the distribution's type is PseudoObservation.  (Obviously this would be updated if we change to one of the names above.)  So now you can do:

```
f ~ dnPseudo( pdAbove(t, age_min) )
f.clamp()
```

For any ordinary random variable `x`, you will get the normal error about a missing argument from `x.clamp()`.

3. We probably need a tutorial on virtual observations
* We can start with an example where we have real data -- for example we can have an observed stratigraphic depth.  We draw the model graph.
* Then we modify the example to suppress the data and use a virtual observation.

4. I updated the help file for dnPseudo.md to reflect the idea that virtualObservation() represents a stratigraphic depth that is missing, in the case of fossil ages.

src/revlanguage/functions/pseudodata/Func_PseudoDataExponential.cpp
```diff
@@ -0,0 +1,129 @@
...
+const ArgumentRules& Func_PseudoDataLikelihoodExponential::getArgumentRules( void ) const
+{
+    static ArgumentRules argumentRules = ArgumentRules();
+    static bool          rules_set = false;
+
+    if ( !rules_set )
+    {
+        argumentRules.push_back( new ArgumentRule( "x", Real::getClassTypeSpec(), "The constrained value.", ArgumentRule::BY_CONSTANT_REFERENCE, ArgumentRule::ANY ) );
+        argumentRules.push_back( new ArgumentRule( "lambda", Real::getClassTypeSpec(), "The rate.", ArgumentRule::BY_CONSTANT_REFERENCE, ArgumentRule::ANY ) );
+        argumentRules.push_back( new ArgumentRule( "shift", Real::getClassTypeSpec(), "The amount to shift the distribution to the right.", ArgumentRule::BY_CONSTANT_REFERENCE, ArgumentRule::ANY, new RealPos(0.0) ) );
```
davidcerny, line 83: In `dnExponential`, we are calling this argument `offset` rather than `shift`. We might want to change the name here for consistency's sake.

help/md/pdConstant.md
davidcerny: Sure, but I also wouldn't tell people that you can get a good pizza in Munich for 1 times €12.50, or 1 times 1 times 1 times 1 times €12.50 for that matter. I would just tell them you can get it for €12.50.

It seems to me that `pdConstant` is a mechanism that allows the inference to be informed by data that carry no information relevant to the inference. It is intrinsically, definitionally useless. If you disagree, that's fine – I might be missing something. But then I would reiterate my original request that you provide an example in the help page of a scenario where you think one would want to use it.

help/md/pseudoObservation.md
davidcerny: I also think it doesn't matter very much. I was more interested in what (if anything) the comparison operator tells us about whether different instances of `pseudoObservation` should be considered unique or all the same, which I think matters from a conceptual standpoint. If they are all the same, to me that sort of indicates that `pseudoObservation` is just a placeholder – a thing you have introduced purely because the RevBayes framework forces us to clamp to _something_, and not because it is well-motivated conceptually. Now that you have allowed "empty" `.clamp()` statements, that has become even clearer.

davidcerny:
I agree with @mikeryanmay that `pseudoObservation` is the most problematic part of this PR, though Ben may well be right that not being able to make sense of it is a symptom of not really having understood `dnPseudo()`, either. And maybe replacing "[p|P]seudo" by "[u|U]nspecified would help a little – but I don't think this is just a naming issue.

I also think that your recent interest in stratigraphic models is somewhat beside the point, Ben, and that this:

> I updated the help file for dnPseudo.md to reflect the idea that virtualObservation() represents a stratigraphic depth that is missing, in the case of fossil ages.

was premature. It seems to rest on the assumption that there is a single actual observed value that we are merely refusing to write down, and a distribution that this value comes from that we are merely refusing to specify. Maybe the unspecified observed value is the depth of the fossil in the stratigraphic column, and the distribution it comes from is given by some sort of probabilistic model accounting for sedimentation rates, unconformities, etc.? If this was true, then yes, it would definitely make it easier to explain to users what `pseudoObservation()` actually stands for.

But it is not true. At least if we also want pseudodata to replace node calibrations. Since I've designed node calibrations before, I actually have some insight into the process, thought it's of course possible that other researchers go about it differently. But just as an example, I'll detail my approximate thought process behind designing my favorite type of node calibration – an offset exponential:

1. The offset or "hard minimum" of the calibration corresponds to the oldest fossil reliably attributable to the clade in question. We pretend that the membership of this fossil within the clade and its minimum age (though not necessarily its maximum age) are known without error, so its choice should be conservative. A younger fossil whose membership in the clade in question is beyond reasonable doubt is preferable to an older fossil whose membership in that clade is merely likely.
2. The 95th (or 97.5th, or 99th, ...) percentile, or "soft maximum", is based on the age of the youngest geological unit in which we would expect to see representatives of the clade in question if the clade had already existed at the time of that unit's deposition, but in which we do not in fact see them. Maybe there are ecologically comparable taxa in the fossil assemblage preserved in that unit, or maybe its paleoenvironment corresponds to later geological units from which the clade is known. The rate of the exponential is calculated from the percentile using the quantile function.
3. The shape of the calibration corresponds to the belief that shorter waiting times (between the actual origin of the clade and its first appearance in the fossil record) are more likely than longer waiting times. It's not a perfect representation of our actual beliefs – the mode of the exponential coincides with the minimum age of the oldest fossil (= the offset), even though we know that the actual origin of the clade must have predated the appearance of that fossil by some margin (because evolving the derived morphological characters that allow us to identify it as such must have taken some time). Around 2010, it was fashionable to solve this problem by using offset lognormal calibrations, where there is a gap between the mode and the offset. The problem is that an offset lognormal is a function with three parameters, and if all we have is a hard minimum and a soft maximum, it's not clear where that third parameter should come from. Obviously, an arbitrary choice is possible but unsatisfactory. I prefer a less realistic and less arbitrary calibration; other researchers may want to strike a different balance.

So, what do we have? We have one concrete but unspecified observation, yes (whatever it is about the offset-fossil that allows us to estimate its age), but we also have a _lack_ of observations (that's what the soft maximum is based on), some background knowledge, and a good amount of pragmatism and simplifying assumptions. I think the whole notion of drawing observations from distributions kind of breaks down at this point. @mikeryanmay's comment here is telling: 

> I've also tried sketching a graphical model and failed, but to be fair I think that's just a general node calibration problem.

So upon further thought, I still think the only solution that makes sense at all is to treat the whole calibration – in the example above, the whole offset exponential function – as a fake observable. Or in other words, pseudodata. That's actually how Heath & Moore (2014) used that word. I.e.,

```
obs_age_ursinae ~ dnPseudo(age_ursinae)
obs_age_ursinae.clamp( pdExponential(lambda=x, offset=y) )
```

I would plead for retaining this interface, even if it's just as syntactic sugar over something that works differently under the hood. Of the two arguments against this:

> First, RevBayes has no infrastructure for updating the value of clamped objects when their parameters change. Second, the parameters `t[i]`, `a[i]`, and `b[i]` are really parameters of the unspecified distribution. So in the end putting the parameters into the distribution was necessary to get the pseudolikelihoods to update when `x` or `a[i]` or `b[i]` changed.

One basically amounts to the fact that it's hard to hammer a pseudodata-shaped peg into a RevBayes-shaped hole – which is fair, but I don't think it's the users who should pay the price. The other carries more weight, but since the whole notion of drawing observations from distributions is not a natural fit for what we are trying to do here, I'd be willing to tolerate some amount of arbitrariness in deciding what should be treated as a draw from a distribution, and what should instead be treated as that distribution's parameters.

`!~ pdExponential(age_ursinae, lambda=x, offset=y)` would work too.

bredelings:
>  It seems to rest on the assumption that there is a single actual observed value that we are merely refusing to write down

There can be any number of observed values, it doesn't have to be a single number.  With the stratigraphic depth, I tried to find a simpler, more concrete example, in order to make it easier to explain.

I am indeed claiming that if you want to multiply extra terms into the probability then they should be interpreted as `Pr(observation | parameters)`.  This isn't something invented by Heath & Moore (2014).

The observation above is indeed something that we are refusing to explicitly put into the model.  However, it doesn't have to be a single number.  It is "all the information that you are using to come up with the function of the parameters".

bredelings:
If you don't have multiple fossils that might be the oldest example of a clade, and you want to use an exponential, then you can also do:
```
oldest_fossil_age[i] ~ t[i] - dnExponential(fossilization_rate)
oldest_fossil_age[i].clamp( observed_age )
```
and you wouldn't need to use pseudodata.  We do this in `tutorials/dating/scripts/tree_BD_nodedate.Rev`, but just uses `dnExponential(1)`.

In order to incorporate additional info, such as time-varying fossilization rates, or multiple "oldest" fossils with some uncertainty about which clade they are in, etc etc one could try to explicitly model them, or suppress them and try to represent them with a single exponential using a procedure like you've mentioned above.  If its possible to make a model, then a model seems preferable to suppressing data and providing a pseudolikelihood.

Now some of this stuff seems like it could be handled by using the FBD instead of node calibrations.  The FBD doesn't require you to _a priori_ assign fossils to a specific clade, for example.  I think under the FBD, you would also explicitly handle the fact that whatever derived characters identify the clade do not arise instantaneously after the speciation.  So you could have something like:

```
feature_origination_time ~ t[i] - dnExponential(feature_origination_rate)
oldest_fossil_age ~ feature_origination_time - dnExponential(fossilization_rate)
oldest_fossil_age.clamp( observed_value )
```

davidcerny:
Apologies for the extremely delayed reply! I think we already discussed a lot of what I'm about to write here during the developer meeting last Thursday, but perhaps it might still be useful to put it in writing.

> There can be any number of observed values, it doesn't have to be a single number. With the stratigraphic depth, I tried to find a simpler, more concrete example, in order to make it easier to explain.

That's a fair point, and phrasing it as an example certainly helps. I'm happy to withdraw that particular objection.

> I am indeed claiming that if you want to multiply extra terms into the probability then they should be interpreted as `Pr(observation | parameters)`. This isn't something invented by Heath & Moore (2014).

Absolutely; that's not in dispute. It's just that, in my opinion, we find ourselves in a rather unusual situation where it's not completely clear what should be treated as a parameter and what should be treated as an observation.

Node calibrations in particular are often based on multiple data points daisy-chained together by assumption-laden reasoning – i.e., basically implicit models. E.g., the calibration fossil comes from formation X, which is assumed to be coeval with formation Y based on faunal correlations; magnetostratigraphic evidence from formation Y allows it to be correlated with formation Z, for which we have actual radiometric data allowing us to put an absolute age on the lower bound of the calibration. Then we go through a similar process to set the upper bound, except that things are even weirder, because our "observation" might actually consist in a _lack_ of observations.

I guess one could formalize this statistically as successive inferences, where the outcome of one inference is treated as a pseudo-observation to be plugged into the next round of inference. (I.e., we are not talking about normal Bayesian updating where the posterior from the previous step becomes the prior for the next step.) If we want to take things even further and collapse this whole chain into a single unspecified distribution to be clamped to unspecified data, the question is what should happen to those intermediate data points, and to the parameters of the models that link them together.
 
I would prefer to treat those parameters as fixed and collapse everything into a single abstract pseudodata object – effectively using the usual RevBayes framework of clamping a distribution to observables only for the last step in the chain. You, on the other hand, are arguing in favor of a neat separation, collecting the parameters of those intermediate models on one side (into `dnPseudo`) and collecting the data on the other side (into `pseudoObservation()`). I don't think this treatment quite works, even though mine is less than ideal, too. But those are difficulties inherent to trying to formalize what's usually an informal reasoning process.

> If you don't have multiple fossils that might be the oldest example of a clade, and you want to use an exponential, then you can also do:
>
> ```
> oldest_fossil_age[i] ~ t[i] - dnExponential(fossilization_rate)
> oldest_fossil_age[i].clamp( observed_age )
> ```
>
>and you wouldn't need to use pseudodata. We do this in `tutorials/dating/scripts/tree_BD_nodedate.Rev`, but just uses `dnExponential(1)`.

Just to clarify, I was aware that what I was outlining could already be done in RevBayes using the approach you suggest. It's just that, in my understanding, the pseudodata approach was supposed to supplant this sort of code. If it's just an alternative to it, then that makes it less problematic but also less useful, IMO.

> Now some of this stuff seems like it could be handled by using the FBD instead of node calibrations.

In some cases, sure, but that's perhaps a little beside the point. When I went over that 3-step procedure in my comment above, I wasn't proposing an actual study and seeking advice on what methodology I should use; I was just describing a common thing that researchers often want to do. I don't think telling them "just use FBD" is going to work, for a variety of reasons – there are multiple papers showing that FBD analyses sometimes yield unrealistically old node ages compared to node-dating, and @hoehna and @acapomorphic just recently had a paper accepted showing that there is a class of scenarios where FBD provides no benefits over node-dating (but may incur additional costs).

I.e., I believe that there are valid reasons for doing what I outlined above, in which case we should (1) continue to support it in RevBayes and (2) make the corresponding interface as intuitive and straightforward as possible.

> So you could have something like:
> 
> ```
> feature_origination_time ~ t[i] - dnExponential(feature_origination_rate)
> oldest_fossil_age ~ feature_origination_time - dnExponential(fossilization_rate)
> oldest_fossil_age.clamp( observed_value )
> ```

I don't see how that provides any tangible benefit over something like a lognormal calibration density. Either you have to fix `feature_origination_rate` to an arbitrary value, thus running into the exact same problem, or you actually have morphological data to estimate it from, in which case you wouldn't be using this kind of setup in the first place.

bredelings:
> Then we go through a similar process to set the upper bound, except that things are even weirder, because our "observation" might actually consist in a lack of observations.

So... did you know that graphical modelers use "observation" and "data" to mean that a random variable has some kind of specific outcome?  The outcome could be T=t, or it could be T>t.  Since the variable is random, you cannot know anything about the outcome until you _observe_.

Its pretty common to have an exponential waiting time `T` until the first event of some process, right?  If you observe that the first event takes place at time `t`, then you "observe" that `T=t`.   By definition, the first event implies that there were no prior events.  I have never heard anybody call this a "lack of observations" -- I would call this a "lack of events".  Likewise, if you have a Poisson process, and observe a collection of event times, then "observing" this collection means that there were no other events that both occurred and were sampled.   That is what is happening with fossil sampling in the FBD.

This seems pretty pervasive in the revbayes documentation.  For example, see the graphical model tutorial: https://revbayes.github.io/tutorials/intro/graph_models -- you'll notice that "observe" comes up a lot.  In general, "observe" and "data" both refer to random variables having a specific outcome.  
* "To specify that a node has an observed value associated with it we say that the node is _clamped_, or fixed, to the observed value."

I would say that we don't have a "lack of observations" -- we are observing a lack of events!

So... I guess I'm wondering...
* if this is such a pervasive issue in revbayes documentation, and in poisson processes, why is the problem coming up on this PR?
* if we use the term "data" instead of "observation" (roughly equivalent) then does that strike you differently?

bredelings:
> Node calibrations in particular are often based on multiple data points daisy-chained together by assumption-laden reasoning – i.e., basically implicit models. E.g., the calibration fossil comes from formation X, which is assumed to be coeval with formation Y based on faunal correlations; magnetostratigraphic evidence from formation Y allows it to be correlated with formation Z, for which we have actual radiometric data allowing us to put an absolute age on the lower bound of the calibration. 

This is kind of the point?

You have (a) all this data and (b) an implicit model, and you are not telling us either of them.  You are instead telling us $`f(t) = Pr_{implicit model}(lots of data|t)`$

This PR proposed writing "fossil data from X, faunal correlation data, magnetostratigraphic evidence from Y and Z, radiometric data from Z" as `suppressedObservation()` and "the implicit model" as `dnSuppressed`.  This paragraph thus seems to me like an argument _for_ the PR... what am I missing?

bredelings:
> I guess one could formalize this statistically as successive inferences, where the outcome of one inference is treated as a pseudo-observation to be plugged into the next round of inference. (I.e., we are not talking about normal Bayesian updating where the posterior from the previous step becomes the prior for the next step.) 

It might be nice to write a formal model for how radiographic info from formation Z tells us about ages in formation X, but this PR is not about doing that.  This PR is about how to get the conditioning correct (i.e. Pr(fossils|time), not Pr(time|fossils)) while NOT doing that.

So, I guess there are two distinct approaches:
1. provide an single likelihood (not a prior!) that claims to incorporate all the data, somehow <---- this PR
2. fully model all the different types of fossil-relevant data (as in stratigraphic depth modeling) <--- not this PR

Additionally, if you DO want to do approach number 2, it doesn't create a chain of inferences using successive models, it creates additional unobserved ("latent")  random variables that are all together in a single model.  If all of the new random variables are also pseudodata, then I think there is no point in doing this.  There would be a point only if you could fully model at least some of them.  For stratigraphic-depth-modeling, people actually do this -- but that is NOT pseudodata, and so is an example of what we are REFUSING to do in this PR.



> If we want to take things even further and collapse this whole chain into a single unspecified distribution to be clamped to unspecified data, the question is what should happen to those intermediate data points, and to the parameters of the models that link them together. 

The single unspecified distribution would then integrate out all those variables, which "collapse[s] this whole chain", leading to a single lump $`Pr_{implicit\ model}(lotsofdata|t)`$.

This is what the paleontologist who provides a node calibration likelihood is claiming to do anyway -- we are just making it explicit -- also also ensuring that we condition on `t`, instead of making "prior on t".

> I would prefer to treat those parameters as fixed and collapse everything into a single abstract pseudodata object – effectively using the usual RevBayes framework of clamping a distribution to observables only for the last step in the chain. You, on the other hand, are arguing in favor of a neat separation, collecting the parameters of those intermediate models on one side (into dnPseudo) and collecting the data on the other side (into pseudoObservation())

No, I am not -- not for this PR at least.  If there are parameters relating different categories of suppressed data, then by simply not mentioning them, you (claim to) integrate over them.  Holding them fixed is a special case of that.  But again, I never proposed to introduce parameters relating different parts of the suppressed data for this PR in the first place.

Again, it seems like what you are proposing we do, is what this PR is already doing?

> I don't think this treatment quite works, even though mine is less than ideal, too. But those are difficulties inherent to trying to formalize what's usually an informal reasoning process.

I think this is moot, this PR is not about doing that.  
* Presumably you are thinking of the proposed idea to decompose the waiting time into "time-until-first-distinguishing-feature" and "time-to-fossilization"?  I think that's interesting, but nothing about this PR requires that anybody do that.  
* The other possibly-confusing thing, is that there might be parameters for fossil age pseudodata, where we say that the fossil is in range $`[a,b]`$ while possibly inferring $`a`$ and $`b`$.  But node calibration pseudodata doesn't need parameters.

bredelings:
> It's just that, in my opinion, we find ourselves in a rather unusual situation where it's not completely clear what should be treated as a parameter and what should be treated as an observation.

I'm not sure I understand this.  Can you explain further?  Although I am wondering if we should just talk on Zoom or something, since that might make explaining things easier.

bredelings:
One idea that occurred to me is that we could remove all the bits (pdLogNormal, pdExponential) that might be used to handle node calibrations from this PR, since it seems that all the objections are to how node calibration is handled.  We could then keep the parts for handling fossil ages, where there have been no objections.

davidcerny:
A lot of points here that should be addressed – I'll do that as soon as possible. Right now, I'm unfortunately swamped with other work, so I'll keep my reply relatively brief. I get the impression that we're perhaps talking past each other to some extent (despite the fact that both of us are clearly making an honest effort not to!), and that we're getting bogged down in details while losing sight of the bigger picture.

If I were to restate my big-picture objection to

```
x ~ dnPseudo( pdBetween(y, a, b) )
x.clamp( pseudoObservation() )
```

I'd do it as follows:

We both agree that calibrations are not probability distributions over node or tip ages, but (pseudo)data that should enter Bayesian inference through the likelihood term rather than the prior term. So far so good! What I, as a researcher who engages in node-dating, present to you as my (pseudo)data is a single parametric curve, which is an ugly conglomerate of data (of different kinds) and assumptions. You agree that you won't force me to show you how that particular sausage got made, i.e., to be explicit about what data and assumptions (= models) I'm using. That's also good!

But then, you don't allow me to clamp to that curve, even though that's what we usually do with data in RevBayes. Instead, you want me to extract the parameters defining the shape of the curve and put them inside a distribution that itself needs clamping. You promise that this will be beneficial for me – I will now be able to sample those parameters, instead of holding them fixed! I'm not sure I need or even want that, because the curve itself, as-is, already represents my best description of the uncertainty associated with the relationship between the age of origin of a clade (or the age of a fossil species) and the age of that clade's (or species') fossil occurrences. Confused, I nevertheless play along.

But then I immediately run into a problem. Since I thought of the curve itself as my (pseudo)data, I'm no longer sure what I should clamp the distribution to. I'm told I should just clamp it to `pseudoObservation()`, or maybe nothing at all. Now I'm even more confused. The old system for setting up tip calibrations, which involved clamping to 0, was similarly opaque, but at least I could follow the logic behind it. The confusion was "just" due to bad interface design. Here, the confusion is conceptual. Before, I was frustrated that RevBayes made me jump through hoops before allowing me to do what I wanted to do; now, I'm no longer sure what I'm doing in the first place.

There were two other participants in this conversation, both of whom either found `pseudoObservation()` confusing themselves, or thought it would be confusing to future users. So I'm clearly not alone in this.

I won't block the PR over this if someone else gives it an approving review, though.

bredelings:
Hi David, OK, thanks for the latest feedback!  I think this gets to the core of the dislike with the current approach.  Let me think about this a bit before responding.

bredelings:
Hi David,

One thing I don't agree with is that the calibrations are "the pseudodata".  I can see calling the whole approach "a pseudodata approach", or calling the likelihood "a pseudodata likelihood", but I don't think this involves labeling anything "the pseudodata".  I had not realized you were thinking that way.

The calibration curve equally mixes the unspecified model and the unspecified data, so it could equally well be called "the pseudomodel". 

Is there a reason that you have decided that the curve is (pseudo)data?

bredelings:
> There were two other participants in this conversation, both of whom either found pseudoObservation() confusing themselves, or thought it would be confusing to future users. So I'm clearly not alone in this.

When trying to design a good interface, reports of raw, uninterpreted user confusion are hard to use.  Sometimes confusion can indicate flaws in the design, but sometimes it does not.  Instead, reports of confusion require some understanding of what type of confusion, why people are confused, and what they are confused about.

I'm assuming that "it is confusing" can mean not only "I am actually confused", but also "I find this unnatural". However

* "confusing" can mean that the user assumed something that wasn't true.  Then when it turned out differently than 
they expected, they were confused.

* "confusing" can mean "it wasn't explained well".  In this case, the solution could be to explain it better, instead of say that the approach itself is confusing.
 
* "confusing" can mean "I am not familiar with it" or "it is a new idea".  For example, windows users find Macs confusing, and Mac users find windows confusing.

* "confusing" could mean "i had to do a lot of thinking in order to understand it".  in any communication, work is required on the part of the explainer, and also on behalf of the listener. 

* "confusing" can be a way of saying "if I'm confused, then we shouldn't do it".  But MCMC is confusing, conditional probabilities are confusing, priors are confusing.  We generally tell people that its their job to learn how to do these things, instead of assuming that because someone is confused we have somehow made a mistake.  This has side-effects -- a lot of people don't want to do Bayesian inference, because "its confusing", so they just use simple R packages that don't require setting a prior.  But we don't assume that MCMC is fundamentally flawed because "it is confusing".

Of course, all of this conversation is a good-faith effort to fix the design problems.  But it would help me if there was more work put in to understand why people are confused, and what they are confused about.

bredelings:
> But then, you don't allow me to clamp to that curve, even though that's what we usually do with data in RevBayes. 

But the curve **isn't** data.  Nor is it "pseudodata".

> Instead, you want me to extract the parameters defining the shape of the curve and put them inside a distribution that itself needs clamping. You promise that this will be beneficial for me – I will now be able to sample those parameters, instead of holding them fixed! 

Not exactly -- I want to use the same approach for pseudodata that isn't node calibrations.  Actually, since the phrase "pseudodata" is both (a) nonstandard and (b) confusing, let's use a more standard term like "virtual evidence".

For simple node calibrations we can indeed compute a likelihood `Pr(suppressed data|node age)` that can be expressed as a 1-dimensional curve.  For but for other cases, such as virtual evidence that node1 is younger than node2, we would have a 2-dimensional curve:  `Pr(suppressed data| node age1, node age2)`.

It sounds like you are evaluating whether we should handle this solely in terms of whether it is "beneficial for me", as someone who cares only about simple node calibrations.  But what about other use cases?  Its not clear to me that it makes sense to wash our hands of them.

We could invent a special form that only works for node calibrations, and not for anything else.  Maybe we should.  On the other hand, RevBayes is supposed to be general software that can express general models that people have not thought of yet.  Additionally, I'm looking at the CladeAge paper, and it seems quite interesting -- and it is not doing a traditional node calibration.  So maybe we shouldn't set the traditional node calibration in stone just because its popular right now.


> But then I immediately run into a problem. Since I thought of the curve itself as my (pseudo)data, I'm no longer sure what I should clamp the distribution to.  I'm told I should just clamp it to pseudoObservation(), or maybe nothing at all. Now I'm even more confused. 

So, the curve is **not** the pseudodata.   I want to say that we observe (aka clamp) the REAL data, but don't say what it is.  It seems like everyone agrees that there is real data, we are observing it, and we are not saying what it is.  But apparently writing that down this is terribly confusing?  I can't tell if you are actually confused, or are talking like a hypothetical user who you think is confused.

 It does kind of feel like I'm trying to be helpful (such as when I added an option to not write `pseudoObservation()` because you didn't like it), and getting yelled at as a result.  I know that this is all in good faith, but it is a bit frustrating on my side too.

For your amusement, I include a link to the amazing skit [7 red lines](https://fredlybrand.com/2014/05/03/seven-red-lines-aka-the-expert-the-transcript/), especially the phrase "That's it, now you've confused everyone".

bredelings:
If I implement the syntax:
```
   !~ vlBelow(t1,t2)

   !~ vlLogNormal(t, lmu, lsigma, shift)
```
would this make people happy?  (Here vl stands for "virtual likelihood" if it matters).

I am going to continue looking into CladeAge-type constraints, and also stratigraphic range models.

davidcerny:
> Is there a reason that you have decided that the curve is (pseudo)data?

I got that interpretation from reading the literature on divergence time estimation. Again, [Heath & Moore (2014: 312)](https://www.taylorfrancis.com/chapters/edit/10.1201/b16965-19/bayesian-inference-species-divergence-times) were the first (to the best of my knowledge) to employ the term:

> Moreover, unlike most priors—which are updated (and typically dominated) by the information in the data—calibration priors are essentially ‘pseudo data’. That is, there is no information in the sequence data (or elsewhere) to update the prior calibration densities.

[Brown & Smith (2017: 345)](https://academic.oup.com/sysbio/article-abstract/67/2/340/4107448) then expanded on it in their "the past sure is tense" paper:

> [...] the interpretation of calibrations as pseudodata holds regardless of whether there is an interaction that leads to a discrepancy from the original user priors.

There have been a few subsequent studies using the term as well, but in increasingly confused and non-rigorous ways that are not particularly helpful for figuring out what statistical interpretation one should attach to calibrations.

> If I implement the syntax:
>
> ```
> !~ vlBelow(t1,t2)
> 
> !~ vlLogNormal(t, lmu, lsigma, shift)
> ```
>
> would this make people happy?

It would certainly make _me_ happy. I think that's an excellent interface that addresses all of my misgivings about the current way of specifying pseudo/virtual likelihoods.

PR #940, opened 2026-02-01
Fix tutorial test and fix rate age proposal

bredelings:
First, fix the integration test script so that it fails when the tutorial tests fail.  When I moved the call to the tutorial tests inside parenthesis, apparently this hides the error if the tutorial script fails.  Oops!  So add `|| exit 1` to ensure we exit on failure.

Second, this revealed that the newly introduced tutorial test for `timefig_dating` was failing.  I tracked this down to a bug in mvRateAgeProposal that was marking the tree already AFFECTED, so that future calls to get affected nodes treated the tree as not AFFECTED, even though it was.  Apparently this was coming from some unused code, so just remove that.  Then we need to update the FBD test output.

This is a prequisite for the `test-tensorphylo` branch, so I've included the changes here in that branch too.


davidcerny (changes requested):
Hi Ben, this looks good, thank you. I noticed that the problematic behavior of `mvRateAgeProposal` is also present in `mvRateAgeSubtreeProposal` – before we merge this, could you remove the lines in question from there as well?

https://github.com/revbayes/revbayes/blob/db9aca3168a4c71199b28d6b2738917c1de79da7/src/core/moves/proposal/tree/RateAgeSubtreeProposal.cpp#L105-L106

bredelings:
Thanks for catching that!  Done.

davidcerny (approved):

PR #942, opened 2026-02-03
Make the test runner more convenient

davidcerny:
This PR fixes some inconsistencies in the bash script we use for running integration tests.

- The help page says that the `-mpi` option is disabled, but it is still listed as a valid usage example in the message we print when one attempts to execute the script without arguments.
- The "USAGE" string in the help page says that options should go before the path to the executable (the first and only required argument), while an optional list of tests to be executed (specified using the `--tests` flag) should go _after_ the path to the executable. However, immediately below, `--tests` is listed as one of the available options. This is bound to cause confusion.
- The list of command-line options also says that the `-t` and `--tests` flags should be synonyms, but only the latter actually works.

This PR:

1. Re-enables the MPI option; it is automatically set to use 4 processes (inherited without change from old commented-out code). However, it still tests again the contents of the `output_expected` directory, since most of our tests currently lack MPI-specific expected output.
2. Makes option parsing position-independent, so recognized flags can be placed either before or after the path to the executable. 
3. Makes `-t` and `--tests` synonyms.

As a result of (2) and (3), if one wants to, say, run just the `basics` and `Mk` tests, the following commands are both valid:

    ```
    ./run_integration_tests.sh -t basics Mk "$PWD/../projects/cmake/build/rb"
    ./run_integration_tests.sh "$PWD/../projects/cmake/build/rb" --tests basics Mk
    ```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -352,7 +352,8 @@ jobs:
            rm -f test_coalescent/output_expected/mcmc_isochronous_EBSP.errout
         fi
 
-        ./run_integration_tests.sh -mpi ${USE_MPI} -windows ${IS_WINDOWS} rb
+        ./run_integration_tests.sh -mpi false -windows ${IS_WINDOWS} rb
+        # ./run_integration_tests.sh -mpi ${USE_MPI} -windows ${IS_WINDOWS} rb
         cd
         echo ::endgroup::
         echo ::group::testiphy
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -317,7 +317,8 @@ jobs:
            rm -f test_coalescent/output_expected/mcmc_isochronous_EBSP.errout
         fi
 
-        ./run_integration_tests.sh -mpi ${USE_MPI} -windows ${IS_WINDOWS} rb
+        ./run_integration_tests.sh -mpi false -windows ${IS_WINDOWS} rb
+        # ./run_integration_tests.sh -mpi ${USE_MPI} -windows ${IS_WINDOWS} rb
         cd
         echo ::endgroup::
         echo ::group::testiphy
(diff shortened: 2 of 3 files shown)


davidcerny:
It looks like the [biogeo](https://github.com/revbayes/revbayes.github.io/tree/source/tutorials/biogeo), [morph](https://github.com/revbayes/revbayes.github.io/tree/source/tutorials/morph), and [morph_ase](https://github.com/revbayes/revbayes.github.io/tree/source/tutorials/morph_ase) tutorials cause the MPI executable to segfault when we try to run it on 4 cores – this is probably worth further investigation. For now, I circumvented this by always executing the test runner with the `-mpi` flag set to `false`. This is slightly different from (but functionally equivalent to) the current solution, which sets the flag correctly but forces the test runner to ignore it.

bredelings:
I wonder if it would simplify things to require that the argument to tests is a single word where the test names are separated by commas.  Then we could collect the argument to --test more simply.

I'm thinking about how we would handle

./run_integration_tests.sh  --tests a,b,c rb -o debugMCMC=1

It seems like we could simply take the first argument that is neither a flag nor an argument as the executable name, and treat all further words as arguments to rb.

This would also allow us to avoid checking if an argument is an executable file with `[ -x  "$arg" ]`.

davidcerny:
> I'm thinking about how we would handle
>
> ./run_integration_tests.sh --tests a,b,c rb -o debugMCMC=1

Hmm, that's a good point. I checked that we can handle simple executable flags like `-i` regardless of whether we place the script flags before or after the executable path:

```
./run_integration_tests.sh -t basics Mk -mpi true "$PWD/../projects/cmake/build/rb" -i
./run_integration_tests.sh "$PWD/../projects/cmake/build/rb" -i -t basics Mk -mpi true
```

But this doesn't work with more complex executable arguments like `-o debugMCMC=1`. Back to the drawing board...

Issue #695, opened 2025-03-16
[BUG] Checkpointing doesn't work w/ multiple `mnExtNewick` monitors


davidcerny:
**Describe the bug**
When running a dating analysis with a relaxed clock model that is unlinked across partitions (i.e., an analysis in which each partition receives a separate set of branch rates), it is sometimes of interest to record the partition-specific rates. The best (and perhaps the only) way to do this that I'm aware of is to record the branch rates as node annotations using a `mnExtNewick` monitor, and specify one such monitor for each partition:

```
monitors.append( mnExtNewick(filename="output/mnExtNewick_test.trees", tree=timetree,
                             branch_rates[1], isNodeParameter=TRUE, printgen=50) )
monitors.append( mnExtNewick(filename="output/mnExtNewick_test.trees", tree=timetree,
                             branch_rates[2], isNodeParameter=TRUE, printgen=50) )
monitors.append( mnExtNewick(filename="output/mnExtNewick_test.trees", tree=timetree,
                             branch_rates[3], isNodeParameter=TRUE, printgen=50) )
```

Unfortunately, such analyses are impossible to restart from a checkpoint, because RevBayes cannot parse the stored trees.

**To reproduce**
Run the two toy scripts from [this zipped folder](https://github.com/user-attachments/files/19267967/mnExtNewick_test.zip). The `mnExtNewick_FBD_test.Rev` script uses the fossilized birth-death model, and throws the following error:

```
   Warning: readIndexFromParameter: Some nodes have an index parameter, and some do not!.  Ignoring indices.
   Error:	Number of tips in the initial tree does not match the number of taxa.
```

The `mnExtNewick_BD_test.Rev` script uses the standard constant-rate birth-death model with uniform sampling, and fails a little more dramatically:

```
   Warning: readIndexFromParameter: Some nodes have an index parameter, and some do not!.  Ignoring indices.
   Age of taxon =0.184223] was above the specified maximum and has been adjusted.
   Age of taxon =0.201474] was above the specified maximum and has been adjusted.
   Age of taxon Ailuropoda_melanoleuca was above the specified maximum and has been adjusted.
   Age of taxon =0.0744232] was above the specified maximum and has been adjusted.
   Age of taxon =0.104139] was above the specified maximum and has been adjusted.
   Age of taxon Tremarctos_ornatus was above the specified maximum and has been adjusted.
   Age of taxon =0.149534] was above the specified maximum and has been adjusted.
[...]

PR #946, opened 2026-02-10
Implement nested comments

bredelings:
This is in response to #695.  I'm not sure if this is a "bug" _per se_, or just unimplemented.

This PR implements Newick "nested comments", which means that if we see a `[` inside a comment then we require a extra ']' before we close the comment.  Preliminary investigation via ChatGPT suggests that "high quality" newick implementations like NCL and DendroPy all implement nested comments, so this should go in.


davidcerny (approved):
> our checkpointing implementation isn't really a general solution

In my opinion, our checkpointing implementation is quite general, in that it simply requires us to be able to read back in what we have written out. Inasmuch as your PR makes it possible to read in and successfully parse something (extended Newick strings produced by `mnExtNewick`) that we couldn't parse before, it helps _make_ it a general solution.

> So if we write out newick comments like `[&x[2]=1]` its possible that other programs will fail to read it.

Yes, it's true that both `ape::read.tree()` and FigTree struggle with this:

```
> a <- readLines("~/Downloads/mnExtNewick_test/output/mnExtNewick_test.ckp")
> b <- read.tree(text = a[2])
> b$node.label
[1] ""            "=0.0401052]" "=0.0558212]" "=0.053361]"  "=0.0659271]" "=0.0290355]" "=0.0179413]"
[8] "=0.0744986]" "=0.270263]" 
```

Screenshot 2026-02-10 at 12 16 55
<br/><br/>

However, to me, that's not really an excuse: _we_, at least, should be able to read in what we write out. This PR helps ensure that's the case, so I'm very much in favor of merging it!

PR #947, opened 2026-02-11
Move tuning fixes and cleanup

### LINKED ISSUE #945: [BUG] Misleading `tuneTarget` argument for moves with no tuning parameter
**Describe the bug**
About 36 different moves* – mostly those that operate on trees or character histories – allow specifying a `tuneTarget` argument at the Rev front end despite the fact that they lack any sort of tuning parameter in their back end. Others, like `mvTreeScale`, seemingly allow the `tuneTarget` value to be specified by the user, but then just disregard it and use their own hardcoded value:

https://github.com/revbayes/revbayes/blob/db9aca3168a4c71199b28d6b2738917c1de79da7/src/core/moves/proposal/tree/TreeScaleProposal.cpp#L237-L249

\* Estimated by running

```
find . -name "*.cpp" -exec grep -l "// this proposal has no tuning parameter" {} \; | wc -l
```

**Expected behavior**
We should only allow specifying `tuneTarget` for those moves that can, in fact, be tuned.

**Additional context**
This problem is due to the fact that pretty much all proposals mindlessly inherit `ArgumentRules` from the `Move` class – they want to inherit `weight`, but end up inheriting `tuneTarget` as well:

https://github.com/revbayes/revbayes/blob/db9aca3168a4c71199b28d6b2738917c1de79da7/src/revlanguage/moves/tree/Move_SPRNonclock.cpp#L107-L109

On the dev_stairwayplot branch, I have implemented an alternative solution that allows us to be more discerning:

https://github.com/revbayes/revbayes/blob/e2db43e8ce47114568e9af5bdc4bff35e4a6230f/src/revlanguage/moves/mixture/Move_AdaptiveReversibleJumpSwitch.cpp#L114-L122

davidcerny:
This PR was originally intended to fix #945, i.e., the fact that many moves allow the user to specify a `tuneTarget` argument that has no effect, because they lack any sort of tuning mechanism. However, it soon became clear there were a number of other problems:

## Tuning issues

1. A lot of moves are capable of being tuned and allow specifying the `tuneTarget` argument, but then proceed to ignore it, since the actual target acceptance rate is hardcoded into the `tune()` function of the corresponding proposal. Most often, that rate happens to be 0.44, which is the default `tuneTarget` set by [`RlMove.cpp`](https://github.com/revbayes/revbayes/blob/master/src/revlanguage/moves/RlMove.cpp), but a few others use 0.23 or 0.234 instead. In those cases, the Rev interface is doubly misleading: not only does it imply that the `tuneTarget` can be changed when that's not actually possible; it also doesn't show the user the real value it's fixed to.

2. A lot of moves are seemingly fine, in that the `getParameterRules()` function of their Rev interface contains something like this:

    ```cpp
    const MemberRules& inheritedRules = Move::getParameterRules();
    move_member_rules.insert( move_member_rules.end(), inheritedRules.begin(), inheritedRules.end() );
    ```
    
    (this allows them to inherit the `tuneTarget` argument from the `Move` class)
    
    and the `tune()` function of the corresponding proposal looks something like this:
    
    ```cpp
    double p = this->targetAcceptanceRate;
    if ( rate > p )
    {
        tuning_parameter *= (1.0 + ((rate - p)/(1.0 - p)) );
    }
    else
    {
        tuning_parameter /= (2.0 - rate/p);
    }
    ```
    
    (so that the acceptance rate approaches an arbitrary value stored in the protected member `targetAcceptanceRate`).
    
    But actually, there is nothing to connect the `tuneTarget` argument specified in the Rev-language front end with the `targetAcceptanceRate` variable in the Core back end. Those few moves that actually do have this connection properly set up, like `mvScale`, do it by passing the `tuneTarget` argument value to the proposal constructor.
    
3. A special shout-out goes to `mvRateAgeBetaShift`, which not only hardcodes the acceptance rate to 0.44 despite seemingly allowing the user to specify an arbitrary `tuneTarget`, but which doesn't even derive from `Proposal`, so it has no `targetAcceptanceRate` member one could set. I.e., the back end for allowing user-specified acceptance rate targets is just not there – not only in practice but even in principle.

## Solutions

1. I removed the `tuneTarget` from the Rev interface of those moves that are not capable of being tuned, by changing the code in `getParameterRules()` from this:

    ```cpp
    /* Inherit weight from Move, put it after variable */
    const MemberRules& inheritedRules = Move::getParameterRules();
    move_member_rules.insert( move_member_rules.end(), inheritedRules.begin(), inheritedRules.end() );
    ```
    
    to this:
    
    ```cpp
    /* Inherit weight (but not tuneTarget!) from Move and put it after the arguments created above */
    const MemberRules& inheritedRules = Move::getParameterRules();
    for (size_t i = 0; i < inheritedRules.size(); ++i)
    {
        if ( inheritedRules[i].getArgumentLabel() == "weight" )
        {
            move_member_rules.push_back( inheritedRules[i].clone() );
        }
    }
    ```
    
    This is somewhat verbose, but hopefully pretty explicit about what it does.
    
2. For the proposals with hardcoded values, I changed the algebraic expression in the `tune()` function to use `p` (where `p = this->targetAcceptanceRate`) rather than a specific number. For those moves that use defaults other than 0.44, I further added the following to `getParameterRules()`, replacing an `ArgumentRule` that is usually inherited from `Move`:

    ```cpp
    /* Provide our own default value for tuneTarget */
    memberRules.push_back( new ArgumentRule( "tuneTarget", Probability::getClassTypeSpec(), "The acceptance probability targeted by auto-tuning.", ArgumentRule::BY_VALUE, ArgumentRule::ANY, new Probability( 0.234 ) ) );
    ```
    
3. When it came to making sure that the moves which allow setting a user-specified `tuneTarget` can actually propagate that value into the back end, I didn't feel like changing several dozen constructors. Instead, I added a member function to the `Proposal` class called `setTargetAcceptanceRate()`, and handled things as follows on the Rev-language side in the `constructInternalObject()` function:

    ```cpp
    double w = static_cast<const RealPos &>( weight->getRevObject() ).getValue();
    bool t = static_cast<const RlBoolean &>( tune->getRevObject() ).getValue();
    double tt = static_cast<const Probability &>( tuneTarget->getRevObject() ).getValue();

    RevBayesCore::MyMoveProposal *prop = new RevBayesCore::MyMoveProposal(<args>);
    prop->setTargetAcceptanceRate(tt);

    value = new RevBayesCore::MetropolisHastingsMove(prop, w, t);
    ```
    
4. I added a new private member called `targetAcceptanceRate` to the `RateAgeBetaShift` class, which is set in the constructor.

## Cleanup

While making the changes above, I also took an ax to the code base, removing the following files:

- Obsolete proposals that haven't been worked on since before the rapture, and have no corresponding Rev interface:

    - [`src/core/moves/proposal/ContinuousCharacterDataScaleProposal.h`](https://github.com/revbayes/revbayes/blob/master/src/core/moves/proposal/ContinuousCharacterDataScaleProposal.h)
    - [`src/core/moves/proposal/ContinuousCharacterDataScaleProposal.cpp`](https://github.com/revbayes/revbayes/blob/master/src/core/moves/proposal/ContinuousCharacterDataScaleProposal.cpp)
    - [`src/core/moves/proposal/scalar/ScaleUpDownProposal.h`](https://github.com/revbayes/revbayes/blob/master/src/core/moves/proposal/scalar/ScaleUpDownProposal.h) – not to be confused with another proposal that was retained, [`src/core/moves/proposal/UpDownScaleProposal.h`](https://github.com/revbayes/revbayes/blob/master/src/core/moves/proposal/UpDownScaleProposal.h) (yeah, those naming decisions are horrendous in themselves)
    - [`src/core/moves/proposal/scalar/ScaleUpDownProposal.cpp`](https://github.com/revbayes/revbayes/blob/master/src/core/moves/proposal/scalar/ScaleUpDownProposal.cpp)
    - [`src/core/moves/proposal/scalar/SlideUpDownProposal.h`](https://github.com/revbayes/revbayes/blob/master/src/core/moves/proposal/scalar/SlideUpDownProposal.h) – not to be confused with another proposal that was retained, [`src/core/moves/proposal/UpDownSlideProposal.h`](https://github.com/revbayes/revbayes/blob/master/src/core/moves/proposal/UpDownSlideProposal.h)
    - [`src/core/moves/proposal/scalar/SlideUpDownProposal.cpp`](https://github.com/revbayes/revbayes/blob/master/src/core/moves/proposal/scalar/SlideUpDownProposal.cpp)
    - [`src/core/moves/proposal/tree/SpeciesTreeNodeSlideProposal.h`](https://github.com/revbayes/revbayes/blob/master/src/core/moves/proposal/tree/SpeciesTreeNodeSlideProposal.h)
    - [`src/core/moves/proposal/tree/SpeciesTreeNodeSlideProposal.cpp`](https://github.com/revbayes/revbayes/blob/master/src/core/moves/proposal/tree/SpeciesTreeNodeSlideProposal.cpp)
    
- A move where the Rev interface doesn't correctly connect to the proposal constructor, and which is missing from the [move register](https://github.com/revbayes/revbayes/blob/master/src/revlanguage/workspace/RbRegister_Move.cpp):

    - [`src/core/moves/proposal/tree/EmpiricalTreeTopologyProposal.h‎`](https://github.com/revbayes/revbayes/blob/master/src/core/moves/proposal/tree/EmpiricalTreeTopologyProposal.h)
    - [`src/core/moves/proposal/tree/EmpiricalTreeTopologyProposal.cpp`](https://github.com/revbayes/revbayes/blob/master/src/core/moves/proposal/tree/EmpiricalTreeTopologyProposal.cpp)
    - [`src/revlanguage/moves/tree/Move_EmpiricalTreeTopology.h`](https://github.com/revbayes/revbayes/blob/master/src/revlanguage/moves/tree/Move_EmpiricalTreeTopology.h)
    - [`src/revlanguage/moves/tree/Move_EmpiricalTreeTopology.cpp`](https://github.com/revbayes/revbayes/blob/master/src/revlanguage/moves/tree/Move_EmpiricalTreeTopology.cpp)
    
- A move where the corresponding proposal is entirely commented out, and which is also missing from the register:

    - [`src/core/moves/proposal/tree/NodeTimeSlideUniformCharacterHistoryProposal.h‎`](https://github.com/revbayes/revbayes/blob/master/src/core/moves/proposal/tree/NodeTimeSlideUniformCharacterHistoryProposal.h)
    - [`src/core/moves/proposal/tree/NodeTimeSlideUniformCharacterHistoryProposal.cpp‎`](https://github.com/revbayes/revbayes/blob/master/src/core/moves/proposal/tree/NodeTimeSlideUniformCharacterHistoryProposal.cpp)
    - [`src/revlanguage/moves/tree/Move_NodeTimeSlideUniformCharacterHistory.h‎‎`](https://github.com/revbayes/revbayes/blob/master/src/revlanguage/moves/tree/Move_NodeTimeSlideUniformAgeConstrained.h)
    - [`src/revlanguage/moves/tree/Move_NodeTimeSlideUniformCharacterHistory.cpp‎`](https://github.com/revbayes/revbayes/blob/master/src/revlanguage/moves/tree/Move_NodeTimeSlideUniformAgeConstrained.cpp)

## Doubts

1. In theory, the changes made in this PR should have no effect on existing tests, as long as these don't specify non-default `tuneTarget`s, or don't try to specify a `tuneTarget` for moves that are not capable of being tuned (which did actually happen in [one case](https://github.com/revbayes/revbayes/blob/master/tests/test_mvSlice/scripts/treeAssembly.Rev), for `mvNNI` and `mvSPR`). However, for `mvAVMVN` ~and `mvRateAgeBetaShift`~, the changes did make a difference, necessitating the replacement of the expected output of the [`mcmc_AVMVN.Rev`](https://github.com/revbayes/revbayes/blob/master/tests/test_MVN_move/scripts/mcmc_AVMVN.Rev) ~and [`FBD_dating_ex3.Rev`](https://github.com/revbayes/revbayes/blob/master/tests/test_FBD/scripts/FBD_dating_ex3.Rev)~ test script~s. For what it's worth, after some joint investigation of the latter move, this is what Claude told me:~
    
    > ~The diagnostic shows different acceptance counts (590/939 vs 559/927), indicating different accept/reject decisions. This could be due to:~
    >
    > ~1. Memory layout differences affecting hash values or pointer comparisons~
    > ~2. Uninitialized memory reads~
    > ~3. Different initialization order~
    >

    > ~Since `targetAcceptanceRate` is only used in `tune()` and doesn't affect the move logic, the divergence likely comes from the object layout change.~
    
2. It looks like beta proposals are tuned in the opposite way compared to all others. For most proposals, tuning works as follows:

    ```cpp
    double p = this->targetAcceptanceRate;
    if ( rate > p )
    {
        tuning_parameter *= (1.0 + ((rate - p)/(1.0 - p)) );
    }
    else
    {
        tuning_parameter /= (2.0 - rate/p);
    }
    ```
    
    For beta proposals, we instead do:
    
    ```cpp
    double p = this->targetAcceptanceRate;
    if ( rate > p )
    {
        tuning_parameter /= (1.0 + ((rate - p)/(1.0 - p)) );
    }
    else
    {
        tuning_parameter *= (2.0 - rate/p);
    }
    ```
    
    I thought this was a mistake at first, but it's almost certainly intentional, as it is a very consistent pattern that's only present in proposals that draw from the beta or Dirichlet (= multivariate beta) distribution. (Most of them helpfully have "beta" somewhere in their name.) ~Well... that's almost true. There are two proposals that use the same pattern, but draw from the inverse Wishart distribution instead:~
    
    - ~[`src/core/moves/proposal/matrix/CorrelationMatrixExpansionProposal.cpp`](https://github.com/revbayes/revbayes/blob/master/src/core/moves/proposal/matrix/CorrelationMatrixExpansionProposal.cpp)~
    - ~[`src/core/moves/compound/CorrelationMatrixReparameterizationMove.cpp`](https://github.com/revbayes/revbayes/blob/master/src/core/moves/compound/CorrelationMatrixReparameterizationMove.cpp)~
    
    ~It's not clear to me if this makes good mathematical sense, or if it's accidentally copied over from the closely related beta proposals like `CorrelationMatrixElementBetaProposal.cpp`.~
    
3. There are several moves which seem to be superficially okay, in that they have a Rev front end that connects to a Core back end, but which are missing from the move register. Two examples I noticed are [`mvFNPRCharacterHistory`](https://github.com/revbayes/revbayes/blob/master/src/revlanguage/moves/tree/Move_FNPRCharacterHistory.cpp) and [`mvNarrowExchangeCharacterHistory`](https://github.com/revbayes/revbayes/blob/master/src/revlanguage/moves/tree/Move_NarrowExchangeCharacterHistory.cpp), but there are almost certainly more. I'm not sure what to do with these. One possibility is that their lack of inclusion in the register is just an oversight, so we should register them and make them available for use. The other possibility is that they were deliberately not registered because they were known not to work as intended, in which case I'd like to remove them from the code base.


bjoelle:
so for doubt 2) based on a quick look through I would agree that it is intentional 
- in the non-beta proposals, a rate too high leads to an increase in lambda, i.e. an increase in the scaling factor, which makes the proposal wider
- in the beta proposals, a rate too high leads to a decrease in delta, which leads to lower a and b, which seems to make the variance of the beta distro higher (so I'm guessing also a wider proposal)
- no idea what's going on in the inverse wishart because the tune variable is commented out and doesn't appear in the rest of the calculation

davidcerny:
Thank you, that's super helpful! And clearly I should have taken a closer look at the two correlation matrix moves. I grepped the `*= (2.0 - rate/p)` string from the proposal files and didn't even notice it was commented out. Luckily, I hadn't made the same mistake earlier, when changing the Rev interface – neither move allows setting the `tuneTarget` argument, correctly reflecting the fact they have no tuning mechanism.

The commented-out body of the `tune()` function has now been removed for clarity.

src/core/moves/characterhistory/DiscreteEventCategoryRandomWalkProposal.cpp
```diff
@@ -206,24 +206,8 @@ void DiscreteEventCategoryRandomWalkProposal::setProposalTuningParameter(double
...
 
 
-/**
- * Tune the Proposal to accept the desired acceptance ratio.
- *
- * The acceptance ratio for this Proposal should be around 0.44.
- * If it is too large, then we increase the proposal size,
- * and if it is too small, then we decrease the proposal size.
- */
 void DiscreteEventCategoryRandomWalkProposal::tune( double rate )
```
bredelings: So, right now it looks like AbstractMove::tune() is pure virtual -- theres no implementation.

One possible cleanup would be to give it an implementation that does nothing.  We could then simply delete implementations like this that do not do anything.

src/core/moves/proposal/matrix/CorrelationMatrixPartialElementBetaProposal.h
```diff
@@ -50,9 +50,9 @@ template <class variableType> class StochasticNode;
         
         StochasticNode<MatrixReal >*             variable;
         
-        double                                   alpha;                                                                             //!< The Beta parameter of the move (larger lambda -> larger proposals).
+        double                                   alpha;                                                                              //!< The Beta parameter of the move (larger lambda -> larger proposals).
```
bredelings: This comment is a bit confusing, though that is of course not the fault of this PR.

I can see why we might want to call the beta parameter "alpha", but the comment about lambda seems like it might be a mistake.  More significantly, for a Beta distribution, a larger value means a SMALLER proposal, in which case the tuning might be backward...

If this is hard to address, we could just add an additional comment pointing out the problem like `// NOTE: Is this comment correct?  Is the move correct?`

src/core/moves/compound/RateAgeBetaShift.cpp
```diff
@@ -24,11 +24,12 @@ namespace views = ranges::views;
 
 using namespace RevBayesCore;
 
-RateAgeBetaShift::RateAgeBetaShift(StochasticNode<Tree> *tr, std::vector<StochasticNode<double> *> v, StochasticNode<RbVector<double> > *sv, double d, bool t, double w) : AbstractMove( w, t),
+RateAgeBetaShift::RateAgeBetaShift(StochasticNode<Tree> *tr, std::vector<StochasticNode<double> *> v, StochasticNode<RbVector<double> > *sv, double d, bool t, double w, double targetAcceptanceRate) : AbstractMove( w, t),
     tree( tr ),
     rates_vec( v ),
     rates( sv ),
     delta( d ),
+    targetAcceptanceRate( targetAcceptanceRate ),
```
bredelings, line 32: ... and initializing this to `0.44` here.  That preserves the information in the previously-fixed value, but allows the value to be tweaked by modifying `targetAcceptanceRate`.  

I guess that depends on whether there's an interface to modify it though...  Hmm.

src/core/moves/compound/RateAgeBetaShift.cpp
```diff
@@ -609,13 +610,14 @@ void RateAgeBetaShift::tune( void )
     {
         double rate = num_accepted_current_period / double(num_tried_current_period);
         
-        if ( rate > 0.44 )
+        double p = this->targetAcceptanceRate;
```
bredelings, line 613: If we don't want to add this field to the object, we could just set something here, like

`double p = 0.44; // target acceptance rate`

bredelings:
Many of the existing proposals ignore the `targetAcceptanceRate` field.  Why might they do that?  In some cases the optimal value is known -- for example I think `0.44` is the optimal acceptance rate for Gaussian proposals on 1D Gaussian posteriors, and `0.234` is the optimal acceptance rate on high-dimensional Gaussian posteriors.  In other cases, the optimal value may not be known, so there is value in letting people tweak it in search of better MCMC performance.  (In practice, I think the code is often using 0.44 even though that is optimal for a different proposal :roll_eyes: .)

Making the class always use `targetAcceptanceRate` improves coherence and makes the acceptance rate always tweakable.  But in order to preserve the information about the default acceptance rate I think we need to also set the targetAcceptanceRate value to the previously fixed value in the constructor.  Right now `targetAcceptanceRate` defaults to `0.44` if I'm understanding this correctly (see `Proposal.h`).  I think we can make the constructor call `Proposal(0.234)` to set `targetAcceptanceRate` to `0.234`.

src/core/moves/compound/RateAgeBetaShift.cpp
```diff
@@ -24,11 +24,12 @@ namespace views = ranges::views;
 
 using namespace RevBayesCore;
 
-RateAgeBetaShift::RateAgeBetaShift(StochasticNode<Tree> *tr, std::vector<StochasticNode<double> *> v, StochasticNode<RbVector<double> > *sv, double d, bool t, double w) : AbstractMove( w, t),
+RateAgeBetaShift::RateAgeBetaShift(StochasticNode<Tree> *tr, std::vector<StochasticNode<double> *> v, StochasticNode<RbVector<double> > *sv, double d, bool t, double w, double targetAcceptanceRate) : AbstractMove( w, t),
```
bredelings, line 27: I'd suggest removing this parameter...

src/core/moves/proposal/tree/BranchLengthScaleProposal.cpp
```diff
@@ -191,22 +191,23 @@ void BranchLengthScaleProposal::setProposalTuningParameter(double tp)
...
- * The acceptance ratio for this Proposal should be around 0.44.
+ * The acceptance ratio for this Proposal should be around 0.234.
  * If it is too large, then we increase the proposal size,
  * and if it is too small, then we decrease the proposal size.
  */
 void BranchLengthScaleProposal::tune( double rate )
 {
     
-    if ( rate > 0.234 )
+    double p = this->targetAcceptanceRate;
```
bredelings, line 203: I think we might need to call `Proposal(0.234)` in the constructor to set `targetAcceptanceRatio` to `0.234`.  Otherwise it will be `0.44`.

src/core/moves/proposal/tree/SpeciesSubtreeScaleBetaProposal.cpp
```diff
@@ -380,22 +380,23 @@ void SpeciesSubtreeScaleBetaProposal::setProposalTuningParameter(double tp)
...
- * The acceptance ratio for this Proposal should be around 0.44.
+ * The acceptance ratio for this Proposal should be around 0.234.
  * If it is too large, then we increase the proposal size,
  * and if it is too small, then we decrease the proposal size.
  */
 void SpeciesSubtreeScaleBetaProposal::tune( double rate )
 {
 
-    if ( rate > 0.234 )
+    double p = this->targetAcceptanceRate;
```
bredelings, line 392: Also here we need to initialize targetAcceptanceRatio to 0.234.

src/core/moves/proposal/tree/SpeciesTreeScaleProposal.cpp
```diff
@@ -314,27 +314,28 @@ void SpeciesTreeScaleProposal::swapNodeInternal(DagNode *oldN, DagNode *newN)
...
- * The acceptance ratio for this Proposal should be around 0.44.
+ * The acceptance ratio for this Proposal should be around 0.234.
  * If it is too large, then we increase the proposal size,
  * and if it is too small, then we decrease the proposal size.
  */
 void SpeciesTreeScaleProposal::tune( double rate )
 {
 
-    if ( rate > 0.234 )
+    double p = this->targetAcceptanceRate;
```
bredelings, line 331: Initialization issue here too

src/core/moves/proposal/tree/TreeScaleProposal.cpp
```diff
@@ -228,22 +228,23 @@ void TreeScaleProposal::setProposalTuningParameter(double tp)
...
- * The acceptance ratio for this Proposal should be around 0.44.
+ * The acceptance ratio for this Proposal should be around 0.234.
  * If it is too large, then we increase the proposal size,
  * and if it is too small, then we decrease the proposal size.
  */
 void TreeScaleProposal::tune( double rate )
 {
     
-    if ( rate > 0.234 )
+    double p = this->targetAcceptanceRate;
```
bredelings, line 240: initialization to 0.234

src/core/moves/proposal/vector/ElementScaleProposal.cpp
```diff
@@ -182,22 +182,23 @@ void ElementScaleProposal::setProposalTuningParameter(double tp)
...
- * The acceptance ratio for this Proposal should be around 0.44.
+ * The acceptance ratio for this Proposal should be around 0.23.
  * If it is too large, then we increase the proposal size,
  * and if it is too small, then we decrease the proposal size.
  */
 void ElementScaleProposal::tune( double rate )
 {
     
-    if ( rate > 0.23 )
+    double p = this->targetAcceptanceRate;
```
bredelings, line 194: Initialization to 0.23

src/core/moves/proposal/vector/MultipleElementScaleProposal.cpp
```diff
@@ -231,22 +231,23 @@ void MultipleElementScaleProposal::setProposalTuningParameter(double tp)
...
- * The acceptance ratio for this Proposal should be around 0.44.
+ * The acceptance ratio for this Proposal should be around 0.23.
  * If it is too large, then we increase the proposal size,
  * and if it is too small, then we decrease the proposal size.
  */
 void MultipleElementScaleProposal::tune( double rate )
 {
     
-    if ( rate > 0.23 )
+    double p = this->targetAcceptanceRate;
```
bredelings, line 243: intialization

src/core/moves/proposal/vector/ShrinkExpandProposal.cpp
```diff
@@ -219,22 +219,23 @@ void ShrinkExpandProposal::setProposalTuningParameter(double tp)
...
- * The acceptance ratio for this Proposal should be around 0.44.
+ * The acceptance ratio for this Proposal should be around 0.23.
  * If it is too large, then we increase the proposal size,
  * and if it is too small, then we decrease the proposal size.
  */
 void ShrinkExpandProposal::tune( double rate )
 {
     
-    if ( rate > 0.23 )
+    double p = this->targetAcceptanceRate;
```
bredelings, line 231: initialization?

src/core/moves/proposal/vector/ShrinkExpandScaleProposal.cpp
```diff
@@ -219,22 +219,23 @@ void ShrinkExpandScaleProposal::setProposalTuningParameter(double tp)
...
- * The acceptance ratio for this Proposal should be around 0.44.
+ * The acceptance ratio for this Proposal should be around 0.23.
  * If it is too large, then we increase the proposal size,
  * and if it is too small, then we decrease the proposal size.
  */
 void ShrinkExpandScaleProposal::tune( double rate )
 {
     
-    if ( rate > 0.23 )
+    double p = this->targetAcceptanceRate;
```
bredelings, line 231: initialization?

src/core/moves/proposal/vector/VectorScaleProposal.cpp
```diff
@@ -137,13 +137,14 @@ void VectorScaleProposal::setProposalTuningParameter(double tp)
 void VectorScaleProposal::tune( double rate )
 {
     
-    if ( rate > 0.23 )
+    double p = this->targetAcceptanceRate;
```
bredelings, line 140: Initialization?

src/core/moves/proposal/vector/VectorSingleElementScaleProposal.cpp
```diff
@@ -134,13 +134,14 @@ void VectorSingleElementScaleProposal::setProposalTuningParameter(double tp)
 void VectorSingleElementScaleProposal::tune( double rate )
 {
     
-    if ( rate > 0.23 )
+    double p = this->targetAcceptanceRate;
```
bredelings, line 137: Initialization?

src/core/moves/proposal/AVMVNProposal.cpp
```diff
@@ -633,23 +633,24 @@ void AVMVNProposal::setProposalTuningParameter(double tp)
...
+ * The acceptance ratio for this Proposal should be around 0.234.
  * If it is too large, then we increase the proposal size,
  * and if it is too small, then we decrease the proposal size.
  */
 void AVMVNProposal::tune( double rate )
 {
 
     // Update proposal variance
-    if ( rate > 0.234 )
+    double p = this->targetAcceptanceRate;
```
bredelings, line 646: Initialization?

bredelings (dismissed):
OK, I finished 😅.  Hopefully the comments are useful.

src/core/moves/compound/RateAgeBetaShift.cpp
bredelings: Although if the alternative is to implement setAcceptanceRate and then call it right after the constructor, maybe this way is fine.

bredelings:
Um, now I see that for the proposals that used `0.234` instead of `0.44` you put the default at the RevLanguage level.. so that's fine.

Anyway, this seems like a nice cleanup to me.  Thanks!  Lots of code being deleted -- less code to maintain!

src/core/moves/characterhistory/DiscreteEventCategoryRandomWalkProposal.cpp
davidcerny: Great idea! I went ahead and did that. I also provided a no-op default implementation for `AbstractMove::setMoveTuningParameter()`, and by analogy, for `Proposal::setMoveTuningParameter()` and `Proposal::tune()`. See commit 69a8163.

src/core/moves/proposal/matrix/CorrelationMatrixPartialElementBetaProposal.h
davidcerny: I wouldn't even call it a "mistake"... In general, these sorts of comments can almost never be trusted, since people often create new header files by copying old ones without bothering to change the comments, especially in what they perceive as boilerplate code. There are dozens of moves operating on all sorts of data structures, from trees to character histories, that all claim to be a simple scaling move, for example.

Fixing all these misleading comments is outside the purview of this PR, but I did verify that the tuning of the proposal in question is correct, contrary to what the comment says – i.e., that a larger `alpha` does in fact correspond to a smaller proposal. All beta moves work this way (see my "doubt 2" and @bjoelle's reply thereto), even if their header file comments incorrectly claim otherwise.

That's preferable to the alternative (a comment that correctly describes the wrong tuning behavior), but just to prevent doubts like this from arising again, I went ahead and fixed the comments describing the tuning parameters of all beta moves. See commit d6db4f1.

src/core/moves/compound/RateAgeBetaShift.cpp
davidcerny: That was my thinking as well. The alternative solution you describe is what I used for moves that derive from `Proposal`, but since `RateAgeBetaShift` does not, I thought adding a parameter to the constructor was simpler.

src/core/moves/compound/RateAgeBetaShift.cpp
davidcerny: Isn't that exactly what the move currently does, though? Hardcode the target acceptance rate to 0.44 inside the `tune()` function?

src/core/moves/compound/RateAgeBetaShift.cpp
davidcerny: Maybe I'm misunderstanding, but did I not provide such an interface?


We allow the user to specify a `tuneTarget` argument on the Rev side (by inheriting it from `Move`):

https://github.com/revbayes/revbayes/blob/d6db4f18863d5ce86fef902fe7a30249b992a8c5/src/revlanguage/moves/tree/Move_RateAgeBetaShift.cpp#L160-L161

We extract its value:

https://github.com/revbayes/revbayes/blob/d6db4f18863d5ce86fef902fe7a30249b992a8c5/src/revlanguage/moves/tree/Move_RateAgeBetaShift.cpp#L66

And pass it as an additional parameter to the modified constructor:

https://github.com/revbayes/revbayes/blob/d6db4f18863d5ce86fef902fe7a30249b992a8c5/src/revlanguage/moves/tree/Move_RateAgeBetaShift.cpp#L92

src/core/moves/proposal/tree/BranchLengthScaleProposal.cpp
davidcerny: It's initialized to the correct value in the Rev-language front end:

https://github.com/revbayes/revbayes/blob/d6db4f18863d5ce86fef902fe7a30249b992a8c5/src/revlanguage/moves/tree/Move_BranchLengthScale.cpp#L126-L127

src/core/moves/proposal/tree/SpeciesSubtreeScaleBetaProposal.cpp
davidcerny: It's initialized to the correct value in the Rev-language front end:

https://github.com/revbayes/revbayes/blob/4983e3d3feecd9274cf1e9d76832ce075937c56e/src/revlanguage/moves/tree/Move_SpeciesSubtreeScaleBeta.cpp#L202-L203

src/core/moves/proposal/tree/SpeciesTreeScaleProposal.cpp
davidcerny: It's initialized to the correct value in the Rev-language front end:

https://github.com/revbayes/revbayes/blob/4983e3d3feecd9274cf1e9d76832ce075937c56e/src/revlanguage/moves/tree/Move_SpeciesTreeScale.cpp#L205-L206

src/core/moves/proposal/tree/TreeScaleProposal.cpp
davidcerny: It's initialized to the correct value in the Rev-language front end:

https://github.com/revbayes/revbayes/blob/4983e3d3feecd9274cf1e9d76832ce075937c56e/src/revlanguage/moves/tree/Move_TreeScale.cpp#L152-L153

src/core/moves/proposal/vector/ElementScaleProposal.cpp
davidcerny: It's initialized to the correct value in the Rev-language front end:

https://github.com/revbayes/revbayes/blob/4983e3d3feecd9274cf1e9d76832ce075937c56e/src/revlanguage/moves/vector/Move_ElementScale.cpp#L146-L147

src/core/moves/proposal/vector/MultipleElementScaleProposal.cpp
davidcerny: It's initialized to the correct value in the Rev-language front end:

https://github.com/revbayes/revbayes/blob/4983e3d3feecd9274cf1e9d76832ce075937c56e/src/revlanguage/moves/vector/Move_MultipleElementScale.cpp#L149-L150

src/core/moves/proposal/vector/ShrinkExpandProposal.cpp
davidcerny: It's initialized to the correct value in the Rev-language front end:

https://github.com/revbayes/revbayes/blob/4983e3d3feecd9274cf1e9d76832ce075937c56e/src/revlanguage/moves/vector/Move_ShrinkExpand.cpp#L170-L171

src/core/moves/proposal/vector/ShrinkExpandScaleProposal.cpp
davidcerny: It's initialized to the correct value in the Rev-language front end:

https://github.com/revbayes/revbayes/blob/4983e3d3feecd9274cf1e9d76832ce075937c56e/src/revlanguage/moves/vector/Move_ShrinkExpandScale.cpp#L157-L158

src/core/moves/proposal/vector/VectorScaleProposal.cpp
davidcerny: It's initialized to the correct value in the Rev-language front end:

https://github.com/revbayes/revbayes/blob/4983e3d3feecd9274cf1e9d76832ce075937c56e/src/revlanguage/moves/vector/Move_VectorScale.cpp#L116-L117

src/core/moves/proposal/vector/VectorSingleElementScaleProposal.cpp
davidcerny: It's initialized to the correct value in the Rev-language front end:

https://github.com/revbayes/revbayes/blob/4983e3d3feecd9274cf1e9d76832ce075937c56e/src/revlanguage/moves/vector/Move_VectorSingleElementScale.cpp#L112-L113

src/core/moves/proposal/AVMVNProposal.cpp
davidcerny: It's initialized to the correct value in the Rev-language front end:

https://github.com/revbayes/revbayes/blob/4983e3d3feecd9274cf1e9d76832ce075937c56e/src/revlanguage/moves/Move_AVMVN.cpp#L383-L384

davidcerny:
@bredelings I tried to address your individual comments as best I could, so here's just a few answers to your more general points:

> In some cases the optimal value is known -- for example I think `0.44` is the optimal acceptance rate for Gaussian proposals on 1D Gaussian posteriors, and `0.234` is the optimal acceptance rate on high-dimensional Gaussian posteriors.

Interesting! I always thought these values were just rules of thumb; I didn't know there was solid math behind them. In that case, though, it's ironic that we currently only allow changing the `tuneTarget` for simple moves like `mvScale` and `mvSlide` where there is no reason to deviate from a known optimum, and don't allow it for more complex moves where the optimum is unknown.

Hopefully, this PR helps change that, even if it does so by erring on the side of giving the user too much freedom (by allowing them to overwrite the default optimum value with their own suboptimal one, if they so choose).

> Making the class always use `targetAcceptanceRate` improves coherence and makes the acceptance rate always tweakable. But in order to preserve the information about the default acceptance rate I think we need to also set the targetAcceptanceRate value to the previously fixed value in the constructor. 

Absolutely...

> Um, now I see that for the proposals that used 0.234 instead of 0.44 you put the default at the RevLanguage level.. so that's fine.

Yes indeed. :)

davidcerny:
Hmm... I would have sworn I had to modify the expected output of the `FBD_dating_ex3.Rev` test to make it pass, but apparently that's not the case. That's good – it means my changes to `mvRateAgeBetaShift` are less invasive than I thought!

davidcerny:
... but it does bring up the question of why the expected output had to be replaced after changing `mvAVMVN`. Actually, there is a simple answer to that. In `AVMVNProposal::tune()`, we originally had the following line:

https://github.com/revbayes/revbayes/blob/db9aca3168a4c71199b28d6b2738917c1de79da7/src/core/moves/proposal/AVMVNProposal.cpp#L648

There is a small arithmetic error in there – the second number is supposed to be $1 - 0.234 = 0.766$, not $0.764$. Goes to show why hardcoded numeric values are a bad idea!

bredelings (approved):

PR #952, opened 2026-02-24
Make `mrcaIndex()`/`getMRCA()` work with `BranchLengthTree`s

davidcerny:
I was mildly horrified at the thought that we don't have a free-standing `getMRCA()` function – I had to grep through the documentation to find out that we do actually have it, but call it `mrcaIndex()`. Unfortunately, it can only be called on `TimeTree`s. I don't think there is a good reason for that.

This PR:

1. Makes `mrcaIndex()` work for both `TimeTree`s and `BranchLengthTree`s;
2. Adds `getMRCA()` as its alias.
3. Adds a help page for the function.

### DIFF
--- a/help/md/mrcaIndex.md
+++ b/help/md/mrcaIndex.md
@@ -2,8 +2,18 @@
 mrcaIndex
 ## title
 ## description
+Returns the index of the most recent common ancestor (MRCA) of a given `clade`
+in a given `tree`.
 ## details
 ## authors
+Michael Landis
 ## see_also
+tmrca
 ## example
+    # Read in a tree
+    tr <- readTrees(text="(((Lemur,Galago),Tarsius),((Pan,Homo),Alouatta));")[1]
+    
+    # Find the node index of the MRCA of Pan and Homo
+    mrcaIndex(tr, clade("Homo", "Pan"))
+    
 ## references
--- a/src/core/datatypes/trees/Tree.cpp
+++ b/src/core/datatypes/trees/Tree.cpp
@@ -2342,7 +2342,7 @@ void Tree::suppressOutdegreeOneNodes(bool replace)
 
 void Tree::unroot( void )
 {
-
+    
     if ( isRooted() == true )
     {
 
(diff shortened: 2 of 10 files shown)


bredelings:
I think maybe not defining MRCA for branch-length trees could be intentional, since MRCA doesn't make sense for unrooted trees.  For unrooted trees you need to specify 2 taxa AND the root (so, three taxa) to get a unique MRCA.  

In theory its possible for  branch-length trees to be either rooted or unrooted, whereas time-trees are always rooted.  I guess you could modify the MRCA function to throw an exception if the `tree.isRooted()` return false...

davidcerny:
Hmm... I don't find that very convincing:

1. No other function of this kind throws an exception when asked to find a MRCA in an unrooted tree. In R, `ape::getMRCA()`, `phytools::findMRCA()`, and `castor::get_mrca_of_set()` all handle unrooted trees just fine. It's been a while since I last used Python with ETE Toolkit, but as far as I can see, their `.get_common_ancestor()` method doesn't discriminate between rooted and unrooted trees, either.

2. The original motivation behind this PR was precisely an attempt to root an unrooted tree. The idea is to: (1) find the MRCA of two tips, (2) extract its index, and (3) re-root the tree on the clade originating with the node we just found. Obviously, this will not work if we can't perform step (1) on an unrooted tree to begin with.

bredelings:
Great, this looks good now.  I'm glad you added the .isRooted() member function, as well as making .unroot() work for non-clock trees.

One issue with our code (not this PR per se) is that if the root has 2 children, then we always remove it when we unroot.  So there's no way to simply mark the tree as unrooted without modifying it.  Eventually we should probably add a bool flag to the `unroot()` method to control this behavior or (better yet) make a separate function to remove degree two nodes.

One other issue is that I've been thinking about how to allow RevBayes to more flexibly operate on and modify trees.  Right now we use these node indices, but any tree operation can change all index.  I think we might need a way to have stable node and branch identifiers... but we probably cannot eliminate the current unstable indices.  So then we would need a way to specify whether a function should return an index or a stable-id.  And getMRCA (unlike mrcaIndex) would not specify which is returned.

Having mentioned both of those things, I'm going to go ahead and approve the PR as-is :-P

bredelings (approved):
Looks good, very nice!

basantakhakurel (approved):

davidcerny:
> One issue with our code (not this PR per se) is that if the root has 2 children, then we always remove it when we unroot. So there's no way to simply mark the tree as unrooted without modifying it.

Hmm... I'm not quite sure why that would be an issue? I can't imagine a situation where we would like to keep a degree-2 node in an unrooted tree, or what interpretation one would attach to such a node.

On the other hand, I was briefly proud of myself for generalizing our `unroot` method to root nodes with $>2$ children, before I realized that node removal is not really what I want to happen in such cases. The difference between a rooted tree with a trichotomy at the root and a strictly bifurcating unrooted tree is just a matter of interpretation – the corresponding Newick strings look exactly the same. In this case, the `unroot` method should therefore also change just the interpretation (by calling `setRooted(false)`) and not the tree itself – i.e., we (almost certainly) don't want to end up with a tree with $2n - 4$ nodes and a degree-4 node.

So the current solution is to make `unroot` always remove a root node with 2 children, but never a root node with more than 2 children. This should behave as expected in all cases I can think of, though there might of course be edge cases I'm missing.

bredelings:
> Hmm... I'm not quite sure why that would be an issue? I can't imagine a situation where we would like to keep a degree-2 node in an unrooted tree, or what interpretation one would attach to such a node.

Two interpretations are (i) having a taxon label there and (ii) dividing a branch into smaller segments that have different Q matrices.  Really its hard to list all possible interpretations.  I'm sure there are many interpretations that I have not imagined yet.  So automatically removing a node that the user has put there could have be dangerous.

PR #959, opened 2026-03-10
Fix ccache misses

bredelings:
Some changes to decrease the number of times we have no ccache to load.

These were from the website repo, and I needed them to avoid rebuilding revbayes from scratch on every website change.

There's also a change to release.yml to avoid getting out of sync with build.yml

The basic idea is that:
* we shouldn't keep multiple caches per branch, because we only have 10Gb of cache space total
* we shouldn't make caches for things like refs/pull/816/merge, but only for actual branches

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -107,7 +107,7 @@ jobs:
     container:
       image: ${{ matrix.image || '' }}
     steps:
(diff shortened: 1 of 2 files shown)


davidcerny (approved):
Looks good to me, and should be useful.

PR #320, opened 2023-01-27
Dev beagle clean

spydr073:
This branch allows for users to choose to use BEAGLE as a backend for likelihood calculations. There are 2 additional dependencies that must be installed to compile with BEAGLE support:  `BEAGLE-LIB` and `Eigen3`.

Implemented:
- Selecting and using BEAGLE enabled devices (cpu, sse, avx, gpu, opencl, cuda)
- Settings for BEAGLE scaling, float/double, threading
- Likelihood calculations for _single_ matrix substitution models (JC, HKY, F81, GTR, etc.)
- Among Site Rate Variation (ASRV)
- Invariable sites (+I)
- Choice of using eigensystems or per-node transition matrices for likelihood calculations (eigen way is prefered)

Partially Implemented:
- Currently there is a problem using ASRV and the per-node transition matrices backend (ASRV currently just works with the eigen backend)
- Mixture Models are partially implemented. I was unable to figure out how to set the category weights for the mixtures properly in BEAGLE. 

Not Implemented:
- Partitioned analyses
- Using BEAGLE with MPI

To compile: 
- ./build.sh -beagle true -eigen true

Commands and options to use in Rev scripts to use BEAGLE:

```
beagleInfo()
setOption("useBeagle", <boolean>)
setOption("beagleDevice", <natural>)
setOption("beagleResource", <beagle_resource>)
setOption("beagleDoublePrecision", <boolean>)
setOption("beagleMaxCPUThreads", <natural>)
setOption("beagleScalingMode", <scaling_mode>)
setOption("beagleDynamicScalingFrequency", <natural>)
```

beagle_resource options:

```
cpu
cpu_sse
cpu_avx
gpu_opencl
gpu_cuda
```

scaling_mode options:

```
manual
auto
always
dynamic
```

When using dynamic scaling, users may also want to set the `beagleDynamicScalingFrequency`.
The default scaling frequency is 100.

PR #960, opened 2026-03-12
Dev beagle

bredelings:
Add BEAGLE functionality.


davidcerny:
The GitHub Actions already fail on the tutorial tests before even getting to the CI test suite, but the latter shows several failures as well:

- `test_BDSTP` – same likelihoods for the initial state (generation 0) but different likelihoods for all subsequent generations.
- `test_DEC` – different likelihoods, including for the initial state (generation 0). Note also that this test currently gives the same output on macOS and Linux, but for this branch, the output diverges between the two platforms.
- `test_FBD` – multiple different subtests fail, but interestingly, the skyline FBD test does not.
- `test_Partition` – only the analysis partitioned by codon produces different output, while the one partitioned by gene does not. So the partitioning problems noted by Killian more than three years ago (#320, #321) still persist, although it doesn't seem to be the case that partitioning is impossible altogether.
- `test_negCladeConst` – fails with the following error:

    ```
           |     Error:	Possible bug: Invar sites with ambiguous chars at site 933 out of bounds! Site was Gap
           |     Error:	Problem processing line 82 in file "scripts/negativeCladeConstraint.Rev"
    ```
    
- `test_propose_bad_index` – different likelihoods once again, but it can take multiple generations for them to diverge.

Issue #237, opened 2022-04-26
[BUG] Cyclic DAG


hoehna:
**Describe the bug**
Re-assignments that cause a cyclic DAG are not prevented and lead to a segfault

**To Reproduce**
> a ~ dnUnif(0,10)
> b := ln(a)
> a := exp(b)

**Expected behavior**
This should throw an error instead of crashing.

bredelings:
Interesting.... I think some languages handle this by making the second `a` refer to a new node.  In that situation, `b` refers to the node of the original `a`, which still exists but is not called `a` anymore.  With those semantics, the DAG ends up being acyclic.

To keep the original semantics, you could analyze the graph to find cycles, which takes some CPU time.  Maybe you have a better idea, but I suppose you could look for cycles only if it is a RE-definition and not a first definition?

hoehna:
Most likely, this happens when someone, like me, forgets to comment out a line of code. I had two versions on how to specify the model with a slightly different parameterization, i.e., prior on a transform or not. I still wanted to monitor the actual rate and its transform.

This means, renaming the variable would have caused me very unexpected and hard to debug behavior.

Yes, this indeed should only happen in a re-definition. Perhaps then we should look for dependencies. Our graphs are usually not that large that this should be problematic, but I can see this becoming a slowdown in loops.

davidcerny:
Fixed by #962.

Issue #243, opened 2022-05-30
[BUG] Segmentation fault when creating a cycle


NathanKolbow:
**Describe the bug**
Creating a simple cycle between deterministic nodes causes a seg fault.

**To Reproduce**
z <- 3
s := z*2
z := s*2

**Expected behavior**
An error that disallows the user from creating a cycle.

**Computer info**
MacOS Monterey 12.4, RevBayes v1.1.1

bredelings:
Thanks for the bug report!  This is a duplicate of #237.  Let's continue the conversation there.

PR #962, opened 2026-03-15
Detect cycles in the model graph

davidcerny:
While investigating #958, I kept running into a segfault that was hard to track down and ultimately turned out to be due to (a non-obvious case of) a cycle in the graphical model. As pointed out already in #237, it would be good if in such a case we could throw an informative exception instead of crashing.

This (partially vibe-coded) solution does exactly that. Cycles are detected as soon as they are created – i.e., in an interactive session, one doesn't have to wait until the final call to `model()` to figure out that something's gone wrong. For the example from #237, we get:

```
> a ~ dnUnif(0,10)
> b := ln(a)
> a := exp(b)
   Error:	Cycle detected in DAG during touch propagation: node 'b' is both a child and an ancestor of 'a'.
```


bredelings:
Hmm.... part of the problem here is that the second `a := ...` modifies the definition of the original `a` node.  What the snippet really should do is introduce a new deterministic node and make the name `a` point to it.  However, this is trick to do with something like `a[1] := ...`.

davidcerny:
Both issue submitters (@hoehna and @NathanKolbow – see #237, #243) stated that the behavior they expected in this situation was for RevBayes to throw an error, rather than try to resolve the cycle by implicit variable renaming. This makes sense to me, since cycles are most likely to result from errors in model setup. So I do not agree that

> the snippet really should [...] introduce a new deterministic node and make the name `a` point to it

and prefer the behavior implemented in this PR.

bredelings:
I don't think you understood what I was saying.  I wasn't expressing a problem with this PR, but I do think that RevBayes has a design flaw in allowing variable definitions to have backwards effects.  We have talked about changing this before, and there are arguments pro and con, independent of the cycles.  The fact that this behavior allows introducing cycles seems like another con.

Changing the general behavior would solve a number of problems where we tell people that they cannot do things that you would do in R or C++, such as writing for loops with local variables:

```
for(i in 1:10)
{
    tmp := x[i]*y[i]                                             # you cannot do this in RevBayes, because RevBayes is weird
    z[i] ~ dnNormal(tmp*tmp + tmp,1)
}
```

That is a second "con".  Right now we tell users that they should learn to avoid this, but in my opinion it would be better if it works.  

Anyway, changing this general behavior would only happen in the far future, if at all. So for the present, the question is only how best to prevent creation of cycles.  Long-term, though, I'm not sure that we should allow the current behavior that introduces cycles in the first place.

bredelings:
Another interesting option would be to warn or error when anything is redefined at all, not just for cycles.  Possibly this behavior could be controlled by an option.  For example, in an interactive session, people might want to revise the definition of a variable.  However, in a script this is almost always a mistake, even if it does _not_ create cycles.

davidcerny:
Okay, yes, it wasn't clear to me that that was the point you were making. Maybe this broader point about design flaws in Rev would be better suited for a discussion thread.

For now, I will just say that while the idea of making it impossible to create cycles in the first place sounds elegant in theory, I'm not sure what it would imply in practice – i.e., what would be the actual behavior users would encounter on inputting clearly malformed code such as:

```r
a ~ dnUnif(0,10)
b := ln(a)
a := exp(b)
```

It still seems to me that there are just two options here – either we throw an exception, or we retain both the initial stochastic node and the subsequent deterministic node, and make a decision about which one gets to keep the name `a`. Since I find the latter behavior undesirable, it's unclear to me how even a fundamental change to Rev semantics could improve on throwing an exception in this particular case.

davidcerny:
In the discussion under #237, there were some concerns about potential performance hits incurred by checking the graph for cycles. To quantify this, I ran the [large model graph test](https://github.com/revbayes/revbayes/blob/master/tests/test_large_model_graph/scripts/mcmc_large_normal_model.Rev) (with 15,000 stochastic nodes) on my machine (an M3 Max MacBook Pro) using both the current tip of development and this branch, and recorded the resulting runtimes:

```
branch        | user   | system | total
--------------|--------|--------|--------
development   | 44.99s | 16.96s | 1:02.06 
detect-cycles | 44.33s | 17.14s | 1:02.22
```

Based on this, the computational cost of the check appears to be negligible. However, perhaps I haven't chosen the most challenging example – the test in question involves a graph that is large but structurally simple.

src/core/dag/DagNode.cpp
```diff
@@ -1100,18 +1146,41 @@ void DagNode::touch(bool touchAll)
 
 /**
  * Tell affected variable nodes to touch themselves (i.e. that they've been touched).
+ * Throws if a cycle is detected (e.g. two deterministic nodes that depend on each other).
  */
 void DagNode::touchAffected(bool touchAll)
 {
-    // touch all my children
+    if ( nodes_currently_touched.count( this ) != 0 )
```
bredelings: Why did you decide to look for cycles here -- inside `touchAffected( )`?  Is it because `touchAffected( )` was the function that crashed when there was a cycle?

src/core/dag/DagNode.cpp
```diff
@@ -13,6 +15,18 @@
 
 using namespace RevBayesCore;
 
+// Anonymous namespace to keep the two sets below private to this file (i.e., no other file can mutate them)
+namespace {
+
+    // Nodes currently in the touch propagation stack (used to detect cycles).
+    thread_local std::set<DagNode*> nodes_currently_touched;
```
bredelings: Adding `thread_local` here seems a bit extreme.  I have never actually seen this used.

It seems like the real problem is that there is global state.  This causes problems with multithreading since the threads share the global state.  Have you considered alternative ways of coding this that don't use global state?

src/core/dag/DagNode.cpp
```diff
@@ -13,6 +15,18 @@
...
+
+    // Nodes currently being visited by isConstant() (used to detect cycles among parameters).
+    thread_local std::set<const DagNode*> is_constant_visiting;
```
bredelings: Why is this called `is_constant_visiting`?  What does `is_constant` have to do with this?

bredelings:
One issue that occurs to me is that trying to make `x[i]` work when `i` is deterministic would create cycles between where `x[i]` depends on `x` and `x` depends on `x[i]`.  However, its possible that implementing `[](x,i)` so that it does NOT depend on `x` would circumvent this problem.

src/core/dag/DagNode.cpp
davidcerny: In part, yes. My understanding (which is admittedly AI-aided, so it should be approached with a healthy dose of skepticism) is that there are two recursive model graph traversals that will loop forever on a cycle, and which therefore represent natural detection points. `DagNode::touchAffected()` is one of them, and it walks _down_ the graph: it propagates the "dirty" flag from a node to its children, recursively. If it encounters a cycle, it will cause a stack overflow. The other such traversal is `DeterministicNode::isConstant()`, which walks _up_ the graph. Since a deterministic node is only constant if all of its parents are constant, and since its immediate parents may themselves be other deterministic nodes, the function walks recursively from a child node to its parents. It will also result in a stack overflow if there is a cycle.

This PR inserts cycle detection code into both of these traversals. I'm not 100% sure that's necessary – I'm having a hard time thinking of a situation where a cycle might be caught by one but not the other. I guess I think of it as an extra line of defense that comes at very little additional cost.

src/core/dag/DagNode.cpp
davidcerny: Yes, I agree, this was not a good solution. I have eliminated the shared global sets (intended to store all nodes currently being visited by a given graph traversal) in favor of per-node flags that indicate whether a given node is currently inside one of the two traversals we monitor for cycles (`touchAffected` and `isConstant`). These flags were implemented as `mutable bool` members of class `DagNode`, so each node owns its own flags, requiring no global state. As a bonus, boolean checks are simpler than set lookups.

A potential complication to account for is that the traversal could throw an exception partway through, in which case we should clear the flag. The current code accomplishes that using a resource-allocation-is-initialization guard pattern, where `CycleGuard` clears the flag in its destructor as the exception propagates up the call stack. Once again, this might be overly cautious – I'm not sure there are any other exceptions `touchAffected` and `isConstant` could throw besides cycle errors, and since cycle errors are fatal, the DAG won't be used again, so the cleanup might not be necessary.

I'm not 100% sure if this, too, could cause problems with multithreading. Do we ever expect two threads to simultaneously traverse the same node?

src/core/dag/DagNode.cpp
davidcerny: See above. `DeterministicNode::isConstant()` performs a traversal of the model graph, and the (now eliminated) `is_constant_visiting` set was supposed to store all the nodes visited by it.

davidcerny:
> One issue that occurs to me is that trying to make `x[i]` work when `i` is deterministic would create cycles between where `x[i]` depends on `x` and `x` depends on `x[i]`. However, its possible that implementing `[](x,i)` so that it does NOT depend on `x` would circumvent this problem.

I hope that's true! But it might be more relevant for PR #964 than for this one.

davidcerny:
Note to self: as pointed out by @hoehna, we still need to ensure that the following works when we run it with the `-c` (`c`ontinue on error) flag:

```r
a ~ dnUnif(0,10)
b := ln(a)
a := exp(b)
a
b
```

Currently, we get another segfault when we ask for the value of `b`. Asking just for the value of `a` is okay-ish: we get an error message stating that `Variable a does not exist`. However, ideally, we would like to preserve both the original assignment of `a` and `b`, so that we can pick up where we left off – give ourselves another chance to construct a valid DAG, so to speak.

@hoehna thinks that a comprehensive solution must ultimately be implemented in the Rev parser, and not the DAG machinery as done here.

src/core/dag/DagNode.cpp
```diff
@@ -1104,14 +1156,9 @@ void DagNode::touch(bool touchAll)
 void DagNode::touchAffected(bool touchAll)
 {
     // touch all my children
-    for (std::vector<DagNode*>::const_iterator it=children.begin(); it!=children.end(); ++it)
+    for (std::vector<DagNode*>::const_iterator it = children.begin(); it != children.end(); ++it)
```
bredelings: Not important, but you _could_ just do `for(DagNode* child: children)`

davidcerny:
Okay, I believe I now have a much more robust solution that puts the cycle detection step in the parser, i.e., `SyntaxAssignment::evaluateContent()`. We use a boolean flag called `slot_was_occupied` to check if a given "slot" (variable) held a value before the attempted (re-)assignment. As before, we try to call `assign()`, which in turns calls `RevVariable::replaceRevObject()`, which finally calls `DagNode::replace()`. We modified this last function to include as its first step a call to the newly defined `dependsOn()` member function, which recursively walks up the model graph – i.e., from a child to its parents – and checks whether the node in question depends on itself (though the function is general enough to allow checking for a dependency on an arbitrary node). If we find out that it does, we throw an exception, which we then catch inside `SyntaxAssignment::evaluateContent()`. We then erase the variable we tried to create, but only if the `slot_was_occupied` flag is false. If it is true, we let the attempted re-assignment fail but keep the variable – associated with its previously assigned value – in the workspace.

This solution has a number of advantages:

(1) After a failed attempt to create a cycle, we can pick up where we left off. The code chunk below now works as intended (as also demonstrated by the updated integration test):

```r
a ~ dnUnif(0,10)
b := ln(a)
a := exp(b)
a
b
```

(2) It is computationally cheap. Here is how long it takes to run the `large_model_graph` test (with 15,000 stochastic nodes) on my machine (M3 Max MacBook Pro) both using the current tip of development (0dcb717) and the branch this PR proposes to merge:

```
                     user                         system                        total
          ---------------------------   ---------------------------   ----------------------------
Attempt | development | detect-cycles | development | detect-cycles | development | detect-cycles
--------|-------------|---------------|-------------|---------------|-------------|---------------
1       | 42.67 s     | 43.50 s       | 17.55 s     | 17.43 s       | 1:01.15     | 1:01.58
2       | 43.51 s     | 43.98 s       | 16.27 s     | 17.46 s       | 59.860      | 1:01.52
3       | 44.18 s     | 43.71 s       | 17.04 s     | 16.96 s       | 1:01.32     | 1:00.74
```

(3) Since we already have a `slot_was_occupied` flag, it would be straightforward to extend the current system to deliver an optional warning even for non-cyclic redefinitions. We'd just have to decide how to make the warnings optional, and perhaps also if we want to distinguish between different types of redefinitions (e.g., giving the same constant node a different value feels less likely to be wrong than redefining a constant node to be deterministic).

src/core/dag/DagNode.cpp
davidcerny: Done.

davidcerny:
As pointed out above, the `large_model_graph` test might not be the most challenging scenario for the sort of recursive search of the model graph that this branch implements, because the DAG in question is so structurally simple. A more interesting example might be obtained by modifying it so that each successive node depends on the previous one:

```r
seed(12345)

mvi = 0
n = 15000 # the number of variables in the model

x[1] ~ dnNormal(0, 1)
moves[++mvi] = mvSlide(x[1])

# make the variables
for(i in 2:n) {
    x[i] ~ dnNormal(x[i-1], 1)
    moves[++mvi] = mvSlide(x[i])
}

# make the model
my_model = model(x)

# make the monitors
monitors[1] = mnModel(printgen = 10, filename = "output/large_normal_model.log")

my_mcmc = mcmc(my_model, monitors, moves, moveschedule = "single")

# run the mcmc
my_mcmc.run(generations = 1000)

# quit
q()
```

Once again, there is no detectable difference in performance between `detect-cycles` and `development`:

```
                     user                         system                        total
          ---------------------------   ---------------------------   ----------------------------
Attempt | development | detect-cycles | development | detect-cycles | development | detect-cycles
--------|-------------|---------------|-------------|---------------|-------------|---------------
1       | 41.67 s     | 41.34 s       | 16.53 s     | 16.41 s       | 58.885 s    | 57.826 s
2       | 40.35 s     | 39.64 s       | 15.43 s     | 15.68 s       | 55.836 s    | 55.380 s
3       | 40.16 s     | 40.38 s       | 16.06 s     | 15.63 s       | 56.271 s    | 56.059 s
```

bredelings:
So, if we use the attached script [order-two.Rev.txt](https://github.com/user-attachments/files/26312241/order-two.Rev.txt), then we see that the first time we define the x[i], we cover all 1000 iterations very quickly.  However, when we redefine `x[i]` it gets slower every iteration, and bogs down completely around iteration 40 or so.

The reason seems to be that the first time around `RevVariable::replaceObject( )` doesn't call `DagNode::replace(...)`   because `rev_object` is null:  https://github.com/revbayes/revbayes/blob/db9aca3168a4c71199b28d6b2738917c1de79da7/src/revlanguage/workspace/RevVariable.cpp#L369 .
Therefore `DagNode::dependsOn( )` is never called.

However, when we redefine `x[i]` it seems that `rev_object` is not null, and so we _do_ call `DagNode::replace( )` -> `DagNode::dependsOn( )`, and we quickly bog down.  Perhaps it is OK to skip the call to `DagNode::replace( )` when  `rev_object == NULL` because this means we are defining a variable for the first time, so there can't be any cycles?

bredelings:
I fixed an issue with the `dependsOn(...)` function that could cause exponential time complexity in some cases.  With this fixed, we still see quadratic time complexity when we have (i) linear chains of variable dependencies and (ii) the variables are redefined.  But it is not as slow as before.  

If we have a chain of 10,000 variables (i.`x[i] ~ dnNormal(x[i-1],sigma)`) then setting up the model a second time takes about 1 second without the cycle detection, and 51 seconds with the cycle detection.

I'm not sure anyone would actually redefine 10,000 variables though.

src/revlanguage/parser/SyntaxAssignment.cpp
```diff
@@ -80,23 +81,24 @@ RevPtr<RevVariable> SyntaxAssignment::evaluateContent( const std::shared_ptr<Env
     // Get variable slot from lhs
     RevPtr<RevVariable> the_slot = lhsExpression->evaluateLHSContent( env, the_variable->getRevObject().getType() );
     
-//    // let us remove all potential indexed variables
-//    removeElementVariables(env, the_slot);
```
bredelings, line 84: Is it OK to remove this?

I guess this handles the case when the variable was initially an array?

src/revlanguage/parser/SyntaxAssignment.cpp
```diff
@@ -80,23 +81,24 @@ RevPtr<RevVariable> SyntaxAssignment::evaluateContent( const std::shared_ptr<Env
...
+    // Remember whether the slot already held a value before the assignment.
+    // If it did, a failed assignment should leave it intact rather than erasing it.
+    bool slot_was_occupied = ( &the_slot->getRevObject() != &RevNullObject::getInstance() );
+
     try
     {
         // now we delegate to the derived class
         assign(the_slot, the_variable);
-        
-//        if ( the_slot->isElementVariable() == true )
```
bredelings, line 91: Is it OK to remove this?

It looks like this handles the case where this variable is `x[3]`.  When we assign it we need to update `x`?

src/revlanguage/parser/SyntaxAssignment.cpp
bredelings: Oh, I guess it was commented out 😬

src/revlanguage/parser/SyntaxAssignment.cpp
bredelings: Oh, I guess it was commented out 😬

bredelings:
So I asked Claude if it understood why the algorithm here was quadratic, and if there was any way to make the order of computation better.

Apparently there are several algorithms in this area.  The simpler ones, such as the Pearce-Kelly algorithm,  use a "topological sort" as a quick check for whether you can add an edges from x to y without creating a cycle. (See https://en.wikipedia.org/wiki/Topological_sorting).  The general question is called "online cycle detection" or "online topological sort".  Here "online" means that you add edges to the graph one-at-a-time and see if each new edge creates a cycle.  Sometimes people also say "incremental cycle" instead of "online".

The Pearce-Kelly algorithm seems to be near-linear for sparse graphs, where each node is connected to only a few edges.  That should be the RevBayes situation.  So when adding an edge from x->y, if order[x] < order[y], then there are no cycles.  If order[x] > order[y], then the two tricks are:
- first, to check if a cycle is created, you look at the descendants of y to see if any of them are equal to x. But STOP whenever you see a node where order[x] > order[node].  So you don't need to look at ALL descendants of y, just some of them.  This group of descendant is "in the band".
- second, if no cycle is created, then you add x->y to the graph.  You need to update the topological order of the nodes "in the band" -- not ALL the nodes.  The order is actually maintained as something like std::list<DagNode*>, and each DagNode needs a new field `double order;` to indicate its order.  Claude recommended using `double` instead of `int` because it allows you to put a node in between `1` and `2` --- you just give it order `1.5`!

Anyway, that's a very short overview.  Claude can probably explain the details if you are interested.

davidcerny:
Many thanks for the pointers to the relevant literature, @bredelings!

I agree that the Pearce-Kelly algorithm – which brings the complexity down from $\mathcal{O}(n^2)$ to $\mathcal{O}(n^{3/2})$ – is probably the best realistic solution to the problem here. [Bhattacharya & Kulkarni (2020)](https://epubs.siam.org/doi/abs/10.1137/1.9781611975994.153) apparently managed to improve on it with an $\mathcal{O}(n^{4/3})$ algorithm, and [Chen et al. (2024)](https://dl.acm.org/doi/abs/10.1145/3618260.3649745) brought the complexity down even further all the way to $n^{1 + \mathcal{O}(1)}$, i.e., very nearly linear-time, but these algorithms are very complex and appear to be cutting-edge research results rather than something that's widely used in practice. (For the Chen et al. algorithm, I couldn't even find any public implementation at all.) In contrast, the Pearce-Kelly algorithm is good enough for Google's [Abseil library](https://github.com/abseil/abseil-cpp/blob/master/absl/synchronization/internal/graphcycles.cc) and for [Tensorflow](https://github.com/tensorflow/tensorflow/blob/master/third_party/xla/xla/service/graphcycles/graphcycles.cc), so it should be good enough for us, too.

I took a crack at implementing the Pearce-Kelly algorithm, with _lots of_ help from Claude Opus 4.7. It seems to be working well, as the (admittedly trivial) cycle in the [relevant test](https://github.com/revbayes/revbayes/blob/detect-cycles/tests/test_DAG/scripts/Test_cycle.Rev) is still detected, and no false positives occur in any of the other tests. I also tested it on the less trivial case from #958, which originally inspired me to revisit this issue, and it detects the cycle there, too:

```
   Error:	Cannot complete this assignment: 'branch_rates[255]' depends on 'branch_rates[255]', so replacing one with the other would create a circular dependency.
```

To get a sense of its performance, I ran it on a [modified version](https://github.com/user-attachments/files/27565195/variable_redefinition_test.txt) of your [`order-two.Rev`](https://github.com/revbayes/revbayes/blob/detect-cycles/tests/test_DAG/scripts/order-two.Rev) test script, where I increased the number of chained variables `N` from 3000 to 10,000, and where I removed the MCMC code (so that any runtime differences have to do strictly with model graph construction). Here is a comparison between the current tips of `detect-cycles` (c2a4f41) and `development` (9edfd16):

```
                     real                          user                        system
          ---------------------------   ---------------------------   ----------------------------
Attempt | development | detect-cycles | development | detect-cycles | development | detect-cycles
--------|-------------|---------------|-------------|---------------|-------------|---------------
1       | 7.381 s     | 18.769 s      | 6.618 s     | 18.633 s      | 0.088 s     | 0.125 s
2       | 6.676 s     | 18.675 s      | 6.593 s     | 18.570 s      | 0.076 s     | 0.099 s
3       | 6.714 s     | 18.878 s      | 6.648 s     | 18.759 s      | 0.058 s     | 0.112 s
```

I remember that you were seeing much more pronounced differences between the runtimes without cycle detection and with the original $\mathcal{O}(n^2)$ cycle detection algorithm (1 s vs. 51 s), so I wonder what the results are going to be for your test script / on your machine, @bredelings.

Finally, note that the new cycle detection algorithm only required changes to `DagNode.[cpp|h]`. The mechanism used in `SyntaxAssignment.cpp` is left untouched. In particular, it still includes the `slot_was_occupied` hook that we can potentially use to emit warnings on redefinition.

bredelings:
Wow, this implementation seems quite clean, simple, and uninvasive.  The complex case is when we need to "repair" the topological order.  I made an ASCII-art figure for this case:

```
Adding a new edge 6->3

      1
    /   \
   /     \
  2       5@
  |       |
  |       |
  3*<---- 6@
  |       |
  |       |
  4*      7
   \     /
    \   /
      8

  deltaF (*) = {3,4} 
  deltaB (@) = {5,6}
  pool = {3,4,5,6}
```

However, it is not nearly as fast as I was expecting.  For development I see about 1 second run times without the edge detection, and about 40-50 second run times with the original edge detection, and about 20-second run times with the Pearce-Kelly edge detection.  I feel like we should be able to get nearly no slowdown with the Pearce-Kelly algorithm, so I want to at least try to improve it before merging this.

It seems like this new code is spending a lot of time in pkAddEge, and a similar amount of time in std::sort. Since we create new nodes with the highest number topological order, ALL nodes will be after the new node.  This results in deltaB containing about 5 nodes, while deltaF contains about 28,000 nodes.  This means that we spend a lot of time adding the nodes into the deltaF vector, and then a lot of time sorting it.

Its possible that Claude is implementing the simplest possible version of Pearce-Kelly.  For example, I think it told me that some versions using `double` for the order number, since you can easily get a number between two other numbers without swapping existing numbers.  It also occurs to me that in order to determine WHETHER OR NOT there is a cycle you can go either forward from the child (deltaF), or backward from the parent (deltaB).  If the algorithm tries both at the same time (in an interleaved fashion perhaps) it would determine that there is no cycle very quickly -- after walking about 5 nodes backwards I think, since DeltaB has 5 nodes.  That leaves the issue of how to repair the topological order without visiting most of the graph nodes.

Anyway, I wonder if you can ask Claude how to improve the implementation to avoid walking so much of the graph.

(BTW, I just canceled my Claude Pro subscription, because I want to try some other models.  I decided to try ChatGPT Plus because I heard that Codex is interesting: https://medium.com/@cdcore/the-tale-of-2-models-opus-4-6-vs-gpt-5-3-codex-129fcb35630f  . So perhaps I can ask it some more questions on Wednesday.)

davidcerny:
Applying a frontier model to this problem made for a pretty interesting experience... I followed your original advice, @bredelings, and simply passed on your feedback to Claude. It actually pushed back on both of your suggestions. 🙂 

Let's start with the second suggestion: here, I find the counterargument clear and convincing, which is why I'm paraphrasing it even though I didn't come up with it myself. If we conduct the search in a bidirectional / interleaved fashion (i.e., if we build `deltaB` and `deltaF` in lockstep), we might improve on the time it takes to detect a cycle, but that's the case that's already very fast. If there is no cycle to be detected, then this doesn't help, because both the forward search and the backward search have to be allowed to run to completion. That's because – as you noted yourself – we need their results (the full `deltaB` and `deltaF` vectors) to perform the reassignment of ordinals that we use to repair the topological-order invariant.

Your other suggestion – if I understand it correctly – was that this repair could itself be improved by denoting the topological rank with `double`s instead of integer-valued ordinals, as you may then be able to just insert a new value between existing ones, as opposed to reassigning all the values already in use. Here, Claude's counterarguments were as follows:

1. This would run the risk of precision exhaustion. A `double` has only 53 bits dedicated to the significand, so after you compute the midpoint of a gap 53 times in a row, you've run out of available distinct values, and have to perform a reassignment anyway.
2. Even more importantly, changing the type of the ordinals doesn't free us from the need to change them not only for a single child node, but also for all its descendants.

However, the model agreed that the most promising way of achieving further efficiencies is to eliminate the `std::sort` calls from `pkAddEdge()`. It's possible to do this if the depth-first searches themselves can be made to collect nodes in (reverse) topological order – a property satisfied by the post-order traversal. I let Claude implement this suggestion, and it indeed yielded a modest but consistent speedup: on the test script from my last comment, it reduced the runtime from more than 18 s to just over 16 s. Again, not that great of an improvement, but a real one, so I pushed it.

(On a related note, I wonder what test script you have been using? You are seeing the Pearce–Kelly cycle detection increase the runtime by a factor of 20, while I was only getting a factor of 3 even before the latest round of changes.)

> I feel like we should be able to get nearly no slowdown with the Pearce-Kelly algorithm

Is that really true? I'm not sure it has a closed-form complexity in the number of nodes (my previous claim of $\mathcal{O}(n^{3/2})$ may have been an oversimplification), but it wouldn't surprise me if it was appreciably worse than linear.

That said, I think I've taken this as far as I can. I'm curious what Codex is going to tell you. Feel free to play around with the branch, and to push anything that improves performance.

bredelings:
I looked at your new version, and it goes from 21s -> 14s.  Nice!

OK, I investigated this further.  Claude and Gemini implied that it was possible to space out the numbers like 0,10,20,30 instead of 0,1,2,3 so that there are "gaps", and it might be possible to simply move the few nodes in deltaB into these "gaps" WITHOUT updating the large number of nodes in deltaF.

However, after asking for how you would actually maintain gaps and move nodes.... it seems it may require a very complicated "order-maintainence structure".  So it may not be practical, and it is also no longer Pearce-Kelly.  Or at least, that is what I think the AIs or telling me.  I'm sorry for wasting your time.

But I did find something else!  I'm trying ChatGPT with "extended" thinking, and I got it to describe further research after Pearce-Kelly.  I found an algorithm that works kind of differently than Pearce-Kelly called BFGT.   It doesn't need to maintain gaps.  The big differences from Pearce-Kelly are:
* the "order" is now called the "level".  This is because multiple different nodes can have the same level.
* the requirement is that level(parent) <= level(child)
* the repair and the cycle detection are both a bit more complicated 
* all new nodes start out with level 1

It is the last point that makes the big difference -- when we redefine x[i], this will do NOTHING instead of renumbering all downstream nodes.

I pushed a change -- the runtime is now 2.07 s, compared to 1.63 seconds for development.

bredelings:
I did a bit more reading on the cycle-detection problem.  It kind of seems like the LLMs are not really able to give good advice on this topic unless you steer them in the right direction.  But the information is out there.  For example: 
https://publications.lib.chalmers.se/records/fulltext/248308/248308.pdf

Anyway, I'm happy with the BFGT version, so if you are also happy with this PR, I'll approve it.

davidcerny:
Very impressive! Here is an updated runtime comparison for my test script and machine:

```
                     real                          user                        system
          ---------------------------   ---------------------------   ----------------------------
Attempt | development | detect-cycles | development | detect-cycles | development | detect-cycles
--------|-------------|---------------|-------------|---------------|-------------|---------------
1       | 7.400 s     | 9.914 s       | 6.730 s     | 8.836 s       | 0.075 s     | 0.092 s
2       | 6.837 s     | 8.981 s       | 6.752 s     | 8.888 s       | 0.077 s     | 0.089 s
3       | 6.871 s     | 8.891 s       | 6.786 s     | 8.809 s       | 0.079 s     | 0.074 s
```

I can also confirm that the algorithm successfully detects the cycle created by the script from #958.

Interestingly, when I asked Claude to give me an overview of online cycle detection techniques, it did list the BFGT algorithm – it scored it as practical to use and "medium-high" in implementation complexity (compared to "medium" for PK), and did mention it as "the next step" to try out if PK's performance proves to be insufficient. So thanks for taking this next step, and for providing a clean and relatively simple implementation!

I took the liberty of adding a comment to `DagNode.cpp` that (1) explains what "BFGT" stands for, which was not clear before; (2) cites the original Bender et al. paper, just so that we give credit where credit is due; and (3) provides a high-level overview of the algorithm, explaining how `addChild()`, `bfgtAddEdge()`, and `dependsOn()` fit together. I hope this is okay.

> Anyway, I'm happy with the BFGT version, so if you are also happy with this PR, I'll approve it.

I was just about to ask you if you think we need a review from a third person, since you ended up being a co-author of the PR, or if you are still comfortable approving it. :) If you are, please go ahead!

bredelings (approved):
I think this is fast enough now.

bredelings:
The comments about the algorithm at the beginning of the file were not quite correct -- there already was a correct description of the repair procedure in front of bfgtAddEdge, so I replaced them with a reference to the later comments.  I could have duplicated the comments, but ended up leaving them in front of the routine that does the work, for ease of reference.

A third reviewer could be nice -- who would it be?  I'm not sure... I have reviewed your code, and you have reviewed mine.  So, I approved, and if you approve of my changes, you can merge.  But if you have a third reviewer who is willing to look, then we could do that as well.

davidcerny:
I think it's probably okay. 🙂 Merging!

PR #964, opened 2026-03-15
Make workspace variable lookup work with deterministic indices

### LINKED ISSUE #963: [BUG] More problems with indexing by deterministic nodes
**Describe the bug**
This issue was first identified by @joelh251 in #958. When `tree` is a stochastic node, node indices such as `tree.parent(i)` become deterministic nodes, and we can't use them to index a vector.

**To reproduce**
The script and data in [this zipped directory](https://github.com/user-attachments/files/26005709/relaxed_brownian_rate_estimation.zip) are completely identical to those in the ["Relaxed Brownian Rate Estimation"](https://revbayes.github.io/tutorials/cont_traits/relaxed_bm.html) tutorial, except that I changed the following line:

```r
tree <- T
```

to

```r
tree ~ dnUniformTimeTree( rootAge=T.rootAge(), taxa=T.taxa() )
```

I.e., `tree` is now a stochastic rather than a constant node. Then, when we try to create a `branch_rates` vector as follows:

```r
for(i in nbranches:1) {

    < some code >

    # compute the rate for the branch
    if ( tree.isRoot( tree.parent(i) ) ) {
       branch_rates[i] := sigma2_root * branch_rate_multiplier[i]
    } else {
       branch_rates[i] := branch_rates[tree.parent(i)] * branch_rate_multiplier[i]
    }

    < other code >
}
```

We get the following error:

```
   Error:	No function named '[]'
```

**Expected behavior**
The code should work regardless of what type of node `tree` is.

**Computer info**
macOS 26.3 (Tahoe), ARM64 architecture, RevBayes 1.3.2, current tip of development (92174d5).

**Additional context**
What apparently happens is that upon encountering the deterministic node, we set `constant_index` to `false` in [`SyntaxIndexOperation::evaluateContent()`](https://github.com/revbayes/revbayes/blob/db9aca3168a4c71199b28d6b2738917c1de79da7/src/revlanguage/parser/SyntaxIndexOperation.cpp#L237), which causes the workspace lookup to fail. We then fall back on `getFunction("[]", args)`, which fails as well, because there is no matching overload.

davidcerny:
Fixes #963. If the base variable is a vector variable, the index type is `Natural` (or derived), and the element variable already exists in the environment, we no longer attempt the failing `[]` dispatch, but rather take the current value of the index, build the same identifier we would use for constant indexing, and look that variable up in the environment.

This feels slightly hacky, but it does solve the issue.

### DIFF
--- a/src/core/functions/vector/SelectVectorElement.h
+++ b/src/core/functions/vector/SelectVectorElement.h
@@ -1,0 +1,112 @@
+#ifndef SelectVectorElement_H
+#define SelectVectorElement_H
+
+#include "RbVector.h"
+#include "TypedFunction.h"
+#include "RbException.h"
+
+#include <cstddef>
+#include <cstdint>
+#include <string>
+#include <vector>
+
+namespace RevBayesCore {
+
+    template <class valueType>
+    class SelectVectorElement : public TypedFunction<valueType> {
+        
+    public:
+        SelectVectorElement(const TypedDagNode<std::int64_t>* index, const std::vector<const TypedDagNode<valueType>*>& elements);
+        virtual                                            ~SelectVectorElement(void);                                       //!< Virtual destructor
+        
+        SelectVectorElement*                                clone(void) const;                                               //!< Create an independent clone
+        void                                                update(void);
+        
+    protected:
+        void                                                swapParameterInternal(const DagNode* oldP, const DagNode* newP); //!< Implementation of swapping parameters
+        
+    private:
+        
+        // Members
+        const TypedDagNode<std::int64_t>*                   idx;
+        std::vector<const TypedDagNode<valueType>*>         elems;
+    };
+}
+
+
+template <class valueType>
+RevBayesCore::SelectVectorElement<valueType>::SelectVectorElement(const TypedDagNode<std::int64_t>* index, const std::vector<const TypedDagNode<valueType>*>& elements) : TypedFunction<valueType>( new valueType() ),
+    idx( index ),
+    elems( elements )
+{
+    if ( idx == nullptr )
+    {
+        throw RbException( "SelectVectorElement: index must not be NULL." );
+    }
+        
+    if ( elems.empty() )
+    {
+        throw RbException( "SelectVectorElement: elements must not be empty." );
+    }
+
+    // DAG wiring: depend on the index and *all* candidate elements. This keeps the DAG structure static while allowing the output to update
+    // when the selected index value changes.
+    this->addParameter( idx );
+
+    for ( const auto* e : elems )
+    {
+        this->addParameter( e );
+    }
+
+    update();
+}
+
+
+template <class valueType>
+RevBayesCore::SelectVectorElement<valueType>::~SelectVectorElement( void )
+{
+    // We don't delete parameters because they are owned by the model/DAG.
+}
+
+
+template <class valueType>
+RevBayesCore::SelectVectorElement<valueType>* SelectVectorElement<valueType>::clone(void) const
+{
+    return new SelectVectorElement<valueType>( *this );
+}
+
+
+template <class valueType>
+void RevBayesCore::SelectVectorElement<valueType>::update(void)
+{
+    // Natural indices in Rev scripts are 1-based; elems is stored 0-based.
+    const std::int64_t k = idx->getValue();
+    if ( k < 1 || k > static_cast<std::int64_t>( elems.size() ) )
+    {
+        throw RbException("Index out of range. Index = " + std::to_string( k ) + ", vector size = " + std::to_string( elems.size() ) + "." );
+    }
+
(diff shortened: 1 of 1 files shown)


src/revlanguage/parser/SyntaxIndexOperation.cpp
```diff
@@ -278,35 +280,46 @@ RevPtr<RevVariable> SyntaxIndexOperation::evaluateContent( const std::shared_ptr
...
+            if ( theParentVar->isVectorVariable()
+                 && indexVar->getRevObject().getTypeSpec().isDerivedOf( Natural::getClassTypeSpec() ) )
             {
-                theMemberMethod->setMemberObject( theParentVar );
-
-                the_var = the_function->execute();
-                the_var->setName( identifier );
-                the_var->setElementVariableState(true);
+                std::int64_t idx_val = static_cast<const Natural&>( indexVar->getRevObject() ).getValue();
+                std::string fallback_id = theParentVar->getName() + "[" + std::to_string( idx_val ) + "]";
```
bjoelle: so it looks like here we're setting the id based on the current value of the node, but I'm not sure if that would update properly when e.g. the tree gets rearranged ?

bredelings: yeah, it seems like we would need to create a deterministic node `[](v, i)`?

bredelings (changes requested):

davidcerny:
You're both 100% correct, and I should have recognized the problem before opening this PR. This is not a viable fix for #963.

I'd be happy to either (1) close this PR, or (2) try to turn it into a more ambitious overhaul of vector indexing that does actually help resolve the problem at hand. That would basically involve implementing what @bredelings suggested above:

- When `constant_index == false`, i.e., when the index is dynamic / model-valued, we would not try to resolve `[]` through `FunctionTable` as we currently do, but rather call a new Core function
- This function should take the vector of candidate nodes (say, `std::vector<const DagNode*>`) and a model-valued index (`TypedDagNode<int64_t>`) as its input, and return that element which corresponds to the current value of the index as its output
- In terms of touch propagation, we would probably want the selected node to be marked as dirty whenever _any_ element node of the vector changes.

Let me know which one of these options do you prefer, @bjoelle @bredelings.

bjoelle:
I feel like this is a common enough issue that we should expect it to pop up again - in fact doesn't it imply that any sort of autocorrelated clock is impossible to make work in RevBayes right now ?
so I would be in favour of fixing it, but admittedly I don't know enough about how the vector structure is currently handled to know how difficult it would be

davidcerny:
Yes, that sounds exactly right – @basantakhakurel told me just yesterday he had run into this issue before while trying to set up an autocorrelated clock.

I'll see if I can make some headway on the solution outlined in my previous comment.

bredelings:
I have known about this issue for a while -- I previously assumed that it could not be simply "fixed" because it is part of the design, not really a bug.  For example, the DAG node `[](x, i)` is performing a deterministic computation to decide what its children are, and is therefore no longer a graphical model!  In contrast, a real graphical model has a fixed graph relating the DAG nodes.  Currently RevBayes uses the (fixed) graph to decide what a node's children and parents are, which is part of deciding which nodes to "touch" and mark as "dirty".

But maybe it is possible to implement this?  I guess what you would need is for `[](x,i)` to be a deterministic node that looks inside the RevVariable structure for `x` at `x->vector_var_elements[i]`.  It is important for `[](x,i)` not to depend on `x`, since that would cause a dependency loop.

It seems like it would be pretty difficult, for two main reasons, of which the second one is more fundamental:

1. Looking at VectorIndexOperator, I see this:
```
VectorIndexOperator(const TypedDagNode< RbVector<valueType> >* v, const TypedDagNode<std::int64_t>* idx);
```
Here we see that the deterministic node only gets a TypedDagNode for the vector `v`, so it can't actually see the RevVariable!  This means that it would not be able to find the DagNode for v[i].  So we have a problem.  Currently, the access goes from RevVariable -> RevObject -> DagNode, but you cannot go in the reverse direction.

So, SyntaxIndexOperation is able to look up the the elements, because it has access to the RevVariable for `x`, but `[](x,i)` is not able to look up the elements, because it only has the DagNode.

2. If  `[](x,i)` has a deterministic index then the children of the DagNode are not constant, but depend on `i`.  We use these children to determine when to flag things dirty.  For example, if `i` changes from `i1` to `i2`, then we must arrange that `[](x,i)` gets marked dirty when `x[i2]` changes, but not when `x[i1]` changes.  However, if a proposal to change `i` from `i1` to `i2` is rejected, then we need to revert this dependency.

But currently, DagNode objects have a fixed set of `DagNode::children( )`, and we do not have any way of making the children change when a deterministic node changes, nor do we have any way of making them change back when an MCMC move is rejected.

Issue #965, opened 2026-03-17
Heap-buffer-overflow in `computeRootLikelihood` after topology moves


ms609:
# Heap-buffer-overflow in `computeRootLikelihood` after topology moves

> Report and debugging led by Claude Opus under human direction

## Summary

Topology moves (`mvNNI`, `mvSPR`) on a `treeAssembly` tree cause a heap-buffer-overflow in `PhyloCTMCSiteHomogeneous<StandardState>::computeRootLikelihood`. The overflow is exactly one element past the buffer allocated by `resizeLikelihoodVectors()`.

The bug is triggered when multiple `dnPhyloCTMC` distributions with variable-dimension rate matrices share one topology. It does not manifest with a single partitioned character matrix using fixed-dimension rate matrices (244/244 such jobs completed; ~94% of the variable-dimension jobs crashed).

Not a recent regression — both a Feb 13 and March 12 build (commit `92174d501`) are affected.

## Reproducer

Save as `nni-bug.Rev`:

```rev
# Reproducer: heap-buffer-overflow in computeRootLikelihood.
# Requires multiple dnPhyloCTMC with variable-dimension rate matrices.
# Crashes within minutes under ASAN; at larger scale (25 taxa, 50 chars)
# crashes without ASAN via glibc heap corruption detection.
#
# Usage: rb nni-bug.Rev [n_taxa] [n_char] [seed]

if (!exists("args")) args = v("10", "5", "42")
n_taxa = args[1]
n_char = args[2]
seed(args[3])

for (i in 1:n_taxa) taxa[i] = taxon("t" + i)

kMax = 6

function bigJC(k) {
  for (i in 1:kMax) {
    for (j in 1:kMax) {
      if (i <= k && j <= k) {
        if (i == j) {
          ret[i][j] = 0.0
        } else {
          ret[i][j] = abs(1.0 / (k - 1.0))
        }
      } else {
        ret[i][j] = 0.0
      }
    }
  }
  return fnFreeK(ret, rescaled = FALSE)
}

jcVec = v(bigJC(2), bigJC(3))
for (i in 3:(kMax - 1)) jcVec.append(bigJC(i + 1))

function rootFq(k) {
  ret = rep(1, k)
  for (i in (k + 1):kMax) ret.append(0.0)
  return Simplex(ret)
}
rfqVec = v(rootFq(2), rootFq(3))
for (i in 3:(kMax - 1)) rfqVec.append(rootFq(i + 1))

[...]

PR #966, opened 2026-03-17
Tree::reroot() must call setRoot() to rebuild nodes[], not assign root pointer directly

ms609:
Bug identified by Claude Opus AI – I follow its logic but can't vouch for the necessity of the fix.

## Problem

`Tree::reroot()` ended with a bare pointer assignment:

```cpp
// set the root
root = &n.getParent();
```

This updates the raw `root` pointer but leaves `nodes[]` and `num_nodes` in
their pre-reroot state — still ordered and indexed relative to the old root.
Any subsequent operation that uses `nodes[]` by index will work on stale data.
This includes likelihood buffer arithmetic, `getNode(i)`,
`orderNodesByIndex()`, and anything downstream of `getNumberOfNodes()`.

## Fix

Replace the bare assignment with:

```cpp
// set the root — use setRoot() so nodes[] and num_nodes are rebuilt,
// not just the raw root pointer.
setRoot( &n.getParent(), reindex );
```

`setRoot()` calls `fillNodesByPhylogeneticTraversal()` to reconstruct
`nodes[]` from the new root, and either reindexes all nodes (`reindex=true`)
or re-sorts by existing indices (`reindex=false`). The `reindex` parameter is
already threaded through `reroot()` from all call sites, so no interface
changes are needed.

## Note

This bug is distinct from the `computeRootLikelihood` buffer overflow fixed in
the companion PR (linked to #965). The `reroot()` code path is not triggered by `NNI` or `SPR`
topology proposals (which call `reverseParentChild` + `setRoot` directly), so
it does not cause the heap-buffer-overflow observed in practice. It is
reachable from other code paths, however, and represents a latent correctness
hazard that would be difficult to diagnose when it eventually manifests.


davidcerny:
This fails the rerooting test, but the new output actually only differs from the expected output in node indices – the topologies themselves are exactly the same. So if you want to make the tests pass simply by replacing the expected output, I wouldn't necessarily object.

That said, I agree with your initial remark:

> can't vouch for the necessity of the fix.

I guess the LLM diagnosis might be right in calling this "a latent correctness hazard" – since we don't reroot trees during the MCMC, the problem wouldn't arise in practice even if it was in fact real.

ms609:
Thanks - I've updated the test expectations.  I don't see a clear reason to bypass the `setRoot` function here, so lean towards merging.

davidcerny (approved):

ms609:
Thanks - before I merge, just to check that the failure on macos xcode 16 intel can be safely overlooked:

```
   mcmc_scm_hrm_RJk.Rev: 
       |     
       |  
       |  Iter        |      Posterior   |     Likelihood   |          Prior   |    elapsed   |        ETA   |
       |  ----------------------------------------------------------------------------------------------------
       |  0           |       -30.3088   |       -35.9794   |        5.67056   |   00:00:00   |   --:--:--   |

   FAIL: mcmc_scm_hrm_RJk.Rev: error 137
```

davidcerny:
Yes, that was apparently just one of the annoying stochastic failures that we get every once in a while. I'm rerunning the corresponding GitHub Action, and all the tutorial tests (including `mcmc_scm_hrm_RJk.Rev`) are now passing.

bredelings:
I guess this makes sense.  Given that the node index indicates BOTH the node AND its parent branch, when we reroot we can't preserve them both.  We preserve the name of the branch, but not the name of the node.  So node names must change.

PR #967, opened 2026-03-17
Fix heap-buffer-overflow when Q/rootFreq dimension changes during MCMC

### LINKED ISSUE #965: Heap-buffer-overflow in `computeRootLikelihood` after topology moves

ms609:
Fixes #965. Proposed by Claude Opus. Manually reviewed by @ms609.  

Requires review by someone with more complete understanding of the code-base.

## Problem

`AbstractPhyloCTMCSiteHomogeneous` allocates its partial-likelihood buffer in
`resizeLikelihoodVectors()` using `num_chars` — the number of character states
recorded at construction time:

```
buffer size = 2 × num_nodes × num_site_mixtures × pattern_block_size × num_chars  (doubles)
```

During MCMC, if a DAG update swaps `homogeneous_rate_matrix` or
`root_frequencies` to an object with a different number of states, `num_chars`
is never updated and the buffer is never reallocated. The inner state loop in
`computeRootLikelihood` iterates over the full frequency vector `f` using
iterator arithmetic (`f_begin` to `f_end`). When `f.size() > num_chars`, each
extra iteration advances the write pointer past the end of the buffer, causing
a heap-buffer-overflow.

The overflow lands exactly one element past the buffer end:

```
WRITE of size 8 at <addr> — 0 bytes to the right of <2 × activeLikelihoodOffset>-byte region
  #0  PhyloCTMCSiteHomogeneous<StandardState>::computeRootLikelihood()
        PhyloCTMCSiteHomogeneous.h:196
  #1  PhyloCTMCSiteHomogeneousConditional<StandardState>::computeRootLikelihood()
        PhyloCTMCSiteHomogeneousConditional.h:423
  #2  AbstractPhyloCTMCSiteHomogeneous<StandardState>::computeLnProbability()
        AbstractPhyloCTMCSiteHomogeneous.h:969
  #5  MetropolisHastingsMove::performMcmcMove()
  #8  MonteCarloAnalysis::burnin()
```

A guarding `assert(f.size() == this->num_chars)` exists at the top of the
loop, but is compiled out by `-DNDEBUG` in release builds, silencing the
mismatch entirely.

### Trigger conditions

The crash requires multiple `dnPhyloCTMC` distributions whose rate matrix or
root-frequency parameter is a DAG-reactive node that can resolve to objects of
different sizes at runtime (e.g. an indexed lookup vector whose index is a
stochastic node). Fixed-dimension rate matrix models are not affected. The bug
is stochastic and typically surfaces within a few thousand MCMC iterations.

Tested with: GCC 11.2, Linux (HPC cluster), ASAN build, RevBayes commit
`92174d501`. Also crashes with the Feb 13 build, confirming this is not a
recent regression.

### `AbstractPhyloCTMCSiteHomogeneous.h`

- Remove `const` from the `num_chars` member — it must be mutable if the
  distribution's rate matrix can change dimension after construction.
- Add a dedicated `homogeneous_rate_matrix` branch in `touchSpecialization`:
  when the DAG notifies the distribution that its rate matrix changed, compare
  `getNumberOfStates()` against the current `num_chars`. If they differ,
  update `num_chars` and call `resizeLikelihoodVectors()` before any
  likelihood computation is attempted.
- Apply the same dimension check in the existing `root_frequencies` branch
  (compare `f.size()` against `num_chars` and resize if needed).

### `PhyloCTMCSiteHomogeneous.h`

- Replace `assert(f.size() == this->num_chars)` with a throwing `RbException`
  in both `computeRootLikelihood` overloads (2-child and 3-child root). The
  exception fires in release builds and produces a legible diagnostic if the
  invariant is ever violated despite the fix above.

## Testing

ASAN build, Hamilton HPC (GCC 11.2), `nni-bug.Rev 10 5 42`:
- Before fix: crashes with heap-buffer-overflow at `PhyloCTMCSiteHomogeneous.h:196`
- After fix: Working correctly, notwithstanding a separate bug in MCMCMC (to be reported separately).  Replacing MCMCMC with MCMC results in successful runs once this patch is applied.

### DIFF
--- a/src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
+++ b/src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
@@ -181,7 +181,7 @@ namespace RevBayesCore {
         std::optional<double>                                               storedLnProb;
         size_t                                                              num_nodes;
         size_t                                                              num_sites;
-        const size_t                                                        num_chars;
+        size_t                                                              num_chars;
         size_t                                                              num_site_rates;
         size_t                                                              num_site_mixtures;
         size_t                                                              num_matrices = 1;
@@ -935,6 +935,16 @@ double RevBayesCore::AbstractPhyloCTMCSiteHomogeneous<charType>::computeLnProbab
 
     // we start with the root and then traverse down the tree
     size_t root_index = root.getIndex();
+
+    // Guard against out-of-bounds root index: this indicates tree/PhyloCTMC desynchronisation
+    if ( root_index >= num_nodes )
+    {
+        throw RbException() << "BUG in dnPhyloCTMC: root node index " << root_index
+                            << " is out of bounds for num_nodes=" << num_nodes
+                            << " (tree reports " << tau->getValue().getNumberOfNodes() << " nodes)."
+                            << " This usually means a topology proposal corrupted the tree's node"
+                            << " indices or the Tree::root pointer became stale.";
+    }
 
     // only necessary if the root is actually dirty
     if ( dirty_nodes[root_index] == true )
@@ -2381,6 +2391,15 @@ void RevBayesCore::AbstractPhyloCTMCSiteHomogeneous<charType>::recursivelyFlagNo
 
     // we need to flag this node and all ancestral nodes for recomputation
     size_t index = n.getIndex();
+
+    // Guard against out-of-bounds index from a corrupted tree state
+    if ( index >= dirty_nodes.size() )
+    {
+        throw RbException() << "BUG in recursivelyFlagNodeDirty: node index " << index
+                            << " >= dirty_nodes.size()=" << dirty_nodes.size()
+                            << " (num_nodes=" << num_nodes << ")."
+                            << " Tree topology proposal likely corrupted a node's index.";
+    }
 
     // if this node is already dirty, the also all the ancestral nodes must have been flagged as dirty
     if ( dirty_nodes[index] == false )
@@ -4253,9 +4272,29 @@ void RevBayesCore::AbstractPhyloCTMCSiteHomogeneous<charType>::touchSpecializati
     }
     else if ( affecter == root_frequencies )
     {
+        // If root-frequency dimension changed (e.g. DAG swap to a different-size simplex),
+        // reallocate the likelihood buffer before flagging nodes dirty.
+        size_t new_num_chars = root_frequencies->getValue().size();
+        if ( new_num_chars != num_chars )
+        {
+            num_chars = new_num_chars;
+            resizeLikelihoodVectors();
+        }
 
         const TopologyNode &root = this->tau->getValue().getRoot();
         this->recursivelyFlagNodeDirty( root );
+    }
+    else if ( affecter == homogeneous_rate_matrix )
+    {
+        // If the rate-matrix dimension changed (e.g. DAG swap to a different-size Q),
+        // reallocate the likelihood buffer before flagging nodes dirty.
+        size_t new_num_chars = homogeneous_rate_matrix->getValue().getNumberOfStates();
+        if ( new_num_chars != num_chars )
+        {
+            num_chars = new_num_chars;
+            resizeLikelihoodVectors();
+        }
+        touch_all = true;
     }
     else if ( affecter == p_inv )
     {
--- a/src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
+++ b/src/core/distributions/phylogenetics/substitution/AbstractPhyloCTMCSiteHomogeneous.h
@@ -4272,29 +4272,8 @@ void RevBayesCore::AbstractPhyloCTMCSiteHomogeneous<charType>::touchSpecializati
     }
     else if ( affecter == root_frequencies )
     {
-        // If root-frequency dimension changed (e.g. DAG swap to a different-size simplex),
-        // reallocate the likelihood buffer before flagging nodes dirty.
-        size_t new_num_chars = root_frequencies->getValue().size();
-        if ( new_num_chars != num_chars )
-        {
-            num_chars = new_num_chars;
-            resizeLikelihoodVectors();
-        }
-
         const TopologyNode &root = this->tau->getValue().getRoot();
         this->recursivelyFlagNodeDirty( root );
-    }
-    else if ( affecter == homogeneous_rate_matrix )
-    {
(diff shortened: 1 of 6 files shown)


bredelings:
So #816 actually should fix this problem as a side effect, because it doesn't pre-allocate a fixed-sized buffer for the conditional likelihoods. Can you check that it fixes the problem for you?

I asked @afmagee to review it, and he was planning to get to it in a week or so.

ms609:
Thanks for the pointer; I'll see whether I can reproduce the pathology under #816.

bredelings:
OK, so it looks like #816 is not sufficient, and we need to handle a dynamically changing `num_chars` also.  It makes sense that we should pull the `num_chars` from the `Q` matrix.

In general, an `assert( )` is good for documenting something that you believe can never be violated.  If it can indeed be violated but it might be depending on user input, then you want a user-facing message instead.  When you get a segfault recompiling in debug mode to enable assertions is a good idea.  Good software should have many many asserts that document assumptions that have been made.

Eventually each branch should have its own PartialLikelihoods object, and we could store the `num_chars` on that.  It would be set from the Q matrix when we compute the likelihoods for tip nodes, and we could assert that the inputs to an internal node agree on the value of `num_chars`.

This kind of avoids asking what the `num_chars` value for a dataset is.  For morphological datasets `num_chars` (really, the number of states) seems ill-defined.  I'm not totally sure how we should handle that.  For data types like DNA, if the `num_chars` does not match the Q matrix, then we should give a user-visible error.  But for morphological data, we should not.

PR #968, opened 2026-03-17
Feature: Bounded random walks

ms609:
> LLM generated (ChatGPT 5.3 Codex) under human direction; manually reviewed

## Summary

Prevent negative proposals for Natural random-walk moves, before downstream deterministic nodes are evaluated.

## Motivation

Previously, random-walk proposals could generate negative intermediate values and rely on later rejection by prior support. Because downstream deterministic nodes are evaluated before accept/reject, this could trigger runtime failures.

## Changes

- Added configurable lower-bound reflection support in integer/geometric walk proposals.
- Wired Natural walk wrapper to enforce lower bound 0.
- Added integration test coverage for bounded Natural walk behavior.

## Implementation details

- Reflection behavior is enabled via optional lower bound in proposal constructors.
- Natural walk now passes a lower bound (0).
- Existing integer/geometric behavior remains backward compatible when no lower bound is provided.

## Validation

- Ran bounded natural-walk integration script with RevBayes binary.
- Exit code: 0
- Trace check: sampled Natural variable never below 0 (minimum observed value = 0, 50 samples) (softened to avoid stochastic fragility).
- Expected-output comparison: no differences.


bredelings:
As a design issue, I guess the question is whether we should modify the proposal, or modify the infrastructure so that it is not needed, or both.  Is there an issue with a script that demonstrates the problem?

@ms609  BTW, tomorrow I'm going to try and devote some attention to the messages you sent, and also to some of the bugs from @pbbreslin1

ms609:
This is associated with the same issue as #967, and I think is now redundant to the proposal there - I'll close this PR when verified.

davidcerny:
> This is associated with the same issue as https://github.com/revbayes/revbayes/pull/967, and I think is now redundant to the proposal there

Did you ever get a chance to verify this, @ms609? Should we close this PR?

ms609:
Not had the chance to verify and am unlikely to in the foreseeable future.  Shall close the PR.

PR #970, opened 2026-03-18
Fix #931: truncate monitor files to checkpoint generation on resume

### LINKED ISSUE #931: [BUG] Checkpoint resume produces duplicate generations in trace files
**Describe the bug**
When an unexpected exit occurs during MCMC running, there would be some generations after last checkpoint in the monitor files. And if I try to resume this run based on that checkpoint, those generations are sampled again and appended, resulting in duplicate generation numbers at the beginning of the resumed segment (with different parameter values).

curiosusJR:
Fixes #931. When MCMC resumes from a checkpoint, monitor files may contain samples written after the last checkpoint but before a crash. On resume those generations are re-sampled and appended, creating duplicates.

- Add `AbstractFileMonitor::truncateAfterGeneration(lastGen)`  that removes any samples past the checkpoint generation before the resumed run appends to the file. Handles both separator (tab/comma) and JSON (`"Iteration":`) formats.
- Override in `NexusMonitor` to handle the Nexus tree format (`tree TREE_<gen>`) and remove the trailing `End;` so new trees can be appended cleanly.
- In `Mcmc::initializeSamplerFromCheckpoint`, call truncation for each existing monitor file after reading the checkpoint generation. Throw an error if `last_generation == 0` (empty or corrupted checkpoint) to prevent silently wiping valid data.

Here is a test suit for confirming this bug fixed. 
[dup.zip](https://github.com/user-attachments/files/26090780/dup.zip)

In this test, mcmc.Rb runs a phylogenetic MCMC analysis in 3 steps:
    - Run 10 generations, saving a checkpoint at generation 7
    - Resume from that checkpoint and run 10 more generations
    - This writes output to 7 different monitor files (tab-delimited logs, JSON logs, tree files, etc.)

The monitor output files could contain duplicate generation numbers — e.g., generations 8, 9, 10 appearing twice. After fixing, there would be no duplication and the nexus monitor file now can drop the redundant 'End;' marker correctly.


davidcerny:
Very impressive, thank you! I'll take a more detailed look at the code itself, but it seems to work exactly as advertised.

src/core/monitors/AbstractFileMonitor.cpp
```diff
@@ -103,6 +106,54 @@ void AbstractFileMonitor::openStream( bool reopen )
...
+        {
+            // separator format: generation is the first field
+            char *end_ptr = NULL;
+            gen = static_cast<std::uint64_t>( std::strtoull(line.c_str(), &end_ptr, 10) );
+            found = (end_ptr != line.c_str());
+        }
+        else if ( line[0] == '{' )
+        {
+            // JSON format: look for "Iteration": <number>
+            size_t it_pos = line.find("\"Iteration\":");
```
bredelings: Is there a reason you don't just read the JSON object and extract the "Iteration" element?  That seems less fragile.

src/core/monitors/AbstractFileMonitor.cpp
```diff
@@ -103,6 +108,74 @@ void AbstractFileMonitor::openStream( bool reopen )
...
+            char *end_ptr = NULL;
+            std::uint64_t gen = static_cast<std::uint64_t>( std::strtoull(line.c_str(), &end_ptr, 10) );
+            if ( end_ptr != line.c_str() && gen > lastGen )
+            {
+                infile.close();
+                std::filesystem::resize_file(working_file_name, pos);
+                return;
+            }
+        }
+    }
```
davidcerny: You might be able to condense this section somewhat by putting the `if (is_json)` conditional inside the `while` loop, rather than outside of it

src/core/monitors/NexusMonitor.cpp
```diff
@@ -105,6 +108,45 @@ void NexusMonitor::monitor(std::uint64_t gen) {
...
+        std::streampos pos = infile.tellg();
+        if ( !safeGetline(infile, line) ) break;
+
+        if ( line == "Begin trees;" ) {
+            in_trees_block = true;
+            continue;
+        }
+
+        // Nexus tree lines: "tree TREE_<gen> = ..."
+        const size_t tree_prefix_len = 10; // length of "tree TREE_"
```
davidcerny, line 129: Is this 100% safe? Will individual lines always begin with this string?

davidcerny (changes requested):
This looks great! My only major request would be to turn the test you have already prepared (i.e., the `dup.zip` folder enclosed with the PR description) into an actual integration test. You could put it in a new `test_checkpointing` folder. If you need any help, let me know!

src/core/monitors/AbstractFileMonitor.cpp
curiosusJR: Oh, I initially thought that placing the `if(...)` outside the loop would reduce the time of checking and thus slightly improve performance, and now I know that modern compilers can automatically optimize this.

davidcerny: Ah – it didn't even occur to me to think of the performance impact of doing it one way or the other! Anyway, the code is fine as-is.

src/core/monitors/NexusMonitor.cpp
curiosusJR: If 'resume running' only depends on RevBayes output, I believe ensuring consistency between the current parser's format and the Nexus tree trace output should be sufficient. I've found that existing Nexus parsers don't support line-by-line operations or generation number comparisons within the tree block. Since implementing a full-scale Nexus parser here seems overkill. So maybe just keep it as-is which is already same with `NexusMonitor::monitor()`, or I propose defining a single constant shared by both the write and read paths to guarantee data integrity and consistency.

tests/test_checkpointing/scripts/mcmc.Rev
```diff
@@ -0,0 +1,42 @@
...
+monitors.append( mnFile(filename="output/file.log", printgen=1, separator=TAB, topology, TL) )
+monitors.append( mnFile(filename="output/file.json", printgen=1, format = "json", topology, TL) )
+
+monitors.append( mnScreen(printgen=10) )
+
+mymcmc = mcmc(mymodel, monitors, moves, nruns=1, combine="none")
+mymcmc.run(generations=10, checkpointFile="output/posterior.checkpoint", checkpointInterval=7, tuningInterval=2)
+
+mymcmc.initializeFromCheckpoint("output/posterior.checkpoint")
+mymcmc.run(generations=10, checkpointFile="output/posterior.checkpoint", checkpointInterval=7, tuningInterval=2)
```
davidcerny: Not a big deal by any means, but for the reasons discussed at #934, do you think you could remove `tuningInterval=2` from `mymcmc.run()`?

tests/test_checkpointing/scripts/mcmc.Rev
```diff
@@ -0,0 +1,42 @@
...
+monitors.append( mnStochasticVariable(filename="output/variable.var",printgen=1) )
+
+monitors.append( mnExtNewick(tree=topology, filename="output/extnewick.trees", isNodeParameter=FALSE, printgen=1, separator=TAB) )
+monitors.append( mnNexus(filename="output/nexus.trees", tree=topology, isNodeParameter=FALSE, writeTaxa=FALSE, printgen=1) )
+
+monitors.append( mnFile(filename="output/file.log", printgen=1, separator=TAB, topology, TL) )
+monitors.append( mnFile(filename="output/file.json", printgen=1, format = "json", topology, TL) )
+
+monitors.append( mnScreen(printgen=10) )
+
```
davidcerny: Careful – the output of the screen monitor (specifically, the runtime and ETA) is non-deterministic, so it might prevent the test from passing.

curiosusJR: Thank you for reminding!

davidcerny (approved):

PR #972, opened 2026-03-18
Add getNode( ) assertion checks

bredelings:
This adds a few assertion checks to ensure that we don't accidentally forget to recompute the nodes[ ] vector again.


ms609 (approved):
Looks good, thanks for implementing!

PR #974, opened 2026-03-19
SSE error message

bredelings:
When running an SSE model with 9 states, we currently find the that probability is Nan and keep resampling initial parameters forever.

Instead:
* when the highest likelihood is zero, don't divide by zero to scale it :grimacing: 
* when the number of characters in the data is larger than the number of characters in the model, quit with an error message.
* when the number of characters in the data is smaller than the number of characters in the model, print a warning.
* assert that the observed data at a tip does not have probability zero


davidcerny:
Hmm... the failing ChromoSSE test now outputs:

```
   Warning: SSE model has 19states, but observed data set has only 19 states!
```

so the check doesn't seem to be quite right. Assuming that it can be fixed, it might be worth adding an extra test based on the issue that motivated this PR.

bredelings:
It was I typo -- somehow I wrote `<<` instead of `<` to  compare two numbers.

tests/test_SSE/scripts/mcmc_BiSSE.Rev
```diff
@@ -0,0 +1,194 @@
...
+                                                  withTips        = true, 
+                                                  withStartStates = false, 
+                                                  filename        = "output/primates_BiSSE_activity_period_anc_states.log") )
+
+# stochastic character map monitor
+monitors.append( mnStochasticCharacterMap(cdbdp    = timetree, 
+                                          filename = "output/primates_BiSSE_activity_period_stoch_map.log", 
+                                          printgen = 1) )
+
+monitors.append( mnScreen(printgen=10, speciation, extinction) )
```
davidcerny: We should remove the screen monitor from the test script

tests/test_SSE/scripts/mcmc_BiSSE_error.Rev
```diff
@@ -0,0 +1,194 @@
...
+                                                  withTips        = true, 
+                                                  withStartStates = false, 
+                                                  filename        = "output/primates_BiSSE_activity_period_anc_states.log") )
+
+# stochastic character map monitor
+monitors.append( mnStochasticCharacterMap(cdbdp    = timetree, 
+                                          filename = "output/primates_BiSSE_activity_period_stoch_map.log", 
+                                          printgen = 1) )
+
+monitors.append( mnScreen(printgen=10, speciation, extinction) )
```
davidcerny: Also here

davidcerny (changes requested):

davidcerny (approved):
Looks good!

PR #976, opened 2026-03-25
Fix hidden virtual functions

bredelings:
In a few cases people have added versions of virtual functions where the type doesn't match, leading to a lot of warnings.  In some of these cases people pass large objects like `std::vector` by value instead of by reference, which wastes time copying the object.


basantakhakurel (approved):
Looks good!

PR #977, opened 2026-03-25
Ensure we start the interactive loop with -i is given

bredelings:
When a `-i` or `--interactive` is given, we should start the interactive loop -- even if there is an error.

So `RevLanguageMain::startRevLanguageEnvironment( )` should not call `std::exit(1)` when we see an error.  Make it pass back any error results to the caller instead, and let the caller decide if/when to quit.

### DIFF
--- a/src/revlanguage/RevLanguageMain.cpp
+++ b/src/revlanguage/RevLanguageMain.cpp
@@ -26,7 +26,7 @@ RevLanguageMain::RevLanguageMain(bool c, bool e, bool q)
 }
 
 
-void RevLanguageMain::startRevLanguageEnvironment(const std::vector<std::string> &expressions, const std::optional<std::string>& filename, const std::vector<std::string> &args)
+int RevLanguageMain::startRevLanguageEnvironment(const std::vector<std::string> &expressions, const std::optional<std::string>& filename, const std::vector<std::string> &args)
 {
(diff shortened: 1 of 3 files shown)


davidcerny (approved):
Neat! This works exactly as intended. For an `example.Rev` script that looks as follows:

```r
x ~ dnNormal(0,1)    # runs
XXXX                 # forced error
4 + 4
```

I now get:

(1)

```
David@Mac ~ % ./rb example.Rev                               
   Missing Variable:	Variable XXXX does not exist
   Error:	Problem processing line 2 in file "example.Rev"
David@Mac ~ %
```

(2)

```
David@Mac ~ % ./rb -c example.Rev
   Missing Variable:	Variable XXXX does not exist
   Error:	Problem processing line 2 in file "example.Rev"
   8
David@Mac ~ %
```

(3)

```
David@Mac ~ % ./rb -i example.Rev 
   Missing Variable:	Variable XXXX does not exist
   Error:	Problem processing line 2 in file "example.Rev"
>
```

(4)

```
David@Mac ~ % ./rb -i -c example.Rev
   Missing Variable:	Variable XXXX does not exist
   Error:	Problem processing line 2 in file "example.Rev"
   8
>
```

The behaviors in (1), (2), and (4) are identical on the current tip of development, but (3) now makes more sense.

Issue #874, opened 2025-10-21
[BUG] Out-of-place prompt symbol in the MPI version on some but not all Linux machines


davidcerny:
**Describe the bug**
I recently compiled the MPI version of RevBayes on an HPC cluster running RHEL 8.4 (Ootpa). I used the following tools and libraries:

- CMake 3.19
- GCC 12.2.0
- boost 1.75.0
- OpenMPI 5.0.2 (also built with GCC 12.2.0)

The compiling and linking steps finished without any obvious issues, and the executable seems to work fine... up to a point. First of all, the command prompt `>` doesn't appear when commands are being typed out, but rather when their output is being shown. So instead of the usual:

```
> 3 + 3
   6
> 3 + 4
   7
> q()
```

I'm seeing:

```
3 + 3
>    6
3 + 4
>    7
q()
```

This is happening regardless of the specified number of cores (i.e., even with `mpirun -np 1`), but not when I launch the program without relying on the `mpirun` utility.

Some experimentation with the [`Test_multipleRules.Rev`](https://github.com/revbayes/revbayes/blob/master/tests/test_convergenceRules/scripts/Test_multipleRules.Rev) test script further revealed other strange behavior:
**============ [EDIT BY DČ FROM 2025-11-08: I can no longer reproduce these issues; see below.] ============**

- When I execute the script using a single process, everything works fine and I get the expected output, regardless of whether I do it from the command line (by supplying the script file name as an argument), or from within an interactive session (using the `source()` function).
- When I execute the script using two processes (one for each MCMC run) from within an interactive session, I get different output, but the run still completes successfully. This still seems okay. However:
- When I execute the script using two processes from the command line, like so:

     ```
     mpirun -np 2 ./build-mpi/rb-mpi ../../tests/test_convergenceRules/scripts/Test_multipleRules.Rev
    ```

    I get stochastic behavior, where the analysis sometimes completes successfully and sometimes crashes after the first ~8,000 iterations with the following error:

    ```
    --------------------------------------------------------------------------
    prterun detected that one or more processes exited with non-zero status,
    thus causing the job to be terminated. The first process to do so was:
    ```

    Moreover, all convergence statistics other than PSRF get stuck:

    ```
       
   Running MCMC simulation
   This simulation runs 4 independent replicates.
   The simulator uses 1 different moves in a random move schedule with 1 moves per iteration
   
   Stopping rules:
       Target value for effective sample size (ESS): > 50
[...]

PR #978, opened 2026-03-26
Add -p flag to print commands from -e expression or script

bredelings:
Sometimes its nice to have revbayes print the commands that its executing from a script, so that you can see what its doing, and when it is doing it. This is conceptually similar to the `-x` flag for bash, which prints all the actions for debugging purposes.  And also its similar to what happens when we are interactive.

This PR adds a flag `-p` or `--print-commands` to do that.  With `-p`, executing the file
```
1+1
x=2
x
```
yields output like:
```
> 1+1
   2
> x=2
> x
   2
```
Running `rb -p -e '1+1' -e 'x=2' -e x` shows exactly the same output.  Running `rb -pi script` shows the same output and then starts an interactive session right underneath.  **If the  script stops with an error, this should make it a bit clearer which line has the error.**

The `-p` flag has no effect on interactive execution, where the terminal is responsible for printing what you type.

The `--help` line for `-p` looks like this:
```
  -p,      --print-commands        Print commands from file or -e expr
```
The entire help from `rb --help` looks like this:
```
Bayesian phylogenetic inference using probabilistic graphical models and an 
interpreted language 

Usage: rb [options]
       rb [options] file [args]
       rb [options] -e expr [-e expr2 ...] [args]

OPTIONS:
  -h,      --help                  Print this help message and exit 
  -v,      --version               Show version and exit 
  -j,      --jupyter               Run in jupyter mode 
  -q,      --quiet                 Hide startup message (if no file or -e expr) 
  -i,      --interactive           Force interactive (with file or -e expr) 
  -c,      --continue              Continue after error (with file or -e expr) 
  -p,      --print-commands        Print commands from file or -e expr
  -s,      --seed UINT             Random seed (unsigned integer) 
  -o,      --setOption TEXT ...    Set an option key=value (See ?setOption for 
                                   the list of available keys and their 
                                   associated values) 

Expressions (one or more '-e <expr>') may be used *instead* of 'file'. 
```

### DIFF
--- a/src/libs/range-v3/meta/meta.hpp
+++ b/src/libs/range-v3/meta/meta.hpp
@@ -3778,7 +3778,7 @@ namespace meta
         /// \ingroup integral
         template <char... Chs>
         constexpr fold<list<char_<Chs>...>, meta::size_t<0>, quote<detail::atoi_>>
-            operator"" _z()
+            operator""_z()
         {
             return {};
         }
--- a/src/revlanguage/RevLanguageMain.cpp
+++ b/src/revlanguage/RevLanguageMain.cpp
@@ -94,7 +94,7 @@ int RevLanguageMain::startRevLanguageEnvironment(const std::vector<std::string> 
     {
         // Should we be using RBOUT here?  It looks weird with the 2 spaces of padding.
         if (echo and rank == 0)
-            std::cerr<<"> "<<expression<<"\n";
+            std::cout<<"> "<<expression<<"\n";
 
         int result = RevLanguage::Parser::getParser().processCommand(expression, RevLanguage::Workspace::userWorkspacePtr());
         
--- a/src/revlanguage/main.cpp
+++ b/src/revlanguage/main.cpp
@@ -44,9 +44,10 @@ struct ParsedOptions
 
     bool script_or_expr() const {return filename or not expressions.empty();}
 
-    bool force_interactive = false;       /* Force interactive if script_or_expr() == true. */
-    bool force_continue_on_error = false; /* Force continue-on-error if script_or_expr() == true. */
-    bool force_quiet = false;             /* Suppress header if script_or_expr() == false */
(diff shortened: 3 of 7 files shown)


davidcerny (commented):
On the single-thread version of RevBayes, this does exactly what it says on the tin. Unfortunately, the new flag doesn't get along with MPI very well. When I execute the following script:

```r
x ~ dnNormal(0,1)
y ~ dnExp(10)
4 + 4
```

like this:

```
mpirun -np 8 ./build-mpi/rb-mpi -p example.Rev
```

I get something like this:

```
> x ~ dnNormal(0,1)
> y ~ dnExp(10)        
> 4 + 4
> x ~ dnNormal(0,1)
> x ~ dnNormal(0,1)
> x ~ dnNormal(0,1)
> y ~ dnExp(10)        
> x ~ dnNormal(0,1)
> 4 + 4
> y ~ dnExp(10)        
> y ~ dnExp(10)        
> 4 + 4
> 4 + 4
> y ~ dnExp(10)        
> 4 + 4
   8
> x ~ dnNormal(0,1)
> y ~ dnExp(10)        
> x ~ dnNormal(0,1)
> 4 + 4
> y ~ dnExp(10)        
> x ~ dnNormal(0,1)
> 4 + 4
> y ~ dnExp(10)        
> 4 + 4
```

I doubt this is fixable without a comprehensive solution to MPI I/O that also addresses #874 – it's basically one more opportunity for a known problem to pop up. The question is whether that's an acceptable trade-off for the added functionality.

bredelings:
I think this is fixed.  Luckily `linenoise` is not involved, we just need to only print the commands from rank 0.  Since they all end with a `'\n'` they should all print fine.

davidcerny (approved):
Glad to be proved wrong on this! I can confirm it is fixed. Approving.

bredelings:
THanks!

PR #981, opened 2026-03-27
Remove the per-line fields parser in `VariableMonitor::combineReplicates()`

### LINKED ISSUE #971: [BUG] combining runs takes too much time
**Describe the bug**
When running mcmc() with `nruns=2` and `combine="mixed"`, the post-run file-merging step is significantly slower than expected. Maybe ~3x slower than setting `combine="none"` then combine runs manually.

**To Reproduce**
[merge.zip](https://github.com/user-attachments/files/26094238/merge.zip)

Here is a 2 runs mcmc with 1000 generations. The Rb script requires an argument for the 'combine' setting, like: `rb mcmc.Rb mixed` or `rb mcmc.Rb none`. 

I also make a comparison script to show the difference in taken time. 

In my test, here is the result: 
```
combine="none", combined manually, takes:

real	0m1.571s
user	0m0.286s
sys	0m1.276s
combine="mixed", combined in revbayes, takes:

real	0m4.266s
user	0m0.510s
sys	0m3.738s
```
There is no difference between manually combined and automatically combined files.

Fewer files can help narrow the gap, though a large generation number is another common cause. In my experience, a large monitor file can take minutes per combing step and may slow down MCMC performance across multiple resume runs.

curiosusJR:
This PR refactors replicate-trace combination in src/core/monitors/VariableMonitor.cpp. 

`VariableMonitor::combineReplicates()` is only used for editing the first 2 columns in trace files. But the previous implementation unnecessarily split each header and data line into a `vector<string>` with `StringUtilities::stringSplit(...)`, then rebuilt the output line column by column. A 20,000 generations mcmc with 20 columns (common in GTR+gamma model) in 3 trace files would creates roughly on the order of a million of temporary std::string field objects while combining, and that would take the most of combining time. 

The new implementation avoids re-parsing the entire line. It only rewrites the first 2columns (`Iteration` and the inserted `Replicate_ID`), and then copies the rest of the row exactly as is.

A test based on the script in #971 on my laptop shows the improvement: 
```
$ time ( rb mcmc.Rev mixed > /dev/null ) && time ( rb mcmc.Rev none > /dev/null ) && time ( ../../projects/cmake/build/rb mcmc.Rev mixed > /dev/null)
( rb mcmc.Rev mixed > /dev/null; )  33.61s user 60.57s system 99% cpu 1:34.79 total
( rb mcmc.Rev none > /dev/null; )  29.46s user 20.47s system 99% cpu 50.022 total
( ../../projects/cmake/build/rb mcmc.Rev mixed; )  28.86s user 21.67s system 99% cpu 50.702 total
```

The largest improvement appears  in system time (60.57s -> 21.67s), and the combining time becomes to less than 1 second, which may indicate that the current modifications are effective in resolving issue #971 .


davidcerny (approved):
I tried to get a sense of how bad the problem really is – and consequently, how big of an improvement this PR provides – by setting up a rather extreme example. I took a pretty large and complex FBD analysis I'd recently finished running, and (1) changed the chain length to 10,000 iterations, (2) changed the sampling frequency to 1 sample per iteration, and (3) added `combine="mixed"` to `mcmc.run()`. This amounted to combining 10 log files, each of them over 50 MB in size (10,000 rows times 489 columns). I originally wanted to provide a table along the following lines:

```
development; combine = "none" (DN)

real ...	user ...	sys ...
real ...	user ...	sys ...

development; combine = "mixed" (DM)

real ...	user ...	sys ...
real ...	user ...	sys ...

fastCombine; combine = "none" (FCN)

real ...	user ...	sys ...
real ...	user ...	sys ...

fastCombine; combine = "mixed" (FCM)

real ...	user ...	sys ...
real ...	user ...	sys ...
```

However, this proved to be infeasible. The FCM analysis finished in a very similar time to DN (about 2 hours 40 minutes), but I had to kill both of the DM analyses I had tried to run, in one case after more than 16 hours – long after they stopped sampling, but before they stopped combining the log files.

So yes, the problem appears to be very bad indeed, and this PR successfully solves it. Approving.

PR #982, opened 2026-03-27
Add new tests for the model graph

bredelings:
Add two new tests based on order-two Markov processes.


davidcerny:
Should we close this now that you've added the tests in question directly to #962? The tests are also failing here, for whatever reason...

davidcerny:
Closed in favor of #962

Issue #558, opened 2024-09-12
[Help Wanted] Get error when doing posteriorPredictiveSimulation() under a mix_branch_length model


curiosusJR:
**Describe the desired behavior.**
What are you trying to do? 
Doing `posteriorPredictiveSimulation` under a mix_branch_length model after `mcmc`.
[MBL_model.txt](https://github.com/user-attachments/files/16977511/MBL_model.txt)

**What is going wrong?**
Copy in code showing what is going wrong.
```
RevBayes version (1.2.5-preview)
Build from remotes/origin/development (rapture-3003-gbcbe1c) on Sat Sep  7 14:54:51 CST 2024

Visit the website www.RevBayes.com for more information about RevBayes.

RevBayes is free software released under the GPL license, version 3. Type 'license()' for details.

To quit RevBayes type 'quit()' or 'q()'.


> source("MBL_model.Rev")
   Processing file "MBL_model.Rev"
   Successfully read one character matrix from file 'seq.nex'

   Running burn-in phase of Monte Carlo sampler for 2000 iterations.
   This simulation runs 2 independent replicates.
   The simulator uses 20 different moves in a random move schedule with 26 moves per iteration


Progress:
0---------------25---------------50---------------75--------------100
********************************************************************


   Running MCMC simulation
   This simulation runs 2 independent replicates.
   The simulator uses 20 different moves in a random move schedule with 26 moves per iteration


Iter        |      Posterior   |     Likelihood   |          Prior   |             TL   |    elapsed   |        ETA   |
-----------------------------------------------------------------------------------------------------------------------
0           |        14.5565   |       -11.5639   |        26.1204   |      0.4444978   |   00:00:00   |   --:--:--   |
...
25000       |        11.1937   |       -12.6542   |        23.8479   |      0.5941138   |   00:00:11   |   00:00:00   |
   Error:	Vector index out of range: 2 of 2
   Error:	Problem processing line 117 in file ""MBL_model.Rev""
>
```

If I remove the 'br_tmp' column in the file 'posterior.var', `posteriorPredictiveSimulation` can be run with no error.

the 'posterior.var' file looks like.
```
[...]

PR #983, opened 2026-03-28
Add proper trace/checkpoint serialization support for `MatrixReal`

### LINKED ISSUE #836: [BUG] dnLKJPartial: can't resurect from checkpoint, segfault when initializing from trace
**Describe the bug**

Models with dnLKJPartial fail to resurect from checkpoint and segfault when initializing from trace

**To Reproduce**

Set up a simple model using dnLKJPartial
```r
nchar <- 2
eta <- 1.0
P ~ dnLKJPartial(eta, nchar)
outdir <- "/tmp/rb/"

moves = VectorMoves()
moves.append( mvCorrelationMatrixRandomWalk(P, weight=1.0) )
# or mvCorrelationMatrixSingleElementBeta(P, weight=1.0)

monitors = VectorMonitors()
monitors.append( mnModel(filename=outdir + "p_only.log", printgen=10) )
monitors.append( mnStochasticVariable(filename=outdir + "posterior.var",printgen=50, append=TRUE) )

mymodel = model(P)
mymcmc = mcmc(mymodel, monitors, moves)

mymcmc.run(generations=200, checkpointFile=outdir + "p_only.ckp", checkpointInterval=20)
```

Now try to reload the checkpoint
```r
mymcmc.initializeFromCheckpoint(outdir + "p_only.ckp")                                                              
   Error:       Could not resurrect object from string value:
   [ [ 1.0000, 0.5513 ] ,
```

Or try to reload the trace
```r
tr = readStochasticVariableTrace(outdir + "posterior.var")                                                          
Segmentation fault (core dumped)
```

Contents of p_only.ckp
```bash
cat /tmp/rb/p_only.ckp
P
[ [ 1.0000, 0.5513 ] ,
  [ 0.5513, 1.0000 ] ]
```

Contents of posterior.var
```bash
cat /tmp/rb/posterior.var 
Iteration       P
0       [ [ 1.0000, 0.4456 ] ,
  [ 0.4456, 1.0000 ] ]
50      [ [ 1.0000, 0.2217 ] ,
  [ 0.2217, 1.0000 ] ]
100     [ [ 1.0000, 0.2366 ] ,
  [ 0.2366, 1.0000 ] ]
150     [ [ 1.0000, 0.4584 ] ,
  [ 0.4584, 1.0000 ] ]
200     [ [ 1.0000, -0.2595 ] ,
  [ -0.2595, 1.0000 ] ]
```

Although the matrices are written with line breaks, manually fixing them does not solve the issue. e.g., after updating `p_only.ckp`

```bash
cat /tmp/rb/p_only.ckp
P
[ [ 1.0000, 0.5513 ] ,  [ 0.5513, 1.0000 ] ]
```

```r
mymcmc.initializeFromCheckpoint(outdir + "p_only.ckp")                                                              
   Error:       Could not resurrect object from string value:
   [ [ 1.0000, 0.5513 ] ,  [ 0.5513, 1.0000 ] ]
```

Manually fixing `posterier.var` likewise leads to a segfault.

**Expected behavior**
The mcmc object should correctly initialize from the checkpoint, and readStochasticVariableTrace should not segfault.

**Screenshots**
Not applicable

**Computer info**
OS: Redhat Enterprise Linux 8. 
RevBayes Version: 1.3.1

**Additional context**
I'm attempting to run MuSSCRat (May and Moore 2020. Syst. Biol.). Also described here: https://www.edwardburress.com/code. The MuSSCRat script can be found here: https://www.edwardburress.com/_files/ugd/9c5270_5234695ff8d345f1bf2107a7e80519e1.txt?dn=musscrat_script.txt

curiosusJR:
Multi-line outputs cause broken on the trace and checkpoint system, which mainly caused by incomplete defined print behavior of `MatrixReal` which was relying on multi-line user friendly output, but did not provide complete storage/restore behavior for trace and checkpoint use. 

To fix this, this PR adds explicit `Printable` and `Serializable` support to `MatrixReal`, defines a canonical single-line storage format for whole-matrix values, and implements string/file resurrection for that format. 

As a part of trace system, it is worth noting that `mnModel` does not currently record `MatrixReal`. However, values ​​from Stochastic nodes in distributions such as `dnLKJPartial()` should, according to the current design, appear in `mnModel`.

Following the existing vector/`Simplex` pattern where appropriate, `mnModel` now flattens matrices into scalar trace columns with names like `P[1][1]`, while `mnStochasticVariable` and checkpointing preserve the matrix as a single whole-object value.

Changes summary: 
 - Implement Printable and Serializable support for MatrixReal 
       - initFromString
      - toJSON
      - printForUser
      - printForSimpleStoring
      - printForComplexStoring
      - writeToFile
 - Define a canonical single-line storage format for whole-matrix values:
    - `[[a,b],[c,d]]`
  - Keep flattened matrix output in row-major order for tabular traces
  - Keep human-readable output in `printForUser`
  - Allow `MatrixReal` values to be restored correctly through `initFromString`
  - Add matrix-aware `mnModel` support for matrix-valued nodes, with flattened headers such as:
    - `P[1][1]`, `P[1][2]`, ...


davidcerny (changes requested):
This works great – thank you! I was particularly impressed by the following:

- We can, but don't have to, employ `mnStochasticVariable` to resurrect variables of type `MatrixReal` – `mnModel` is fully sufficient.
- If we add `stochasticOnly=TRUE` to `mnModel`, the matrix is still correctly recorded. This [used to be](https://github.com/revbayes/revbayes/pull/900) a problem for vectors, where only the individual elements are stochastic (while the vector as a whole is deterministic). As pointed out in the PR cover letter, `MatrixReal` resembles `Simplex` more than vectors in this respect.

At first I was a bit skeptical of using different representations in `mnStochasticVariable` and `mnModel`, but that's exactly what we already do with vectors, so it makes sense to do it for matrices as well.

As a result, the only request I'd make is (as usual) to add a test. You could either use the script from the original issue (#836), or you could modify the [`mcmc_mvBM.Rev`](https://github.com/revbayes/revbayes/blob/master/tests/test_mvBM/scripts/mcmc_mvBM.Rev) along the following lines:

```r
monitors[++mni] = mnModel(filename="output/primates_mvBM.log", printgen=10, separator=TAB)
monitors[++mni] = mnStochasticVariable(filename="output/primates_mvBM.var", printgen=10, append=TRUE)

mymcmc = mcmc(mymodel, monitors, moves)
mymcmc.burnin(generations=5000,tuningInterval=250)
mymcmc.run(generations=2500, checkpointFile="output/primates_mvBM.ckp", checkpointInterval=500)

mymcmc.initializeFromCheckpoint("output/primates_mvBM.ckp")
mymcmc.run(generations=2500)
```

src/core/dag/TypedDagNode.cpp
```diff
@@ -81,6 +82,56 @@ bool TypedDagNode<Simplex>::isSimpleNumeric(void) const
 {
     return true;
 }
+
```
bredelings, line 85: Maybe add a simple comment explaining why isSimpleNumeric is true -- I presume this means "contains only numbers" versus "is a single number"?

src/core/datatypes/math/MatrixReal.cpp
```diff
@@ -603,6 +654,170 @@ size_t MatrixReal::size( void ) const
 }
 
 
+void MatrixReal::initFromString(const std::string &s)
```
bredelings, line 657: I wonder if we should use JSON to initialize and resurrect from strings.
If so, I wonder if we should do that now or later.

curiosusJR: JSON is indeed a better format for reading and writing, but I think maybe also keep this function. Considering in PPS, the model still needs to initialize from mnStochasticVariable, and if it maintains the support for the delimiter file format, then initializeFromString here still seems necessary?

curiosusJR: Actually, I'm not sure if I understand correctly, but mnStochasticVariable is designed to fully recover the internal state, right? If so, preserving only the structured format seems like a better design. I think `initFromString` also kind of cause the bug in #558 .

src/core/dag/TypedDagNode.cpp
curiosusJR: I make it return true because this is requirement of output in mnModel. (ref: after the 'Step 3' in comments of ModelMonitor::resetDagNodes ). I also refered the RbVector and Simplex, both of them did so, although there is a comment on `isSimpleNumeric` in `src/core/dag/DagNode.cpp` said vectors/matrices should not be true.

src/core/datatypes/math/MatrixReal.cpp
curiosusJR: Oh wait, I am sorry, did you mean using JSON to parse matrix string? I can do that.

davidcerny (approved):

Issue #519, opened 2024-07-23
[Feature Request] Tests for the checkpointing system


bjoelle:
Context : there have been quite a few issues recently tied to the checkpointing, and most of it seems to be tied to newly introduced proposals which had bugs that were not detected by the test suite

Proposed resolution : since the checkpointing system is a widely used feature for anything aside from toy analyses, we should have tests for that system specifically

davidcerny:
I'd be happy to help with adding this to the test suite. I guess there's no point in having a dedicated checkpointing test, since that would necessarily only cover a narrow subset of moves. Maybe a more far-reaching change would be helpful, where all of our current tests that involve running MCMC have their final lines changed to something like this:

```
# Run the MCMC
mymcmc.run(generations=30, checkpointInterval=30, checkpointFile="output/state.state")

# Initialize from a checkpoint file
mymcmc.initializeFromCheckpoint("output/state.state")

# Run the MCMC again
mymcmc.run(generations=30)

# Quit RevBayes
q()
```

This would have to be coupled with a careful examination of the expected output. For example, if the current version generates flat traces after restart for some parameters (as we've seen with one of our recent issues), we don't want future versions to past the test just because they suffer from the same problem!

Then there is the problem that checkpointing is still only a partially supported functionality; it's not implemented for `powerPosterior` analyses and it doesn't work with MPI. (I've recently talked about this with @brpetrucci – to me, this is currently the single most limiting feature of RevBayes, in the sense that it restricts the range of analyses I'm able to run.) Some of our tests contain an `output_expected_mpi` directory in addition to the regular expected output, though I don't think we actually check its contents.

davidcerny:
Hmm... so unfortunately, there is a reason why we don't do this. The test runner can deal with the `*_mcmc.state` and `*_moves.state` files, but the `*.state` files are apparently architecture- and even compiler-specific.

davidcerny:
Hmm... thinking of it further, this might not really be a fundamental obstacle. We just wouldn't include the `*.state` file in the expected output.

PR #987, opened 2026-03-29
Fix dnMixture crash during checkpoint restore and posterior predictive simulation

### LINKED ISSUE #558: [Help Wanted] Get error when doing posteriorPredictiveSimulation() under a mix_branch_length model

curiosusJR:
Fixes #558 

## Bug description
When restoring dnMixture state from a trace (mnStochasticVariable) or checkpoint, setValue() tries to reconstruct the index by `==` – comparing the restored value against every component:
```
  for (index = 0; index < vals.size(); ++index)
  {
      if ( vals[index] == *v )
          break;
  }
```
 This fails for two independent reasons:
  1. Ordering. The mixture node can be restored before its component parameters. The components still hold stale values, so no match is possible.
  2. Precision. Values round-trip through string serialization (double -> string -> double). The restored value and the component value are not bitwise equal, so `==` always fails, especially real numbers have different print precision with vectors by default.

When no match is found, index overflows to vals.size() (e.g. 2 of 2). Later, `touchSpecialization()` accesses parameter_values->getValue()[index] — out of bounds, crash: "Vector index out of range: 2 of 2".

`MixtureDistribution` samples by picking an internal allocation index from a categorical distribution, then returning component_values[index]. So the index is the true latent variable; the value is derived from it. And the index cannot be recovered simply from the value, it requires more information, which causes the current problem.

Based on the above considerations, explicitly recording the actual variables, which is the hidden state index here, where for the model recovery, and recording the numerical values ​​​​where focus on the numerical analysis might be a good solution.

## PR summary
Interface (Distribution.h): adds two virtual methods with no-op defaults:
  - getHiddenStateString() — returns "" by default (no hidden state)
  - setHiddenStateFromString(s) — no-op by default

MixtureDistribution overrides both, delegating to existing getCurrentIndex() / setCurrentIndex():
  - getHiddenStateString() → std::to_string(getCurrentIndex())
  - setHiddenStateFromString(s) → setCurrentIndex(std::stoul(s))

Writing (StochasticVariableMonitor, Mcmc::checkpoint): for nodes whose distribution reports hidden state, writes the hidden state as a quoted string (e.g. '1') instead of the numeric value. User-facing monitors (mnModel, mnScreen) are unchanged — they still write values.

Reading (StochasticNode::setValueFromString): checks the string format:
  - Quoted ('1') → hidden state → calls setHiddenStateFromString, bypassing setValue() and the broken == search
  - Plain (0.594) → value → original ressurectFromString + setValue() path

Safety (MixtureDistribution::setValue): if the == search fails and index overflows, prints a warning and clamps index to 0 instead of silently overflowing. This only probably happens when restoring from old trace files without quoted indices (new traces bypass setValue() entirely). The impact is minor: in PPS the index is not used (simulation depends on the value, not the index); in MCMC it causes a slightly wrong acceptance rate until the next allocation move corrects the index. A warning is sufficient — exiting would be too harsh for a self-correcting state.

--- 

\* It is worth pointing out that the same class of bug (hidden state lost during serialization) exists in other distributions but is not addressed here. (maybe like ReversibleJumpMixtureConstantDistribution, EmpiricalDistribution and UniformPartitioningDistribution)
\* Considering the index is the true latent variable, maybe refactor dnMixture to be purely index-based — make the stochastic node hold a size_t index as its value, with the actual mixture value derived through a deterministic lookup node. Or just using index instead of value in some other case.


davidcerny:
Hi @curiosusJR, could you possibly add the script from issue #558 (or maybe a simplified version thereof) as an integration test? As pointed out by @hoehna, having a test would be useful even if we ultimately decide to address this problem in some other way.

src/core/distributions/MixtureDistribution.h
```diff
@@ -277,7 +295,13 @@ void RevBayesCore::MixtureDistribution<mixtureType>::setValue(mixtureType *v, bo
             break;
         }
     }
-    
+
+    if ( index >= vals.size() )
+    {
+        std::cerr << "Warning: mixture allocation index could not be reconstructed from value. Hidden state not restored." << std::endl;
+        index = 0;
```
davidcerny: Should we maybe throw an exception here, rather than just a warning?

davidcerny:
A few of us discussed this PR during our last monthly meeting (May 5), and three different proposals for solving this problem were put forward:

1. Let's go with the solution implemented in this PR, i.e., recording the latent variable corresponding to the index.
2. Use the original solution based on the `vals[index] == *v` comparison, but allowing for a tolerance instead of requiring strict equality.
3. Use an even more relaxed comparison where the `index` simply points to that component which is numerically closest to the restored value.

Solution (2) would be somewhat attractive if the problem that this PR attempts to solve was strictly checkpointing-related. However, while we discussed it as such, that's not the case: as shown by the original issue (#558), the problem also affects posterior predictive simulations, where the category values are extracted not from a specialized checkpoint file, but from a standard parameter trace produced by `mnStochasticVariable`. Such a trace records the values with a much lower precision (7 sig figs) than a checkpoint file, so the tolerance would potentially have to be rather huge. Solution (3) gets around this problem, but feels (to me, at any rate) very hacky.

Moreover, neither solution (2) nor solution (3) addresses the first of the two failure modes mentioned in the PR cover letter:

> Ordering. The mixture node can be restored before its component parameters. The components still hold stale values, so no match is possible.

As a result, I think I'm convinced that the solution put forward in this PR – i.e., recording the latent index – really is the best way to handle this. If my previous feedback is addressed, I'd be happy to approve.

bredelings:
Recording the latex index seems best to me as well.  The question is more about how to do so,

This PR introduces a general infrastructure for storing extra information besides just the value at a stochastic node - the "hidden state".  The hidden state seems to be basically the state of the distribution object that backs the stochastic node, minus the value.  So... could we just serialize the distribution object -- including both its state and its value -- instead of trying to divide this into "value" and "hidden"?  For a "stochaministic node", where the "hidden" state might determine the value, perhaps we should really store ONLY the "hidden" state, and the regenerate the value from the hidden state when we reload the hidden state (I haven't thought this through carefully).

So, I guess I prefer a different implementation where we simply serialize the underlying distribution, including both value and hidden state.  That is logically simpler as well.

But the current implementation is not bad, so perhaps we should not let the perfect become the enemy of the good.  I'm not against merging the current approach, since it doesn't seem to block progress toward more complete serialization, and in some ways documents what is missing.  Thoughts?

davidcerny:
@bredelings I'm definitely always in favor of not letting the perfect become the enemy of the good. 🙂 

> So... could we just serialize the distribution object -- including both its state and its value -- instead of trying to divide this into "value" and "hidden"? For a "stochaministic node", where the "hidden" state might determine the value, perhaps we should really store ONLY the "hidden" state, and the regenerate the value from the hidden state when we reload the hidden state (I haven't thought this through carefully).
>
> So, I guess I prefer a different implementation where we simply serialize the underlying distribution, including both value and hidden state. That is logically simpler as well.

I agree that these ideas for more complete serialization may be worth exploring as a potential improvement on our current checkpointing system. However, I'm not sure they would represent a viable solution to the other one of the two problems that this PR is intended to address, i.e., restoring the hidden state from a trace written by a regular monitor – something we need to do to set up posterior predictive simulations.

In a way, the PPS problem is analogous to our past attempts to provide infrastructure for re-initializing a sampler from a trace, which we abandoned (for good reasons; see #314, #646).

davidcerny (approved):
One of the newly added tests is failing, but that's just because it includes the primary checkpoint file (`p.ckp`) in the expected output. This is a known problem; since checkpoint files write out parameter values with greater precision than we normally use, they tend to exhibit platform-specific differences in the last few decimal places (see #519). Note that this problem doesn't extend to the auxiliary checkpoint files (`p_mcmc.ckp`, `p_moves.ckp`); nor does it affect the trace produced by resuming an analysis from a checkpoint. That's why checkpointing tests still work even if they don't include the main checkpoint file in the expected output.

I also took the liberty of removing the `mnScreen` monitor from the [`MLB_model.Rev`](https://github.com/revbayes/revbayes/blob/fix_MBL/tests/test_mixtures/scripts/MBL_model.Rev) test. After making these changes, the PR looks good to me!

PR #990, opened 2026-04-02
Make meson always build rb-help2yml

bredelings:
A few months ago (?) we made `projects/cmake/build.sh` always build `rb-help2yml`.  Let's make meson also do the same.  (This doesn't mean that program ever gets run, but at least it is always available)

I included a few other tweaks to `projects/meson/build.sh`.  (I don't actually use this script, but it does seem to work, so lets improve it).


basantakhakurel (approved):
Looks good to me!
Thanks!

Issue #936, opened 2026-01-20
[Feature Request] Remove the `range-v3` library


davidcerny:
This is certainly not the most pressing problem we have on our hands, but the `range-v3` library keeps triggering compiler warnings, at least with Clang on macOS:

```
/Users/runner/work/revbayes/revbayes/src/libs/range-v3/range/v3/utility/compressed_pair.hpp:110:20: warning: 'compressed_tuple' is deprecated: ranges::compressed_tuple is deprecated. [-Wdeprecated-declarations]
  110 |             return compressed_tuple<bind_element_t<Args>...>{
      |                    ^
[  5%] Building CXX object core/CMakeFiles/rb-core.dir/moves/proposal/tree/EmpiricalTreeTopologyProposal.cpp.o
/Users/runner/work/revbayes/revbayes/src/libs/range-v3/concepts/concepts.hpp:653:27: note: expanded from macro 'CPP_auto_fun'
[  5%] Building CXX object revlanguage/CMakeFiles/rb-parser.dir/moves/tree/Move_TreeScale.cpp.o
  653 | #define CPP_auto_fun(X) X CPP_AUTO_FUN_IMPL_
      |                           ^
/Users/runner/work/revbayes/revbayes/src/libs/range-v3/concepts/concepts.hpp:656:47: note: expanded from macro 'CPP_AUTO_FUN_IMPL_'
  656 | #define CPP_AUTO_FUN_IMPL_(...) (__VA_ARGS__) CPP_AUTO_FUN_RETURNS_
[  5%] Building CXX object revlanguage/CMakeFiles/rb-parser.dir/moves/tree/Move_NarrowExchange.cpp.o
      |                                               ^
[  5%] Building CXX object core/CMakeFiles/rb-core.dir/moves/proposal/tree/IndependentTopologyProposal.cpp.o
/Users/runner/work/revbayes/revbayes/src/libs/range-v3/concepts/concepts.hpp:661:9: note: expanded from macro 'CPP_AUTO_FUN_RETURNS_'
  661 |         CPP_AUTO_FUN_SELECT_RETURNS_,                                           \
      |         ^
/Users/runner/work/revbayes/revbayes/src/libs/range-v3/range/v3/utility/compressed_pair.hpp:110:20: note: in instantiation of template type alias 'compressed_tuple' requested here
[...]

PR #927, opened 2026-01-10
Switch from c++17 to c++23

bredelings:
Enable c++23.

Issue #794, opened 2025-07-23
[BUG] Core dump in fossil only phylogenies with tuningInterval = 0


NiklasHohmann:
**Describe the bug**
I run a fossil only inference using FBD with rho = 0.0 to model observing fossils an interval [t_1, t_2], but knowing the taxa survive after t_2 plus there is no special sampling effort at t_2. I experienced significant slowdown during the tuning interval (to the point where the progress bar does not move for 24 h, but CPU usage remains at 100 %, see #795 ). Turning off the tuning interval by setting `tuningInterval = 0` results in the message `Floating point exception (core dumped)`

**To Reproduce**

Minimum running example for the core dump. Removed all moves on the tree but the error persists.

[fossils_cont_rho0_1.csv](https://github.com/user-attachments/files/21387967/fossils_cont_rho0_1.csv)

```
taxa <- readTaxonData( "fossils_cont_rho0_1.csv")

moves = VectorMoves()

fbd_tree ~ dnFBDP(origin=2.0, lambda=1.0, mu=0.3, psi=1.0, rho=0.0, taxa=taxa)

branch_rates_morpho ~ dnExponential(lambda = 10.0)
moves.append( mvScale(branch_rates_morpho, weight = 16.0))
Q_morpho := fnJC(2)

phy_morpho ~ dnPhyloCTMC(tree = fbd_tree, branchRates = branch_rates_morpho, Q = Q_morpho, type = "Standard", coding = "all")
mymodel = model(phy_morpho)
monitors = VectorMonitors()
monitors.append( mnScreen(printgen=1000) )

mymcmc = mcmc(mymodel, monitors, moves, nruns = 2)
mymcmc.burnin(generations=2000,tuningInterval=0)
mymcmc.operatorSummary()
mymcmc.run(generations=10000)
mymcmc.operatorSummary()
q()
```

**Expected behavior**
The burnin runs without tuning of the moves

**Computer info**
Running on WSL2 (Ubuntu 24.04.2 LTS), revBayes 1.3.0

**Additional context**
Potentially connected to #792

davidcerny:
I can't reproduce this is – the script runs fine for me both with `tuningInterval=0` and with `tuningInterval=100` (macOS 15.5, current tip of development: 0ea58fb). Just as a quick note, though, I wasn't able to execute the script exactly as-is, because `readTaxonData()` didn't correctly parse the provided CSV file:

```
[...]

PR #991, opened 2026-04-04
Checkpointing for power posterior analyses

### LINKED ISSUE #986: [Feature Request] Implement checkpointing for `powerPosterior` analyses
Marginal likelihood estimation using power posterior sampling is famously very computationally intensive, and since we have to converge onto the posterior before starting to trace our way back to the prior, one should expect it to be quite time-consuming as well. Because our implementation of power posterior sampling employs smart parallelization (see [Höhna et al. 2021](https://peerj.com/articles/12438/) for details), many users will likely want to run it on HPC clusters, which may impose wall-clock runtime limits that are not sufficient for the analysis to finish. It would therefore be great to provide a checkpointing functionality to get around this problem. However, while both `mcmc()` and `mcmcmc()` have an `.initializeFromCheckpont()` method (as well as `checkpointFile` and `checkpointInterval` arguments in their `.run()` method), `powerPosterior()` does not.

**Describe alternatives you've considered**
Some (rather tortuous) workarounds for the lack of a power posterior checkpointing functionality are described at #697.

**Additional context**
Implementing this should be relatively easy, as the problem is very similar to checkpointing MC^3 analyses (see #670). As in the latter case, we would probably delegate most of the work to the individual chains, and only provide an extra checkpoint file to store the power to which we are raising a given chain's likelihood.

davidcerny:
This should be ready, I think, but as the description below makes clear, there's a lot of moving parts here. If people could subject it to some stress testing, that would be great.

Adressing #986 proved to be far more challenging than I anticipated. This is because `powerPosterior()` is rather different from `mcmc()` or `mcmcmc()`: the last two functions assign a separate sampler to every run, or to every chain of every run, but when we carry out a power posterior analysis, all stones / powers / betas share the same sampler (for a good reason).

Here are some details about the system I implemented:

## What can be checkpointed

Here, a basic refresher can be useful. A power posterior analysis consists of up to four distinct stages, which we can illustrate using the following example:

```r
pow_p = powerPosterior(mymodel, moves, monitors, "analysis.out", cats=10, sampleFreq=10)
pow_p.burnin(generations=10000, tuningInterval=200)
pow_p.run(generations=1000)
```

1. A global pre-burnin phase, specified by `pow_p.burnin()` and lasting 10,000 generations in the example above. This one is equivalent to `mcmc.burnin()` – we use it to tune our moves to help the analysis converge onto the posterior. No samples are collected, and we use a separate generation counter.
2. A stone-specific pre-burnin phase. This one serves to do an extra bit of tuning when switching between the powers to which we are raising the likelihood (which makes sense, since we may want to make bolder moves for powers closer to 0). By default, this pre-burnin doubles the specified stone length, so in the example above, it would consist of 1000 generations for each stone. However, we could change this using `pow_p.run(..., preburninGenerations=<>)`. No samples are collected during the stone-specific pre-burnin, and another separate generation counter is used.
3. A stone-specific burnin phase. This one serves to discard samples collected during the transition from one likelihood power to another, before we have fully converged to the next power posterior. This phase has no direct counterpart in normal MCMC analyses – in those, you usually have to identify and discard your burnin manually, after the fact. By default, 25% of the specified stone length is treated as burnin, so in the example above, it would consist of 250 generations for each stone. However, we could change this using `pow_p.run(..., burninFraction=<>)`. The burnin phase shares its generation counter with the sampling phase, so the first entry in the "Iteration" column of our log files will actually be 250: generations 0–240 have been discarded as burnin.
4. A stone-specific sampling phase. Here, we are simply collecting samples from a given power posterior. Its length is specified by `pow_p(generations=<>}`, so in the example above, it would consist of 750 generations for each stone: the 1000 specified generations minus the 25% reserved for burnin. As noted above, the generation counter is shared between the burnin phase and the sampling phase.

Of course, since stones can be run sequentially, stage 2 of one stone can be executed after stages 3 and 4 of another stone – these are "stages" in a conceptual rather than strictly temporal sense of the word.

One of the changes made in this PR is that we now distinguish between these four stages in the screen output. Executing the example code now looks as follows:

Screenshot_red
<br/><br/>

After this lengthy preamble, then answer is: we can restore a `powerPosterior` sampler from stages 1, 3, or 4. We cannot resume the stone-specific pre-burnin from a checkpoint, just like we cannot do it with the pre-burnin in MCMC and MC^3 analyses. However, I did implement a mechanism to allow resuming the global pre-burnin, because the sampler is quite likely to spend a lot of time in this stage – potentially more than in all the remaining stages combined, if the analysis is highly parallelized.

## How to checkpoint

Resuming an analysis that was interrupted in its global pre-burnin stage is simple:

```r
pow_p.burnin(generations=10000, tuningInterval=200, checkpointFile="prelim.ckp",
             checkpointInterval=200)
pow_p.initializeFromCheckpoint("prelim.ckp")
pow_p.burnin(tuningInterval=200)
```

One difference here compared to what we'd do in an MCMC or MC^3 analysis is that we specify the `generations` argument in the first call to `.burnin()` but not in the second one (i.e., upon resumption). This is because the analysis automatically records the planned length of 10,000 generations specified by the first call in a special file (`prelim_burnin_info.ckp`) and, once resumed, only runs as many generations as needed to reach this target. However, we can also do things in a way that's more reminiscent of MCMC or MC^3 checkpointing, and explicitly set the `generations` argument in the second call as well:

```r
pow_p.burnin(generations=10000, tuningInterval=200, checkpointFile="prelim.ckp",
             checkpointInterval=200)
pow_p.initializeFromCheckpoint("prelim.ckp")
pow_p.burnin(generations=10000, tuningInterval=200)
```

This will override the originally planned length, and cause the global pre-burnin stage to be _extended_ by 10,000 generations, regardless of how many generations were completed before the interruption.

As it turns out, this trick also allows us to seed the power posterior sampler with checkpoint files produced by a regular MCMC or MC^3 analysis. That's because the global pre-burnin stage only really requires two files: the base checkpoint file storing the variable names and values (`prelim.ckp`), and an additional file storing the move info (`prelim_moves.ckp`). Both of these will also be produced by a regular MCMC or MC^3 analysis, if one has set up checkpointing for it. The `prelim_burnin_info.ckp` file, which is unique to power posterior analyses, is not required as long as the `generations` argument is explicitly specified. Therefore, if your workflow involves running both MCMC and power posterior analyses on the same dataset, under the same model, and with the same moves, you might be able to skip a lengthy global pre-burnin stage altogether:

```r
pow_p.initializeFromCheckpoint("mcmc.ckp") # load checkpoint files from a pre-existing analysis
pow_p.burnin(generations=10)               # short burnin just to get the sampler started
pow_p.run(generations=1000)                # proceed to actual power posterior sampling
```

How about checkpointing and restarting individual stones, i.e., power posteriors? Let's assume we set up an analysis as follows:

```r
pow_p = powerPosterior(mymodel, moves, monitors, "analysis.out", cats=50, sampleFreq=10)
pow_p.burnin(generations=10000, tuningInterval=200)
pow_p.run(generations=1000, checkpointFile="analysis.ckp", checkpointInterval=10)
```

and let's further assume that we run it on 8 parallel workers. We can define a "worker" either as a single process, or as a group of `procPerLikelihood` processes if we set this argument to something other than the default value of 1.

A table printed at the beginning of the analysis will inform us that the 50 stones will be distributed among the 8 processes as follows:

```
Process | Step 1 | Step 2 | Step 3 | Step 4 | Step 5 | Step 6 | Step 7
--------|--------|--------|--------|--------|--------|--------|--------
      1 |      1 |      2 |      3 |      4 |      5 |      6 |
      2 |      7 |      8 |      9 |     10 |     11 |     12 |
      3 |     13 |     14 |     15 |     16 |     17 |     18 |
      4 |     19 |     20 |     21 |     22 |     23 |     24 |     25
      5 |     26 |     27 |     28 |     29 |     30 |     31 |
      6 |     32 |     33 |     34 |     35 |     36 |     37 |
      7 |     38 |     39 |     40 |     41 |     42 |     43 |
      8 |     44 |     45 |     46 |     47 |     48 |     49 |     50
```

If the analysis was interrupted in its last step (step 7), you'd want to resume just 2 stones: 25 and 50. You can do that easily by specifying the optional `stones` argument of the `.initializeFromCheckpoint()` method, which can either be a single index, or a vector thereof:

```r
pow_p.initializeFromCheckpoint("analysis.ckp", [25, 50])
pow_p.run(generations=800)
```

Now, it could happen that the original analysis was specifically interrupted at generation 200 of step 7. Since we run each stone for 1000 generations, and since `burninFraction` equals 0.25 by default, this means that the interruption took place within the 250-generation-long burnin phase. What happens in this case is that the command above will cause the analysis to complete the remaining 50 generations of burnin, followed by sampling for 750 generations (numbered 250 through 1000) – exactly as though no interruption had taken place in the first place. This is because the `*_stone_info.ckp` checkpoint file stores a variable called `planned_burnin` (in this case set to 250), and the resumption mechanism can compare its value to the generation counter stored in `*_mcmc.ckp` file.

This is convenient, but one consequence of doing things this way is that the `burninFraction` argument of the resumed run is ignored: if it were not, we would discard the first 200 generations of the resumed run (0.25 times 800), and not just 50. When that happens, we warn the user:

```
Running power posterior analysis ...
Stone pre-burnin: -, burnin: +, sampling phase: *
Step  1 / 10		      Finishing a previously scheduled and interrupted burnin stage; burninFraction of the current run() is ignored.
```

The screen output of the resumed analysis then follows the now-familiar pattern:

```
++++**************************************
```

There is a key case we haven't covered yet: what if we want to not just resurrect an interrupted stone, but also to use it as a starting point for a whole sequence of stones that we originally didn't get around to running at all? One can imagine handling this in multiple ways. For example, we could store the original scheme assigning stones to processes (cores), and consult it upon resumption to figure out what to do. The idea is superficially appealing, but runs into considerable practical difficulties. Can we really guarantee – or to put it another way, should we require – that the resumed analysis (will) be always run under the same conditions as the original one? Shouldn't you, for example, be free to take the files from an interrupted analysis run on 128 cores on an HPC cluster, and finish that analysis on your laptop with 8 cores?

To allow this freedom, we instead pass a nested vector (vector of vectors) to `.initializeFromCheckpoint()` via the `stones` argument. The elements of the outer vector represent stone sequences, each one of which is assigned to one parallel worker. Each inner vector then represents individual stones to be executed sequentially by a given worker. If the user specifies more stone sequences than there are available resources (i.e., if the length of the outer vector exceeds `floor( num_processes / procPerLikelihood )`), we throw an exception.

Going back to the example above, if the analysis was interrupted during step 2, we could restore it on 8 workers as follows:

```r
pow_p.initializeFromCheckpoint( "analysis.ckp",
                                [ 2:6, 8:12, 14:18, 20:25, 27:31, 33:37, 39:43, 45:50 ] )
```

Note that the second argument uses a condensed notation for the following nested vector:

```
[ [ 2, 3, 4, 5, 6 ],
  [ 8, 9, 10, 11, 12 ],
  [ 14, 15, 16, 17, 18 ],
  [ 20, 21, 22, 23, 24, 25 ],
  [ 27, 28, 29, 30, 31 ],
  [ 33, 34, 35, 36, 37 ],
  [ 39, 40, 41, 42, 43 ],
  [ 45, 46, 47, 48, 49, 50 ] ]
```

If the analysis was instead interrupted during step 6, we could write:

```r
pow_p.initializeFromCheckpoint( "analysis.ckp",
                                [ [6], [12], [18], 24:25, [31], [37], [43], 49:50 ] )
```

A word of caution: when using the range-based notation (with the colon), it makes a huge difference whether the range is enclosed in square brackets (`[]`) or not. The following:

```r
pow_p.initializeFromCheckpoint( "analysis.ckp", 1:5 )
```

means "resurrect stones 1 through 5, and leave it up to the inner machinery of `powerPosterior` to decide how to distribute them across available cores." (Ideally, we would run this on 5 cores, or 5 times `procPerLikelihood` cores.) Checkpoint files must be available for all 5 specified stones. In contrast,

```r
pow_p.initializeFromCheckpoint( "analysis.ckp", [1:5] )
```

means "resurrect stone 1, and then run stones 2 through 5 on the same core, using the last sample from the previous stone as the starting point for the next stone". In this case, we only require checkpoint files for stone 1 – we don't expect to have them for stones 2–5, and if we do for some reason have them, we will ignore them.


src/core/analysis/PowerPosteriorAnalysis.cpp
```diff
@@ -322,43 +510,156 @@ void PowerPosteriorAnalysis::runStone(size_t idx, size_t gen, double burnin_frac
...
+            sampler->writeMonitorHeaders( false );
+        }
+        // Do not re-print the checkpoint generation before the next cycle (avoids duplicate monitor lines).
+        if ( not stone_resumes_from_checkpoint )
+        {
+            sampler->monitor(k_start);
+        }
+    }
+    
+    // Open the output file: append if resuming, overwrite if starting fresh.
```
curiosusJR, line 834: The output file is written directly through a plain ofstream, outside the monitor system, so `AbstractFileMonitor::truncateAfterGeneration()` is never called on it, may resulting in duplicated generations after checkpoint resume.

curiosusJR: Maybe `truncateAfterGeneration()` could be extracted into a shared function in RbFileManager?

davidcerny: That's a great point... an alternative solution would be to let the writing of `.out` files be handled by an existing monitor, which we would implicitly set up in the constructor. We don't do this sort of thing – automatically creating a monitor directly in the analysis object, behind the user's back – very often, but there _is_ a precedent for it, since `ValidationAnalysis` sets up a `StochasticVariableMonitor` in this way.

However, there might be no existing monitor that lets us print exactly what we need here (three columns named `state`, `power`, and `likelihood`). So we would have to create one first.

davidcerny: Hmm... your solution was better, so I implemented it, @curiosusJR 🙂

bjoelle (commented):
two comments from me
* I wonder if it would be a good idea to print the repartition of stones at the beginning either in the main output or directly to the checkpoint file - otherwise I worry that it will be very cryptic for the user which stones should be restarted
* as far as I can tell none of the SS tests actually use the new checkpointing, no ?

davidcerny:
> I wonder if it would be a good idea to print the repartition of stones at the beginning either in the main output or directly to the checkpoint file - otherwise I worry that it will be very cryptic for the user which stones should be restarted

Do you mean printing it at the beginning of the original analysis, i.e., the first time we call `pow_p.run(...)`? If so, then yes, I agree that could be helpful. We could print a table similar to the one I included in the PR cover letter (and in the `powerPosterior()` documentation), except perhaps transposed to make it more legible when the number of CPU is large. We could then show which process (or a group of processes if `procPerLikelihood` is greater than 1 – we'd have to settle on a terminology that's general enough to cover all these cases) runs which stones, and in what order.

(One could argue that this information is already implicit in the progress bars – e.g., if you know you've specified 50 stones, and your progress bar shows 7 steps, that tells you you're using 8 cores, each handling 6 or 7 stones. But it's still non-trivial for the user to extract from this the exact assignment of stones to processes, so I agree we should make that more transparent.)

What this wouldn't tell the user, though, is which stones need to be restarted. It could help, but it would still have to be combined with an examination of the output directory: the user would have to check which stones are missing output files altogether (meaning that at the very least they haven't advanced to stage 4, i.e., the sampling phase), and which output files have fewer generations in them than specified. Unfortunately, as detailed above, I don't see any auto-magic way around this that doesn't result in an unacceptable loss of generality (e.g., not being able to run all stones for longer after they have completed their originally specified number of generations, like one can do with the regular MCMC).

> as far as I can tell none of the SS tests actually use the new checkpointing, no ?

Yes, that's correct, right now the PR doesn't contain any tests for the newly added functionality. I'll fix that. :)

bjoelle:
@davidcerny yes, I meant printing the table - we could even only do that if checkpointing is turned on
I think which steps haven't completed should be visible from the progress bar itself, but maybe I misunderstand

davidcerny:
Yes, that's a good point, the progress bars should help a lot in that respect. I'll get to work on the table printing!

ms609:
This looks like great progress to me!
My main comment would be that checkpointing in "pre-burnin" phases would be a high-value extension of this work.

With your model analysis that comprises 10000 generations in step 1, then 1000 generations in step 2, then 1000 during which checkpoints are made, then the first 11/12ths of a run must complete before the first checkpoint is created.  This means that the current implementation (as I understand it) would be most useful in runs that are already very close to completion when they time out.

In this instance, checkpointing would be twice as useful simply from a timing perspective if a single checkpoint could be created to allow resumption after step 1, and thus still had 2/12ths instead of 1/12th of its runtime to go.
(Such a checkpoint might also facilitate an increase in the number of 'stages' after an initial run?)

The utility of a "post-global-pre-burnin" checkpoint would be even greater if this would allow the global preburnin to execute as a separate job from individual stones.  Presently, for an 8-chain analysis, I need to reserve a hard-to-come-by 128-core slot, and watch 120 of those cores sit idle during the global pre-burnin.  Optimally I would submit a "pre-burn-in" SLURM job using resources tailored to that phase, then submit either one 128-core job to compute all my stones, or perhaps better multiple dependent jobs that each compute a modest number of stones independently.

bredelings:
BTW, note that `std::ranges::to<T>` is first supported in g++14.  g++12 is from 2022, and g++12 is from 2023.  So, as we drop older compilers we get to use incrementally more c++23 features. :man_shrugging: 

Now, `ranges::to<T>` IS available in the ranges-v3 library.

davidcerny:
Thanks for all the feedback, and apologies for the late reply!

@bjoelle

It was a bit of a pain, but we now do print a table showing which stones are assigned to which MPI workers (where a "worker" could either be a single process, or a group of `procPerLikelihood` processes), and in what order they are executed. I tried to test it on as many cases as I could, and I'm quite happy with the results.

This is what it looks like for 50 stones distributed among 8 processes:

```
The 50 requested stones will be executed in 7 steps using 8 processes as follows:

Process | Step 1 | Step 2 | Step 3 | Step 4 | Step 5 | Step 6 | Step 7
--------|--------|--------|--------|--------|--------|--------|--------
      1 |      1 |      2 |      3 |      4 |      5 |      6 |
      2 |      7 |      8 |      9 |     10 |     11 |     12 |
      3 |     13 |     14 |     15 |     16 |     17 |     18 |
      4 |     19 |     20 |     21 |     22 |     23 |     24 |     25
      5 |     26 |     27 |     28 |     29 |     30 |     31 |
      6 |     32 |     33 |     34 |     35 |     36 |     37 |
      7 |     38 |     39 |     40 |     41 |     42 |     43 |
      8 |     44 |     45 |     46 |     47 |     48 |     49 |     50
```

A single-step analysis with 16 stones distributed among 16 processes:

```
The 16 requested stones will be executed in 1 step using 16 processes as follows:

Process | Step 1
--------|--------
      1 |      1
      2 |      2
      3 |      3
      4 |      4
      5 |      5
      6 |      6
      7 |      7
      8 |      8
      9 |      9
     10 |     10
     11 |     11
     12 |     12
     13 |     13
     14 |     14
     15 |     15
     16 |     16
```

And the opposite extreme, where a large number of stones is executed sequentially on a single process:

```
The 128 requested stones will be executed in 128 steps using 1 process as follows:

Process | Step 1 | Step 2 | Steps 3--126 | Step 127 | Step 128
--------|--------|--------|--------------|----------|----------
      1 |      1 |      2 |       3--126 |      127 |      128
```

Here is the same 128-stone analysis distributed across 10 processes, because why not:

```
The 128 requested stones will be executed in 13 steps using 10 processes as follows:

Process | Step 1 | Step 2 | Steps 3--11 | Step 12 | Step 13
--------|--------|--------|-------------|---------|---------
      1 |      1 |      2 |       3--11 |      12 |
      2 |     13 |     14 |      15--23 |      24 |      25
      3 |     26 |     27 |      28--36 |      37 |      38
      4 |     39 |     40 |      41--49 |      50 |      51
      5 |     52 |     53 |      54--62 |      63 |      64
      6 |     65 |     66 |      67--75 |      76 |
      7 |     77 |     78 |      79--87 |      88 |      89
      8 |     90 |     91 |     92--100 |     101 |     102
      9 |    103 |    104 |    105--113 |     114 |     115
     10 |    116 |    117 |    118--126 |     127 |     128
```

And here is what we get when we interrupt that same analysis during step 3, and restart it as follows:

```r
pow_p.initializeFromCheckpoint( "output/model_uniform_run_all.ckp",
                                [ 3:12, 15:25, 28:38, 41:51, 54:64, 67:76, 79:89, 92:102, 105:115, 118:128 ] )
pow_p.run(generations=200)
```

```
The 108 requested stones will be executed in 11 steps using 10 processes as follows:

Process | Step 1 | Step 2 | Step 3 | Step 4 | Step 5 | Step 6 | Step 7 | Step 8 | Step 9 | Step 10 | Step 11
--------|--------|--------|--------|--------|--------|--------|--------|--------|--------|---------|---------
      1 |      3 |      4 |      5 |      6 |      7 |      8 |      9 |     10 |     11 |      12 |
      2 |     15 |     16 |     17 |     18 |     19 |     20 |     21 |     22 |     23 |      24 |      25
      3 |     28 |     29 |     30 |     31 |     32 |     33 |     34 |     35 |     36 |      37 |      38
      4 |     41 |     42 |     43 |     44 |     45 |     46 |     47 |     48 |     49 |      50 |      51
      5 |     54 |     55 |     56 |     57 |     58 |     59 |     60 |     61 |     62 |      63 |      64
      6 |     67 |     68 |     69 |     70 |     71 |     72 |     73 |     74 |     75 |      76 |
      7 |     79 |     80 |     81 |     82 |     83 |     84 |     85 |     86 |     87 |      88 |      89
      8 |     92 |     93 |     94 |     95 |     96 |     97 |     98 |     99 |    100 |     101 |     102
      9 |    105 |    106 |    107 |    108 |    109 |    110 |    111 |    112 |    113 |     114 |     115
     10 |    118 |    119 |    120 |    121 |    122 |    123 |    124 |    125 |    126 |     127 |     128
```

To further help the user figure out which stones they need to restart, I also changed the progress monitor. In PR #673, I implemented a system that would have produced the following output for 128 stones distributed among 10 processes:

```
Running power posterior analysis ...
Steps    1--10 / 128		*********************************
Steps   11--20 / 128		*********************************
Steps   21--30 / 128		*********************************
Steps   31--40 / 128		*********************************
Steps   41--50 / 128		*********************************
Steps   51--60 / 128		*********************************
Steps   61--70 / 128		*********************************
Steps   71--80 / 128		*********************************
Steps   81--90 / 128		*********************************
Steps  91--100 / 128		*********************************
Steps 101--110 / 128		*********************************
Steps 111--120 / 128		*********************************
Steps 121--128 / 128		*********************************
```

These "steps" accurately capture the cumulative number of stones that have been or are being executed at any given moment, but not their indices, and they do not correspond to the steps from the table. Moreover, the system doesn't really account for restored analyses. I have fixed these issues, and the corresponding output now looks as follows:

```
Running power posterior analysis ...
Stone pre-burnin: -, burnin: +, sampling phase: *
Step  1 / 13		---------------------------------++++++++*************************
Step  2 / 13		---------------------------------++++++++*************************
Step  3 / 13		---------------------------------++++++++*************************
Step  4 / 13		---------------------------------++++++++*************************
Step  5 / 13		---------------------------------++++++++*************************
Step  6 / 13		---------------------------------++++++++*************************
Step  7 / 13		---------------------------------++++++++*************************
Step  8 / 13		---------------------------------++++++++*************************
Step  9 / 13		---------------------------------++++++++*************************
Step 10 / 13		---------------------------------++++++++*************************
Step 11 / 13		---------------------------------++++++++*************************
Step 12 / 13		---------------------------------++++++++*************************
Step 13 / 13		---------------------------------++++++++*************************
```

Between this and the table, the user will be able to identify the stones that failed to execute if the analysis was interrupted at a particular step.

@ms609

> With your model analysis that comprises 10000 generations in step 1, then 1000 generations in step 2, then 1000 during which checkpoints are made, then the first 11/12ths of a run must complete before the first checkpoint is created. This means that the current implementation (as I understand it) would be most useful in runs that are already very close to completion when they time out.

This would be the case if all 10 stones from my model analysis were executed simultaneously, by distributing them among 10 processes. My example actually assumed that they are all executed sequentially (which is why the progress monitor in the screenshot shows 10 lines instead of just one). In this scenario, things aren't quite so bad: only the first 11/30 of a run must be completed before the first checkpoint is created, and the probability that at any given moment you find yourself in a "checkpointable" stage of the analysis goes up from 1/12 to 1/3.

However, your general point still stands. If you are fully exploiting the parallelization of power posterior sampling and executing, say, 128 stones all at the same time on an HPC cluster (something I've definitely done before), then the new functionality will be of very limited benefit to you.

For me, though, the main implication of this is not so much that we should be able to checkpoint the stone-specific pre-burnins: that would increase the checkpointable portion of the analysis from 1/3 to 2/3 of its total length in my example, but only from 1/12 to 2/12 in yours, and we can always shorten this stage by manipulating the `preburninGenerations` argument. (Interestingly, the original paper by Höhna et al. (2021) doesn't envision a stone-specific pre-burnin stage at all: it only proposes a _process_-specific pre-burnin, which combines our actual stages 1 and 2.) It's the global pre-burnin – i.e., converging onto the posterior – that represents the really hard part where we can't easily cut corners, and which it would therefore be useful to checkpoint. I.e., I agree with your point that it would be great to

> allow the global preburnin to execute as a separate job from individual stones.

There are good reasons (I think) why we don't bother checkpointing the (pre-)burnin stage of a regular MCMC analysis, but those don't quite apply here.

> Presently, for an 8-chain analysis, I need to reserve a hard-to-come-by 128-core slot, and watch 120 of those cores sit idle during the global pre-burnin.

Here I'm a little confused – what do you mean by running an "8-chain" power posterior analysis?

@bredelings

Right – I hadn't realized that we only committed ourselves to supporting a subset of C++23 features with PR #927, although it was obviously for a good reason.

I dislike ranges-v3, though (and would like to [see it removed](https://github.com/revbayes/revbayes/issues/936)), so I decided I didn't really need that one-liner _that_ badly. :stuck_out_tongue:

ms609:
@davidcerny By 8-chain I mean that MCMCMC uses a cold chain + 7 heated chains.  Perhaps I'm getting confused and Power posterior doesn't use heated chains?

davidcerny:
That's correct; it doesn't. It doesn't even support multiple runs.

ms609:
That's interesting; does that represent an optimistic view of how likely MCMC is to converge without tempering?  Insofar as MCMCMC tends to accelerate convergence relative to MCMC, this makes me nervous about whether power posterior samplers are really likely to be "burnt in"... (This is probably a tangent from the main PR - perhaps better to discuss elsewhere?)

davidcerny:
Hmm, I actually think it's still kind of related. If we were to implement checkpointing for the `.burnin()` method of `powerPosterior` analyses (i.e., the method responsible for the global pre-burnin), you could in theory take checkpoint files from any regular analysis you like (including an MCMC analysis with multiple runs, or an MC^3 analysis with multiple heated chains) and use them to initialize power posterior sampling from a plausible starting point. Höhna et al. (2021) already envisioned this possibility:

> We performed an additional performance analysis where we omitted the pre-burnin phase (running the MCMC sampler on the posterior distribution, see Fig. 2) but kept the burnin phase for each power posterior distribution. This scenario could be realistic when one has already performed a full posterior probability estimation and only wants to compute the marginal likelihoods for model selection. In this case, the samples from the posterior distribution can be used to specify starting values of the power posterior analysis.

Unlike checkpointing during stages 3 and 4, for which you need _more_ information than a standard MCMC / MC^3 analysis would give you (you need all of that plus the power to which you are raising the likelihood), checkpointing during stage 1 would actually require _less_ information: there are no monitors to restart, and we don't really care about the generation counter (stored in the `*_mcmc.*` checkpoint file) either. All we would need are the parameter values (stored in the base checkpoint file) and the proposal data (stored in the `*_moves.*` file). Again, you could just grab these from any analysis you like, as long as you remembered to run it with checkpointing turned on.

I'm now thinking about what the best interface would be for this. I could overload the `.initializeFromCheckpoint()` method even further, and make it resume the global pre-burnin phase if no `stones` argument is specified. This might require some checks, or at least emitting a warning, to make sure that this is in fact what the user intended to do. Or I could split it into two methods, `initializeBurninFromCheckpoint()` and `initializeRunFromCheckpoint()`.

Any thoughts?

ms609:
I often want to run MCMC and PP on the same dataset-model combination.

Starting PP from MCMC output seems like a big efficiency win.  Even without a checkpoint, simply seeding the PP run with the parameter values obtained during MCMC seems like it could be a big timesaver.

davidcerny:
This is ready for another round of reviews, I think. Some comments:

- I addressed the previous reviews by @bjoelle and @ms609, and changed the cover letter accordingly.

    - We now print a table showing the assignment of individual "stones" (power posteriors) to parallel workers (processes / groups of processes), and the progress monitor has been changed to use the word "step" with the same meaning it has in the table. Therefore, figuring out which stones need to be resurrected should be a simple matter of examining the standard output.
    - We can now checkpoint and resume not only stages 3 and 4 (stone-specific burnin + stone-specific sampling phase), but also stage 1 (global pre-burnin). In fact, we can use checkpoint files from an arbitrary analysis (MCMC or MC^3) to initialize this stage of the `powerPosterior` sampler, as long as these analyses used the same model and moves. See the PR cover letter for details.
    
- As I already pointed out above, burnin checkpointing presents an interesting challenge: while the checkpointing of individual power posteriors involves all the work performed by the usual MCMC checkpointing mechanism, and then some, in this case, we do not want to write out and parse the `*_mcmc` file storing the generation counter. Writing out this file would be harmless enough, but MCMC's `initializeSamplerFromCheckpoint()` function uses it to reset existing monitors, or to create new ones if it doesn't find them – something we don't want. To address this problem, I refactored our existing checkpointing implementation. Instead of keeping `initializeSamplerFromCheckpoint()` and `checkpoint()` as pure virtual functions in the base class (`MonteCarloSampler`) and only implementing them in the derived classes (`Mcmc`, `Mcmcmc`), I provided bare-bones implementations in the base class, and delegated the extra work (such as writing out and parsing `*_mcmc`) to the derived classes. But the bare-bones implementations are still accessible if you need them – and for `powerPosterior.burnin()`, they are exactly what we need.

- I apologize if the diff is somewhat hard to read. `PowerPosteriorAnalysis.cpp` alone has ballooned from ~450 lines to ~1,100 lines, with most existing functions refactored and many new ones added.

- I tested the checkpointing functionality in the following settings:
    
    - `.burnin()` and `.run()` on the single-thread version
    - `.burnin()` and `.run()` on the MPI version (10 processes)
    - `runOneStone()` on the single-thread version
    - `runOneStone()` on the MPI version (10 processes)
    
    Everything seems to be working as it should, but more stress testing is certainly welcome.

bjoelle (commented):
looks fine to me but I didn't have that many comments last time so I will let the other people check that it fits with what they wanted
thanks for adding the documentation and tests !

ms609:
Sounds promising.  I'll try to find time to test this out in the wild over the next week.

davidcerny:
@ms609 Thank you! :pray: I'm very curious as to how it's going to perform in practice.

ms609:
I've queued up some big jobs on the remote and will let you know how I get on once they execute, terminate, and resume.

ms609:
I've reaching the stage of resuming runs from checkpoints - trying to juggle this between other tasks so not had time to explore in depth.

At least [one run](https://github.com/neo-trans/07206_hg_b_kv) I've resumed from a checkpoint seems to be printing `*` to stdout rather a lot.  My slurm job's `.out` file reached 390GB – too big for me to open.  Claude tells me:

> The entire file is pure * characters — no text at all. This is RevBayes printing its mnScreen separator line at every MCMC step when running under MPI with powerPosterior, because the separator (\r-overwritten progress) gets written as a new line to a file instead of overwriting in-place. 128 processes × every generation = millions of * lines.
>
> Clear diagnosis: every model script adds mnScreen, and marginal.Rev passes that monitor vector straight to powerPosterior. With 128 MPI processes all printing to the same stdout file, you get millions of * separator lines at every MCMC step — printgen=300 is apparently ignored in this context.
>
> The fix: in marginal.Rev, clear monitors after sourcing the model script and before powerPosterior. The .pp output file is written by powerPosterior directly, so monitors are diagnostic only.

This was a job that had behaved "normally" on a previous version of RevBayes (just too slowly to complete in the 3d job limit).

After applying this workaround I restarted the run and there seems to have been an initial barrage of content - once I'd noticed that the job had begun, it'd been running a couple of minutes, and the `.out` file was already too large for `nano` to open.  But I deleted the `.out` file and no more content has been added to it, and the job is continuing to run.

davidcerny:
Oh no, that's not good 😬 

Thank you for sharing the repo with me; that should be very useful for diagnosing the source of the issue. Could you please also share the Rev script you used to restart the analysis?

For what it's worth, I'm pretty sure Claude is hallucinating here. The asterisks almost certainly come from the burnin progress monitor, which is completely independent of `mnScreen()` (it's hardcoded into the sampler itself, rather than being part of the monitor vector that gets passed to the sampler as an argument).

ms609:
I launched with [marginal.Rev](https://github.com/neo-trans/07206_hg_b_kv/blob/main/marginal.Rev).

Hallucination is likely – I need to get round to building a proper harness to brief Claude for areas where the RevBayes documentation is thin; it's not been doing a great job with some of the other errors I've been trying to untangle today.

ms609:
Am I right to think this is a breaking change for `mcmcmc` checkpointing?

My mcmcmc (**not** marginal / power posterior) resumption scripts (of the form of [mcmcmc.Rev](https://github.com/neo-trans/07206_hg_b_kv/blob/main/mcmcmc.Rev)) have started failing with "<script>_run_1.ckp not found"; my "should I be resuming from a checkpoint" guard used to check `if (fileExists(scriptBase + "_run_1_chain_0.ckp")) mymc3.initializeFromCheckpoint(...`; for runs that were started with the `development` branch, these `_run_1_chain_0.ckp` files exist in the absence of a master `_run_1.ckp`.

From a usability/readability and future-proofing perspective, I wonder whether there's a case for including a native function that might replace my `fileExists(scriptBase + "_run_1_chain_0.ckp")` check, which would also defend against future changes to the checkpointing API.

davidcerny:
> I launched with [marginal.Rev](https://github.com/neo-trans/07206_hg_b_kv/blob/main/marginal.Rev).

Oh wow! That's a very smart set-up to automatically identify which stones have to be resumed. Wish I had thought of that, frankly!

That said, I can't reproduce the problem with `*` getting printed to the screen billions of times. The output I'm getting looks very ordinary, and just the way it's supposed to:

Screenshot 2026-05-26 at 15 32 25
</br></br>

I tried asking an LLM (Composer 2.5) to read through `PowerPosteriorAnalysis.cpp` and identify the conditions under which something like that might happen, but it failed – in the typically unhelpful manner of questioning the accuracy of your report.

Do you happen to know when exactly the asterisks started to proliferate? Was it when you were (1) restarting the global pre-burnin, or (2) restarting existing stones?

> Hallucination is likely – I need to get round to building a proper harness to brief Claude for areas where the RevBayes documentation is thin; it's not been doing a great job with some of the other errors I've been trying to untangle today.

I've been using Claude from within the Cursor IDE, and that combo works reasonably well for me. There are certainly tasks that are above its pay grade, but overall it's doing a pretty good job.

> Am I right to think this is a breaking change for mcmcmc checkpointing?

That definitely wasn't my intention, but you're right! Apparently, development only produces chain-level checkpoint files, not run-level checkpoint files (which makes sense). However, this branch has accidentally introduced extra run-level files. The existing checkpoint files are completely untouched by the change, which is why the [integration test](https://github.com/revbayes/revbayes/blob/development/tests/test_MC3/scripts/Test_heat_tuning_and_checkpointing.Rev) for MC3 checkpointing is still passing, without the need for any changes to the script itself or to its expected output. Let me try and fix this.

By the way – there is no need to keep this branch up to date with development by merging the latter into it. Once it's ready to be merged, I can just rebase it on development and force-push.

davidcerny:
The accidental breaking change for MC3 checkpointing should now be fixed.

ms609:
> I've been using Claude from within the Cursor IDE, and that combo works reasonably well for me. There are certainly tasks that are above its pay grade, but overall it's doing a pretty good job.

Yes, once it has the full codebase to orient against, I suspect it can find its way around – but when running in a repo that's just launching / debugging analyses, my guess is that it's falling back on its memories of whatever documentation/code it was trained on – once I point it to the RevBayes source it is generally less wayward.  I ought to formalize how it refers to the source, and help it to navigate the source without burning tokens.

> Do you happen to know when exactly the asterisks started to proliferate? Was it when you were (1) restarting the global pre-burnin, or (2) restarting existing stones?

The stone files seem to be a mixture of "column headings only" and "just a few (~4) rows exist", which seems to have been the state before I got "starred out", so I suspect but cannot guarantee case (2).

davidcerny:
Hmm. I'll keep investigating, but I can't reproduce the problem using the current state of the repo, no matter how many cores I use.

ms609:
I'll see whether I can roll back my changes to reproduce the exact original failure state.

ms609:
I got Claude to roll the repo state back and create a "clean" repo that reproduces the pathology on our HPC: https://github.com/neo-trans/powerposterior-stall-repro

I can't vouch for the quality of the AI commentary, but hopefully you are able to reproduce the failure state from that repo

davidcerny:
Thank you for putting this together, @ms609! Claude really threw the kitchen sink at it, but the `evidence` folder proved to be useful. It shows very clearly that the problem actually took place when you tried to extend the global pre-burnin phase by 0 generations. Clearly, you ran into some sort of division-by-zero problem there, and it's one that only occurs when you compile RevBayes using GCC – Clang is able to optimize it away, which is why the screenshot I posted above shows the correct, expected outcome (no asterisks, as opposed to infinite asterisks). In this way, it's reminiscent of issue #794, which I also could not reproduce using Clang.

After some more digging, it turned out the issue was actually completely unrelated to the changes made in this PR – it came from [`ProgressBar.cpp`](https://github.com/revbayes/revbayes/blob/development/src/core/io/ProgressBar.cpp), and if we hadn't noticed it earlier, it was only because no one had thought of running burnin for 0 generations before. That said, I have now fixed it.

ms609:
Super, thank you (and well done!) for tracking this down!

davidcerny:
Of course! Let me know what you think the next steps should be.

ms609:
I've had a little audit of progress so far.  The Stage 3/4 resume seems to be validated.  But my extremely long-running jobs don't have time to complete the global pre-burnin before hitting their wall-clock limits.

I got Claude to assess the state of play, with reference to the revbayes source.  Its diagnosis - which it has tested with a toy dataset - is that stage 1 doesn't record how far it's got along the global preburnin.  As such, `burnin(g)` runs `g` *fresh* generations from the restored state.


Here's a distillation of Claude's summary:


The stage-1 recipe above

```r
pow_p.burnin(generations=10000, …, checkpointFile="burnin.ckp", checkpointInterval=200)
pow_p.initializeFromCheckpoint("burnin.ckp")
pow_p.burnin(generations=10000, …)        # runs 10000 MORE, not "finish to 10000"
```

cannot complete a fixed-length global pre-burnin across a chain of sub-burnin-length jobs: every resume restarts a full `g`, so a burnin that does not fit in one allocation never terminates and the analysis never reaches the stones.

Contrast stages 3–4, which *do* resume-to-target: a restarted stone "completes the remaining N generations … as though no interruption had taken place" because `planned_burnin` (`*_stone_info.ckp`) is compared against the `*_mcmc.ckp` counter.

- **Source** (`PowerPosteriorAnalysis::burnin`): `for (k=1; k<=generations; k++)` — exactly `generations` every call; resume loads state only (`baseInitializeSamplerFromCheckpoint`)

## RevBayes-side workaround

Wrap stage-1 `burnin()` in a loop that persists the running total to disk and stops at the target:

```r
cumBurnin = 0
if (fileExists(ckpCounter)) { cumBurnin = readDelimitedDataFile(ckpCounter, FALSE, "")[1][1] }
while (cumBurnin < burninGen) {
  if (fileExists(ckpBurnin)) { power_p.initializeFromCheckpoint(ckpBurnin) }   # reload each pass:
  thisBurnin = increment                                                       # burnin() clears its
  if (burninGen - cumBurnin < increment) { thisBurnin = burninGen - cumBurnin } # resume flag on return
  power_p.burnin(generations = thisBurnin, tuningInterval = tuneInt,
                 checkpointFile = ckpBurnin, checkpointInterval = ckpInterval)
  cumBurnin = cumBurnin + thisBurnin
  write(cumBurnin, file = ckpCounter, append = FALSE)   # AFTER return ⇒ conservative lower bound
}
write("done", file = ckpBurninDone)
```

Tested with toy dataset; now deployed with three real ones.

## Suggested C++ solution

Persist the completed global-pre-burnin generation count behind the existing single-writer gate (re-enable the burnin `*_mcmc.ckp` counter, or add a `*_burnin_info.ckp` analogous to `*_stone_info.ckp`); load it in `initializeFromCheckpoint`; and have `burnin(generations)` treat `generations` as a **target total** (`for k = completed+1 … generations`) — i.e. resume-to-target, exactly as stones already do. This removes the stage-1/stage-3–4 asymmetry.

davidcerny:
> stage 1 doesn't record how far it's got along the global preburnin. As such, `burnin(g)` runs `g` _fresh_ generations from the restored state.

My first instinct was to say that this is a feature, not a bug, because this is how re-initialization from a checkpoint also works for `mcmc.run()` and `mcmcmc.run()`. It's desirable, because one of the functions of the checkpointing system is not just to resurrect an analysis that was interrupted before reaching its target length, but also to extend an analysis whose originally specified target length proved to be insufficient.

However, I agree that in the case of the global pre-burnin, things are a bit more complicated:

- When do we want to extend an analysis? Typically, when an inspection of the parameter trace makes it clear that the analysis has not yet reached stationarity, or collected enough samples to obtain sufficiently precise parameter estimates. But during the pre-burnin stage, we are not collecting samples or writing out a trace, so there is virtually no way to conclude we haven't run it for long enough, even if that was in fact the case.
- With `mcmc()` or `mcmcmc()`, the user knows how many generations were run prior to the checkpoint (from multiple sources, actually – the screen monitor, parameter traces, or the checkpoint files themselves), and can use this info to estimate how many more generations are needed. This is not the case with the global pre-burnin, where the only hint as to the number of generations completed so far is the progress bar (which does not always reliably appear due to #874).

That said, what would it take to implement a system for "resuming-to-target"?

- We would indeed have to bring back the `*_mcmc.*` counter in some shape or form, but instead of utilizing the system that's already in place in `Mcmc.cpp`, we would have to re-implement our own. That's because when the `Mcmc` sampler encounters a `*_mcmc.*` checkpoint file but no existing parameter trace, it starts to write one from scratch – which we don't want.
- We would also need an extra checkpoint file to record the `planned_preburnin` variable. But since normal `mcmc()` or `mcmcmc()` do not produce such a file, we would lose the ability to use a pre-existing analysis to initialize the global pre-burnin, unless we're willing to write such a file manually.

Note also that Claude's summary is not 100% accurate:

> Contrast stages 3–4, which _do_ resume-to-target: a restarted stone "completes the remaining N generations … as though no interruption had taken place" because `planned_burnin` (`*_stone_info.ckp`) is compared against the `*_mcmc.ckp` counter.

Stage 3 (stone-specific burnin) does resume-to-target, but stage 4 (stone-specific sampling phase) does not. For the latter, this:

```r
pow_p.initializeFromCheckpoint( "analysis.ckp", 1:32 )
pow_p.run(generations=g)
```

will also run an extra `g` generations regardless of the number of generations completed so far, and I think that makes perfect sense for the reasons outlined above. So the "stage-1/stage-3–4 asymmetry" will _not_ be eliminated in any case: the only question is whether stage 1 should behave more like stage 3, or more like stage 4.

Overall, I'm leaning slightly toward retaining the current behavior, but it's not a particularly strong preference.

P.S. Your (or Claude's) "RevBayes-side workaround" looks interesting, but it relies on an `increment` variable that's not properly initialized in the code chunk you showed.

ms609:
Another answer to "When do we want to extend an analysis" is "When a job was terminated before it converged".

In the present case, I have power posterior jobs that take >3 days to complete the global pre-burnin, and my HPC has a 3-day time limit for jobs.  The "resume" functionality would therefore make jobs possible that are currently impossible. As an added bonus, able to schedule day-long or 12h-long jobs and know they will resume from checkpoint without excessive data loss would allow tasks to be queued to make use of idle processor time, rather than waiting until a core is known to be available for a 3-day window. It's this feature that got me most excited about the new functionality!

davidcerny:
> The "resume" functionality would therefore make jobs possible that are currently impossible. As an added bonus, able to schedule day-long or 12h-long jobs and know they will resume from checkpoint without excessive data loss would allow tasks to be queued to make use of idle processor time, rather than waiting until a core is known to be available for a 3-day window. It's this feature that got me most excited about the new functionality!

You can do all of that already – the resume functionality for the global pre-burnin _is_ part of this branch. What you can't do is write a single clever script, and then "run and forget" – you need to have some idea of how long your pre-burnin will actually take to finish on a given machine. This might involve a short preliminary MCMC run to give you an ETA.

davidcerny:
That said, one obvious disadvantage of the current system is that if an analysis is interrupted during stage 1, resumed from a checkpoint, and run for an additional `g` generations, one might never actually know how many generations in total it spent in the global pre-burnin – a fairly basic piece of information that one would want to report!

So maybe "resume-to-target" _is_ a better solution after all.

ms609:
I'm not sure I'm quite following – I think we may be talking past each other here.

My situation is: I want to run a X000-generation pre-burnin.  I know that running X000 generations takes >7d, the longest run time available on my HPC.  I wish to run the analysis in 3d chunks (the longest slots available on the "shared" queue). Hence I know that I will need to resume the pre-burnin from a checkpoint.

I think you are telling me that the current system doesn't know how many generations have been spent in the global pre-burnin.  If so, then any user in my situation will need to keep track of how many generations have been completed, in order to determine how many additional burnin generations to request.  I would find it helpful if RevBayes could provide a robust mechanism to determine how many generations have already been conducted.  And from this it feels like it should be straightforward to extend the functionality to allowing the user to specify a target burnin length, and then to compute how many generations are required in order to meet that target and proceed to the next step of the analysis.

ms609:
A positive update: I've completed a long-running job that *did* manage to complete its preburnin within the 3d walltime.  This has made a job that was previously intractable succeed: so a ringing endorsement for the post-global-preburnin review machinery, thank you! 🎊

ms609:
## Follow-up: "resume can't complete" gap in the stone phase (stages 3–4)

I encountered an issue analogous to the stage-1 (global pre-burnin) finding above in the stone sampling phase (built from 8f71e5f05, running on a 128-process MPI in repo https://github.com/neo-trans/07204_hg_kv/tree/d4bddefbd380d6a43fd07c0cadc5ddd5b8ed53b7)

Summary: a `powerPosterior` run interrupted by a wall-time limit before every stone has been started never completes; but instead produces a rogue marginal likelihood from an incomplete power schedule.

---
Edited Claude diagnosis:

### Symptom
After a few resume installments: the `.pp` covers fewer than `cats` distinct powers, per-stone sample counts are ragged, stepping-stone and path-sampling estimates diverge by thousands of log units, and analytic SEs are `nan` (a stone left with a single sample).

`summarizeStones()` truncates the output file and appends whatever per-stone files exist, warning to stderr for the missing ones without throwing an error.  I wouldn't have been aware of the issue if I hadn't noticed the pathological values.

### Mechanism
`runStone()` works as expected: it checks `stone_resumes_from_checkpoint` and either resumes from the checkpoint or does a fresh pre-burnin.

The gap is in `runAll()` and the flat-resume contract:

- **Fresh run:** `runAll()` iterates the full `0 .. powers.size()` block and calls `runStone()` for every stone.
- **Flat resume** (`initializeFromCheckpoint(ckp, vector<size_t>)`): `runAll()` builds `resurrection_indices` *only* from `ckp_stone_file` and iterates that set. Never-started stones aren't in `ckp_stone_file`, so they're skipped on every resume. Passing a non-checkpointed index into the flat overload throws `"Stone <i> was not initialized from checkpoint."`

So there's no way to ask a flat-resumed analysis to also sample the stones it never reached; a stone phase that doesn't fit one allocation never completes.

### A concrete bug in the nested overload (Claude's workaround)

The nested `initializeFromCheckpoint(ckp, vector<vector<size_t>>)` *can* express mixed resume/fresh, so Claude drove completion from the Rev script with it. But it requires the caller to (a) know `worker_count` (`num_processes / processors_per_likelihood`), which isn't exposed to Rev, and (b) supply **exactly `worker_count` sequences** — fewer triggers `std::bad_array_new_length`, because `runStone()` and `printStoneAssignmentToWorkers()` index `resume_stone_sequences[i]` for all `i` in `[0, worker_count)` unconditionally (`PowerPosteriorAnalysis.cpp` ~L746, ~L515), whereas `getStepNumber()` correctly bounds by `.size()`. Claude worked around
it by padding to exactly `worker_count` lanes (promoting already-complete stones as extra lane heads). With that workaround a 128-stone / 32-process analysis now resumes and completes correctly across wall-time limits (a chopped run that had sampled 36/128 stones resumed as "27 incomplete + 65 missing across 32 lanes; 36 complete retained" and proceeded to fill every stone) — but the padding is a
sharp edge that shouldn't be necessary.

### Potential solutions

1. **Let flat-resume fresh-start missing stones** — in `runAll()`'s resume branch, iterate the full stone set (as the fresh branch does) and drop the `runStone()` throw for non-checkpointed stones in flat mode; `runStone()` already resumes-or-starts per stone, and the existing block distribution keeps it MPI-parallel with no caller-side worker arithmetic.

Or 

2. **Add a resume-to-target / `complete()` helper** that does the above and, like the stones' `planned_burnin` accounting, treats the requested generation count as a target total so repeated resumes converge to a uniform sample count instead of over-extending early stones.

And independently: guard the nested path against `resume_stone_sequences.size() < worker_count` (bound those loops by `.size()` and treat absent lanes as empty) so it degrades gracefully instead of an OOB.

ms609:
Using the workarounds, I'm now seeing gnarly long-running runs consistently proceed to completion ☺

davidcerny:
Hi @ms609, that's great to hear, thank you for the update!

I was actually just about to (finally) reply to your previous messages, even though I probably won't get around to making further changes to the code until next week. Apologies for dropping the ball on this for so long!

> I think you are telling me that the current system doesn't know how many generations have been spent in the global pre-burnin. If so, then any user in my situation will need to keep track of how many generations have been completed, in order to determine how many additional burnin generations to request. I would find it helpful if RevBayes could provide a robust mechanism to determine how many generations have already been conducted. And from this it feels like it should be straightforward to extend the functionality to allowing the user to specify a target burnin length, and then to compute how many generations are required in order to meet that target and proceed to the next step of the analysis.

Yes, I think you're right. Not being able to tell how many generations the global pre-burnin has been run for is a pretty serious problem, and resume-to-target is a reasonable way of getting around it.

The checkpoint interface is complicated enough as-is, but I wonder if it makes sense to make the `generations` argument of the `.burnin()` method optional. Then we'd have the following cases:

- Pre-resumption, `generations` not specified: error
- Pre-resumption, `generations` specified: normal behavior
- Post-resumption, `generations` not specified: run to target
- Post-resumption, `generations` specified: run as many generations as requested, even if the number exceeds (or fails to reach) the target stored in the checkpoint file

This would allow the user to extend the global pre-burnin if needed.

One last issue with this system is that it requires at least one extra checkpoint file (presumably storing both the generation counter and the `planned_preburnin` variable). This would make it harder to start a `powerPosterior()` analysis from a sample collected by a regular `mcmc()` analysis. Let me think a little more about what can be done here.

As for your second report, I'm not 100% sure it documents a real bug. From the repo you linked, it looks like you built an extremely complicated and fragile system in your `marginal.Rev` script that's supposed to handle the resumption of power posterior analyses automatically, without any user intervention. To be honest, I think the attempt is somewhat misguided. Unlike `mcmc()`, where it's easy to do something like:

```
if ( fileExists("analysis.ckp") ) {
    mymcmc.initializeFromCheckpoint("analysis.ckp")
} else {
    mymcmc.burnin(generations=50000, tuningInterval=100)
}
```

power posterior analyses, with their multiple stages and automated parallelization scheme, just don't lend themselves easily to this kind of solution; some degree of user intervention is probably necessary.

> So there's no way to ask a flat-resumed analysis to also sample the stones it never reached

Well, yes: that's not what flat resume (i.e., what happens when a simple vector is passed to the `stones` argument of `.initializeFromCheckpoint()`) is for. If some stones have never even entered the burnin phase (phase 3), you need to use nested resume (performed by passing a vector of vectors to the `stones` argument) instead. Your AI goes on to acknowledge this, so I'm not sure why it's complaining about flat resume in the first place.

> The nested `initializeFromCheckpoint(ckp, vector<vector<size_t>>)` _can_ express mixed resume/fresh, so Claude drove completion from the Rev script with it. But it requires the caller to (a) know `worker_count` (`num_processes / processors_per_likelihood`), which isn't exposed to Rev

Indeed, and this is a good example of why full automation is going to be difficult for power posterior checkpointing. You _do_ need to know the number of cores you are using for your analysis, and the AI is correct to point out that this piece of information is external to your Rev script. I don't really see any way around this.

> [Nested resume requires the user to] (b) supply *exactly `worker_count` sequences* — fewer triggers `std::bad_array_new_length`

Hmm, okay, I don't remember ever running into this, but that does sound like a genuine bug. The intended behavior is to handle the scenario where there is fewer than `worker_count` stone sequences as follows:

https://github.com/revbayes/revbayes/blob/c036a45d929ee412c0317afe70e7437366c0aa44/src/core/analysis/PowerPosteriorAnalysis.cpp#L364

davidcerny:
## Resume-to-target for global pre-burnin

I've updated the checkpointing interface for the global pre-burnin in line with my comment above:

> - Pre-resumption, `generations` not specified: error
> - Pre-resumption, `generations` specified: normal behavior
> - Post-resumption, `generations` not specified: run to target
> - Post-resumption, `generations` specified: run as many generations as requested, even if the number exceeds (or fails to reach) the target stored in the checkpoint file

The documentation, tests, and PR description have been updated accordingly.

> One last issue with this system is that it requires at least one extra checkpoint file (presumably storing both the generation counter and the `planned_preburnin` variable). This would make it harder to start a `powerPosterior()` analysis from a sample collected by a regular `mcmc()` analysis. Let me think a little more about what can be done here.

This indeed proved to be doable; it just required thinking a little more about when the file in question (which I ended up calling `*_burnin_info.ckp`) is actually required upon resumption. Consultation with Grok 4.5 proved enlightening here.

1. `generations` not specified, file available: normal "resume-to-target" behavior
2. `generations` not specified, file not available: throw an error:

    https://github.com/revbayes/revbayes/blob/8800ac0bd1597e02954acb574d2959a2f6d745d9/src/core/analysis/PowerPosteriorAnalysis.cpp#L186-L189
    
3. `generations` specified, file available: the recorded target is first disregarded and then rewritten in accordance with what the user specified

4. `generations` specified, file not available: global pre-burnin is initialized from the `*.ckp` and `*_moves.ckp` files (potentially produced by a regular `mcmc()` or `mcmcmc()` analysis) and run for the specified number of generations

--------------------------------------------------

## Resume can't complete in the stone phase

It seems to me that this problem, as [originally reported](https://github.com/revbayes/revbayes/pull/991#issuecomment-4840709387), can be broken down into three parts:

1. "There's no way to ask a flat-resumed analysis to also sample the stones it never reached". This is fully intentional, because fresh-starting missing stones is not what flat resume is for – it's a job for nested resume. I disagree with the claim that using nested resume here represents a "workaround". It's simply the right solution for this situation.

2. The user has to "supply exactly `worker_count` sequences". This was indeed a bug! It's now been fixed – pretty much in the way the original comment suggested.

3. The user has to "know `worker_count` (`num_processes / processors_per_likelihood`), which isn't exposed to Rev". I agree it would be nice if there was a way to find out from within a Rev script how many parallel workers the script is being executed on. I might open a feature request for this. However, since issue (2) has been solved, it's currently beyond the scope of this PR.

--------------------------------------------------

@ms609 Let me know what you think of these latest changes when you get a chance. I hope we're close to the finish line now!

ms609:
I'm afraid I'm struggling to get my head back into the details, but from my perspective: I've able to use the code in this branch, along with Claude-generated scaffolds, to resume and complete multiple long-running analyses across a wide range of interruption conditions.  From my perspective, I'd be happy to see this merged (even if I might be back with suggestions for improvements in future!).

Thanks again for putting this together – it's been great to finally be able to back up some significant analyses with Bayes Factor information!

ms609 (approved):

PR #994, opened 2026-04-06
implementing usual datatypes for mnStochasticCharacterMap

mikeryanmay:
This is just a small mod to mnStochasticCharacterMap to enable the usual datatypes for CTMC models, mirroring mnJointConditionalAncestralState. 

(accidentally made this pull request a few minutes ago against master! trying again against development. sorry to anyone that got some confusing alerts!)


src/revlanguage/monitors/Mntr_StochasticCharacterMapping.cpp
```diff
@@ -125,8 +128,43 @@ void Mntr_StochasticCharacterMapping::constructInternalObject( void )
...
-        std::string data_type = ctmc_sn->getValue().getDataType();
-        if (data_type == "Standard") {
+        if (data_type == "AA" || data_type == "Protein") {
+            RevBayesCore::StochasticCharacterMappingMonitor<RevBayesCore::AminoAcidState>* m;
+            m = new RevBayesCore::StochasticCharacterMappingMonitor<RevBayesCore::AminoAcidState>( ctmc_sn, (std::uint64_t)print_gen, file_name, is, sd, sep, idx - 1 );
+            m->setAppend( app );
+            m->setPrintVersion( wv );
+
+            delete value;
+            value = m;
```
davidcerny: These steps are the same in every branch, so they should ideally be factored out of the conditional... although I guess that would require writing some sort of lambda, since `m` doesn't have a unique type at compile time.

src/revlanguage/monitors/Mntr_StochasticCharacterMapping.cpp
```diff
@@ -143,6 +181,16 @@ void Mntr_StochasticCharacterMapping::constructInternalObject( void )
...
+        } else if (data_type == "Binary" || data_type == "Restriction") {
+            RevBayesCore::StochasticCharacterMappingMonitor<RevBayesCore::BinaryState>* m;
+            m = new RevBayesCore::StochasticCharacterMappingMonitor<RevBayesCore::BinaryState>( ctmc_sn, (std::uint64_t)print_gen, file_name, is, sd, sep, idx - 1 );
+            m->setAppend( app );
+            m->setPrintVersion( wv );
+
+            delete value;
+            value = m;
+        } else {
+            throw RbException( "Incorrect character type specified. Valid options are: AA, DNA, NaturalNumbers, PoMo, Protein, RNA, Standard, Binary/Restriction" );
```
davidcerny: We seem to allow two additional datatypes in `dnPhyloCTMC`: `Codon` and `Doublet`. I have no idea if anyone would ever want to put these into a stochastic map monitor, though.

davidcerny (dismissed):

davidcerny:
> accidentally made this pull request a few minutes ago against master! trying again against development. sorry to anyone that got some confusing alerts!

No worries! For what it's worth, you can [change the base branch of a PR even after you have opened it](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-base-branch-of-a-pull-request). I think the bigger problem was that your previous branch didn't contain the commit you wanted to include 🙂

bredelings:
I'm filling in some of the bits to make this a "cold fix" rather than a "hot fix".

It seems like the GLHBDSP and the CDBDP only implement this for natural numbers.  Do the other character types not make sense, or are they just not implemented?

src/revlanguage/monitors/Mntr_StochasticCharacterMapping.cpp
bredelings: I did this.

src/revlanguage/monitors/Mntr_StochasticCharacterMapping.cpp
bredelings: Fixed.

bredelings:
I addressed both comments by:
* using template lambdas
* adding a template function `apply_to_character_type(lambda_operation, character_type)` that looks at the `character_type` string and applies the operation to the right C++ type for the character.
* using these in a few other files -- notably `Mntr_AncestralState.cpp`, which didn't work for amino acids either.

Hopefully this decreases code duplication, and also ensures that we don't skip character types like amino acids and codons.

davidcerny (approved):
Looks good!

PR #975, opened 2026-03-21
Fix state lables in NexusWriter

basantakhakurel:
Fixes #969 

### Problem
When concatenating character matrices with different state spaces, the NexusWriter only used the state labels from the first matrix. 
This meant that the `symbols=` field in the NEXUS file was incomplete (missing the complete state labels) causing the written file to be unreadable. 

### Changes
**`src/core/io/NexusWriter.cpp`**
- Instead of reading state labels from only the first character, now iterates over all included characters and collects a union of all state symbols, ensuring completeness.

**`src/revlanguage/datatypes/phylogenetics/characterdata/RlAbstractHomologousDiscreteCharacterData.cpp`**
- Previously only returned the states from the first matrix. Now builds a union similar to NexusWriter fix to return the complete state labels.

Added a new test script (`test_writeNexus`) that checks if the nexus files are being written properly. 

(Also added .zed in `.gitignore`)

PR #995, opened 2026-04-11
writeNexus expected output

basantakhakurel:
Thanks to @davidcerny for pointing this out.
I forgot to include the expected output when I added a test for the `writeNexus()` function in PR #975.
This PR adds the expected output.


davidcerny (approved):
Thanks so much for addressing this so promptly, @basantakhakurel 🙂 

1. You could potentially add the Nexus files themselves to the expected output as well, which would require prefixing their current file paths with `output/`. But I guess it's not strictly necessary, since we just want to know whether the files satisfy certain properties, rather than what they look like.
2. You've already merged development into this branch once, and it looks like you might have to do it once more. That's no biggie, of course, but it might be more elegant to use interactive rebase and graft your commits onto the current tip of development...

PR #997, opened 2026-04-13
Remove Dollo model

bredelings:
The Dollo model seems somewhat broken.  It also appears to have no users, and no tests.  Should we remove it?  It will survive in the git history should anyone want to resuscitate it later on.


src/revlanguage/workspace/RbRegister_Dist.cpp
davidcerny: Also need to remove this...

davidcerny:
I would be in favor of this. Things that we know don't work the way they should, as well as "I might get back to this later"-style unfinished experiments, have no place on the master branch if we want RevBayes to be mature software that users can rely on.

That said, it might be useful to briefly describe _how exactly_ the distribution is broken.

bredelings:
That's kind of why this is a draft.  Figuring out exactly how it is broken would take more work :-P  I'm not sure how deeply it is broken.

davidcerny:
Oh, sorry, that was unfortunate wording on my part. :sweat_smile: I just meant specifying how the breakage manifests itself on the surface. Do we get a segfault? Wrong likelihoods? Or is it that we can't figure out what it does / should be doing?

(@basantakhakurel might have some insight into this, too...)

basantakhakurel:
For me, it is unclear how we set up the rate matrix and the CTMC when using the Dollo model. I tried multiple combinations to understand the mechanisms, but I often got segfaults. But yeah, I think it is best removed right now.

bredelings:
Ah, I see.  So I guess there were some SEGFAULTs.  I guess the question is whether we should just remove it, or merge some version of the `dollo` branch from @basantakhakurel, which fixes some SEGFAULTs and adds some tests that don't crash.

The code has some weird stuff (e.g. https://github.com/revbayes/revbayes/blob/9e1bfe8dd2edb8f5563d25a6afba3d9840cd17e6/src/core/distributions/phylogenetics/substitution/PhyloCTMCSiteHomogeneousDollo.cpp#L213), and might just not be reliable with no users and no maintainers.  So if @basantakhakurel thinks that in its current state it would be best to remove it, I would lean that way too.

FWIW, it also reduces the burden to updating other parts of the PhyloCTMC code.

basantakhakurel (approved):

Issue #686, opened 2025-03-07
[BUG] Broken virtual overrides in the birth-death process


bredelings:
```
In file included from ../../git/src/core/distributions/phylogenetics/tree/birthdeath/FossilizedBirthDeathSpeciationProcess.cpp:10:
../../git/src/core/distributions/phylogenetics/tree/birthdeath/AbstractBirthDeathProcess.h:29:61: warning: ‘virtual double RevBayesCore::AbstractBirthDeathProcess::computeLnProbabilityDivergenceTimes() const’ was hidden [-Woverloaded-virtual=]
   29 |         double                                              computeLnProbabilityDivergenceTimes(void) const;                                        //!< Compute the log-transformed probability of divergence times.
      |                                                             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from ../../git/src/core/distributions/phylogenetics/tree/birthdeath/FossilizedBirthDeathSpeciationProcess.cpp:13:
../../git/src/core/distributions/phylogenetics/tree/birthdeath/FossilizedBirthDeathSpeciationProcess.h:59:57: note:   by ‘double RevBayesCore::FossilizedBirthDeathSpeciationProcess::computeLnProbabilityDivergenceTimes()’
   59 |         double                                          computeLnProbabilityDivergenceTimes(void);                            //!< Compute the log-transformed probability of the current value.
      |                                                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[3/6] Compiling C++ object librb-revlanguage.a.p/src_revlanguage_workspace_RbRegister_Dist.cpp.o
In file included from ../../git/src/core/distributions/phylogenetics/tree/birthdeath/BirthDeathProcess.h:7,
                 from ../../git/src/core/distributions/phylogenetics/tree/birthdeath/ConstantRateBirthDeathProcess.h:4,
                 from ../../git/src/revlanguage/distributions/phylogenetics/tree/Dist_bdp.h:4,
                 from ../../git/src/revlanguage/workspace/RbRegister_Dist.cpp:164:
../../git/src/core/distributions/phylogenetics/tree/birthdeath/AbstractBirthDeathProcess.h:29:61: warning: ‘virtual double RevBayesCore::AbstractBirthDeathProcess::computeLnProbabilityDivergenceTimes() const’ was hidden [-Woverloaded-virtual=]
[...]

Issue #499, opened 2024-07-12
[BUG] `dnUPP` causing MCMC to fail


davidcerny:
**Describe the bug**
Because of the "stochaministic nature" of `dnUPP()` (the uniform partitioning distribution), changes in parameter values are not correctly propagated through the model DAG. This may result in analyses that either run seemingly fine but produce invalid results, or crash when RevBayes is compiled with the `DEBUG_MCMC` flag. With a sufficiently complex model, the distribution can cause MCMC to fail for seemingly unrelated reasons even when this flag is not enabled; see #293 for an example.

**To reproduce**
Run [this script](https://github.com/user-attachments/files/16185736/dnUPP_test.txt). (One absolute path has to be changed; the `bears_cytb.nex` file is the one from our test suite and [dating tutorials](https://revbayes.github.io/tutorials/dating/global)). This is a heavily stripped down version of the script from #293. When I run it after enabling `DEBUG_MCMC`, I get:

```
   Error:	Issue in 'UPP-Allocation' on 'branch_rates[18]' after move because posterior of -8152.37 and -8149.6/-8149.6. The move was rejected.
   Error:	Problem processing line 123 in file "/Users/David/Downloads/dnUPP_test.Rev"
```

**Expected behavior**
Reversible-jump MCMC analyses should be able to employ `dnUPP` without crashing.

**Computer info**
macOS 14.5 (Sonoma), ARM64 architecture, RevBayes 1.2.5-preview, latest version of the `development` branch (c46173f).

**Additional context**
This bug may have affected several published papers: 

- Freund FD, Freyman WA, Rothfels CJ. 2018. Inferring the evolutionary reduction of corm lobation in _Isoëtes_ using Bayesian model-averaged ancestral state reconstruction. _Am. J. Bot._ 105(2): 275--286. doi:[10.1002/ajb2.1024](https://doi.org/10.1002/ajb2.1024)
- Paluh DJ, Stanley EL, Blackburn DC. 2020. Evolution of hyperossification expands skull diversity in frogs. _Proc. Natl Acad. Sci. USA_ 117(15): 8554--8562. doi:[10.1073/pnas.2000872117](https://doi.org/10.1073/pnas.2000872117)

Corresponding Rev scripts:

- https://github.com/wf8/isoetes/blob/master/upp_mcmc.Rev
- https://github.com/wf8/isoetes/blob/master/upp_sim.Rev
[...]

PR #998, opened 2026-04-14
Fix compiler warnings

davidcerny:
This PR should reduce the number of warnings we get when compiling RevBayes, which might in turn make it easier to locate genuine compile errors during developer experimentation. Most of the changes should hopefully be pretty self-explanatory, so just a few brief comments:

- Unused variables were usually just removed (e.g., there was a couple that seemed to have been left over from turning `while` loops into `for` loops), but in a few monitors, there was commented out control flow that relied on them. In those cases, I commented them out as well.

- A more interesting case (caught only by GCC 15) occurred in `UniformPartitioningDistribution` and `CTMCProcess`, where broken overrides of the `setValue()` function with incorrect type signatures resulted in dead code that was never instantiated. In `CTMCProcess`, the logic behind the original code was clear, so I decided to fix the function. However, in `UniformPartitioningDistribution`, I wasn't quite sure what a fix should look like, so I just commented out the broken override. Since analyses relying on `dnUPP` don't seem to work anyway (#407, #499), I think we should consider disabling the distribution, just like we disabled `dnFBDRP` and `dnFBDSP` in #449. (There are published papers that employed it in their analyses, so we probably shouldn't remove it altogether.)

The only compiler warnings that still pop up under default settings come from:

- Broken virtual overrides in the disabled FBD-range distributions (#686)
- Additional broken overrides of `computeDataAugmentedCladogeneticLnProbability()` that snuck into the code base with PR #904
- The libraries. TensorPhylo, NCL, and range-v3 (see #936) all seem to be responsible for a few. The TensorPhylo problem in particular should be easy to fix (by replacing a failing assert with an exception), but this is probably not the place to do it.


bredelings (approved):
Very nice!  There are some real bugs fixed here, in addition to reducing warnings.  I think `abs(x)` converts `x` to integer, for example, whereas `std::abs(x)` does not.

Note the lex.yy.cpp is a generated file, so when the file is regenerated, the warning may come back, unless we remember to put the [[maybe_unused]] in a second time.

PR #1002, opened 2026-04-15
Refactor scaling

bredelings:
This branch modifies the scaling method that RevBayes uses to prevent underflow.  It also
* avoids Inf/-Inf in the F81 rate matrix
* avoids Inf/nan in the BetaSimplex proposal
* makes the PartialLikelihood object non-copyable, to prevent accidental copies.
* allows using AVX -- previously we always used SSE even when AVX was available
* checks AVX availability by looking at the `__AVX__` macro
* _updates/fixes_ the hand-written AVX code  in PhyloCTMCSiteHomogeneousNucleotide
* _removes_ the hand-written SSE code in PhyloCTMCSiteHomogeneousNucleotide, as it is slower than the plain C code.
* aligns the partial likelihood buffers to 32-byte alignment, for a slight speed increase
* avoids SEGFAULTS with AVX when the partial likelihood buffers are not 32-byte aligned

## Avoiding log( ) operations
Previously, when we rescaled a node, we would
* compute the __maximum__ probability at each site across all mixture categories
* __divide__ by the maximum to bring the highest probability up to 1
* take the __log__ of the maximum and add it to to `log_scale[site]` 
* at the root, we __add__ a factor of `log_scale[site]` to the total log-probability

The `log( )` operation is very slow compared to multiplication and addition.  The division is a bit slow also.

The new method (from Alexis Stamatakis)
* checks if the __maximum__ at each site across all mixture  categories is below 2^-256
* if so, then we 
    * __multiply__ all the probabilities at each site by 2^256
    * __increment__ `scale[site]` to account for this modification
* at the root, we __add__ a factor of `scale[site] * log(2^-256)` to the total log-probability

## Benchmarks

Scaling has the biggest effect with DNA data or binary data.  Since we have only 4 letters, the amount of work done per site is small, so taking a log is relatively large compared to that.

| scaling | branch | time (no avx) | time (with avx) |
| :---------:|:---------:| ------:| ------:|
| yes | development | 8.52 | 8.45 |
| yes | this PR | 2.98 | 2.09 |
| no | development | 2.54 | 2.42 |
| no | this PR | 1.96 | 1.58|

We get about a 4-fold speed increase for DNA + scaling!

This means that scaling is now about 52% slower than no scaling without AVX, and 32% slower with AVX.

(Note that AVX instructions are x86_64-only, and need to be explicitly enabled with gcc and clang.  Adding `-march=native`  should work, but then the executables may not run on other computers.  We should probably build the release executable with something like `-march=x86_64-v3`, which should enable AVX, and should work on chips after around 2013 or so.)

NOTE: this is the first part of a series of patches speeding up the likelihood calculator.


benchmarks/Results.md
```diff
@@ -0,0 +1,69 @@
+# Benchmark results
+
+This is an unofficial place to collect results.
+The results are not official.
+
+## Machines
+
+### Ben | centromere | 4/17/2026
+
```
davidcerny, line 9: I assume "centromere" is the name of your cluster or something? It might be more informative to record the CPU used.

src/core/distributions/phylogenetics/substitution/PhyloCTMCSiteHomogeneousNucleotide.h
```diff
@@ -45,12 +45,7 @@ namespace RevBayesCore {
 
 #include <cmath>
 #include <cstring>
-#if defined ( SSE_ENABLED )
-#include <xmmintrin.h>
-#include <emmintrin.h>
-#include <pmmintrin.h>
-#elif defined ( AVX_ENABLED )
-#include <xmmintrin.h>
+#if defined (__AVX__)
```
davidcerny, line 48: Hmm... this is obviously not new to this PR, but I'm wondering why the AVX code is placed here rather than in `PhyloCTMCSiteHomogeneous.h`. Is it because the handwritten code can provide efficiencies only if `nChars` is known at compile time?

I also find it strange that despite its name, `PhyloCTMCSiteHomogeneousNucleotide` is still templated over `charType`... but I guess that's because we distinguish between DNA and RNA types?

davidcerny (approved):
This looks great! For what it's worth, here's what the speedup looks like on an ARM64 CPU (M3 Max) when building with `-DCMAKE_CXX_FLAGS="-march=native"`:

```
time ../projects/cmake/build/rb -o useScaling=false bench-unrooted.Rev ../tests/data/primates_and_galeopterus_cytb.nex DNA 100000

#### development

4.055
4.055
 4.089
4.290
4.608

#### refactor-scaling

3.925
3.931
 3.932
4.008
4.693

time ../projects/cmake/build/rb -o useScaling=true bench-unrooted.Rev ../tests/data/primates_and_galeopterus_cytb.nex DNA 100000

#### development

8.694
8.726
 8.746
8.756
8.778

#### refactor-scaling

4.756
4.797
 4.797
4.800
4.825
```

src/core/distributions/phylogenetics/substitution/PhyloCTMCSiteHomogeneousNucleotide.h
bredelings: The AVX code requires `num_chars` to be exactly 4.  Eventually maybe we can pull this kind of thing back into `PhyloCTMCSiteHomogeneous` and make specializations for `num_chars` = 2,3,4,5,6,7,8 or something like that.  The current AVX code would be part of the specialization for `num_chars==4` then.

> I also find it strange that despite its name, PhyloCTMCSiteHomogeneousNucleotide is still templated over charType... but I guess that's because we distinguish between DNA and RNA types?

I think so.  I think we should be able to decrease the amount of templating eventually.  The actually partial-likelihood calculation routines don't depend on `charType`.

benchmarks/Results.md
bredelings: These are informal numbers that I just put here so that I can compare future results with these ones.  I'm not trying to compare between CPUs.

bredelings:
Thanks for the ARM numbers.  Supposedly `-march=native` has a big effect on Intel/AMD because the default architecture is really old and doesn't contain a lot of vector instructions.  In comparison, it has little effect on ARM, because the default architecture is newer and has more vector instructions.  Or something like that.

At some point, I'd like to try using Eigen for some of this stuff.  It has code to vectorize for both Intel/AVX and ARM/NEON, so its possible it could get more speedups for Macs.  Its also possible it would not.  If Eigen can give us code that is as fast as the hand-written AVX, then possibly we could remove it, which would make it much easier to cleanly handle `num_chars != 4`.

benchmarks/Results.md
davidcerny: Sure, but if we are going to keep these notes in a public repo, they should probably make sense to other people too.

PR #1006, opened 2026-04-17
Make tests less fragile

bredelings:
Currently, small changes to likelihoods in the 14th or 15th decimal place can make big changes to an MCMC run.  It turns out that a lot of this is because we check if the _difference_ between two log likelihood is exactly zero, and decide to draw a random number or not based on that.

So, currently if the current and proposed likelihoods are 15.0000000000001 and 15, then we draw a random number.  But if they are 15 and 15, then we do not.  This check turns out to be really fragile.  Since drawing a random number affects all downstream random numbers, this turns out to be a big part of why different platforms yield different MCMC traces (but not all of it).

So this PR proposes to always draw one, in order to keep the random number streams in sync, even if we don't really need it.  After this change, the tests on Mac that disagree with Linux drops down to:

```
>>>> Test failed: CBDSP
	mismatch: mcmc_CBDSP.errout
>>>> Test failed: FBD
	mismatch: mvFossilTimeSlideUniform_multiple_runs.errout
>>>> Test failed: propose_bad_index
	mismatch: primates_and_galeopterus_cytb_RJ_posterior_samples.log
	mismatch: model_average_primates_cytb_debug.errout
	mismatch: primates_and_galeopterus_cytb_RJ_posterior_samples_debug.log
```

However, a large number of tests are different before and after this PR, so update the test results.

NOTE: this PR is a prerequisite for #1002


davidcerny:
Interesting that such a seemingly tiny change can have such a huge impact!

> After this change, the tests on Mac that disagree with Linux drops down to:
>
> ```
> >>>> Test failed: CBDSP
> 	mismatch: mcmc_CBDSP.errout
> >>>> Test failed: FBD
> 	mismatch: mvFossilTimeSlideUniform_multiple_runs.errout
> >>>> Test failed: propose_bad_index
> 	mismatch: primates_and_galeopterus_cytb_RJ_posterior_samples.log
> 	mismatch: model_average_primates_cytb_debug.errout
> 	mismatch: primates_and_galeopterus_cytb_RJ_posterior_samples_debug.log
> ```

If so, should this PR perhaps also reduce the amount of special-casing we currently perform in `build.yml` and `release.yml`?

bredelings:
Yeah, I can do that.  I was hoping that I would eventually figure out how to remove ALL the special cases, but we can certainly remove the unneeded ones now.

It make take a few tries to get this exactly right.

davidcerny (approved):

PR #1007, opened 2026-04-19
More options for resolving multifurcations in `BranchLengthTree`s

### LINKED ISSUE #992: No starting state with computable probability for the stepwise inference tutorial
**Bug report: Stepwise Bayesian inference of phylogeny tutorial fails at Step 2 (dnBranchRateTree initialization)**

**RevBayes version:** v1.3.0
**OS:** macOS (Apple Silicon 26.3.1)
**Tutorial:** Stepwise Bayesian inference of phylogeny
**Tutorial last modified:** April 21, 2025

I ran the tutorial using the original example dataset without modification.

Step 1 (gene-tree inference with `dnUniformTopologyBranchLength`) runs successfully and produces:

photinus_COI.trees
photinus_COI_MAP.tre

Step 2 (ultrametric dating using:

phis ~ dnEmpiricalSample( dnBranchRateTree(time_tree, branch_rate_prior, root_branch_fraction) )

fails during initialization with:

Unable to find a starting state with computable probability after 1000 tries

Error: Problem processing line XXX in file mcmc_ultrametric.Rev

I verified:

* taxa match between files
* tutorial scripts unchanged
* using nruns=1 instead of nruns=2 does not fix the issue
* using only a single sampled tree in the .trees file still produces the same error
* initialization from sampled trees instead of MAP tree does not fix the issue

Because the failure occurs even with one posterior tree, the issue appears to be related to dnBranchRateTree or dnEmpiricalSample initialization rather than the empirical tree sample itself.

Please let me know if additional diagnostics would help.

davidcerny:
When the `.resolveMultifurcations()` method inserts new branches into a `BranchLengthTree` to break up polytomies, it currently sets their length to 0. This PR provides a new argument, `nonzeroBranchLengths`, which instead allows the newly inserted nodes to "steal" one half of the length of their parent's subtending branch (or, if the parent is the root, one half of the length of their sibling's subtending branch) for themselves.

For the motivation behind this, see my comment under issue #992.

### DIFF
--- a/src/core/datatypes/trees/TopologyNode.cpp
+++ b/src/core/datatypes/trees/TopologyNode.cpp
@@ -1904,7 +1904,7 @@ void TopologyNode::renameNodeParameter(const std::string &old_name, const std::s
 }
 
 
-void TopologyNode::resolveMultifurcation(bool resolve_root)
+void TopologyNode::resolveMultifurcation(bool resolve_root, bool halve_branch_length)
 {
     if (children.size() < 3) return;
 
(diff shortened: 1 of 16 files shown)


davidcerny:
Closing this after a conversation with @bredelings. Rooting an unrooted tree by inserting an extra node should not really be the job of `.resolveMultifurcations()`, but of a separate method, and we might want this method to be more flexible than the solution adopted here (for example, by allowing the user to decide exactly _where_ along a given branch the root node should be inserted).

PR #1008, opened 2026-04-22
Remove pointer arithmetic

bredelings:
This PR is intended to be merged after #1002.  After #1002 is merged, I will change the merge base to `deleopment`, but right now I've set the merge base as #1002 so that you can see changes relative to #1002.

The goal is to remove pointer arithmetic in favor of array indexing.  Pointer arithmetic does:
```
for(int i=0;i<n;i++)
{
   sum += *x * *y;
   x++;
   y++;
}
```

whereas array indexing does
```
for(int i=0;i<n;i++)
{
    sum += x[i] * y[i];
}
```


bredelings:
Close in favor of #1026

PR #1009, opened 2026-04-23
Remove pointer arithmetic2

bredelings:
Second pass at refactoring the likelihood calculator to avoid pointer arithmetic.  This time I am checking that all the modifed code is covered by at least one test.

Also:
* make sumMarginalLikelihoods return `vector<vector<double>>` instead of a pointer.  With modern C++, we don't copy the vector -- since its an r-value reference (`&&`), we steal its internal storage instead of copying.
* don't use this->num_chars in loop bounds are array indexing, because the compiler can't prove that some other function didn't change this value behind its back.  Instead copy these to a stack variable that the compiler can tell isn't changing.
* use __restrict__ on pointers so that the compiler know that the pointers to the left and right child can't point to the same thing.


bredelings:
Closed in favor of #1026

PR #1016, opened 2026-05-07
Add badges with download counters & other info to the README

davidcerny:
At the developer meeting this past Tuesday, @hoehna suggested tracking the number of downloads for different versions of RevBayes. We can get these numbers pretty easily for executable downloads, and display them using GitHub badges. While I was at it, I added a few other badges, too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,21 @@
 # RevBayes 
 
+[![RevBayes v1.0 releases](https://img.shields.io/github/downloads/revbayes/revbayes.archive/total.svg?style=social&logo=github&label=Downloads:%20v1.0)](https://github.com/revbayes/revbayes.archive/releases)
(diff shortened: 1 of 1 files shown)


basantakhakurel (approved):

bredelings:
This looks really nice -- but note that people who go to revbayes on github won't see it unless they select the development branch.  

This is maybe another argument to rename development to main or master, and have only one protected branch.  I was going to do this, but then chickened out because I hadn't thought through all the implications.

davidcerny:
Well, they _will_ see it once we merge development into master 😛 A bigger issue is that the links point to specific tags, so the badges are one more thing we'll have to update every time we do a release...

basantakhakurel:
I was thinking of putting it in the website in the downloads page.

PR #1018, opened 2026-05-11
Fix reading of RNA datasets

### LINKED ISSUE #1017: [BUG] Cannot read an RNA alignment in Phylip format
**Describe the bug**
When I pass a Phylip file with an interleaved RNA alignment to `readDiscreteCharacterData()`, an exception is thrown:

```
   Error:	Cannot add a state to a RNA character!
```

**To reproduce**
Here is the file I was trying to read in: [Aberer_et_al_2015_dataset_150.phy.txt](https://github.com/user-attachments/files/27600118/Aberer_et_al_2015_dataset_150.phy.txt)

**Expected behavior**
We should be able to read this file.

**Computer info**
macOS 26.3 (Tahoe), ARM64 architecture, RevBayes v1.4.0-dev, current tip of development (9edfd16).

**Additional context**
After a cursory examination of [`RnaState.cpp`](https://github.com/revbayes/revbayes/blob/master/src/core/datatypes/phylogenetics/character/RnaState.cpp), I thought the error might be thrown because the ambiguity code `N` (unknown nucleotide) is not a valid symbol in our RNA alphabet. However, when I inserted an extra line to `RnaState::addState()` to print out the symbol whose addition we were requesting, I got a `C`. So I'm not quite sure what's happening.

davidcerny:
This closes #1017 by patterning the member functions of `RnaState` after those of `DnaState`.


basantakhakurel:
Looks good to me, and can confirm it works.
Is there a test that can be added? Something like reading in the interleaved RNA dataset and showing the chartype?

basantakhakurel (approved):

davidcerny:
Good idea – added.

PR #1019, opened 2026-05-12
Implementing stop function (Resolves #1011)

### LINKED ISSUE #1011: [Feature Request] Implement a `stop()` function
The Rev language already provides basic control flow facilities such as loops, conditionals, and even user-defined functions. One fairly basic feature we don't have is the ability to halt execution with a user-specified message, similar to R's `stop()` function or the `raise` keyword in Python or Ruby:

```r
if (input == 1) {
    < ... >
} else if (input == 2) {
    < ... >
} else {
    stop("Invalid input detected.")
    # or: raise "Invalid input detected."
}
```

Since Rev is (mostly) implicitly typed, a simple R-like solution might work better for our purposes than the typed exception hierarchies of languages like Python. In fact, a Rev function acting as a thin wrapper around one of the `RbException` constructors should be sufficient, and probably easier to implement than a new keyword.

**Describe alternatives you've considered**
The closest we can currently get to the above is probably

```r
if (input == 1) {
    < ... >
} else if (input == 2) {
    < ... >
} else {
    print("Invalid input detected.")
    q()
}
```

However, we might not always want to quit the session upon encountering an error.

**Additional context**
An obvious extension would be to implement some form of try-catch exception handling.

basantakhakurel:
This PR implements the `stop()` function as requested in #1011.
It provides a way to halt Rev script execution with a user-specified message without terminating the entire RevBayes session.


bredelings:
Have you checked that for non-interactive scripts, this exits with 1, while `q()` exits with 0?

You should be able to do something like
```
if rb script.Rev ; then
    print "success!"
else
   print "FAIL!"
fi
```

bredelings:
Long term (maybe not here), we might want to derive a new exception class for Quit and Stop exceptions, instead of just recording a single exception type on the RbException.  Then we could do something like

```
try {
....
}
catch(RbQuitException& e)
{
   // this will be 0 if this exception came from q(), or 1 if it came from stop()
   std::exit(e.exit_code);
}
catch (RbException&e)
{
   // other exception types
   ...
}
```

Then we could also allow `q(status=exit_code)` like in R.

davidcerny:
This is great! It works exactly as intended, and it also satisfies 2 of the 4 desiderata listed by @bredelings in his comment under the original issue:

- The exit code is 1 (implying failure) rather than 0 (implying success);
- The message specified by the user is prefixed with a keyword indicating what happened (in our case, `Stop:\t`).

Of the other two things that R's `stop()` does but ours does not, storing the traceback would probably require _a lot_ of extra work for fairly minimal gains. If the goal is to locate where the statement was triggered, we provide an alternative mechanism for doing that by pointing to the line in question (as long as we are executing a script, which is the most realistic use case):

```
   Stop:	Wrong value of a.
   Error:	Problem processing line 9 in file "./test.Rev"
```

As it turns out, `stop("message")` sends `"message"` to stdout rather than stderr, but that's consistent with other types of exceptions, such as `Missing variable`. Maybe we can change that down the road, but it's beyond the scope of this PR.

@basantakhakurel Can you rebase the branch on the current tip of development? I'm happy to approve once that has been done.

bredelings:
It seems like this should perhaps print "Error: message", instead of "Stop: message"?

davidcerny:
@bredelings Then it would look like this:

```
   Error:	Wrong value of a.
   Error:	Problem processing line 9 in file "./test.Rev"
```

In my opinion, that's a little confusing – it's not clear if we encountered one error or two.

basantakhakurel:
> It seems like this should perhaps print "Error: message", instead of "Stop: message"?

I was debating whether to use "Raise: message" or "Stop: message"; I went with "Stop". 
I think having two Error messages in consecutive lines would make it less elegant

davidcerny (approved):

bredelings:
Apparently this is the normal way that we print errors:
```
   Error:	Cannot complete this assignment: '<unnamed> (stochastic node)' depends on 'a', so replacing one with the other would create a circular dependency.
   Error:	Problem processing line 3 in file "tmp.Rev"
```
The second line isn't really the error content, its the error *location*, so perhaps it shouldn't be labelled "error" -- maybe "Note: "?  I think that's kind of what gcc does, and it seems rather clearer to me - but we could change that later.

bredelings:
See also #1024 for some possible future improvements in the error messages.  I like "Error: " more than "Stop: ", because it makes it clear that the message is an error and not just a stopping condition.  Thus, we should see things like "Stop: reached 1000 iterations".  But the Pr is fine as is.

Thanks @basantakhakurel !

davidcerny:
@bredelings If you look at [`RbException.cpp`](https://github.com/revbayes/revbayes/blob/master/src/core/utils/RbException.cpp), you'll see that "Error: " is just the default. We only use it when we have nothing more specific to say about the type of the exception that has occurred (such as "Mathematical Error", "Missing Variable", etc.). In this case, we do have something more specific to say about it: it's a _user-specified_ exception.

In short, I think "Stop: " is fine.

bredelings:
All of the other things specifically say that something is wrong, not just that we have stopped, though.  Unlike the other things, "stop" doesn't actually imply that an error occurred -- it just implies that we have stopped.  If we just print "Stop:", someone is going to do something like "Stop: successfully finished all iterations".

Additionally, I think there should be an option to NOT print the error location on a user-defined error message, for cases like `stop("The  script needs 3 arguments, but you have only specified 2")`.  In such cases the location where we issue the `stop` from is not the location of the error.  And then there would be NO line that says "Error".

And I think we should stick closer to the R design here.  Which also doesn't print an error location.

So, I still don't like it.  But I won't hold up the PR.

Issue #889, opened 2025-11-10
[BUG] BDSTP's `swapInternalParameter()` function only handles rates, not timelines


davidcerny:
**Describe the bug**
The distribution for the birth-death-sampling-treatment process (BDSTP) has a number of protected variables that are pointers to the vectors of rates, event probabilities, and times:

https://github.com/revbayes/revbayes/blob/cbe72b0372cf2c869d951293b383181bacdb9e99/src/core/distributions/phylogenetics/tree/birthdeath/BirthDeathSamplingTreatmentProcess.h#L99-L120

The [`swapParameterInternal()`](https://github.com/revbayes/revbayes/blob/cbe72b0372cf2c869d951293b383181bacdb9e99/src/core/distributions/phylogenetics/tree/birthdeath/BirthDeathSamplingTreatmentProcess.cpp#L1915) function handles the first two categories of parameters, but not the last.

**Computer info**
The problem is present both in the latest release (v1.3.1) and on the current tip of development (5b5774c).

**Additional context**
Fixing this issue may be necessary for solving #887, but it is not sufficient (I've checked).

davidcerny:
Note that this will only be an issue if the timelines are treated as free parameters that we need to update during the MCMC – i.e., if the times at which the FBD rates (speciation, extinction, etc.) are allowed to shift are themselves being estimated. This is a theoretical possibility, but to the best of my knowledge, no such analysis has ever been conducted.

Issue #864, opened 2025-10-09
[Help Wanted] / [Feature Request] Joint fossil tip time slide moves / equal tip age constraint


mcranium:
Dear development team,

**Background**
I am trying to set up an FBDP analysis that includes fossil insects. The aspect that separates fossil insects from many, but certainly not all, other studied organisms is that they are quite rare in the fossil record, but if they appear, there are often tens if not hundreds or thousands of species per fossil site. Some of the fossil sites are not well dated, leading to large uncertainty intervals. Those large uncertainty intervals then lead to problems with the analysis, both regarding the performance and the possible results.

<img width="3030" height="1143" alt="Image" src="https://github.com/user-attachments/assets/483ebbf1-25b5-4b18-85a7-0a098174edfb" />

**What is going wrong?**
If several species from the same site are included and they can be seen as coming from the same point in time, each species receives its own uniform distribution to draw from, leading to an unnecessary high number of parameters. The larger the uncertainty interval is, the larger the discrepancy between the tip time estimates will get. lf sampled ancestors are allowed, it can happen that for a species an ancestral species is recovered, which lived at the same time and place and is morphologically distinct.

**Describe the desired behavior.**
The desired implementation should allow to specify which species have equal ages. Ideally, this could be done using a table read in by `readTaxonData`, for example by adding a column like `age_group`. Then, one should be able to pass this information to `mvFossilTipTimeSlideUniform`, which should perform a joint move. Note that this move has to check, whether the proposed move would make any of the tips older than its parent; in such case another value should be drawn from the distribution until the move does not lead to negative branch lengths.
Alternatively, this could be handled similarly as the relative age constraints that are already implemented in RevBayes.

**Anything else you've tried?**
Inspired by suggestions from my colleague Brendon Boudinot, I have tried to construct a workaround by trying to assign new tip ages through `tree.setAge(nodeID, proposed age)`. However, this does not seem to work even when `proposed_age` is a deterministic node originating from a `dnUniform`.

My colleague Ziad Khoury managed to set up a workaround that utilizes a small interval of e.g. 1 million years around one of the species from the same age, in which sampling the other species will lead to a non-zero likelihood. However, this workaround does not seem to be a good solution, since it is very inefficient when the uncertainty interval is large and the arbitrary interval in which same aged fossils are allowed is small. While this will lead to better results, a good solution should reduce the number of to-be estimated parameters, instead of adding further distributions to sample from.

[...]

PR #1021, opened 2026-05-12
Slide fossil sites

mcranium:
I added the function `mvFossilSiteSlideUniform` to solve an issue for which I found no solution within the Rev language.

There are situations where you might want to move several fossil tips for a FBD analyis, but at the same time the tips should not move independently. An example for this would be a badly dated fossil site, for which it is clear that the fossils are of the same age. If one runs a standard analysis, the fossil tips move independently, making the convergence harder and the end result unsatisfactory.

 I adapted the code from `mvFossilTimeSlideUniform` in order to accomplish this and registered the function to be available via the Rev language.

I used the clade object for the fossil sites (an alias "site" for the clade object could help to make the user code easier to read, but not implemented here). The usage of the function is:
```
moves.append( mvFossilSiteSlideUniform(fbd_tree, origin=origin_time, clade=fossil_site_a, weight=5.0) )
```
I prepared an example script and demo data for testing, adapted from the FBD Morphology tutorial.
[mvFossilsiteTimeSlideUniform_demo.zip](https://github.com/user-attachments/files/27645758/mvFossilsiteTimeSlideUniform_demo.zip)

From what I can tell, it get's the job done and produces sane results.

I hope that this functionality is not too niche to make it into RevBayes. For me, personally (research on fossil insects), having the option to constrain fossil tips to be contemporary is absolutely essential.

Please note that this was my first ever contribution in C++ and I can't say that it was all smooth sailing, coming from R and Python. Any help is much appreciated.

Best,
Mario


src/core/moves/proposal/tree/FossilSiteTimeSlideUniformProposal.cpp
```diff
@@ -0,0 +1,364 @@
...
+        max_age = taxon_max_age;
+    }
+    else
+    {
+        // adjust max age given provided variable
+        double provided_max_age = max->getValue();
+        max_age = provided_max_age;
+    }
+
+    double max_ages[clade.size()];
```
mikeryanmay: This looks to be creating an array of variable length, which I think is not valid syntax for all compilers (and is not supported by any standard as far as I can tell, but I could be wrong). Array size is usually fixed at compile time. std::vector is the normal way to handle things with sizes that are determined at runtime.

src/core/moves/proposal/tree/FossilSiteTimeSlideUniformProposal.cpp
```diff
@@ -0,0 +1,364 @@
...
+        max_ages[i] = max_age;
+    }
+
+    max_age = *min_element(max_ages, max_ages + clade.size());
+    
+    // now we store all necessary values
+    stored_age = my_age;
+    
+    double size = max_age - min_age;
+    assert(size >= 0); //otherwise the while will hang forever
```
mikeryanmay: Tripping the assert here will stop your run. Asserts are intended for when your program encounters a state that should terminate the program. I think your goal is to ensure a proposal happens only if size >= 0, in which case an if statement would probably suffice.

src/core/moves/proposal/tree/FossilSiteTimeSlideUniformProposal.cpp
```diff
@@ -0,0 +1,364 @@
...
+ *
+ * The summary just contains the current value of the tuning parameter.
+ * It is printed to the stream that it passed in.
+ *
+ * \param[in]     o     The stream to which we print the summary.
+ */
+void FossilSiteTimeSlideUniformProposal::printParameterSummary(std::ostream &o, bool name_only) const
+{
+    
+    o << "delta = ";
```
mikeryanmay, line 279: Little bit confusing that delta is the name of the variable at the language level but it's called lambda here. I see that's also true in the fossil tip proposals so it makes sense to be consistent.

src/core/moves/proposal/tree/FossilSiteTimeSlideUniformProposal.cpp
```diff
@@ -0,0 +1,364 @@
...
+ * where complex undo operations are known/implement, we need to revert
+ * the value of the variable/DAG-node to its original value.
+ */
+void FossilSiteTimeSlideUniformProposal::undoProposal( void )
+{
+    
+    // undo the proposal
+    Tree& tau = tree->getValue();
+    TopologyNode* node = &tau.getNode(node_index);
+    node->setAge( stored_age );
```
mikeryanmay: If I understand correctly, I think you need to reset the ages for all the nodes in the "clade" when the proposal is rejected, right?

mikeryanmay:
Overall looks good, but I've provided some comments on some of the C++ syntax (I've stopped short of providing fixes in case you want the exercise). I think there's also a significant bug in the rejection step.

Mechanically and empirically I understand the motivation for this proposal, since fossils can be collected in the same interval. However, I'm a little worried about delegating this behavior to MCMC proposals rather than to the model itself. If you use this proposal with the standard FBD model, for example, each fossil is still sampled independently at rate psi (or phi or whatever you want to call it) at exactly the same moment. However, because the FBD model says each even is independent, the density of more than one event at exactly the same moment should in principle be zero. I guess what I'm saying is the correct treatment of multi-fossil samples is different from just using an FBD with coincident ages, but I understand that to be out of scope for this particular pull request.

mikeryanmay:
You may also consider adding a test script to ensure your code doesn't break downstream! See [here](https://revbayes.github.io/developer/tests/)

mcranium:
Alright, I will try to solve the code issues on my own. How do the corrections for pull requests work? Do I make another pull request for this?

> Mechanically and empirically I understand the motivation for this proposal, since fossils can be collected in the same interval. However, I'm a little worried about delegating this behavior to MCMC proposals rather than to the model itself. If you use this proposal with the standard FBD model, for example, each fossil is still sampled independently at rate psi (or phi or whatever you want to call it) at exactly the same moment. However, because the FBD model says each even is independent, the density of more than one event at exactly the same moment should in principle be zero. I guess what I'm saying is the correct treatment of multi-fossil samples is different from just using an FBD with coincident ages, but I understand that to be out of scope for this particular pull request.

I am still trying to understand this aspect. As far as I understand it, my function only changes the age values for each MCMC iteration. In the tutorial, it says that in the FBD process the lineages produce fossils at a constant rate. I assume, in RevBayes, mechanistically, this is done the other way round (otherwise there would not be the need for a function like `mvFossilTipTimeSlideUniform`. The for loop with the `F[i].clamp( 0 )` command sets the likelihoods to zero if the current fossil age is outside of the specified range. Then, i assume the psi parameter is used to either set the likelihood of finding the fossil to zero or not. Is this correct? If my understanding is correct. I don't really see the problem yet. When the psi parameter comes into play, the contemporary fossils are treated independently.

Maybe I am looking at this problem from a wrong angle.

Aside from this: Am I correct in assuming that if `mvFossilSiteTimeSlideUniform` were to be used alongside a single `mvFossilTipTimeSlideUniform` call (without specifying a single fossil tip), in some few (depending on how many other fossils there are) MCMC iterations the current ages of two fossils that are treated as contemporary in `mvFossilSiteTimeSlideUniform` would differ?

davidcerny:
Hi @mcranium, I just wanted to quickly answer this question of yours:

> Alright, I will try to solve the code issues on my own. How do the corrections for pull requests work? Do I make another pull request for this?

No, you can just keep adding commits to the branch for which you opened this pull request (i.e., `slideFossilSites` in your own fork of the repo), and they will automatically show up here.

I also wanted to thank you for being proactive and addressing an issue (#864) you yourself drew attention to! It's always great when someone does that. I'm hoping to take a more detailed look at your PR over the weekend, and to try it out in practice.

I basically agree with @mikeryanmay that what you are doing is well-motivated empirically, which is why it has its counterparts in other software (for example, in PyRate's "site-linking" option, or the [`palaeo` package](https://peerj.com/articles/9368/) for BEAST 2), but a bit problematic mathematically. What you might need for proper site-linking is the FBD-burst process described by [Magee & Höhna (2021)](https://www.biorxiv.org/content/10.1101/2021.01.14.426715v1), which is implemented in RevBayes by means of the `Lambda` / `Mu` / `Phi` parameters passed to `dnFBDP`.

bredelings:
One use of virtual evidence would be to specify that we have observed that times t1, t2, and t3 are all equal.  When this is eventually done, I guess it could look like:
```
!~ vlkAllEqual([v1,v2,v3], tol=1.0e-6)
```
This could then be coupled with a transition kernel that moves them all together.

If t1, t2, and t3 were all stochastic nodes (which they usually aren't, at least in RevBayes), then in theory we could also do something like:
```
v2.clamp(v1)
v3.clamp(v1)
```
... but only if we implement the ability to clamp to a deterministic node, and not a constant.

As to whether this fits the model, I think  there is a time-honored tradition of using a bad model that is made somewhat usable by conditioning on observed data, even if the event conditioned on has probability 0 under the model.  

That said, a better model would be nice.  Presumably in reality, different fossils from the same formation have similar, but not identical ages?  Also, if different fossils were from the same formation, perhaps you could determine their relative depths within that formation, and say that the ages are some function of depth within the formation?  This would guaranteed that the ages are between the max-age and min-age for the formation, even if the max-age and the min-age also need to be estimated.

davidcerny:
@bredelings

> Presumably in reality, different fossils from the same formation have similar, but not identical ages?

If different fossils are merely from the same _formation_, then yes, those can easily be several million years apart from each other. But what you often get are different fossils from the same bone bed, in which case you do actually know them to be contemporaneous – sometimes even perfectly contemporaneous, as is the case when the fossils belong to organisms that died in the same event (mud slide, crossing a flooded river, etc.).

bredelings:
OK, cool.  Then presumably even the episodic model, with a high extinction rate for a short period, would not actually handle finding multiple fossils with the exact same time?

One interested idea would be a biogeographic range model where you say that the fossilization rate depends on the area as well as the time.   If you observe a fossil, you know which range it is in, which makes `Pr(fossil observation|range) = 0` if the range does not include the area where the fossil was collected.

davidcerny:
> Then presumably even the episodic model, with a high extinction rate for a short period, would not actually handle finding multiple fossils with the exact same time?

Yes, which is why I proposed using not an episodic / skyline model, but rather the FBD-burst model, in which different kinds of events (birth, death, sampling) may take place in instantaneous tree-wide bursts, _in addition to_ occurring at a constant or piecewise-constant background rate. Consulting the paper I linked would have made the distinction clear.

Before issuing specific recommendations or criticisms, it is usually useful to familiarize oneself with what the people in a given research area are actually doing. They have valuable domain knowledge, and it often turns out that they have already given careful consideration to the issues and caveats one just thought of.

src/core/moves/proposal/tree/FossilSiteTimeSlideUniformProposal.cpp
mcranium: I adopted this from [FossilTipTimeSlideUniformProposal.cpp](https://github.com/revbayes/revbayes/blob/990c15a899d16a52e1c90c149b7fb3ced8b0d6bf/src/core/moves/proposal/tree/FossilTipTimeSlideUniformProposal.cpp#L274)
Maybe it would make sense to clean this up in both functions independently of this PR?

src/core/moves/proposal/tree/FossilSiteTimeSlideUniformProposal.cpp
mcranium: Right, and this is not needed in [FossilTipTimeSlideUniformProposal.cpp](https://github.com/revbayes/revbayes/blob/990c15a899d16a52e1c90c149b7fb3ced8b0d6bf/src/core/moves/proposal/tree/FossilTipTimeSlideUniformProposal.cpp#L292) because there in each MCMC iteration, only one node (fossil tip) gets changed, right?

src/core/moves/proposal/tree/FossilSiteTimeSlideUniformProposal.cpp
mcranium: I adopted this from [FossilTipTimeSlideUniformProposal.cpp](https://github.com/revbayes/revbayes/blob/990c15a899d16a52e1c90c149b7fb3ced8b0d6bf/src/core/moves/proposal/tree/FossilTipTimeSlideUniformProposal.cpp#L225)
Is this then not equally bad there? I tested it with the bears dataset and if the max and min ages are the same, the analysis just struggles to compute likelihoods
```
   Drawing new initial states ... 
   Could not compute lnProb for node 'F[6]': lnProb = inf
   0
   
   
   Drawing new initial states ... 
   Error:	Unable to find a starting state with computable probability after 1000 tries
   Error:	Problem processing line 191 in file "FBD_FossilSiteTimeSlideUniform_tutorial.Rev"
```
If the ages are the wrong way round, `readTaxonData` catches it before the analysis can even start.

mcranium:
Did I understand it correctly, that even if one used the FBD-burst process from [Magee & Höhna (2021)](https://www.biorxiv.org/content/10.1101/2021.01.14.426715v1), a function like `mvFossilSiteSlideUniform` is still needed to "link" the sites (i.e. to change the proposed ages of two fossil tips from one site), and that the standard FBD model would technically work but does not accurately describe what is happening because of assumption of a constant fossilization rate?

I tried to fix the code issues and will continue with my attempt of an integration test.

Thank you very much for the support and the interesting discussion!

By the way, my specific use case is a fossil site, where the age of the entire site has a high uncertainty range of several million years (lake sediments where it is hard to establish stratigraphic relationships to other sites) and the fossils are probably a few hundred thousand years apart at most. In this case a simplifying assumption of the fossils being contemporary should be justifiable (and likely necessary).

davidcerny:
> Did I understand it correctly, that even if one used the FBD-burst process from [Magee & Höhna (2021)](https://www.biorxiv.org/content/10.1101/2021.01.14.426715v1), a function like `mvFossilSiteSlideUniform` is still needed to "link" the sites (i.e. to change the proposed ages of two fossil tips from one site), and that the standard FBD model would technically work but does not accurately describe what is happening because of assumption of a constant fossilization rate?

That's exactly right, yes! In fact, about a year ago, I asked @afmagee for advice on how to use the FBD-burst model with sampling bursts corresponding to well-sampled Lagerstätten, and his reply – if I may take the liberty of quoting it here – was:

> The problem you face is that sampling bursts are “at this exact time, tips come into existence.” (“Exact” meaning “close enough to equal in double precision given that you may be writing a tree with only 6 decimals of branch length precision.”)
>
> Perturb any one sampling time away from that and it becomes part of the continuous time sampling model, not the burst model.
>
> So, if the continuous time sampling rate is 0, you’ll fall off the cliff into likelihood 0.
>
> If the serial sampling/fossilization rate is not zero, without a move capable of setting times to be exactly equal to a Phi time, they’ll all end up under the continuous time model
>
> That’s the needle you have to thread. It’s doable, it’s just potentially a nightmare

I.e., you will notice that the one practical obstacle to what I was attempting to do was precisely the absence of a move like the one you are now introducing!

My only minor quibble with what you wrote would be that what makes the standard FBD model less than ideal in this scenario is not so much the assumption of a _constant_ fossilization rate as the assumption of fossilization being a continuous-time process, for which the probability of two events occurring at exactly the same time is infinitesimally small. You can relax the rate constancy assumption – yielding the episodic or skyline FBD model – and still run into the same problem, as @bredelings pointed out.

mcranium:
I think I understand this a bit better now, but I am still very much confused. If I understand this correctly, the problem is that with the FBDP we actually simulate fossilization events within each MCMC iteration. Then, $\psi$ is a fossilisation rate. If a simulated fossilization event along a certain lineage falls within the occurrence time uncertainty interval, the likelihood of the tree gets calculated as usual (with the likelihood of the fossil occurrence being `1`), if it does not occur within the interval, the likelihood of the entire tree is forced to `0` by the likelihood of the fossil occurrence being `0`.

Does this then mean that the FBDP can not handle fossils with an age fixed to a specific value? 

If one specified a very narrow interval for a fossil, would this lead to a bad performance because in most of the iterations the fossil is sampled outside of the interval?

Wouldn't it make sense, then, to use `mvFossilSiteSlideUniform` in conjunction with a Birth-Death model where there are two sampling *probabilities* ($\rho$ for the extant species and [some other Greek letter] for the fossils), but where the fossil sampling events are not simulated? Is this what is happening in [Stadler 2010](https://doi.org/10.1016/j.jtbi.2010.09.010)? But there, $\psi$ is also referred to a sampling rate rather than a sampling probability.

I know that a pull request is probably not the correct location to teach theoretical foundations, but I am a bit lost here and unsure whether the function that I proposed will actually be useful.

bredelings:
> Before issuing specific recommendations or criticisms, it is usually useful to familiarize oneself with what the people in a given research area are actually doing. They have valuable domain knowledge, and it often turns out that they have already given careful consideration to the issues and caveats one just thought of.

Note that my comment that you referenced ("Then presumably...") was neither a recommendation, nor a criticism, but a question.  I did look at the paper, but got confused and was trying to clarify.

And indeed that makes more sense.

src/core/moves/proposal/tree/FossilSiteTimeSlideUniformProposal.cpp
mikeryanmay: I think if this treatment is inherited from [FossilTipTimeSlideUniformProposal.cpp](https://github.com/revbayes/revbayes/blob/990c15a899d16a52e1c90c149b7fb3ced8b0d6bf/src/core/moves/proposal/tree/FossilTipTimeSlideUniformProposal.cpp#L225), then it's not necessary to resolve for this PR. I guess I'm unclear on what types of failures this is meant to catch and whether those errors should kill the program vs result in a rejected proposal. Let's think about that after this PR is finished.

src/core/moves/proposal/tree/FossilSiteTimeSlideUniformProposal.cpp
mikeryanmay: Agreed.

mikeryanmay:
> I think I understand this a bit better now, but I am still very much confused. If I understand this correctly, the problem is that with the FBDP we actually simulate fossilization events within each MCMC iteration. Then, ψ is a fossilisation rate. If a simulated fossilization event along a certain lineage falls within the occurrence time uncertainty interval, the likelihood of the tree gets calculated as usual (with the likelihood of the fossil occurrence being `1`), if it does not occur within the interval, the likelihood of the entire tree is forced to `0` by the likelihood of the fossil occurrence being `0`.
> 
> Does this then mean that the FBDP can not handle fossils with an age fixed to a specific value?
> 
> If one specified a very narrow interval for a fossil, would this lead to a bad performance because in most of the iterations the fossil is sampled outside of the interval?
> 
> Wouldn't it make sense, then, to use `mvFossilSiteSlideUniform` in conjunction with a Birth-Death model where there are two sampling _probabilities_ ( ρ for the extant species and [some other Greek letter] for the fossils), but where the fossil sampling events are not simulated? Is this what is happening in [Stadler 2010](https://doi.org/10.1016/j.jtbi.2010.09.010)? But there, ψ is also referred to a sampling rate rather than a sampling probability.
> 
> I know that a pull request is probably not the correct location to teach theoretical foundations, but I am a bit lost here and unsure whether the function that I proposed will actually be useful.

It's not so much that the FBD model simulates fossilization events. It more that the probability function for the FBD assumes the ages of the fossils are exactly t[i], then we use MCMC proposals to average over all possible values of t[i] within the occurrence interval. 

The conceptual/mathematical issue with using the standard FBD like this is that, under the imagined model (not paleontological reality), each lineage produces fossils independently at rate ψ (whether or not it is constant over time). Mathematically, the probability density of two independent fossilization events at the exact some instant is zero. So, there's a conceptual gap between the standard FBD and the phenomenology you are trying to describe. Magee and Hoehna (linked above) is one way to address that.

mcranium:
> The conceptual/mathematical issue with using the standard FBD like this is that, under the imagined model (not paleontological reality), each lineage produces fossils independently at rate ψ (whether or not it is constant over time). Mathematically, the probability density of two independent fossilization events at the exact some instant is zero. So, there's a conceptual gap between the standard FBD and the phenomenology you are trying to describe. Magee and Hoehna (linked above) is one way to address that.

Could someone point me towards a possible implementation using the generalization of the FBDP model that was implemented in `dnFossilizedBirthDeath` when the Magee and Hoehna preprint was published? My understanding is: yes, this allows for tree wide events ($\Phi$ for sampling, using the notation of Magee and Höhna), but at the breaks between predefined time intervals. How could this help me, when I don't know the time of the sampling event? Specifying many such breaks within the age uncertainty interval of the contemporary fossils? Also for the $\Phi$ parameter the documentation of `dnFossilizedBirthDeath` says: 
>The probability of sampling taxa at sampling events (at present only if input is scalar).

Also, in this approach, there is still the serial sampling rate $\phi$. Wouldn't multiple contemporary fossils then still be a problem?

davidcerny:
> I know that a pull request is probably not the correct location to teach theoretical foundations, but I am a bit lost here and unsure whether the function that I proposed will actually be useful.

Some of the more theoretical conversation could perhaps be moved over to the [Discussions](https://github.com/revbayes/revbayes/discussions), but I'm also fine continuing it here, if you prefer to keep everything in one place.

> Could someone point me towards a possible implementation using the generalization of the FBDP model that was implemented in `dnFossilizedBirthDeath` when the Magee and Hoehna preprint was published? My understanding is: yes, this allows for tree wide events ($\Phi$ for sampling, using the notation of Magee and Höhna), but at the breaks between predefined time intervals. How could this help me, when I don't know the time of the sampling event?

_In principle_, the times of the sampling bursts do not have to be predefined; you could try to infer them. If you type `?dnFBDP` into the Rev console, you'll see that the `PhiTimeline` argument (i.e., the time line for sampling bursts) accepts real positive values wrapped in model graph nodes of `<any>` kind – i.e., not just constant nodes, but stochastic nodes as well. If one has, say, three sampling bursts, each of which corresponds to a Lagerstätte whose age is constrained but not known exactly, one could then imagine doing something like the following:

```r
lagerstaette_timeline[1] ~ dnUniform(146, 151)
lagerstaette_timeline[2] ~ dnUniform(120, 125)
lagerstaette_timeline[3] ~ dnUniform(47, 49)

time_tree ~ dnFBDP(...,
                   Phi = 1.0, # we are sure that sampling bursts did in fact occur -- just not when
                   PhiTimeline = lagerstaette_timeline,
                   ...)
```

Now, would this actually work? Based on issue #889, I have my doubts. But that's a technical issue, not a fundamental obstacle.

Now, where your move comes in is that it allows all fossils from Lagerstätte 1 to be moved within the [146, 151] Ma time range together, all fossils from Lagerstätte 2 to be moved within the [120, 125] Ma time range together, etc. This _should_ make it possible to estimate the age of these Lagerstätten, and of the corresponding sampling bursts. But perhaps @afmagee can correct me if I got anything wrong here.

> Also, in this approach, there is still the serial sampling rate $\phi$. Wouldn't multiple contemporary fossils then still be a problem?

No. In this model, the contemporary fossils would be treated as arising by a different mechanism (instantaneous bursts) than the others (a continuous-time process). You could not use them to estimate the serial sampling rate $\phi$, but the mere presence of that rate in the model doesn't render their existence impossible if you also have bursts.

davidcerny:
Now, regarding the more technical aspects of the PR, I don't quite like the current interface where the fossils whose ages are to be linked are wrapped in a `Clade` object:

```r
ursavus = clade("Ursavus_brevirhinus","Ursavus_primaevus")
moves.append( mvFossilSiteSlideUniform(fbd_tree, origin=origin_time, clade=ursavus, weight=5.0) )
```

My problem with this is that the fossils from the same site don't necessarily have to form a monophyletic group. Now, our `Clade` objects don't always have to correspond to monophyletic groups, but that's an abuse of terminology I'd like to hide from the users as much as possible.

I think you could perhaps draw inspiration from moves such as `mvUpDownScale` or `mvAVMVN`, which support the following interface:

```r
up_down_move = mvUpDownScale(weight=5.0)
up_down_move.addVariable(speciation_rate_at_present, TRUE)
up_down_move.addVariable(extinction_rate_at_present, TRUE)
moves.append( up_down_move )
```

To me, adding taxa to the move using the `.addVariable()` method would make more sense than pretending they form a clade.

mcranium:
> My problem with this is that the fossils from the same site don't necessarily have to form a monophyletic group. Now, our `Clade` objects don't always have to correspond to monophyletic groups, but that's an abuse of terminology I'd like to hide from the users as much as possible.

I completely agree.

My initial idea was to implement an alias in the Rev language, so that it would be possible to use the `Clade` object internally but do not expose the name "clade" to the user in the context of this function:
```r
fossil_site_1 = coevals("Ursavus_brevirhinus","Ursavus_primaevus") 
moves.append( mvFossilSiteSlideUniform(fbd_tree, origin=origin_time, coevals=fossil_site_1, weight=5.0) )
```
The benefit of this type of Rev code would look more familiar to R users, for which  and more in line with the idea of making the Rev language as similar to R as possible. Of course, this is a design choice and if you say, that the `.addVariable` solution is more elegant, I'll try to implement it.

One other thing: It was suggested that I create an integration test. While this does not sound particularly hard, I just wanted to ask, whether you have some recommendations for testing `mvFossilSiteSlideUniform`, so that it proves its functionality, but does not add excessive computation to the testing process.

mikeryanmay:
> > My problem with this is that the fossils from the same site don't necessarily have to form a monophyletic group. Now, our `Clade` objects don't always have to correspond to monophyletic groups, but that's an abuse of terminology I'd like to hide from the users as much as possible.
> 
> I completely agree.
> 
> My initial idea was to implement an alias in the Rev language, so that it would be possible to use the `Clade` object internally but do not expose the name "clade" to the user in the context of this function:
> 
> ```r
> fossil_site_1 = coevals("Ursavus_brevirhinus","Ursavus_primaevus") 
> moves.append( mvFossilSiteSlideUniform(fbd_tree, origin=origin_time, coevals=fossil_site_1, weight=5.0) )
> ```
> 
> The benefit of this type of Rev code would look more familiar to R users, for which and more in line with the idea of making the Rev language as similar to R as possible. Of course, this is a design choice and if you say, that the `.addVariable` solution is more elegant, I'll try to implement it.

I think of `Clade` as a fancy vector---it contains objects of the same type (like a vector) PLUS it has some extra functionality that makes sense for a set of such objects to have. That special functionality will usually manifest as methods used like `clade.getAge()`, etc. I think ultimately you would want to have something like a `Coevals` or `Collection` type, which could integrate nicely with an extended FBD that handled coevals explicitly.

Having said that, in the current use case it's not clear to me that you gain anything over using a vector of taxa, since no added functionality is required. Perhaps the argument could just be `taxa[]`? 

> One other thing: It was suggested that I create an integration test. While this does not sound particularly hard, I just wanted to ask, whether you have some recommendations for testing `mvFossilSiteSlideUniform`, so that it proves its functionality, but does not add excessive computation to the testing process.

We already have many tests, each of which is a short MCMC run, on the order of 10 iterations with minimal datasets. Adding one more will not be a computational problem! You could probably base your test on the existing FBD test scripts over at [https://github.com/revbayes/revbayes/tree/development/tests/test_FBD](test_FBD). You might even put your own test in that directory since we already have tests for the uniform slide proposal.

EDIT: To clarify, the tests aren't intended to prove correctness of the proposals. The integration tests are just to make sure changes don't break existing functionality (whether or not the existing functionality is mathematically correct). So you're not on the hook for proving the correctness of your proposal, at least as far as the tests are concerned.

davidcerny:
I wanted to write more or less the same thing, but @mikeryanmay beat me to it.

Adding `coevals` as an alias for `clade` would solve the immediate problem, but the underlying issue would still be there: users might discover that these two names stand for the same type of object, and find it confusing, since they don't _sound_ like they should be referring to the same thing.

> The benefit of this type of Rev code would look more familiar to R users, for which and more in line with the idea of making the Rev language as similar to R as possible.

I'm not sure that's what the goal is... My understanding is that Rev draws inspiration from previous probabilistic DSLs, as well as from the two languages most commonly used by computational biologists: R and Python. The whole `object.method()` syntax is quite un-R-like, for example, but very natural for someone with a Python background.

I agree with @mikeryanmay that the simplest change to make is to supply a vector of `Taxon` objects instead of a single `Clade` object. Using the `.addVariable()` and `.removeVariable()` methods would make your move more in line with our existing moves operating on multiple stochastic variables at once, but I don't insist on it if it's too onerous to implement.

> While this does not sound particularly hard, I just wanted to ask, whether you have some recommendations for testing `mvFossilSiteSlideUniform`, so that it proves its functionality, but does not add excessive computation to the testing process.

Again, I agree with the suggestion to just use the script from your demo, put it in the `test_FBD` folder, and decrease the number of iterations.

I was wondering if there was a way to change your script to prove that the move is doing what it's supposed to be doing, i,e., always keeping the ages of certain taxa equal to one another. My initial idea was to use something like the following loop:

```r
trace = readTreeTrace("output/bears.trees")
tol = 1e-6

for (i in 1:trace.size()) {
    age_1[i] = trace.getTrees()[i].getClade( [taxon("Ursavus_brevirhinus")] ).getAge()
    age_2[i] = trace.getTrees()[i].getClade( [taxon("Ursavus_primaevus")] ).getAge()
    print( abs(age_1[i] - age_2[i]) < tol )
}
```

which should always output `TRUE`. But I don't think the `.getAge()` method is the right solution here; I suspect it returns the starting point of a tip's branch rather than its end point (which is what we are interested in). Maybe we need a `.heightAboveTheRoot()` method... but that's beyond the scope of this PR.

mcranium:
I finally managed to change the clade argument to a vector of taxon objects, as suggested above. I also added a test case, using an adapted version of my example code. Sorry for the delay.

tests/test_FBD/scripts/FBD_mvFossilSiteTimeSlideUniform.Rev
```diff
@@ -0,0 +1,198 @@
...
+moves.append( mvFossilTimeSlideUniform(fbd_tree, origin_time, tip="Tremarctos_ornatus", weight=5.0) )
+moves.append( mvFossilTimeSlideUniform(fbd_tree, origin_time, tip="Ursus_americanus", weight=5.0) )
+moves.append( mvFossilTimeSlideUniform(fbd_tree, origin_time, tip="Ursus_arctos", weight=5.0) )
+moves.append( mvFossilTimeSlideUniform(fbd_tree, origin_time, tip="Ursus_maritimus", weight=5.0) )
+moves.append( mvFossilTimeSlideUniform(fbd_tree, origin_time, tip="Agriarctos_spp", weight=5.0) )
+moves.append( mvFossilTimeSlideUniform(fbd_tree, origin_time, tip="Ailurarctos_lufengensis", weight=5.0) )
+moves.append( mvFossilTimeSlideUniform(fbd_tree, origin_time, tip="Ballusia_elmensis", weight=5.0) )
+moves.append( mvFossilTimeSlideUniform(fbd_tree, origin_time, tip="Indarctos_vireti", weight=5.0) )
+moves.append( mvFossilTimeSlideUniform(fbd_tree, origin_time, tip="Kretzoiarctos_beatrix", weight=5.0) )
+moves.append( mvFossilTimeSlideUniform(fbd_tree, origin_time, tip="Zaragocyon_daamsi", weight=5.0) )
```
davidcerny: This is perfectly fine as-is. If you wanted to be a little more concise (and perhaps prevent the possibility of accidentally forgetting to place a move on a certain taxon's age), you could also do something like this:

```r
nonsitelinked_fossils = fossils
nonsitelinked_fossils.erase( imaginary_fossil_site )

for(i in 1:nonsitelinked_fossils.size())
{
    moves.append( mvFossilTimeSlideUniform(fbd_tree, origin_time,
                                           tip=nonsitelinked_fossils[i],
                                           weight=5.0) )
}
```

tests/test_FBD/scripts/mvFossilSiteTimeSlideUniform.Rev
```diff
@@ -0,0 +1,198 @@
...
+# Monitor the number of sampled ancestors in the FBD-tree #
+num_samp_anc := fbd_tree.numSampledAncestors();
+
+                    
+# Monitor the age of the extant bears #
+
+# Monitor the ages of the two jointly moved fossils
+age_Indarctos_arctoides := tmrca(fbd_tree, clade("Indarctos_arctoides"))
+age_Indarctos_punjabiensis := tmrca(fbd_tree, clade("Indarctos_punjabiensis"))
+within_site_age_diff := age_Indarctos_arctoides - age_Indarctos_punjabiensis
```
davidcerny, line 137: This is really smart – well done!

src/core/moves/proposal/tree/FossilSiteTimeSlideUniformProposal.cpp
```diff
@@ -0,0 +1,376 @@
...
+ *
+ * A Uniform-simplex proposal randomly changes some values of a simplex, although the other values
+ * change too because of the renormalization.
+ * First, some random indices are drawn. Then, the proposal draws a new somplex
+ *   u ~ Uniform(val[index] * alpha)
+ * where alpha is the tuning parameter.The new value is set to u.
+ * The simplex is then renormalized.
+ *
+ * \return The hastings ratio.
+ */
```
davidcerny: Not a big deal by any means, but this comment has nothing to do with what the function actually does... let's just delete it.

src/core/moves/proposal/tree/FossilSiteTimeSlideUniformProposal.h
```diff
@@ -0,0 +1,79 @@
...
+     * This node-age proposal is a Uniform-sliding proposal on rooted subtrees without changing the topology.
+     * That is, we pick a random fossil node.
+     * Then, we pick an age between the parent and the present drawn from a Uniform distribution centered around the current age.
+     *
+     *
+     * @copyright Copyright 2009-
+     * @author The RevBayes Development Core Team (Sebastian Hoehna)
+     * @since 2012-07-12, version 1.0
+     *
+     */
```
davidcerny, line 25: Again, this is very nitpicky, but since the comment has been copy-pasted here from another proposal, it doesn't accurately describe _this_ proposal. I'd suggest either rewriting it or just deleting it altogether.

src/revlanguage/moves/tree/Move_FossilSiteTimeSlideUniform.cpp
```diff
@@ -0,0 +1,234 @@
...
+    static MemberRules move_member_rules;
+    static bool rules_set = false;
+    
+    if ( rules_set == false )
+    {
+        
+        move_member_rules.push_back( new ArgumentRule( "tree", TimeTree::getClassTypeSpec(), "The tree on which this move operates.", ArgumentRule::BY_REFERENCE, ArgumentRule::STOCHASTIC ) );
+        move_member_rules.push_back( new ArgumentRule( "origin", RealPos::getClassTypeSpec() , "The variable for the origin of the process giving a maximum age.", ArgumentRule::BY_REFERENCE, ArgumentRule::ANY, NULL) );
+        move_member_rules.push_back( new ArgumentRule( "max", RealPos::getClassTypeSpec() , "The variable for the maximum age of this fossil tip.", ArgumentRule::BY_REFERENCE, ArgumentRule::ANY, NULL) );
+        move_member_rules.push_back( new ArgumentRule( "min", RealPos::getClassTypeSpec() , "The variable for the minimun age of this fossil tip.", ArgumentRule::BY_REFERENCE, ArgumentRule::ANY, NULL) );
```
davidcerny: Maybe "of this fossil site" rather than "of this fossil tip"?

src/revlanguage/moves/tree/Move_FossilSiteTimeSlideUniform.h
```diff
@@ -0,0 +1,56 @@
...
+#include <string>
+
+namespace RevLanguage {
+    
+    
+    /**
+     * @brief Rev wrapper class for the FossilTipTimeSlideUniform move.
+     *
+     * This class is the Rev wrapper class for the FossilTipTimeSlideUniform move,
+     * a move that proposes a node age.
```
davidcerny, line 17: `FossilSiteTimeSlideUniform`, not `FossilTipTimeSlideUniform`

davidcerny (commented):
This looks really good now, thank you so much for your continued work on this! My only remaining request would be to include an additional file in the `output_expected` folder. Ideally, the expected output of a test should include not just the files written by the monitors, but also the standard output and standard error, redirected to the same file. You can generate this file as follows:

1. Remove the `mnScreen` monitor from the test script (since its output is not reproducible);
2. Run the test on your local machine as follows:

```sh
cd tests
../projects/cmake/build/rb ./test_FBD/scripts/bears_mvFossilSiteTimeSlideUniform.Rev &> ./test_FBD/output_expected/bears_mvFossilSiteTimeSlideUniform.errout
```

src/core/moves/proposal/tree/FossilSiteTimeSlideUniformProposal.cpp
davidcerny: I think I'm responsible for the confusion, apologies. 😬 At one point I was trying to enforce the convention of using `delta` as the name of the tuning parameter for sliding moves and `lambda` as the name of the tuning parameter for scaling moves, but I never really went all the way with it.

tests/test_FBD/scripts/FBD_mvFossilSiteTimeSlideUniform.Rev
mcranium: Thanks, that is a brilliant idea!

src/core/moves/proposal/tree/FossilSiteTimeSlideUniformProposal.h
mcranium: I adjusted this. If the wording isn't spot on, feel free to make changes directly.

davidcerny: Looks great to me!

davidcerny (changes requested):
Thank you for taking the time to address all the feedback!

@mcranium and I have been in touch via email, and an interesting question came up during our discussion: is the newly added move merely necessary, or actually sufficient to ensure that the ages of the site-linked taxa will remain identical throughout the analysis? As far as I can tell, this depends on the full move mix. If `mvFossilSiteSlideUniform` is the _only_ move used to update the ages of such taxa, then they will indeed stay the same. However, one can imagine setups where this is not the case:

- If `mvCollapseExpandFossilBranch` is used, it can tie the age of a site-linked fossil to an internal node, which can then get moved by `mvNodeTimeSlideUniform`;
- Tip age moves such as `mvFossilTipTimeUniform` can also be used without the `tip` argument, in which case they simply randomly pick a fossil every time they are performed. As a result, they might update one site-linked fossil but not the others.

In these cases, the ages of the site-linked taxa can still diverge as more samples are collected. The only truly robust solution would be not to treat the ages of the site-linked fossils as independent stochastic variables in the first place, but I'm not sure that's something we can currently do in RevBayes – as pointed out by @bredelings, it might require the ability to clamp to a deterministic node.

I think this caveat is important enough that it should be mentioned in the documentation. My final request for this PR is therefore to provide a short help file for the move. The [`mvFNPR`](https://github.com/revbayes/revbayes/blob/development/help/md/mvFNPR.md) or [`mvRateAgeBetaShift`](https://github.com/revbayes/revbayes/blob/development/help/md/mvRateAgeBetaShift.md) help files might serve as a good template.

mcranium:
Well, this was a very stupid mistake. Everything actually worked as it should have. The only mistake I made was to not consider that the summary tree functions would not care about the within-site age constraints, unless `conditionalAges=TRUE`. Intuitively, I thought at least the MAP tree would keep the constraints, which is why I was looking elsewhere for explanations. Thank you all for your help and patience!

davidcerny (approved):
I may have been a little unclear when I requested that a help page be added in my last review – I wasn't referring to the header file in the Core; I meant that a new markdown file should be added to `help/md`. I have now taken the liberty of doing it myself. With that, I think the PR is complete, so I'm approving.

> Well, this was a very stupid mistake. Everything actually worked as it should have. The only mistake I made was to not consider that the summary tree functions would not care about the within-site age constraints, unless `conditionalAges=TRUE`. Intuitively, I thought at least the MAP tree would keep the constraints, which is why I was looking elsewhere for explanations. Thank you all for your help and patience!

I don't think that was a stupid mistake at all. It can happen when an analysis includes the `mvCollapseExpandFossilBranch`, as I noticed in my previous comment. However, I was wrong about the mechanism by which the problem arises, and your observation about summary trees provides a welcome correction. The problem is not that `mvCollapseExpandFossilBranch` could (either on its own or in conjunction with `mvNodeTimeSlideUniform`) move the age of one site-linked fossil independently of the others. Even if you include this move, the ages of all site-linked fossils are still guaranteed to be exactly identical in every single individual MCMC sample. However, what can differ is their status as either a tip or a sampled ancestor, so when the time comes to summarize the posterior sample, their ages may be averaged over different subsets of samples – and thus seemingly diverge.

> Thank you all for your help and patience!

Thank _you_!

PR #1022, opened 2026-05-13
Allow MCMC to discover non-zero probability states using transition kernels

bredelings:
This PR adds the following new features:
*  **ADDS** a new type LogDensity to hold log-densities that counts the number of -inf, +inf, and nan.  The -inf count includes the number of violated constraints.
* **MODIFIES** Metropolis-Hastings to accept a new move if we decrease the number of nans, +inf, or -inf, in that order.
* **ADDS** a new argument `nsearch` to the `mcmc( )` function.  If nsearch==0 then nothing changes, but if nsearch>0 then it wlll run MCMC during initialization to try and find a finite value for the log probability.
  * It gives up if there is no improvement for nsearch iterations.
  * Currently it does this *after* all the ntries resamples from the prior have failed to yield a finite probability.

Additionally, in order to make the code actually succeed in solving constraints, it also includes a lot of work on bug fixes.  The PR
* **FIXES** a number of MCMC moves not to propose impossible states, such a node with 2 sampled-ancestor children.
* **FIXES** the code to redraw initial states not to break MCMC moves that follow.
* **MODIFIES** the exp(Q*t) code to handle very small negative values like -1.0e-15 by changing them to +1.0e-15, not zero.  This seems to help with the first test case, but otherwise affects only a single output file among all tests.
* **ADDS** a debugging option `logZeroPr`.  With `-o logZeroPr=1`, some messages are printed about why a zero probability termwas created. With `-o logZeroPr=2` the _values_ of parameters affecting the PDF are printed above a PDF with zeros in it.

## Tests

The PR adds 4 tests.  The PR can successfully find initial conditions on all 4 tests some of the time, but the success rate varies.
* initialization_test 1: success rate = 5/10 for 2 chains with 100 search iterations
    * why is the probability of the observed data zero?  possibly too many branch rates are very small, like 1.0e-88
    * someone (else) could look further...
* initialization_test 2: success rate = 6/10 for 10 chains with 200 search iterations
    * the 4 times we failed was we got an exception like: `Specified clade age 153.859 lower than max tip age`
* initialization_test 3: success rate = 10/10 for 1 chain with 150 search iterations
* initialization_test 4: success rate = > 3/10 for 2 chains with 2000 search iterations
  * **to be added**. 

## History

I tried rebasing the previous PR, but I was having trouble.  So this version of the PR was implemented by adding a slightly improved version of LogDensity.h, and changing function signatures to return LogDensity.  A lot of the hard work is here.  If this part is not merged, then I will need to redo this PR from scratch when I next look at it.

 After I got that to compile, I then started bring over individual bits of the previous PR and checking that they passed tests.  So I generally brought over the old code in pieces, but added in a different order, and possibly improved.
 
After that, I did some debugging work to (i) solve probability calculation errors and (ii) improve the constraint search.

## Documentation

Updating the code on writing a density to return LogDensity should not be too hard.  Probably the only thing that needs to change is that the return type is updated, since LogDensity can be constructed from a `double`.  Most user densities do not need to handle more than one -Inf.  

However, the advanced instructions are: if there are multiple constraints that can be violated, then instead of writing `if (failure condition) return -Inf`, the function should ideally do something like `if (failure condition) logProb += -Inf`.  Then it is possible to distinguish two constraint violations from one constraint violation. 

## Future Work

This PR introduces the framework and some working test cases.  However, once the framework is in place, further work can be done (not only by me) to ensure that more probability densities take advantage of the framework.

For example: a number of probability densities _return_ -Inf or Nan in multiple places.  Instead, the densities should _add_ these values, so that violation of multiple  constraints looks worse than violating one constraint.  Making these functions return LogDensity (in cases where they do not already) is relatively easy.  However, determining when to added -Inf requires knowledge of the code that I sometimes don't have.  It is not always clear how these densities should continue after an error if they do not return immediately at the first error.

Other future work will involve:
* improving logging so that it is easier for (very) advanced users to debug why a full solution is not being found
* improving MCMC transition kernels -- for example making the SPR move less constrained -- which should improve constraint search and also regular MCMC
* experimenting with new test cases to see if/when/why the constraint solving works or doesn't work

Another question is: should the constraint solving be activated by default?  After 1000 initialization attempts?

## ???
* I might be able to split out the matching distance code... but I'm not sure.  I'll have to check how good it is.  I was trying to make the hardest test case work, and still not quite succeeding.  This kind of seems beyond the scope of the pr.


davidcerny:
Thank you for all your work on this, Ben! I'm definitely planning to take a more thorough look at the code, but for now, just a few quick suggestions / comments:

1. I see that both the total number of commits and the individual commit messages differ from the original branch / PR (`LogDensity`, #835), so this is not a simple case of rebasing the old branch on development. Could you please briefly describe the differences between the two implementations?
2. This not only changes the existing code base, but – if the system is to keep working as intended going forward – it also requires that new distributions and moves be implemented in a particular way: the `computeLnProbability()` method of the former and the `doProposal()` method of the latter both need to have `LogDensity` as their return type. That's an important piece of info that we might need to add to our [developer's guide](https://revbayes.github.io/developer/implementation).
3. If I remember correctly (and this was for the old branch, not the current one), your system managed to find viable initial states for 2 out of 3 scripts that I had sent you, all of which I had previously found impossible to run without providing a handcrafted starting tree. Do you think it might be worth adding these as tests? That way, we can make sure that potential future changes to the system will not render it less powerful.
4. The function for computing matching distances seems like a very useful addition, to the extent that I wonder if it would make sense to give it a Rev interface and expose it to the user.

davidcerny:
Seeing the new tests demonstrating that the `LogDensity` system allows RevBayes to find viable initial states for clock analyses with monophyly ([`initialization_test_2.Rev`](https://github.com/revbayes/revbayes/blob/LogDensity2/tests/test_zero_prob/scripts/initialization_test_2.Rev)) and backbone ([`initialization_test_3.Rev`](https://github.com/revbayes/revbayes/blob/LogDensity2/tests/test_zero_prob/scripts/initialization_test_3.Rev)) constraints, it occurred to me to try and see if it could do the same for backbone-constrained nonclock analyses, where the need for user-crafted starting trees has been a long-standing issue (#384). Apparently, it could not. This doesn't detract from its usefulness in other cases, obviously – it just means that backbone trees are broken on a bit more fundamental level.

davidcerny:
It looks like the Windows tests keep timing out – any idea why that might be the case, @bredelings?

bredelings:
I noticed this, but I don't know why yet.  It looks like it might specifically be windows-based tests that use tensorphylo?

I have a guess that maybe the change with makes Stochastic and Deterministic nodes start out with `touched=false` could be the cause... but I don't know why.  In order to check this I need to build tensorphylo for windows and run with it under wine, which I haven't had time to do yet...

PR #1026, opened 2026-05-13
Remove pointer arithmetic

bredelings:
This PR converts for loops to use normal array addressing (i.e. `x[i] * y[i]`) instead of modifying the pointers (i.e. `(*x) * (*y) ; x++ ; y++`).  Its possible that pointer modification was necessary to generate good code in the 1990s, but it just makes code harder to read and modify.

This pass doesn't try to do other cleanups at the same time, which makes it simpler.

Further improvements to PhyloCTMC will start from this branch.


mikeryanmay (approved):
A simple cleanup but with big implications. The code will be a bit easier to maintain, but more importantly we'll be able to experiment with data types that support array indexing but not pointer arithmetic (eg Eigen vector/matrix types), enabling some potential improvements (eg better matrix algebra routines). 

All the tests pass, and I played around with some runs on my own and am convinced everything is working properly. I vote ready to merge!

PR #1030, opened 2026-05-16
More fun with badges

davidcerny:
This PR

- Splits the commit-counting badge in two, so that instead of counting commits between the penultimate release and the current tip of development, we count them separately (1) between the penultimate release and the latest release, and (2) between the latest release and the current tip of development.
- Adds an item to the [release checklist](https://github.com/revbayes/revbayes/blob/development/doc/release-workflow.md) to remind us to update the links underlying these badges with every new release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,16 +3,19 @@
 [![RevBayes v1.0 releases](https://img.shields.io/github/downloads/revbayes/revbayes.archive/total.svg?style=social&logo=github&label=Downloads:%20v1.0)](https://github.com/revbayes/revbayes.archive/releases)
 [![RevBayes later releases](https://img.shields.io/github/downloads/revbayes/revbayes/total.svg?style=social&logo=github&label=Downloads:%20v1.1%20and%20later)](https://github.com/revbayes/revbayes/releases)
 
-[![Release](https://img.shields.io/github/v/release/revbayes/revbayes.svg?style=plastic)](https://github.com/revbayes/revbayes/releases/tag/v1.4.0)
-[![Release date](https://img.shields.io/github/release-date/revbayes/revbayes.svg?style=plastic&color=orange)](https://github.com/revbayes/revbayes/releases/tag/v1.4.0)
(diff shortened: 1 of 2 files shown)


bredelings:
I asked chatgpt if it could suggest some improvements.  It says we can compute the latest release so that we don't have to manually update it.  I then played around this this a bit, because it is a rapid way to see what the badges would look like.

[RevBayes v1.0 releases](https://github.com/revbayes/revbayes.archive/releases) [RevBayes later releases](https://github.com/revbayes/revbayes/releases)
[Development branch](https://github.com/revbayes/revbayes/tree/development)  [New commits](https://github.com/revbayes/revbayes/commits/development) [Build status](https://github.com/revbayes/revbayes/actions/workflows/build.yml)
[Latest release](https://github.com/revbayes/revbayes/releases/latest) [New commits](https://github.com/revbayes/revbayes/compare/v1.3.2...v1.4.0) [Release date](https://github.com/revbayes/revbayes/releases/latest)  [License](https://github.com/revbayes/revbayes/blob/development/LICENSE) [Code of conduct](https://github.com/revbayes/revbayes/blob/development/CODE_OF_CONDUCT.md) [Language](https://en.cppreference.com)

Note that only 2 badges now mention a version number.

bredelings:
OK, after playing around with the badge possibilities, my only real suggestions are:
* to have a separate line that says "development" somewhere, and mentions the number of commits that are on development
* maybe use the URLs from ChatGPT that auto-discover what the previous release is called.  Sadly we can't do that for every badge without using some scripting.

The color and style changes are just "bike-shedding" from ChatGPT.  (https://en.wikipedia.org/wiki/Law_of_triviality)

davidcerny:
I know, right? The stakes couldn't possibly be any lower. 😛 But hey, it's going to be the best-looking bike shed of all time!

On a more serious note, I think I have now addressed both of your suggestions.

bredelings (approved):

PR #1032, opened 2026-05-19
fixing BDSTP when rho=0 and samples are at present

hoehna:
This pull request fixes three issues:

1. There was an issue in our BDSTP implementation that when rho=0 but there are samples at the present, the likelihood was non-zero, but obviously it should be!
2. Our approach for offsetting a tree didn't work when the tree contained sampled ancestors.
3. The BDSTP was shifting the times of the tree so that the youngest tip has age zero. This is mathematically incorrect as as the probability of extinction (and probability of survival) depends on how much time is left until the present (time zero).


src/core/distributions/phylogenetics/tree/birthdeath/BirthDeathSamplingTreatmentProcess.cpp
```diff
@@ -411,6 +411,18 @@ double BirthDeathSamplingTreatmentProcess::computeLnProbabilityTimes( void ) con
...
+        // make sure that the sampling probability at the present was > 0 if there are samples at the present
+        else if ( i == 0 && global_timeline[0] < DBL_EPSILON )
+        {
+            // get the number of samples at present
+            int T_i = int(event_tip_ages[0].size());
+            
+            if ( T_i > 0 )
+            {
+                return RbConstants::Double::neginf;
+            }
```
davidcerny, line 420: Is it worth adding a warning before we return?

bredelings:
Hmm... there is also this code in `countAllNodes( )`:

```
else if ( n.isTip() && !n.isFossil() )
{
    if (phi_event[0] >= DBL_EPSILON)
    {
        num_extant_taxa++;
        event_tip_ages[0].push_back(0.0);
    }
    else
    {
        serial_tip_ages.push_back(0.0);
    }
}
```

Does this put the extant tips into `serial_tip_ages` when the sampling probability at the present is zero?  If so, then the above code might not register them.

src/core/distributions/phylogenetics/tree/birthdeath/BirthDeathSamplingTreatmentProcess.cpp
bredelings: What is the use case?

hoehna: 1. I don't think a warning would help if the actual probability should be zero. We are also not checking if you have some variable sites but put your proportion of variable sites to 0.0 in the PhyloCTMC. I do understand your logic, and that it would be nice for the user, I just wonder if we always should report reason why a probability is actually -Inf or NaN?
2. The use case is if we have a phylogeny with fossils, but also at least on extant tip and the user setup an FBD model with the extant-sampling probability being zero. In my view, this must mathematically give a zero probability of observing such a phylogeny (you cannot sample an extant tip if the sampling probability is zero!).

hoehna:
> Hmm... there is also this code in `countAllNodes( )`:
> 
> ```
> else if ( n.isTip() && !n.isFossil() )
> {
>     if (phi_event[0] >= DBL_EPSILON)
>     {
>         num_extant_taxa++;
>         event_tip_ages[0].push_back(0.0);
>     }
>     else
>     {
>         serial_tip_ages.push_back(0.0);
>     }
> }
> ```
> 
> Does this put the extant tips into `serial_tip_ages` when the sampling probability at the present is zero? If so, then the above code might not register them.

The problem with the `serial_tip_ages` didn't occur in my tests. Might be a different issue if at all.

src/core/distributions/phylogenetics/tree/birthdeath/BirthDeathSamplingTreatmentProcess.cpp
davidcerny: Yeah, that's a good point. A warning message could in theory prevent users from trying to set up models that are impossible given their own data, but it would be a bit arbitrary to throw one here if we are not doing it across the board.

davidcerny (approved):
As expected, the changes made on this branch affect the output of FBD analyses of entirely extinct clades (i.e., analyses of datasets where the age range of the youngest taxon doesn't extend to 0). This can be seen with the [`FBD_no_degree_3_nodes.Rev`](https://github.com/revbayes/revbayes/blob/development/tests/test_FBD/scripts/FBD_no_degree_3_nodes.Rev) and [`mvFossilTimeSlideUniform_multiple_runs.Rev`](https://github.com/revbayes/revbayes/blob/development/tests/test_FBD/scripts/mvFossilTimeSlideUniform_multiple_runs.Rev) integration tests, and I can also see it in my analysis of an unpublished plesiosaur dataset: 

Current tip of `development` (7fed5b3):

```
Iter        |      Posterior   |     Likelihood   |          Prior   |       root_age   |    elapsed   |        ETA   |
0           |       -14162.9   |       -13220.8   |       -942.176   |          246.8   |   00:00:00   |   --:--:--   |
50          |       -9625.43   |       -9017.09   |       -608.344   |       250.4301   |   00:08:05   |   --:--:--   |
100         |       -9452.59   |       -8843.72   |       -608.871   |       251.4718   |   00:14:50   |   1631:25:10   |
Attempting to set new age of taxon Pistosaurus_longaevus incompatible with age range
150         |       -9371.44   |       -8775.17   |        -596.27   |       248.6557   |   00:30:27   |   2232:29:33   |
200         |        -9357.1   |       -8759.19   |       -597.912   |       249.8668   |   00:37:20   |   2052:42:40   |
Attempting to set new age of taxon Pistosaurus_longaevus incompatible with age range
250         |       -9315.14   |        -8712.2   |       -602.942   |       246.9552   |   00:44:09   |   1941:51:51   |
300         |       -9262.98   |       -8687.47   |        -575.51   |       247.3244   |   00:51:20   |   1881:22:00   |
350         |       -9313.12   |        -8703.2   |       -609.912   |       246.0237   |   00:58:37   |   1841:15:40   |
Attempting to set new age of taxon Pistosaurus_longaevus incompatible with age range
400         |       -9279.34   |       -8683.38   |       -595.968   |       245.6393   |   01:06:06   |   1816:38:54   |
450         |          -9273   |       -8675.71   |       -597.295   |       244.3114   |   01:13:42   |   1800:19:37   |
500         |       -9254.59   |       -8696.93   |       -557.669   |       245.5589   |   01:21:25   |   1789:48:35   |
```

Versus `fix_BDSTP` (fe9bc46):

```
Iter        |      Posterior   |     Likelihood   |          Prior   |       root_age   |    elapsed   |        ETA   |
0           |       -25914.1   |       -25943.5   |        29.4221   |          246.8   |   00:00:00   |   --:--:--   |
50          |         -10119   |       -9294.64   |       -824.351   |       249.3826   |   00:06:16   |   --:--:--   |
100         |        -9962.9   |        -9171.2   |         -791.7   |       247.9178   |   00:12:21   |   1358:17:39   |
150         |       -9937.69   |       -9137.46   |       -800.225   |       250.1205   |   00:20:50   |   1527:25:50   |
200         |       -9894.65   |       -9091.62   |       -803.031   |       248.1712   |   01:00:03   |   3301:44:57   |
250         |       -9889.46   |       -9059.19   |       -830.263   |       248.2932   |   01:05:41   |   2888:58:19   |
300         |        -9898.1   |       -9015.99   |       -882.111   |       250.2705   |   01:11:14   |   2610:42:06   |
Attempting to set new age of taxon Pistosaurus_longaevus incompatible with age range
Attempting to set new age of taxon Pistosaurus_longaevus incompatible with age range
350         |       -9822.66   |        -8979.9   |       -842.768   |       247.0566   |   01:16:50   |   2413:28:52   |
400         |       -9768.93   |       -8948.46   |        -820.47   |       247.8675   |   01:22:21   |   2263:15:09   |
Attempting to set new age of taxon Augustasaurus_hagdorni incompatible with age range
450         |       -9732.33   |       -8899.08   |       -833.256   |       248.0902   |   01:27:51   |   2145:58:48   |
500         |       -9701.64   |       -8896.95   |       -804.691   |       246.5329   |   01:33:19   |   2051:24:41   |
```

Getting rid of `offset` and `getOffset()` might seem like a radical solution, and since code is usually put in place for a reason, I understand that there might be some apprehension about this. That said, I find the reasoning behind the removal convincing, and since the removed code was accompanied by comments indicating it was of doubtful validity anyway (`// @TODO: @ANDY: Double check the offset works`), this change doesn't seem unreasonable to me.

@bredelings I'd be curious about your thoughts on this!

bredelings:
> The problem with the serial_tip_ages didn't occur in my tests. Might be a different issue if at all.

It turns out we already have a test in the tree that shows that it is an issue: `test_initialTree/bdstp_with_initialTree.rev`. This test confirms that in fact the likelihood is still not zero with extant tips and Phi=0.

bredelings (changes requested):
The code that checks for extant tips seems not to work in its current form.

bredelings:
@davidcerny My non-exhaustive investigation seems to indicate that the offset code didn't really make sense and should indeed be removed.

bredelings (approved):
OK, it looks like the bug I found is fixed now.

PR #1033, opened 2026-05-20
Add flake.nix

sriram98v:
## Summary

This PR adds a `flake.nix` to support building RevBayes and entering a development environment using the [Nix package manager](https://nixos.org/).

**What works:**
- `nix build` — builds the `rb` binary as a Nix package
- `nix develop` — enters a shell with all build dependencies available (`cmake`, `ninja`, `boost`, `zlib`), so you can run the existing `build.sh` workflow manually

**Tested on:** Linux x86_64 (NixOS, `nixos-unstable`)

**Intended use:** primarily a developer convenience shell; the package output (`nix build`) is a bonus for reproducible builds but is not intended as an official distribution mechanism.

**On the hard-coded `1.4.0` version:** the Nix sandbox has no `git` access, so `GitVersion.cpp` must be generated manually in `preConfigure`. The version string is set to match the last release; it can be bumped at release time alongside the rest of the version bookkeeping.

---

_Apologies for not following the contribution guidelines on the RevBayes developer page (branching off `development`, keeping the branch up to date, and including a PR description) — I had not seen them before submitting. I will follow the correct workflow for any future contributions._


basantakhakurel:
Thanks for the contribution. 
Could you please add a PR description explaining what this change is intended to provide?

From the diff, it looks like this adds Nix flake support for building RevBayes and entering a development shell, but the PR body is empty, so it is hard to review the intended workflow and scope. In particular, it would help to know:

- which commands are expected to work, e.g., `nix build`, `nix develop`, or both;
- which platforms/systems this was tested on;
- whether this is intended as a user-facing package, a developer convenience shell, or both;
- whether the hard-coded `1.4.0` version is intentional;

Also, this branch appears to be behind the current tip of `development`, so please update/rebase it before merging.

davidcerny:
Hi Sriram, welcome to RevBayes! And thanks for adding a cover letter – that's very helpful.

> The version string is set to match the last release; it can be bumped at release time alongside the rest of the version bookkeeping.

If that's the case, you might want to add a line about this to our existing [release checklist](https://github.com/revbayes/revbayes/blob/development/doc/release-workflow.md).

sriram98v:
Thanks\! I've added flake.nix to the version bump step in doc/release-workflow.md (alongside meson.build and RbVersion.cpp).

davidcerny (approved):

PR #1035, opened 2026-05-21
Add a `rep()` overload for `Probability`

### LINKED ISSUE #1034: [BUG] `rep()` missing an overload for `Probability`
**Describe the bug**
The `rep()` function can be used to create vectors of the same type as the first one of its arguments:

```r
x <- 1            # type(x) returns Natural
a <- rep(1, 3)    # type(a) returns Natural[]
x <- -3           # type(x) returns Integer
a <- rep(-3, 3)   # type(a) returns Integer[]
x <- 4.7          # type(x) returns RealPos
a <- rep(4.7, 3)  # type(a) returns RealPos[]
x <- -2.3         # type(x) returns Real
a <- rep(-2.3, 3) # type(a) returns Real[]
```

However, for `Probability`, this doesn't work:

```
> x <- 0.8
> type(x)
   Probability
> a <- rep(0.8, 3)
> type(a)
   RealPos[]
```

**To reproduce**
See above.

**Expected behavior**
We should be able to go from `Probability` to `Probability[]` as well.

**Computer info**
macOS 26.3 (Tahoe), ARM64 architecture, RevBayes v1.4.0-dev, current tip of development (c7b2b34).

davidcerny:
Fixes #1034.

### DIFF
--- a/src/revlanguage/workspace/RbRegister_Basic.cpp
+++ b/src/revlanguage/workspace/RbRegister_Basic.cpp
@@ -340,6 +340,7 @@ void RevLanguage::Workspace::initializeBasicGlobalWorkspace(void)
         addFunction( new Func_replicate<Integer>()          );
         addFunction( new Func_replicate<Real>()             );
         addFunction( new Func_replicate<Natural>()          );
+        addFunction( new Func_replicate<Probability>()      );
         addFunction( new Func_replicate<RealPos>()          );
         addFunction( new Func_replicate<RlString>()         );
         addFunction( new Func_replicate<RlBoolean>()        );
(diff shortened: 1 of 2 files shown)


basantakhakurel (approved):
LGTM, thanks for adding this. 
The change just registers the existing `rep()` implementation for `Probability`, so the behavior is consistent with other types.
I checked the likely edge cases with downstream conversions and don't see anything concerning. 
Particularly, a `Probability`-typed value now produces a `Probability[]`, and that vector can still be used in contexts like `simplex()`. 
A small thing to consider might be adding a test to lock in the new overload, but I don't think that needs to block approval, as the current tests are passing.

PR #1036, opened 2026-05-22
Order of state labels for getStateDescriptions

mikeryanmay:
It seems that commit f27f13f18 introduced a small bug that the caused character data method getStateDescriptions() to return state labels in lexicographic order. This broke the relationship between state labels for binary area occupancy and range labels for biogeographic models when they were recoded with `formatDiscreteCharacterData`.

The bug seems to arise from using a std::set to track unique state labels, but this data type is lexiographically ordered by default. The set was being used to check for unique labels, which makes sense, so I maintained that as a method for tracking unique labels. I added a std::vector to keep track of new state labels (that don't appear in the standard set) so that the original relationship between binary areas and ranges is preserved.

EDIT: I should note this bug only affects plotting code, as far as I can tell. Internal DEC-type recodings still recognize the correct map between binary and range state labels.


bredelings:
This looks pretty reasonable to me.  Can we add a test so this kind of thing doesn't regress in the future?

basantakhakurel:
I have modified the writeNexus test to consider the `formatDiscreteCharacterData` function as well. Does that seem sufficient? 
I guess in the future we should have a separate test for Discrete Character Data manipulation within RevBayes.

bredelings:
yeah, that seems good!

bredelings (approved):

PR #1038, opened 2026-05-25
Add pull request template

bredelings:
This PR adds a PR template.  This attempts to clarify what a PR cover letter should include, and also adds some limitations about what we require for AI-assisted contributions.

We actually had old PR templates that didn't work.  Unlike issue templates, github doesn't have a UI for selecting what *type* of PR you are opening, so the multiple templates were not used.

I like this template, but I also see it as a starting point for further evolution.  Our issue template took a bit of time to get to where it is now.

I accidentally left in the statement about closing the PR if people delete the checklist.  I would not actually close a PR by somebody that I know without this checklist.  However, for people who have never interacted with the project before, its probably best that they complete the checklist.   So, maybe we should leave that line in.  Not as a statement of actual project policy, but as an encouragement to newcomers not to ignore the checklist.


basantakhakurel (commented):
LGTM 
Do we really want emojis there in the template?

bredelings:
To me emojis are primarily useful as ways to quickly identify something based on color/shape instead of black text.  So, for the usage here, the reviewers eyes can quickly see the PR categories.  But its a small win, at best.  I mostly copied it from another PR template.

Shall I take them out?  I don't have strong feelings.

basantakhakurel:
Fair, they do look nice. I don't mind them either, just wanted to point out, given the discussion in Slack.

basantakhakurel (approved):

.github/pull_request_template.md
```diff
@@ -0,0 +1,40 @@
...
+<!-- How can we be sure that this change works? What tests have you added to ensure that it won't break in the future? -->
+
+## Additional context
+<!-- Add any other context about the problem here. -->
+
+## Relevant issues
+<!-- e.g. "Fixes #123 -->
+
+## TODO
+<!-- Things that should be addressed before merging -->
```
afmagee, line 28:
```suggestion
<!-- Things that should be addressed before merging. If checked, consider marking this a Draft PR. -->
```

Maybe more than is needed

.github/pull_request_template.md
```diff
@@ -0,0 +1,40 @@
...
+<!-- HOW does this PR attempt to solve this problem? -->
+
+## Design reasoning
+<!-- WHY do you attempt to solve it this way? WHAT other strategies did you consider, and WHY did you decide to pursue this strategy instead of them? -->
+
+## Testing
+<!-- How can we be sure that this change works? What tests have you added to ensure that it won't break in the future? -->
+
+## Additional context
+<!-- Add any other context about the problem here. -->
```
afmagee, line 22: Is this section mandatory or optional?

.github/pull_request_template.md
```diff
@@ -0,0 +1,40 @@
...
+<!-- WHAT problem does this PR solve? -->
+
+## Strategy
+<!-- HOW does this PR attempt to solve this problem? -->
+
+## Design reasoning
+<!-- WHY do you attempt to solve it this way? WHAT other strategies did you consider, and WHY did you decide to pursue this strategy instead of them? -->
+
+## Testing
+<!-- How can we be sure that this change works? What tests have you added to ensure that it won't break in the future? -->
```
afmagee, line 19:
```suggestion
<!-- How can we be sure that this change works? What tests have you added to ensure that it won't break in the future? For more on testing in RevBayes, see: https://revbayes.github.io/developer/automated_builds/ and https://revbayes.github.io/developer/tests/-->
```

Maybe overkill, but if we're aiming for newbie friendliness, it might help?

.github/pull_request_template.md
```diff
@@ -0,0 +1,40 @@
...
+<!-- If this checklist is deleted from the PR submission it will be immediately closed -->
+- [ ] I understand the code I am submitting.
+- [ ] I have added tests that prove my fix/feature works
+- [ ] I have run this code locally and verified it fixes the issue.
+- [ ] New and existing tests pass locally
+- [ ] Documentation was updated where necessary
+- [ ] I used an AI/LLM tool to assist with this PR
+  - *If checked, I affirm that:*
+      - *I personally understand every line of code submitted and agree to act as the sole human guarantor of this change.*
+      - *I wrote the cover letter myself, and I will answer reviewer questions in my own words without copy-pasting AI output.*
```
afmagee, line 40: I'm not sure I see the value in writing the entire cover letter one's self. One of the things "AI" tools are _good_ at is summarizing code changes and walking through things. At the extreme, we might only ask that "Context" be 100% human, with the others being human checked, edited, and corrected. But I suppose that makes this complicated and simple is good.

I also think that "I agree to answer questions myself" and what level of human supervision the cover letter gets are separate matters?
```suggestion
      - *I wrote the cover letter myself.*
      -  *I will answer reviewer questions in my own words without copy-pasting AI output.*
```

afmagee (approved):
I think this looks pretty good. I left a few small suggestions, but at the higher level, I think it would be really good if we could point to at least one gold-standard example PR that covers all these. (But I won't let perfect bet the enemy of good, and we could always add that later.)

.github/pull_request_template.md
bredelings: I think optional.

.github/pull_request_template.md
bredelings: Hmm, this is a good point -- for newbies pointing to how tests work could be important.  Unfortunately, those two pages are 5/6 years out of date...

.github/pull_request_template.md
bredelings: Yeah, its true that these are separate points.  

I also kind of agree that SOME AI generated content could be helpful.  However, I also think that if the user can't explain context + strategy + design reasoning in their own words then we would be in trouble.

I initially took this from a requirement for the PR to contain a concise 3-sentence summary with one sentence each about context + strategy/design + checking/tests.  Supposedly AIs are bad at this, especially the concise summary part.

.github/pull_request_template.md
afmagee: Might want to note that?

.github/pull_request_template.md
afmagee: Touché. Perhaps we just point to where the kinds of tests live, and/or relatively comprehensible examples of each?

.github/pull_request_template.md
afmagee: I agree that these are very important things people should be able to do.

I also agree that, in general, these are playing to the AI relative weaknesses. But my limited experience with these tools, and their recent trajectory, makes me think that as written some of these are actually relatively straightforward fodder for top of the line models, and aren't guarantors of quality human supervision.

- Context, especially in a concise manner[^1], is definitely not something I'd expect them to be amazing at, since this _should_ happen prior to the AI tool being deployed. Unless, of course, this is an "AI scanned the toolbase and chose a topic" PR. Even still, for some sorts of PRs, I could see it being relatively reverse-engineerable.
- Strategy and design are iffier.
  - The simple case is trivial. Things like, "I needed a move that did X, no move does X, but _that_ move does something similar, so I copied and modified it" are common enough and not helpful ways to distinguish human from AI thinking. But also probably not places we care so much?
  - The more complex stuff is still not a slam dunk, though. "How does code X accomplish task Y" strikes me as very much in current AI capacities, it's not much of a leap from "what does this code do" which they can be shockingly good at, nor from asking it to walk you through what code does, which I've seen touted as a strength. The behind-the-scenes "why this way, what else did you think about" start getting back to things that should come before the code, but I could see being kind of reverse-engineered. On the other hand, AI code often chooses very dumb or weird ways to do things, so I still think it's potentially useful.
- Checking and tests are something I've found AI tools can be very bad at writing, but also very good at writing. So unless we're asking for the "why" behind the test, I don't know that this guarantees much. And I will also note that I have found testing to be a relatively advanced topic for most backgrounds that come into scientific computing.

[^1]: This incredibly long-winded, bullet-pointed list is entirely human made.

afmagee: TL;DR: If we're thinking purely from a "this is what the human needs to be able to do" perspective, I think this is fine. If we're expecting this to _guarantee_ humans were involved, caveat emptor.

bredelings:
So... I have some other changes I want to make... but this version is approved.

I'm going to merge this, and then submit additional changes.  So some of @afmagee's suggestions will be address by the second PR.

.github/pull_request_template.md
bredelings: Yeah, I think I'm not really trying to institute a CAPTCHA here.  To some extent, if the AI can actually answer all the questions well, then its much less of a drain on developer resources.

The request for a concise summary on the new PR may partially function like a CAPTCHA, but the main value is to make reviewing less of a burden.

davidcerny:
Hmm... RevBayes development can clearly move very fast sometimes. I didn't even get a chance to review this before it got merged. (Obviously, two other people did, so this is on me.)

I do want to go on record as saying that this is not a good idea – i.e., a PR template is not the right tool for the job – and the execution is not great, either.

(1) We have merged over 500 PRs from 43 contributors without the need for a template, so I'm skeptical that there is a real problem this PR addresses.

(2) Ironically, the PR – whose purpose is at least in part to protect reviewers from having to spend their attention on AI slop – bears hallmarks of AI slop itself. There is the typical saccharine cutesiness (a caterpillar emoji for a bug fix, yay!) and redundant verbosity. Do we really need both "Strategy" _and_ "Design reasoning" sections? What purpose could the "New and existing tests pass locally" check box possibly serve? First of all, the existing tests are _not_ passing locally for me or anyone else using a platform other than Linux, because several of them have platform-specific output, and will only pass locally on Linux machines. Second, why should I care if the tests pass locally for the submitter? I want them to past on all the platforms and with all the compilers tested by our build script. GitHub Actions already tell me whether or not that's the case – the check box communicates absolutely no useful information to me as a reviewer.

(3) The biggest problem is enforceability. Will we stop accepting PRs that don't adhere to the template? Then we can say goodbye to some of our most active developers; good luck making them jump through all these hoops. I don't think that's the intention (after all, we don't automatically close issues that don't adhere to the issue template, either), but if we don't do that, then we can't very well close substandard PRs on the grounds that they don't adhere to the template, either.

> I would not actually close a PR by somebody that I know without this checklist. However, for people who have never interacted with the project before, its probably best that they complete the checklist. So, maybe we should leave that line in. Not as a statement of actual project policy, but as an encouragement to newcomers not to ignore the checklist.

Maybe I'm just old-fashioned, but I'm of the opinion that if we are to have rules, they should be enforced either across the board or not at all. Selectively enforced rules are definitionally unfair, and can discourage new contributors. The template would be more honest if it was explicitly labeled as something to be used by first-time contributors. Then again, the PR that motivated the creation of this template was actually the _second_ contribution by the submitter... so maybe instead of "first-time contributors", we really mean "newbies", and we're back to the problem of rules that only apply to those who we choose to apply them to.

PR #1039, opened 2026-05-25
Improve PR template

bredelings:
## PR Type
- 🚦 Infrastructure

## Summary
The main change for the new PR template is to ask for a 1-3 sentence summary at the top explaining what the problem is, how its being solved, and how we know the solution works.  This should hopefully decrease cognitive load on PR reviewers.

I also tried to incorporate some of @afmagee's suggestions -- mention draft PRs, split cover letter point from the respond-in-your-own-words point

There are a variety of other small changes.

It is interesting to note that this PR doesn't really fit the template :man_shrugging:

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -6,35 +6,49 @@
 - 📚 Documentation
 - 🚦 Infrastructure
 
+## Summary
+<!-- In 1-3 sentences:
+1. What problem does this solve? 
+2. What strategy was chosen? 
+3. How was it checked? 
+-->
+
 ## Context 
-<!-- WHAT problem does this PR solve? -->
+<!-- If the problem is not trivial, please describe the problem this PR solves. -->
 
-## Strategy
-<!-- HOW does this PR attempt to solve this problem? -->
+## Approach
+<!-- If the solution is not trivial, please describe
+ * HOW does this PR attempts to solve this problem
+ * WHY you attempt to solve it this way
+ * WHAT other strategies did you consider, and WHY you decided to pursue this strategy instead.
+-->
 
-## Design reasoning
-<!-- WHY do you attempt to solve it this way? WHAT other strategies did you consider, and WHY did you decide to pursue this strategy instead of them? -->
+## More info
+<!-- Optionally, add more explanation or information here. -->
 
 ## Testing
-<!-- How can we be sure that this change works? What tests have you added to ensure that it won't break in the future? -->
-
-## Additional context
-<!-- Add any other context about the problem here. -->
+<!-- Briefly, How can we be sure that this change works? -->
+<!-- Consider: What tests have you added to ensure that it won't break in the future? -->
+<!-- See the tests/ directory for examples of existing tests. -->
 
 ## Relevant issues
 <!-- e.g. "Fixes #123 -->
 
-## TODO
-<!-- Things that should be addressed before merging -->
+## TODO before merging
+<!-- If non-empty, consider marking this a draft PR. -->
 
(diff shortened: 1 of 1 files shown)


afmagee (approved):
Big fan all around, thanks @bredelings!

I especially appreciate the new "I've cleaned up whatever dumb stuff the AI did" check box

afmagee:
Whether we approve of the specifics in any given PR or not, I feel like the whole breaking the mold/thinking outside the box bit of _not_ having all the sections is pretty strong ~proof~ evidence a human got involved in the process.

davidcerny:
I'm copy-pasting my comment from #1038, since I find it equally relevant here:

> I do want to go on record as saying that this is not a good idea – i.e., a PR template is not the right tool for the job – and the execution is not great, either.
> 
> (1) We have merged over 500 PRs from 43 contributors without the need for a template, so I'm skeptical that there is a real problem this PR addresses.
> 
> (2) Ironically, the PR – whose purpose is at least in part to protect reviewers from having to spend their attention on AI slop – bears hallmarks of AI slop itself. There is the typical saccharine cutesiness (a caterpillar emoji for a bug fix, yay!) and redundant verbosity. Do we really need both "Strategy" _and_ "Design reasoning" sections? What purpose could the "New and existing tests pass locally" check box possibly serve? First of all, the existing tests are _not_ passing locally for me or anyone else using a platform other than Linux, because several of them have platform-specific output, and will only pass locally on Linux machines. Second, why should I care if the tests pass locally for the submitter? I want them to past on all the platforms and with all the compilers tested by our build script. GitHub Actions already tell me whether or not that's the case – the check box communicates absolutely no useful information to me as a reviewer.
> 
> (3) The biggest problem is enforceability. Will we stop accepting PRs that don't adhere to the template? Then we can say goodbye to some of our most active developers; good luck making them jump through all these hoops. I don't think that's the intention (after all, we don't automatically close issues that don't adhere to the issue template, either), but if we don't do that, then we can't very well close substandard PRs on the grounds that they don't adhere to the template, either.
> 
> > I would not actually close a PR by somebody that I know without this checklist. However, for people who have never interacted with the project before, its probably best that they complete the checklist. So, maybe we should leave that line in. Not as a statement of actual project policy, but as an encouragement to newcomers not to ignore the checklist.
> 
> Maybe I'm just old-fashioned, but I'm of the opinion that if we are to have rules, they should be enforced either across the board or not at all. Selectively enforced rules are definitionally unfair, and can discourage new contributors. The template would be more honest if it was explicitly labeled as something to be used by first-time contributors. Then again, the PR that motivated the creation of this template was actually the _second_ contribution by the submitter... so maybe instead of "first-time contributors", we really mean "newbies", and we're back to the problem of rules that only apply to those who we choose to apply them to.

bredelings:
I'm going to merge this PR, as I think it is better than the previous template.  The current template is not perfect -- but this is actually because I cobbled it together quickly from pieces of existing templates from other projects, not because it was AI generated.  Hopefully further changes continue to improve it.

> Do we really need both "Strategy" and "Design reasoning" sections?

I think different sections were excessive -- merged here.  But these are very much distinct questions, and this was my idea, not from an AI.

Removing the "Do the tests pass locally?" question could be a win.  I'll leave that for a subsequent PR.

Regarding rules, I think we should treat this template very much the way we treat the issue template -- which is that we don't reject anything for not fitting the template.  I softened the hostile language about PRs being closed if they don't keep the checklist.  But further improvements may be beneficial.

Perhaps I should have written an enraged response... but I hope I did not.

PR #1045, opened 2026-05-28
log function default to base 10

prilau:
Currently, there is no default value for the base when calling `log()`. I have set base 10 as the default.

I modified `test_basics.Rev` to test `log()` function with and without a specified base.


davidcerny (approved):
Looks great – thanks!

ms609 (commented):
* Is the function documentation updated in line with this change?

* Why 10?  My intuition would be to expect `log` to default to a natural log, so as to be symmetric to `exp()`. As I'd have to think twice to work out how to express `log(x, base = e)`, I'd lean towards base $e$ on from a user experience perspective
* How does a base of 10 compare with (i) other software? (R uses natural logs); (ii) the base of likelihoods that are reported elsewhere, e.g. in monitors?  (Using a different base here by default would introduce confusion).

mikeryanmay:
This makes sense to me, especially because `ln` already exists for the natural log. 

Just an aside: Many languages actually implement log10 and log2 with fixed bases to reduce confusion (and I suppose sometimes use different algorithms for these different operations). I wonder if we can do something like that easily using aliases rather than making whole new revlanguage functions?

davidcerny:
@ms609 As pointed out by @mikeryanmay, we have a separate `ln()` function for the natural logarithm, so using $e$ as the default base would render `log()` somewhat redundant.

> I'd have to think twice to work out how to express `log(x, base = e)`

You can just do `log(x, base = exp(1))`, much like you would in R.

Note also that default argument values are included in the documentation automatically, so on this branch, you get:

```
   Arguments
   
          x : A positive number.
              Type:       RealPos, <any>, const reference
       base : The base of the logarithm.
              Type:       RealPos, <any>, const reference
              Default:    10
```

bredelings:
So... it looks like in basically every language except Excel and PostgeSQL, `log(x)` means the natural logarithm.

**natural log**: C, C++, Fortran, Python, R, MATLAB, Julia, Java, JavaScript, Go, Ruby, PHP, Perl, Lua, VBA, SAS, Wolfram Language, SQL Server, MySQL, MariaDB, Haskell

**base 10:** Excel, Google Sheets, PostgreSQL, hand calculators

**rust:**  x.log(base) requires an explicit base; natural log is x.ln() 

Also in math `log(x)` means natural log.  I think this is going to cause a lot of confusion.  The lack of a default base has bit me in the past, but maybe that is better than quietly doing an unexpected operation.

Can we at least wait to merge this until we get more of a consensus here?

bredelings (changes requested):
I think this might cause too much havoc.  We may at some point add the functions log1p and expm1, and then if log means log10 we would be in trouble.

bredelings:
Would people be for/against:
* making `log(x)` default to natural log

And optionally:
* adding `log10(x)` -- apparently called the "common logarithm"
* adding `log2(x)`

davidcerny:
> So... it looks like in basically every language except Excel and PostgeSQL, log(x) means the natural logarithm.
>
> **natural log:** C, C++, Fortran, Python, R, MATLAB, Julia, Java, JavaScript, Go, Ruby, PHP, Perl, Lua, VBA, SAS, Wolfram Language, SQL Server, MySQL, MariaDB, Haskell

I'm not sure this is super relevant to us, because very few (perhaps none?) of these languages have a separate `ln()` function to go with their `log()` function. Only Rust, Kotlin, Pascal, and some SQL dialects do, apparently.

> Would people be for/against:
>
> making `log(x)` default to natural log

I would be against it, as it would render `log()` and `ln()` redundant with respect to each other.

If we cannot agree on what the default base should be, then perhaps we should just retain the existing behavior (i.e., requiring the base to be explicitly specified).

prilau:
After further discussion, I agree that setting the default base to `10` may cause confusion to some users, especially those whose primary languages do not have a `ln()` function. I also share the sentiment with @mikeryanmay and @davidcerny that setting the default base to `e` it would make `ln()` (which is a great function!) obsolete. Given the lack of consensus on what default to set, I am happy fall back to the status quo of requiring the users to specify the base.

For `log10()`, I can implement it and open another PR. Not sure if `log2()` is necessary though

If there's no objection, I will close this PR later today

bredelings:
> I would be against it, as it would render log() and ln() redundant with respect to each other.

Can you explain why you see that as a problem?  I think we have a number of cases where we have changed the name of something, but kept the old name as a synonym for backwards compatibility.  Here we would basically move from `ln(x)` -> `log(x)`, but keep `ln(x)` for backwards compatibility.  What we gain is fitting users expectations based on the vast majority of programming languages.

I would be happy to eliminate `ln()`, but it might break old code. We could, however, replace `ln()` with `log( )` in any documentation.  We could in theory add a deprecation message to `ln()`, but I don't know if that is worth it.

mikeryanmay:
I don't think there's a fundamental problem with having functions that are redundant. What matters more is that things behave as intuitively as possible. Someone using `ln(x)` knows what they are getting into, someone using `log(x)` is very likely expecting the natural log, which is enough to convince me that is the right default.

davidcerny:
> Can you explain why you see that as a problem?

Hmm, maybe my preferences here are not entirely rational, which is probably a sign that I should re-examine them. Two arguments I can think of:

- The fact that we use `ln()` rather than `log()` for the natural logarithm feels like a deliberate design decision precisely _because_ it deviates from what most programming languages do. So I'm a little wary of reversing it many years later on the grounds that it deviates from what most programming languages do. 
- As far as I can see, we cannot actually turn `ln()` into a real alias of `log()`, in the way that `dnBDP` is an alias of `dnBirthDeath` – that would require us to accept nonsensical notation such as `ln(x, base = 10)`. So, to maintain both backward compatibility and notational correctness, we'd have to have not a single function with an alias (which is okay), but two functions that are different but nevertheless redundant in practice (much less elegant).

basantakhakurel (commented):
I'd like to weigh in on the `log()` default base discussion. While I understand the perspective from computer science and pure math, where `log` often defaults to `e`, I would advocate for keeping the default at base 10 for RevBayes.

I agree that the 'other languages use base e' argument is valid, but the users of those languages are mostly statisticians, mathematicians, or programmers who are used to working in natural-log space.
RevBayes is focused on evolutionary biologists, and I think we should optimize the default with our specific users in mind. 
Thinking from a student's perspective, I would expect "log of 1000" to be 3 and not 6.908. 
In most other scientific contexts, and on any scientific calculator, the log function defaults to base 10.

Our own tutorials do not really use log, and specify the base. Instead, every instance of a logarithm with base `e` is explicitly written as `ln()`. Users who want the natural log already have a dedicated and clearly named function for it.  
As long as we explicitly state in the function's help documentation that it defaults to base 10, I don't think there will be any confusion.

Also, log with base 10 is literally called the '[Common logarithm](https://en.wikipedia.org/wiki/Common_logarithm)' :)

bredelings:
It is true that some people assume `log` means `log10`.  In addition to high-school math before calculus, it seems like it might still be true in engineering?  Also slide rules and log tables use the common logarithm!

So, sure, `log10` is going to be the assumption for some people.  At some point I used to assume `log` meant log10, and it was confusing for a bit, and then I got over it.

But I suspect evolutionary biologists who are using RevBayes are also going to use other computer languages -- such as R and python.  In which case they already can't just assume `log(x)` returns `log10(x)`.  By making revbayes the only language that is different, we are actually making things MORE confusing -- now they have to remember which computer languages have log(x) as natural log, and which ones do not.  The question of what `log(x)` means is not something they can avoid.

I'm also not quite sure that "evolutionary biologists" all lean towards log == log10.  Clearly @ms609 doesn't lean that way.  I've queried a number of eco/evo-type evolutionary biologists, and so far 4/4 say log(x) == natural log.  
 
> me: "Some people are claiming that `log(x)` as natural log is an extra hurdle for evolutionary biologists, since having `log(x)` as natural log is the domain of mathematicians, statisticians, and computer programmers."
> response: "this is plainly not true, since this is the default behavior for R and most of phylogenetic comparative methods are in R"

bredelings:
Most projects larger than a certain size don't use absolute consensus.  If you have enough people, then at least 1 person is going to disagree with any given decision, and thus nothing would ever happen.

My impression is that most larger projects that like consensus use "semi-consensus", which usually means that something can go ahead if most people agree with it, and the people who are opposed to it are not really strongly opposed.  A good public example of semi-consensus is the C++ standards committee: See https://isocpp.org/std/standing-documents/sd-4-wg21-practices-and-procedures#consensus . The explanations and concepts there are pretty good.

I tried to assess levels of support for leaving leaving this unresolved versus going ahead with making `log(x)` mean natural log.  Only one person not involved in the original discussion responded.  So it appears to me that among the people who actually care, we remain divided.  This suggests not making any changes.  Additionally, I don't think anyone is "Strongly Against" the status quo.  @prilau suggested closing the issue; I will let her do that.

PR #1048, opened 2026-05-29
Allow specifying a fixed burnin fraction in convergence rules

davidcerny:
## PR Type
- New Feature

## Summary
Our convergence rules (`srGelmanRubin`, `srGeweke`, `srMinESS`, `srStationarity`) automatically determine the burnin fraction to be discarded in order to optimize the test statistic, but this can be too computationally demanding for very long traces with lots of monitored parameters. Here, I implement a new option, `burninMethod="fixed"`, that simply discards a constant fraction of samples collected so far. I have modified an existing test to check that it works as intended.

## Approach
The `burninMethod` argument is still set to `"ESS"` by default, since the computational cost is trivial in most cases. When the `"fixed"` option is chosen instead, the newly added `burnin` parameter is used to specify what fraction of the chain should be discarded before the test statistic is computed. By default, `burnin` is set to 0.25. If the user explicitly sets `burnin` but uses `burninMethod="ESS"` or `burninMethod="SEM"`, a warning is displayed to let the user know that the argument will be ignored.

I considered throwing an exception when `burninMethod="fixed"` and `burnin` is not specified, but I think a default value makes more sense.

## Additional notes
The fact that `burnin` has a default value (i.e., it is never null) but we want to display a warning if the user explicitly sets it while using an incompatible `burninMethod` leads to something of a conundrum – we don't have an established way of distinguishing between arguments that are non-null because they have a default, and arguments that are non-null because they have been explicitly set by the user. This PR solves this problem by exploiting the fact that when `Function::processArguments()` fills in an empty slot using its `ArgumentRule` default, the synthesized variable is named with a leading dot. Since we do not allow variable names with leading dots in the user space, this criterion is reliable.

## Testing
I have modified an existing test: `test_convergenceRules/scripts/Test_multipleRules.Rev`.

## Relevant issues
See #1020 for a case where this option comes in handy.

## Checklist
- I have added or updated tests for this change, or explained why they are not needed.
- I have added or updated documentation where necessary.

### AI/LLM Assistance
- I used Composer 2.5 to assist with this PR. It helped me figure out that instead of a more extensive refactoring of the Core back ends of the individual convergence rules, which I originally thought would be necessary, it's more elegant to just create a new class deriving from `BurninEstimatorContinuous`.

### DIFF
--- a/help/md/srGelmanRubin.md
+++ b/help/md/srGelmanRubin.md
@@ -13,16 +13,27 @@ be calculated when two or more independent runs are performed, and its value
 tends to unity (1) as the runs converge.
 
 The number of samples to be removed as burnin before calculating the test
-statistic is determined using the `burninMethod`. Different burnin lengths are
-tested, increasing from 0 to 50% (for `ESS`) or 100% (for `SEM`) of the length
-of the trace in increments of 10 samples. If the `ESS` option is chosen
-(default), effective sample sizes (ESS) are calculated for all monitored
-parameters after removing the number of samples corresponding to each candidate
-burnin length. The best burnin length for a given parameter is the one that
-maximizes its ESS value. If the `SEM` option is chosen, the standard error of
-the mean (SEM) is calculated instead, and the best burnin length for a given
-parameter is the one that minimizes its SEM value. In both cases, the final
-burnin length is set to the maximum of the parameter-specific burnin lengths.
+statistic is determined using the `burninMethod`. If the `"ESS"` (default) or
+`"SEM"` options are chosen, different burnin lengths are tested, increasing
+from 0 to 50% (for `"ESS"`) or 100% (for `"SEM"`) of the length of the trace in
+increments of 10 samples. The `"ESS"` option calculates effective sample sizes
+(ESS) for all monitored parameters after removing the number of samples
+corresponding to each candidate burnin length. The best burnin length for a
+given parameter is the one that maximizes its ESS value. The `"SEM"` option
+instead calculates the standard error of the mean (SEM), and the best burnin
+length for a given parameter is the one that minimizes its SEM value. In both
+cases, the final burnin length is set to the maximum of the parameter-specific
+burnin lengths.
+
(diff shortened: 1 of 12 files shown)


bjoelle:
very nitpicking comment - I think the default burnin for other functions (e.g. readTrace) is set at 0.25 so I would use the same value

davidcerny:
@bjoelle Great point, and not nitpicky at all! Fixed now.

prilau:
Looks nice, thanks!

Just some small things: 
- Can you update the description of this PR regarding the default value?
- I was also wondering if it makes sense to give a more descriptive name for the argument, something like `burninFraction` or `FixedBurninFraction`, instead of just `burnin`
- The ability to distinguish 'non-null by default' and 'non-null due to user-specified value' using leading dot is practical. Can you leave a one-line comment in the code to explain why this approach works, just for future reference? I assume this approach is not widely used in RevBayes. If that's not the case, ignore this comment :)

davidcerny:
Thanks for the review, @prilau!

1. Yup, done.
2. I have opted not to make this change. Calling the argument `burnin` is more consistent with `readTrace()` and `readTreeTrace()`, and the argument description already makes it sufficiently clear that it refers to a fraction.
3. Yes, to the best of my knowledge, it's an implicit convention that hasn't been exploited in this way anywhere else in the code base, which actually made me a little nervous about relying on it. For that reason, I agree that it's a good idea to include a comment explaining what's being done, and why. It has now been added, although I couldn't fit it on one line. :)

basantakhakurel:
Thanks for adding this.
I agree with having `burnin` as the argument, as it is the one used for other cases as well.

> @bjoelle Great point, and not nitpicky at all! Fixed now.

I noticed that the default in the code is set to 0.25, but not in the help docs. Once that is changed, it looks good to me.

davidcerny:
> I noticed that the default in the code is set to 0.25, but not in the help docs. 

Ah, yes, of course! Fixed.

basantakhakurel (approved):

PR #1050, opened 2026-06-02
fix(dag topological sorting): fix DAG sorting bug reported in #1047

### LINKED ISSUE #123: Feature request: Generalized distancs with dnPhyloBrownianREML
I am using dnPhyloBrownianREML to calculate the likelihood for trees built using continuous data. There is possibly a high degree of correlation among the characters. [Felsenstein (1973)'s](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC1762641/) paper on REML PIC, which dnPhyloBrownianREML is based upon, states that the use of generalized distances (Mahalanobis distances) can be used to transform the characters to independent characters with unit variance (see section "Correlated Characters"). This would allow the analysis of correlated characters.

Are generalized distances already built into dnPhyloBrownianREML (or any members of the PhyloBrownian family?)? If not, would it be possible to be implemented?

Also, is there any more documentation on the "dnPhyloBrownian..." series of distributions? I am having some trouble understanding the distinction between them. Thank you very much!

### LINKED ISSUE #1047: [BUG] Model::getOrderedStochasticNodes can return children before parents
**Describe the bug**

`Model::getOrderedStochasticNodes(node, ordered, visited)` (`src/core/analysis/Model.cpp:298–328`) executes:

```
  visit(node):
      if visited[node]: return
      visited[node] = true
        for p in node.parents: visit(p)     # ...(i)
        if node.isStochastic: ordered.append(node)     # ...(ii)
        for c in node.children: visit(c)      # ...(iii)
```

The bug lives in step (iii). Worked example: stochastic nodes `P`, `A`, `C` with `parents(A) = {P}` and`parents(C) = {A, P}`.

Calling `visit(A)`:
(i) `visit(P)`: P has no parents, (ii) `ordered = [P]`, (iii) iterate P's children → `visit(A)`, returns immediately (already visited), then `visit(C)` runs: (i) C's parents A and P are both visited (skipped); (ii) `ordered = [P, C]`; (iii) C has no children. Control returns to A's frame. (ii) `ordered = [P, C, A]`.

Output: `[P, C, A]`. C precedes A despite A being a parent of C.

The bug is silent for most distributions where redrawValue only mutates the node's own value. It only happens when a child distribution's redrawValue also mutates a parent's value. `ValidationAnalysis::burnin`, `Mcmc::initializeSampler`, `Mcmc::redrawStartingValues`, `BootstrapAnalysis::runAll`, and `HillClimber::initializeSampler` all iterate over this returned order and call redraw per node, resulting in a silent state inconsistency.

**Proposed fix**

Remove step (iii) and have the public function iterate over all nodes once:

```cpp
  std::vector<DagNode*> Model::getOrderedStochasticNodes()
  {
      std::vector<DagNode*> ordered;
      std::set<const DagNode*> visited;
      for (auto n : nodes)
          getOrderedStochasticNodes(n, ordered, visited);
      return ordered;
  }

  void Model::getOrderedStochasticNodes(const DagNode* n,
                                        std::vector<DagNode*>& ordered,
                                        std::set<const DagNode*>& visited)
  {
      if (visited.count(n)) return;
      visited.insert(n);
      if (!n->isConstant())
          for (auto p : n->getParents())
              getOrderedStochasticNodes(p, ordered, visited);
      if (n->isStochastic())
          ordered.push_back(const_cast<DagNode*>(n));
  }
```

wnsplim:
## PR Type

- 🐛 Bug Fix

## Summary

`Model::getOrderedStochasticNodes` was able to emit a child node before its parent when both shared the same parental node. Solves issue #1047. Fixed by using a valid topological sorting algorithm. `tests/test_DAG/scripts/ordered_stoch_nodes_shared_parent.Rev` replicates the test devised in #1047 and validates that the fix results in desired behavior.

## Context 

With `parents(A)={P}` and `parents(C)={A,P}`, `model(A)` placed `A` at `nodes[0]`. The recursive DFS seeded at `A` recursed into `P`'s children before unwinding to emit `A`, producing the order `[P, C, A]`. Loops that called `redraw()` in that order drew `C` against the pre-redraw `A` while validation logged the post-redraw `A` as truth. See also #1047.

## Approach

Use a valid topological sorting algorithm. In #1047, I and @bredelings proposed a fixed DFS algorithm; per @bredelings's suggestion, "in the future by adding a more standard algorithm for topological sorting", Kahn's topological sorting algorithm is implemented here (which will be more robust to stack overflow), but I will be happy to roll back to the DFS approach if this feels like overkill.

## Testing

`tests/test_DAG/scripts/ordered_stoch_nodes_shared_parent.Rev` reproduces the issue (#1047)'s P--A--C model: fixed binary `Var = 1.09707`; buggy binary (current development branch) `Var = 3.02056`.

## Relevant issues

Fixes #1047

## Checklist

### AI/LLM Assistance
- [x] I used an AI/LLM tool to assist with this PR: Claude Code Opus 4.7 helped track the root cause of SBC coverage drift. It also assisted in drafting Kahn's algorithm implementation. All code was checked both logically and numerically by me.
 
- *I am the primary author of the PR cover letter (e.g., it wasn't AI/LLM-generated).*
- *I have cleaned up the AI-generated code (e.g., removing verbose comments)*


bredelings:
Hi @wnsplim!  Thanks, the fix looks good.  Presumably some of the tests are failing because Mcmc also calls getOrderedStochasticNodes when (i) initializing the sampler and (ii) redrawing starting values.  This changes the logs, but presumably that is just because variables are now initialized / redrawn in a different order.  So we should be able to just replace the old log output with the new log output.  The trick is that it is the Linux log output that is standard, so on a Mac this may not work.

davidcerny (approved):

wnsplim:
@bredelings I cloned the branch on the Ubuntu machine and replaced the files; tests seem to pass now?

PR #1052, opened 2026-06-02
Make it simpler to `exclude` vector variables from `mnModel`

### LINKED ISSUE #1051: [BUG] Cannot `exclude` vector variables from `mnModel` without enumerating their elements
**Describe the bug**
One unintended consequence of PR #900 (merged prior to the v1.3.2 release) is that we can no longer exclude a vector variable as a whole from the scalar parameter monitor (`mnModel`) simply by referring to its name. Instead, we have to exclude it element by element.

**To reproduce**
Replacing the following line of the [`mvFossilTimeSlideUniform_checkpointing.Rev`](https://github.com/revbayes/revbayes/blob/development/tests/test_FBD/scripts/mvFossilTimeSlideUniform_checkpointing.Rev) test script:

https://github.com/revbayes/revbayes/blob/245613727a80724f194f20d1d48bfd6985e6ca9e/tests/test_FBD/scripts/mvFossilTimeSlideUniform_checkpointing.Rev#L209

by:

```r
monitors.append( mnModel(filename="output/mvFossilTimeSlideUniform_checkpointing.log", printgen=10,
                         exclude=["branch_rates"]) )
```

demonstrates that the attempted exclusion has no effect. To actually achieve what we are trying to do, we need some extra code:

```r
for (i in 1:branch_rates.size()) {
    to_exclude[i] = "branch_rates[" + i + "]"
}

monitors.append( mnModel(filename="output/mvFossilTimeSlideUniform_checkpointing.log", printgen=10,
                         exclude=to_exclude) )
```

**Expected behavior**
We should be able to exclude a vector variable by passing its name to the `exclude` argument.

**Computer info**
macOS 26.3 (Tahoe), ARM64 architecture, RevBayes v1.4.0-dev, current tip of development (2456137).

**Additional context**
I was alerted to this problem by @HaoqingDu.

davidcerny:
## PR Type
- Bug Fix

## Summary
We are currently unable to `exclude` vector variables from the `mnModel` monitor unless we do it element by element: e.g., for a vector of length 3 called `x`, we need to specify `exclude=["x[1]", "x[2]", "x[3]"]`. This PR makes it possible to use the much simpler `exclude=["x"]` notation by excluding not only exact matches, but also indexed descendants such as `x[1]` or `x[1][1]`. I have modified two existing tests to check that the approach works as intended.

## Additional notes
The new solution is flexible enough to still allow excluding individual vector elements (e.g., only `x[2]`), should the user so desire.

## Testing
I have modified two existing tests: `test_FBD/scripts/mvFossilTimeSlideUniform_checkpointing.Rev`, which now excludes a simple vector, and `test_Partition/scripts/mcmc_Partition_codon.Rev`, which covers the case of excluding a nested vector.

## Relevant issues
Fixes #1051.

## AI/LLM Assistance
I used Composer 2.5 to assist with this PR. I checked the logic of the generated code, wrote a description for the newly added `ModelMonitor::isExcluded()` function, and moved this function out of an anonymous namespace within the `.cpp` file (where the LLM originally wanted to place it).


bjoelle (approved):
looks good to me

PR #1056, opened 2026-07-08
Make backbone-constrained nonclock analyses work without an initial tree

### LINKED ISSUE #384: [BUG] Segfault w/ backbone constraints for nonclock trees unless initial tree is specified

davidcerny:
## PR Type
- Bug Fix

## Summary

This PR enables conducting nonclock tree inference under backbone constraints (i.e., enforcing partial splits) without requiring the user to supply a compatible initial tree. Multiple changes had to be made to the existing code to make this possible (see "Approach", as well as the individual commit messages for details); of these, probably the most important one was to ensure that the `simulateUnrootedTree()` function gets called when needed. I added two new tests to demonstrate that the changes have the desired effect.

## Approach
Solving the larger issue involved fixing the following problems:

- The `is_rooted` variable should be set depending on the properties of the base distribution, rather than hardcoded to `true`. (Fixing this actually involved reverting a previous, poorly justified change in the opposite direction.) The value of the variable matters, because it determines whether we simulate a new tree using `simulateRootedTree()` or `simulateUnrootedTree()`. The former will never produce topologies that are valid under nonclock tree distributions such as `dnUniformTopologyBranchLength`.
- User-supplied initial trees, as well as initial trees generated by cloning the backbone tree and attaching the remaining taxa to it, can be "bare topologies" with `NaN` branch lengths, but these yield `-inf` log likelihoods when evaluated by `UniformTopologyBranchLengthDistribution::computeLnProbability()`. A new function called `assignBranchLengths()` has been introduced to generate valid branch lengths by drawing from the prior.
- The `active_backbone_clades` bitsets are sometimes only partially updated after DAG touches and can therefore become stale, potentially leading to `matchesBackbone()` failing when it shouldn't (or, for that matter, succeeding when it shouldn't). We fix this by forcing a full refresh in `computeLnProbability()` and in the copy constructor.

## Testing
I added two new tests to check that nonclock inference under backbone constraints works both with and without an initial tree:

- [`Test_backbone_BranchLengthTree.Rev`](https://github.com/revbayes/revbayes/blob/nonclock-backbone-constraints/tests/test_constraints/scripts/Test_backbone_BranchLengthTree.Rev)
- [`Test_backbone_initial_BranchLengthTree.Rev`](https://github.com/revbayes/revbayes/blob/nonclock-backbone-constraints/tests/test_constraints/scripts/Test_backbone_initial_BranchLengthTree.Rev)

## Relevant issues
Fixes #384.

## AI/LLM Assistance

This PR basically started out as an experiment on my part to see whether I can use Composer 2.5 – not a frontier model, but still quite capable when used with the Cursor harness – to fix a non-trivial issue that has been open for a long while now.

Originally, I got back an undifferentiated mass of changes, one of which (an attempt to convert backbone splits into clade constraints) made no conceptual sense (though I could see it being acceptable in a pinch as a quick-and-dirty way of generating a starting tree). I prompted the model to explain the changes to me one by one, decided whether or not to accept them based on the explanations given, and converted those changes that I did accept (which, to be fair, were almost all of them) into individual commits, which should hopefully make life easier for the reviewers of this PR.

One change I did _not_ accept was an attempt to refactor much of `simulateUnrootedTree()` into a separate function which then gets run in a loop for a fixed number of iterations (the AI suggested 10,000). To me, this appeared to duplicate the functionality of `Mcmc::initializeSampler()`, which already uses multiple tries (1000 by default) to find a viable initial state. In the worst case scenario, this duplication could result in 10 million separate calls to `simulateUnrootedTree()`. Also, the test above already works with 1 attempt instead of 10,000, so this seemed like complete overkill.


src/core/distributions/phylogenetics/tree/UniformTopologyBranchLengthDistribution.cpp
```diff
@@ -164,6 +165,28 @@ UniformTopologyBranchLengthDistribution& UniformTopologyBranchLengthDistribution
...
+    {
+        TopologyNode &node = psi.getNode(i);
+        if ( node.isRoot() )
+        {
+            continue;
+        }
+        
+        double bl = node.getBranchLength();
+        
+        // each branch length that is NaN or zero is reset to an independent draw from the prior
```
basantakhakurel: NaN, zero or negative?

src/core/distributions/phylogenetics/tree/TopologyConstrainedTreeDistribution.cpp
```diff
@@ -238,6 +246,7 @@ double TopologyConstrainedTreeDistribution::computeLnProbability( void )
 {
     using namespace RbConstants;
 
+    std::fill( dirty_nodes.begin(), dirty_nodes.end(), true );
```
basantakhakurel, line 249: Could `active_backbone_clades` be utilized here? I wonder if that would make it slightly performative?

basantakhakurel (changes requested):
Looks good. Thanks!
The concern I had (which could not be an issue) is that when the backbone tree has an actual polytomy, and the branch lengths are drawn from an exponential distribution.
And two things that might be worth adding/updating:
- For the tests, adding constraints in the timetree context.
- Updating the help for `dnConstrainedTopology` to reflect the possibility of its usage in unrooted analyses (right now it still gives the impression that it can only be applied to rooted trees).

davidcerny:
@basantakhakurel Thank you for the review! I'll address all of it later this week, but for now I just wanted to say that this:

> The concern I had (which could not be an issue) is that when the backbone tree has an actual polytomy

is an excellent point I didn't think of. It's true that there are two main ways of using backbone constraints. The one I was thinking of relies on an incompletely sampled binary backbone tree, which enforces _partial_ splits. E.g., for a set of taxa labeled `t1`, `t2`, ..., `t10`, it could look like this:

```
(t1, (t7, (t4, t10)));
```

However, one can also specify a completely sampled multifurcating backbone, which is actually another way of enforcing clade constraints (i.e., _full_ splits):

```
(t1, t2, t5, (t7, t9, (t3, t4, t6, t8, t10)));
```

Let me see if the second way works too, and add another test if it does.

src/core/distributions/phylogenetics/tree/UniformTopologyBranchLengthDistribution.cpp
davidcerny: Fixed

src/core/distributions/phylogenetics/tree/TopologyConstrainedTreeDistribution.cpp
davidcerny: If I understand your comment correctly (perhaps I don't!), that would basically amount to going back to the original code, which tried to figure out which bitsets should be updated by using `fireTreeChangeEvent()`, which in turn calls `recursivelyFlagNodesDirty()`. However, the problem with this solution is that if nodes get reindexed (which can be done by the nonclock SPR move, for example), the entries corresponding to non-dirty nodes become stale – they end up referring to a different node than intended.

It's true that forcing a full refresh does increase the computational cost compared to the original incremental update, but I wouldn't worry about that too much. Clade bookkeeping is very cheap compared to likelihood evaluation, so I wouldn't expect any detectable increase in wall-clock runtime.

davidcerny:
> The concern I had (which could not be an issue) is that when the backbone tree has an actual polytomy, and the branch lengths are drawn from an exponential distribution.

It's not entirely clear to me from this sentence what your specific concern was, but your comment definitely pointed me in the right direction. It turned out that we indeed could _not_ handle multifurcating backbone trees, so I had to add some extra code to `TopologyConstrainedTreeDistribution::simulateUnrootedTree()` to fix that. I added two new tests to demonstrate the validity of the fix:

- [`Test_backbone_multifurcating_BranchLengthTree.Rev`](https://github.com/revbayes/revbayes/blob/nonclock-backbone-constraints/tests/test_constraints/scripts/Test_backbone_multifurcating_BranchLengthTree.Rev)
- [`Test_backbone_multifurcating_initial_BranchLengthTree.Rev`](https://github.com/revbayes/revbayes/blob/nonclock-backbone-constraints/tests/test_constraints/scripts/Test_backbone_multifurcating_initial_BranchLengthTree.Rev)

> For the tests, adding constraints in the timetree context.

It's beyond the remit of this PR, which focuses on nonclock trees only, but I did indeed originally try to do this. However, I had to abandon the attempt. For the dataset I picked to design the tests, at least, the time tree analysis failed to find a viable initial state during the default 1000 attempts without a user-supplied starting tree. This, as far as I can tell, is not due to any actual bug (in contrast to the nonclock case tackled by this PR), but simply to the fact that finding good initial states for very complex models is intrinsically difficult. Ben's PR #1022, once it is ready, might help with that. When I did try to provide my own starting tree, I ran into issue #1057. Since the initialization failed in both cases, there was nothing left to test.

Note that the case of time trees with constraints is already covered by several of our tutorial tests: `dating`, `fbd`, `mcmc_troubleshooting`.

> Updating the help for `dnConstrainedTopology` to reflect the possibility of its usage in unrooted analyses (right now it still gives the impression that it can only be applied to rooted trees).

This was a little tricky, because even though there is a single Rev name (`dnConstrainedTopology`) and a single underlying Core distribution (`TopologyConstrainedTreeDistribution`), the latter has two Rev language wrappers (`Dist_ConstrainedTopology` and `Dist_ConstrainedUnrootedTopology`). I had to modify the system for generating and rendering help pages to display the info for both overloads rather than just one, but I'm quite happy with the outcome:

Screenshot 2026-07-15 at 14 16 18

basantakhakurel (approved):
LGTM! Thanks for addressing my comments!

PR #1059, opened 2026-07-10
Some more `dnBDSTP` bug fixes

### LINKED ISSUE #889: [BUG] BDSTP's `swapInternalParameter()` function only handles rates, not timelines

davidcerny:
## PR Type
- Bug Fix

## Summary
This PR fixes some minor issues with `dnBDSTP`. Primarily, it adds handlers for the eight timeline variables (as well as one previously missing event probability variable) to `BirthDeathSamplingTreatmentProcess::swapParameterInternal()`, which helps prevent scenarios in which the distribution keeps a stale pointer to the old timeline nodes while the parameter list points at the new ones – something that could plausibly happen after cloning the model (e.g., on MCMC initialization) or variable re-assignment in Rev. I also fixed an unrelated bug where the internal variable corresponding to the timeline for speciation bursts was initialized with the parameter storing the timeline for sampling bursts.

## Testing
I couldn't think of how to design a test for these changes.

## Relevant issues
Fixes #889.

### AI/LLM Assistance
The original commit (adding the handlers for the timeline variables) was typed [with my own human hands](https://xkcd.com/1530/), but it was Composer 2.5 that told me `heterogeneous_R` was missing from `swapParameterInternal()` as well, and alerted me to the typo in the constructor.


src/core/distributions/phylogenetics/tree/birthdeath/BirthDeathSamplingTreatmentProcess.cpp
```diff
@@ -1941,6 +1942,46 @@ void BirthDeathSamplingTreatmentProcess::swapParameterInternal(const DagNode *ol
...
+    // Timeline parameters
+    if (oldP == interval_times_global)
+    {
+        interval_times_global = static_cast<const TypedDagNode<RbVector<double> >* >( newP );
+    }
+    else if (oldP == interval_times_speciation)
+    {
+        interval_times_speciation = static_cast<const TypedDagNode<RbVector<double> >* >( newP );
+    }
+    else if (oldP == interval_times_extinction)
```
hoehna: I don't think this should be an else-if, but instead an if. The reason is that if a user passes the same variable for both the speciation and extinction timelines, then this will only swap the parameter for the first hit and not for both. Please also check for other timelines.

hoehna (changes requested):
Please replace the else-if with simple if to guarantee that the same parameter can be used multiple times.

davidcerny:
The point about using individual `if`s instead of long `else if` ladders in `swapParameterInternal()` is a very good one. @hoehna and I had a further discussion about this, and there are cases where it makes a rather dramatic difference. In [`GammaDistribution.cpp`](https://github.com/revbayes/revbayes/blob/development/src/core/distributions/math/GammaDistribution.cpp), for example, replacing the two individual `if`s by an `else if` construction leads to a segfault on the following simple script:

```r
seed(12345)

alpha <- 1.0
x ~ dnGamma(alpha, alpha)
x.lnProbability()

alpha <- 2.0
x.lnProbability()
```

To keep things safe, I replaced all the `else if`s in `BirthDeathSamplingTreatmentProcess::swapParameterInternal()` by individual `if`s. I also added a test inspired by the script above.

hoehna (approved):
Seems good now and fixes the issue.

PR #1064, opened 2026-07-19
Test/windows git checkout

### LINKED ISSUE #1063: [BUG] Windows release builds show empty git branch and commit in the startup banner
**Describe the bug**
While testing the new release executables, I noticed that the branch and commit fields are empty in the Windows executable. 
The startup banner is supposed to print a line like this:
```
Build from tags/v1.4.1 (3a504e) on Fri Jul 17 08:58:20 UTC 2026
```
On the win64 release, the branch and commit fields are empty:
```
Build from  () on Fri Jul 17 08:57:26 UTC 2026
```
I tested it in Linux using Wine and verified it in Windows as well (see attached screenshots below).

**To Reproduce**
1. Download a win64 release tarball for v1.3.2 or above.
2. Extract and run `bin/rb.exe`.

**Expected behavior**
The Windows binary's "Build from" line should match the branch and commit information of the Linux and macOS binaries built from the same commit.

**Screenshots**
Using Wine in Ubuntu 26.10:

Image

On Windows:

Image

**Computer info**
- Affected: v1.3.2, v1.4.0, and v1.4.1
- Not affected: v1.3.1 (and earlier)
- Tested via Wine (wine-10.0) on Ubuntu 26.10 and also on Windows 10.

**Additional context**

basantakhakurel:
## PR Type

- 🚦 Infrastructure?

## Summary

The Windows release build's startup banner is missing its git branch and commit information (`Build from () on ...` instead of `Build from tags/v1.4.1 (3a504e) on ...`).
This is caused by the Windows job's `container: image: ubuntu 25.04` having no `git` preinstalled when `actions/checkout` runs, so checkout was falling back silently to a tarball download without the `.git` directory. This PR installs `git` before checkout and marks the workspace as a safe directory so later steps can run git commands.

## Approach

Added a step that installs `git` before `actions/checkout` runs.

## Additional notes
None.

## Testing

- Verified against a workflow run (`release.yml`) triggered by a tag.
- `actions/checkout` performed a real `git init` and `git fetch` instead of falling back to a tarball download.
- The built `rb.exe`'s startup banner shows the correct `Build from tags/zz-test-windows-git-checkout-1 (ed0709) on Sun Jul 19 11:03:32 UTC 2026` line, matching the Linux build from the same commit.

## Relevant issues
Fixes #1063.

## TODO before merging
None.

## Checklist
- [ ] I have added or updated documentation where necessary.

### AI/LLM Assistance
- [x] I used an AI/LLM tool to assist with this PR to guide me with the git commands necessary for creating a tag and triggering the release action.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -89,7 +89,7 @@ jobs:
     steps:
     - name: Install git (containers have none preinstalled)
       if: matrix.image != ''
-      run: 
+      run: |
         apt-get update && apt-get install -y git
         git config --system --add safe.directory '*'
     - uses: actions/checkout@v4
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -363,7 +363,7 @@ jobs:
     name: Notify Slack
     needs: [build_singularity_and_upload, build_and_upload]
     runs-on: ubuntu-latest
-    if: always()
+    if: false
     steps:
     - uses: technote-space/workflow-conclusion-action@v3
     - uses: 8398a7/action-slack@v3
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -363,7 +363,7 @@ jobs:
     name: Notify Slack
     needs: [build_singularity_and_upload, build_and_upload]
     runs-on: ubuntu-latest
-    if: false
+    if: always()
     steps:
     - uses: technote-space/workflow-conclusion-action@v3
     - uses: 8398a7/action-slack@v3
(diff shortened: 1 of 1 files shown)


basantakhakurel:
I forgot to mention it, but `build.yml` should also have the same issue. Since the executable is not really used by users, I decided not to change that one.

davidcerny (approved):
Neat little fix, especially since Windows accounts for a surprisingly high proportion of our total executable downloads!

One change you could consider (though I certainly don't insist on it) would be to clean up the commit history with interactive rebase, so that you don't have a later commit undoing the Slack notification removal introduced in an earlier one.

PR #1066, opened 2026-07-24
Dev soft bounds

hoehna:
## PR Type
- 🐛 Bug Fix

## Summary
Our previous code didn't implement the soft-bounded distributions "correctly". They had the shape of batman, and now we use the corrected implementation. We also now enable soft bounds only for one side.


davidcerny:
In case others might wonder what is "the shape of [B]atman" (more of a Höhna Lab in-joke than a serious description), the problem was that, at some point, the density function of the soft-bounded uniform distribution looked like this:

Bad_soft_uniform
</br></br>

rather than this:

Good_soft_uniform
</br></br>

That said... I'm not sure that this is still a problem, and that the PR description is accurate. One issue is that the newly added test, although it does _cover_ `dnSoftBoundUniformNormal`, doesn't actually directly test for the issue in question. One could devise a much simpler test along the following lines:

```r
x ~ dnSoftBoundUniformNormal( 13.6, 25, sd = 1.0, p = 0.95 )
x.clamp(10)   # lower normal tail
x.probability()

x.clamp(13.6) # transition from normal to uniform
x.probability()

x.clamp(19.3) # midpoint of the uniform
x.probability()

x.clamp(25)   # transition from uniform to normal
x.probability()

x.clamp(28)   # upper normal tail
x.probability()
```

However, this test gives the same results on this branch (`dev_soft_bounds`) and the current tip of `development`. This is perhaps unsurprising, because commit [5164fdb](https://github.com/revbayes/revbayes/pull/1066/commits/5164fdb8cf03ae99a12bc8288d154cb5a602206b), in spite of its message, doesn't touch anything having to do with the actual math. It seems to me that this PR adds a _feature_ (one-sided soft-uniform distributions), but doesn't fix any _bugs_.

help/md/dnSoftBoundUniformNormal.md
```diff
@@ -11,14 +11,12 @@ Sebastian Hoehna
...
 dnUniform
 ## example
-	p ~ dnBeta(1.0,1.0)
-	x ~ dnBernoulli(p)
-	x.clamp(1)
-	moves[1] = mvSlide(p, delta=0.1, weight=1.0)
-	monitors[1] = screenmonitor(printgen=1000, separator = "	", speciation)
+	calib ~ dnSoftBoundUniformNormal( 13.6, 25, sd = 1.0, p = 0.95, "upper" )
+	moves[1] = mvSlide(calib, delta=0.1, weight=1.0)
+	monitors[1] = screenmonitor(printgen=1000, separator = TAB, calib)
```
davidcerny: There is no such thing as `screenmonitor`. Please make sure that the example is a self-contained chunk of code that can actually be run.

davidcerny (changes requested):

help/md/dnSoftBoundUniformNormal.md
hoehna: right, now the example actually works

davidcerny (dismissed):
So there are several issues with this PR. One, which I already pointed out above, is that the actual content doesn't fit the description, since the PR doesn't change the distribution's density function, and consequently still suffers from the very problem it claims to be solving. This is easily verified by the following test, which gives the same result on the current tips of development (69c2220) and dev_soft_bounds (4f76d5b), and which shows that as we approach the soft bounds from the tail side, we get higher probability densities than between the bounds themselves (the bat-eared shape issue):

```
> x ~ dnSoftBoundUniformNormal( 13.6, 25, sd = 1.0, p = 0.95 )
> x.clamp(13.59)
> x.lnProbability()
   -1.563346
> x.clamp(13.599)
> x.lnProbability()
   -1.563296
> x.clamp(13.59999999)
> x.lnProbability()
   -1.563296
> x.clamp(14.01)
> x.lnProbability()
   -2.484907
```

It looks like the commit that was actually intended to fix the issue ([ca04244](https://github.com/revbayes/revbayes/commit/ca042443f92962761d9efe43391778e7e4188448) on the old, now-deleted dev_clock_rate_mix branch) was accidentally omitted from this PR. That commit does change the density function, and does indeed get rid of the bat ears:

```
> x ~ dnSoftBoundUniformNormal( 13.6, 25, sd = 1.0, p = 0.95 )
> x.clamp(13.59)
> x.lnProbability()
   -4.607868
> x.clamp(13.599)
> x.lnProbability()
   -4.607818
> x.clamp(13.59999999)
> x.lnProbability()
   -4.607818
> x.clamp(14.01)
> x.lnProbability()
   -2.484907
```

The problem is... these (log) densities don't look quite right, either. As we approach the bounds from the tail side, I would expect the density to asymptotically approach the value that it takes between the bounds. Here, the value taken between the bounds is −2.484907, which makes sense, since (25 − 13.6) * exp(−2.484907) = 0.95. However, the value that we approach from the tail side is much lower than that: −4.607818. I.e., the distribution now looks something like this:

Image
<br/><br/>

So I have two change requests:

1. Fix the density function so that it has the right shape.
2. Use a simpler, more focused integration test. The code chunk I gave above would be sufficient, in my opinion. The current "test" is a random analysis script that simply happens to contain `dnSoftBoundUniformNormal`, so it doesn't test anything in particular, and didn't even allow us to notice that the PR is not doing what it claims to be doing.

davidcerny:
Hmm, so I was somewhat unfair in my criticism. The real problem is not with commit [ca04244](https://github.com/revbayes/revbayes/commit/ca042443f92962761d9efe43391778e7e4188448); it's with the fact that there is no way to eliminate the discontinuities at `min` and `max` if both `p` and `sd` are treated as free parameters, which is what we do both in the test script and in the help file. When we only specify one, the code introduced in [ca04244](https://github.com/revbayes/revbayes/commit/ca042443f92962761d9efe43391778e7e4188448) correctly calculates the other so that the normal "tails" attach to the uniform "core" without discontinuities.

In light of this, I have added the commit to this branch. All that remains to be done is (in my opinion) to decide how to handle the interactions between `p` and `sd`. I can think of several options:

1. Preserve the current behavior, i.e., quietly return a distribution with discontinuities in its p.d.f.
2. Return it, but emit a warning alerting the user to the presence of discontinuities in its p.d.f.
3. Make it impossible to specify `p` and `sd` at the same time.

src/revlanguage/distributions/math/Dist_softBoundUniformNormal.cpp
```diff
@@ -50,9 +52,39 @@ RevBayesCore::SoftBoundUniformNormalDistribution* Dist_SoftBoundUniformNormal::c
...
+    }
+    else if ( bound == "lower" )
+    {
+        soft_bounds = RevBayesCore::SoftBoundUniformNormalDistribution::LOWER;
+    }
+    
+    if ( sd->getRevObject() == RevNullObject::getInstance() && prob->getRevObject() == RevNullObject::getInstance() )
+    {
+        throw RbException("You need to either provide the 'sd' or 'prob' parameter for the SoftBoundUniformNormal distribution.");
+    }
```
davidcerny: The parameter is actually called `p`, not `prob`

src/revlanguage/distributions/math/Dist_softBoundUniformNormal.cpp
```diff
@@ -129,20 +161,26 @@ std::string Dist_SoftBoundUniformNormal::getDistributionFunctionName( void ) con
...
-        distNormMemberRules.push_back( new ArgumentRule( "sd" , RealPos::getClassTypeSpec()    , "The standard deviation of the normal distribution.", ArgumentRule::BY_CONSTANT_REFERENCE, ArgumentRule::ANY  ) );
-        distNormMemberRules.push_back( new ArgumentRule( "p"  , Probability::getClassTypeSpec(), "The probability of being within the uniform distribution.", ArgumentRule::BY_CONSTANT_REFERENCE, ArgumentRule::ANY ) );
+        dist_member_rules.push_back( new ArgumentRule( "min", Real::getClassTypeSpec()       , "The min value of the uniform distribution.", ArgumentRule::BY_CONSTANT_REFERENCE, ArgumentRule::ANY  ) );
+        dist_member_rules.push_back( new ArgumentRule( "max", Real::getClassTypeSpec()       , "The max value of the uniform distribution.", ArgumentRule::BY_CONSTANT_REFERENCE, ArgumentRule::ANY  ) );
+        dist_member_rules.push_back( new ArgumentRule( "sd" , RealPos::getClassTypeSpec()    , "The standard deviation of the normal distribution. You need to either specify this value, or it can be computed based on the probability of being withing the uniform boundaries.", ArgumentRule::BY_CONSTANT_REFERENCE, ArgumentRule::ANY, NULL  ) );
```
davidcerny: Typo: within, not "withing"

davidcerny:
I do realize my long conversation with myself under this PR might appear a little weird, but:

> All that remains to be done is (in my opinion) to decide how to handle the interactions between `p` and `sd`. I can think of several options:
>
> 1. Preserve the current behavior, i.e., quietly return a distribution with discontinuities in its p.d.f.
> 2. Return it, but emit a warning alerting the user to the presence of discontinuities in its p.d.f.
> 3. Make it impossible to specify `p` and `sd` at the same time.

After thinking about this further, we really need to go with option (3). Otherwise, the bat ear problem can never be eliminated completely: it will re-appear as soon as the user chooses the wrong combination of parameters (specifically, a large tail probability `1 - p` combined with a small standard deviation `sd`). To wit:

```
> x ~ dnSoftBoundUniformNormal( 13.6, 25, sd = 0.05, p = 0.8 )
> x.clamp(18)
> x.lnProbability()
   -2.656757
> # makes sense, since exp(-2.656757) * (25 - 13.6) = 0.8
>
> x.clamp(13.25)
> x.lnProbability()
   -24.72579
> # so far so good
>
> x.clamp(13.59)
> x.lnProbability()
   -0.2457914
> x.clamp(13.5999) 
> x.lnProbability()
   -0.2257934
> # bad! we are getting the bat ears again
```

This means that the scripts associated with [Darlim & Höhna (2024)](https://www.biorxiv.org/content/10.1101/2024.09.25.614964v1.full) will no longer work, as they specify both parameters at the same time. However, since the work in question is a preprint (and based on a now-deleted branch, anyway), perhaps that's acceptable.

prilau:
While waiting to compile this branch, I was just wondering if this is a better representation of how this distribution should look like... or did I understand it wrong?

Screenshot 2026-08-27 at 09 48 52

davidcerny:
Yes, that's exactly what it should look like – my hand-scribbled figures had more exponential-looking tails, I guess, but they should indeed be half-normal.

prilau (changes requested):
It looks good!

Do you think it is possible, when creating the distribution, to throw an exception in case the bat ear case happened unexpectedly? e.g., check if `P(value in the soft bound) > P (value in the core)`
Not a must, just a suggestion

davidcerny:
Now that it's impossible to specify `p` and `sd` simultaneously, we have made that scenario impossible by construction, so a check of that kind would be redundant, I think. This is the type of situation where we need to perform a very careful check once (namely, that we got the math right), rather than a routine check at every function call.

src/core/distributions/math/SoftBoundUniformNormalDistribution.cpp
```diff
@@ -91,17 +76,67 @@ SoftBoundUniformNormalDistribution* SoftBoundUniformNormalDistribution::clone( v
...
     }
     else
     {
-        return log(prob->getValue()) + RbStatistics::Uniform::lnPdf( min->getValue(), max->getValue(), *value);
+        throw RbException( "The sd and p parameters of SoftBoundUniformNormal distribution cannot be provided simultaneously." );
+    }
+    
+    if ( *value < min_val )
+    {
+        if ( soft_bound == BOTH )
```
prilau: Would it be more consistent if we use `side_factor` here, then we can use one if-clause to include both `boundary="both"` and `boundary="lower"` (similar for `line 123-130`)

prilau (changes requested):
See my comment above (though I quoted the wrong lines, should be the lines below)

prilau (approved):
Great, thanks!

PR #1067, opened 2026-07-24
Adding a delay option to moves

hoehna:
## PR Type
- 🆕 New Feature

## Summary
I have added a new feature that lets you allow to delay a move. Specifically, you can set any move to only start working after a given number of iterations. Yeah!


src/revlanguage/moves/RlMove.cpp
```diff
@@ -93,7 +94,9 @@ const MemberRules& Move::getParameterRules(void) const
     
     if ( !rules_set )
     {
+
         move_member_rules.push_back( new ArgumentRule( "weight", RealPos::getClassTypeSpec(), "The weight determines the relative frequency with which this move will be attempted. For details, see the description of the 'moveschedule' parameter on the documentation page for 'mcmc()'.", ArgumentRule::BY_VALUE, ArgumentRule::ANY, new RealPos( 1.0 ) ) );
+        move_member_rules.push_back( new ArgumentRule( "delay", Natural::getClassTypeSpec(), "The of iterations before using.", ArgumentRule::BY_VALUE, ArgumentRule::ANY, new Natural( 0L ) ) );
```
davidcerny, line 99: The word "number" is missing between "The" and "of". The description of the argument should also make it clear where we start counting from – the initialization of the sampler, or the beginning of the sampling phase?

src/core/moves/AbstractGibbsMove.cpp
```diff
@@ -177,3 +177,19 @@ void AbstractGibbsMove::printSummary(std::ostream &o, bool current_period) const
...
+
+
+/**
+ * Tune the move.
+ * This is a dummy implementation because Gibbs move cannot be tuned.
+ */
+void AbstractGibbsMove::tune( void )
+{
+
+}
```
davidcerny, line 195: I deliberately deleted this in PR #947, as I see no reason to retain these particular no-op functions. So I'm opposed to restoring them.

src/core/moves/proposal/IndependentPriorProposal.h
```diff
@@ -149,6 +151,21 @@ void RevBayesCore::IndependentPriorProposal<valueType>::undoProposal( void )
...
+}
+
+
+/** Tune the proposal */
+template <class valueType>
+void RevBayesCore::IndependentPriorProposal<valueType>::tune(double r)
+{
+
+    // nothing to do
+}
```
davidcerny, line 167: Same story; please delete these.

davidcerny (commented):
Some thoughts on this:

1. ~~It's not clear to me how the `delay` parameter interacts with the distinction between the (pre-)burnin phase and the sampling phase. Does it start counting from the moment the sampler was initialized, or from the beginning of the sampling phase (i.e., generation 0, as recorded in the monitors)?~~

2. ~~In the former case, it's possible – and in the latter case, even inevitable – that one part of the sampling phase will be conducted without a particular move, and another part with it. I wonder whether this is a good idea, and how it might relate to the problem described in #934 – i.e., will an MCMC run with delayed moves still satisfy the detailed balance property?~~

3. Has the argument been added to _all_ the moves? (I guess this is up to the reviewers to verify, so I'm happy to look into it later.)

4. ~~This _almost_ passes the tests, except for 3 BDSTP / FBD ones. It's true that these particular tests are notoriously fragile, but it's still unexpected and concerning that the mere addition of a new argument to existing moves causes them to fail. We'll need to look into this further.~~

    ~~(I'm also curious what these particular three failing scripts – [`BDSTP.Rev`](https://github.com/revbayes/revbayes/blob/development/tests/test_BDSTP/scripts/BDSTP.Rev), [`FBD_dating_ex3.Rev`](https://github.com/revbayes/revbayes/blob/development/tests/test_FBD/scripts/FBD_dating_ex3.Rev), and [`mvFossilTimeSlideUniform_multiple_runs.Rev`](https://github.com/revbayes/revbayes/blob/development/tests/test_FBD/scripts/mvFossilTimeSlideUniform_multiple_runs.Rev) – have in common. I thought it was maybe the inclusion of a particular move, like `mvRateAgeBetaShift` or `mvCollapseExpandFossilBranch`, but that doesn't seem to be the case.)~~

5. We need a test script that sets the `delay` argument to a nonzero value.

davidcerny:
Actually, the cause of the failing tests wasn't that hard to find. Even though move constructors gained a new `delay` argument of type `size_t`, they were called in several places using their old form, i.e., with one fewer argument than they now have. This caused the `tune` argument to be misinterpreted as `delay`, and its `bool` value of `true` to be cast as a `size_t` value of 1. The nonzero delay then caused the test results to diverge.

davidcerny:
Okay, so to answer my own question, `delay` specifically refers to sampling-phase generations, since it relies on comparisons with the `generation` counter that is not incremented during the burnin. If your analysis includes a 200-generation burnin, and you set `delay=100` in a particular move, that move will be deactivated for the entire duration of the burnin phase _and_ the first 100 generations of the sampling phase.

One consequence of this is that delayed moves can _only_ be tuned during the sampling phase.

Frankly, I'm not sure adding this feature is a good idea. Any thoughts on this, @bredelings @mikeryanmay @jhuelsenbeck?

bredelings:
Hmm... I guess part of the issue here is that there is no real explanation of why we want this, and no examples that show why we need it. It seems like one possible solution to.... something?  But we don't know what that something is, so its not clear if this is the right solution.

In addition the change is very invasive, and small typos could lead to substantial bugs.

One way to  change the bug cost is to use stronger typing.  In C++ this is a little awkward, but basically we want to say that not all `size_t` values are the same type:

```
struct delay_t
{
   size_t amount;
};
```

Then we can change all the functions that want a delay to take a delay_t, and we could call them with `delay_t{100}` instead of just `100`.

Anyway, that doesn't mean this is a good idea, just that if we decide to go ahead with this, we could prevent some bugs.

bredelings:
Also, yeah, it seems it should skip the first `delay` moves, regardless of whether they occur during the burnin or during the post-burning phase.

Hmm... it seems like one could accomplish the same thing by running a burnin period for `delay` iterations, that has a different move set.  That sounds cleaner because it doesn't require modifying the moves themselves.

davidcerny:
Hmm, @bredelings convinced me that the problems described in #934 don't arise here, since move delays don't really affect the asymptotic behavior of the chain. But as I said in my previous comment, the fact that any nonzero delay also skips the entire burnin phase means that delayed moves cannot be tuned unless one does it during the sampling phase, which _may_ violate detailed balance.

The two main questions that remain for me are the following:

1. What problem does this actually solve?
2. Is this really the most elegant solution to that problem? Wouldn't it be better, for example, to allow `mcmc.burnin()` and `mcmc.run()` to use different move vectors?

PR #1068, opened 2026-07-24
removing rates from affected nodes

hoehna:
## PR Type
- 🐛 Bug Fix

## Summary
Our rate age beta shift move has a bug when rates are counted multiple times.


davidcerny:
The change made by this PR may have fixed a genuine bug when it was [first proposed in April 2023](https://github.com/revbayes/revbayes/commit/446ca8b29d438f18da8cf579684172b47d38b223), but it has become fundamentally incompatible with other changes that have been made to `mvRateAgeBetaShift` since then – especially by PR #688. If the bug still persists, it will need to be solved in some other way.

PR #1069, opened 2026-07-24
Ownership of values in distributions

### LINKED ISSUE #1057: [BUG] Occasional segfault in backbone-constrained FBD analyses unless initial tree is set twice
**Describe the bug**
We can use `dnConstrainedTopology` to impose backbone constraints, i.e., enforce the presence of specific partial splits. Some of these can make it very difficult to find a viable initial state for the MCMC using a stochastic simulator, since the chances of a randomly generated tree containing all the required partial splits are minuscule. To deal with such cases, the distribution allows the user to supply their own tree that satisfies the constraint via the `initialTree` argument.

Occasionally, an attempt to employ this system in an FBD analysis leads to a segfault. This behavior is purely stochastic; it may or may not occur even if the random seed is fixed. When it does occur, this is what the LLDB stack trace looks like:

```
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x5fe1d0215b2bfe25)
    frame #0: 0x0000000100c6ec24 rb`RevBayesCore::TypedDistribution<RevBayesCore::Tree>::setValue(this=0x0000000c88969400, v=0x0000000c8898b6c0, (null)=false) at TypedDistribution.h:197:9
   194 	    // free memory
   195 	    if (value != v)
   196 	    {
-> 197 	        delete value;
   198 	    }
   199 	    
   200 	    value = v;
Target 0: (rb) stopped.
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x5fe1d0215b2bfe25)
  * frame #0: 0x0000000100c6ec24 rb`RevBayesCore::TypedDistribution<RevBayesCore::Tree>::setValue(this=0x0000000c88969400, v=0x0000000c8898b6c0, (null)=false) at TypedDistribution.h:197:9
    frame #1: 0x00000001015dcdd0 rb`RevBayesCore::AbstractRootedTreeDistribution::setValue(this=0x0000000c88969400, v=0x0000000c8898b6c0, f=false) at AbstractRootedTreeDistribution.cpp:775:30
    frame #2: 0x000000010164a638 rb`RevBayesCore::BirthDeathSamplingTreatmentProcess::setValue(this=0x0000000c88969400, v=0x0000000c8898b6c0, f=false) at BirthDeathSamplingTreatmentProcess.cpp:1996:37
    frame #3: 0x00000001015c875c rb`RevBayesCore::TopologyConstrainedTreeDistribution::redrawValue(this=0x0000000c8920fc00, c=MCMC) at TopologyConstrainedTreeDistribution.cpp:632:24
    frame #4: 0x00000001015c58b0 rb`RevBayesCore::TopologyConstrainedTreeDistribution::TopologyConstrainedTreeDistribution(this=0x0000000c8920fc00, base_dist=0x0000000c88969400, c=size=0, t=0x0000000c8898b600, age_check_precision=4) at TopologyConstrainedTreeDistribution.cpp:115:5
    frame #5: 0x00000001015c6344 rb`RevBayesCore::TopologyConstrainedTreeDistribution::TopologyConstrainedTreeDistribution(this=0x0000000c8920fc00, base_dist=0x0000000c88969400, c=size=0, t=0x0000000c8898b600, age_check_precision=4) at TopologyConstrainedTreeDistribution.cpp:65:1
    frame #6: 0x0000000100fa852c rb`RevLanguage::Dist_ConstrainedTopology::createDistribution(this=0x0000000c8898b780) const at Dist_ConstrainedTopology.cpp:104:67
    frame #7: 0x0000000100c683f0 rb`RevLanguage::TypedDistribution<RevLanguage::TimeTree>::createRandomVariable(this=0x0000000c8898b780) const at RlTypedDistribution.h:97:70
    frame #8: 0x0000000100d4de18 rb`RevLanguage::SyntaxStochasticAssignment::assign(this=0x0000000c88a60d00, lhs=0x000000016fdfbab0, rhs=0x000000016fdfb5c8) at SyntaxStochasticAssignment.cpp:51:26
    frame #9: 0x0000000100d59ae8 rb`RevLanguage::SyntaxAssignment::evaluateContent(this=0x0000000c88a60d00, env=std::__1::shared_ptr<RevLanguage::Environment>::element_type @ 0x0000000c891f3480 strong=6 weak=0, dynamic=false) at SyntaxAssignment.cpp:91:9
    frame #10: 0x0000000100d65bc0 rb`RevLanguage::Parser::execute(this=0x0000000102055498, root=0x0000000c88a60d00, env=std::__1::shared_ptr<RevLanguage::Environment>::element_type @ 0x0000000c891f3480 strong=6 weak=0) const at Parser.cpp:157:24
    frame #11: 0x0000000100d5250c rb`yyparse() at grammar.y:259:37
    frame #12: 0x0000000100d6725c rb`RevLanguage::Parser::processCommand(this=0x0000000102055498, command="fbd_tree ~ dnConstrainedTopology(tree_dist, backbone=backbone_tree, initialTree=scaled_init_tree)", env=std::__1::shared_ptr<RevLanguage::Environment>::element_type @ 0x0000000c891f3480 strong=6 weak=0) at Parser.cpp:446:22
    frame #13: 0x000000010021a1d4 rb`RevClient::execute_file(filename=0x000000016fdfe330, echo=false, continue_on_error=false) at RevClient.cpp:386:38
    frame #14: 0x0000000100d6bd1c rb`RevLanguageMain::startRevLanguageEnvironment(this=0x000000016fdfe655, expressions=size=0, filename= Has Value=true , args=size=0) at RevLanguageMain.cpp:112:13
    frame #15: 0x0000000100004ba0 rb`main(argc=2, argv=0x000000016fdfee38) at main.cpp:256:21
    frame #16: 0x000000018ae93e00 dyld`start + 6992
```

As far as I can tell, the problem seems to be with the ownership of `value`, which is what I also remember @hoehna recently telling me in connection with his attempt to make `dnConstrainedTopology` work for things like `dnCDBDP`.

The problem can be circumvented by passing the same initial tree to both `dnConstrainedTopology` _and_ the base distribution (i.e., `dnFBDP` or `dnBDSTP` in this case). Passing it _just_ to the base distribution is not enough: while it won't trigger a segfault, the analysis will fail to find a viable initial state.

**To reproduce**
[This zipped directory](https://github.com/user-attachments/files/29842700/FBD_backbone_test.zip) contains a script triggering the segfault, as well as the associated data files. One absolute path has to be changed.

**Expected behavior**
Backbone-constrained FBD analyses with an initial tree passed to `dnConstrainedTopology` should reliably work without segfaults.

**Computer info**
macOS 26.5.1 (Tahoe), ARM64 architecture, RevBayes v1.4.0-dev, current tip of the [nonclock-backbone-constraints](https://github.com/revbayes/revbayes/tree/nonclock-backbone-constraints) branch (af08930) and of development (1f76654).

hoehna:
## PR Type
- 🐛 Bug Fix

## Summary
We do have an internal bug about the ownership of values in distributions. Generally, each distribution owns their value. However, for efficiency reasons (and others), there are two distributions that do not own their values: ConstrainedTopologyDistribution and MixtureAnalytical. Therefore, these distributions need to tell that the values are not deleted, which leads to segmentation faults.


src/core/distributions/AnalyticalMixtureDistribution.h
```diff
@@ -132,6 +146,14 @@ double RevBayesCore::AnalyticalMixtureDistribution<mixtureType>::computeLnProbab
...
     dirty = false;
     
+    mixtureType* val_ptr = this->value;
+    for ( size_t i=0; i<base_distributions.size(); ++i )
+    {
+        if ( val_ptr != (&base_distributions[i]->getValue()) )
+        {
+            throw RbException("Bug in AnalyticalMixture: Distributions work on different values!");
+        }
+    }
```
davidcerny, line 156: It's not clear to me whether this is just leftover test code for a case that we don't expect to ever actually happen, or a signpost for a known bug. If it's the former, let's delete it. If it's the latter, let's disable the distribution altogether until we have a proper fix.

davidcerny (commented):
I can confirm this fixes issue #1057, so even though it is marked as a draft, the PR seems essentially good to go to me. All it needs is a test, which I'm happy to add.

src/core/distributions/AnalyticalMixtureDistribution.h
davidcerny: @hoehna confirmed that this exception was included to _locate_ the bug that this branch fixes, so it should no longer be necessary. However, since ownership sharing still seems rather fragile, we use an `assert` to document the contract, and catch any potential regression in debug builds:

```cpp
assert( val_ptr == &base_distributions[i]->getValue() );
```

To make sure that we don't end up with an empty loop and an unused pointer (`val_ptr`), we wrap the diagnostic loop in `NDEBUG`.

davidcerny (changes requested):
I think I've taken this as far as I can, but there's still more work to be done.

First, although the PR does solve the segfault issue from #1057 (and also a related segfault issue with `dnMixtureAnalytical`, which we didn't know about because it has never been reported, and because our test suite doesn't cover that distribution), adding the script from #1057 as an integration test proved to be surprisingly difficult. That's because it sets up a pretty complex FBD analysis that yields different output on different platforms. These differences actually affect the `.log` and `.trees` files – there is a divergence in the move sequence, caused by the fact that at a certain point a move gets accepted on one platform but not another. This is different from the platform sensitivity seen in one of the existing FBD tests ([`mvFossilTimeSlideUniform_multiple_runs.Rev`](https://github.com/revbayes/revbayes/blob/development/tests/test_FBD/scripts/mvFossilTimeSlideUniform_multiple_runs.Rev)), which always produces the same move sequence and logs, and it's only the standard output (specifically, the sign of `NaN` values, which should really be considered a printing issue more than anything else) that ends up being inconsistent.

With a lengthy investigation assisted by Cursor Grok 4.6, I was able to pinpoint the origin of the issue. First, the fossil age slide moves use a `while`-loop to reflect values outside the window about its boundaries. With enough "bounces", we always end up inside the window eventually – but it turns out that floating-point rounding can then bring us outside the window _again_, just by a tiny bit. The loop will then bounce one more time on one machine but not the other, resulting in an extra reflection that turns the minuscule floating-point error into a very noticeable difference. Second, a lot of node age moves propose a new value using a calculation that goes like this: `(max - min) * (uniform draw) + min`. This multiply-add operation is performed with one rounding on some CPUs and two roundings on others, and this is enough for the results to diverge.

I solved these issues (again, with a lot of help from Grok) as follows: we now use a closed-form reflection formula that has been packaged into a separate function, `RbMath::Helper::reflectIntoInterval()`. This new implementation makes sure that if the final rounding leaves the value just outside `[min, max]`, we set it equal to the nearer endpoint (`min` or `max`). To get rid of the node age interpolation issue, we use the fused multiply-add (FMA) from the C++ standard library, `std::fma`, which forces `a*b + c` to be performed with a single rounding. This is enough to eliminate differences between different flavors of Unix, but alas, things are more complicated on Windows. Due to a known mingw-w64 bug, `std::fma` calls an incorrect software implementation that is not fused. If you have a machine that supports a hardware-level FMA, you can get around this, which is how I eventually made the tests pass. We actually perform a runtime check of whether we have the right CPU – if we do, we rely on the hardware FMA that gives the same answer as Unix; if not, we fall back to the default, unfused multiply-add. Compared to a compile-time check, this has the advantage of ensuring that a RevBayes binary built on a Windows machine that supports the FMA instruction will not crash if you execute it on an older machine without it.

If people want me to, I could split out these changes into a separate PR, but I figured that since they affect a test this PR is adding, I'd include them here.

--------------------------------------------------

Now, with regard to the rest of the PR, I still see several issues:

1. [ ] The `owns_value` machinery is not consistently applied. Even though `TopologyConstrainedTreeDistribution` is explicitly named as one of the two non-owners in the cover letter, it never calls `setOwnsValue(false)` on itself.

2. [ ] The only class whose `redrawValue()` function checks the flag is `ConstantRateBirthDeathProcess`. If we used other distributions as mixture components – e.g., `BirthDeathSamplingTreatmentProcess` – we would still allow a non-owner to free a shared tree.

3. [ ] `AnalyticalMixtureDistribution::setValue()` has an extra argument, so it doesn't properly override the virtual (`TypedDistribution::setValue()`). The code only compiles because it is never instantiated, but the fact that it is never instantiated means that when a mixture is clamped, its own `value` pointer is updated, but the component distributions keep pointing at the old object that just got deleted. We should instead have a proper two-argument override that first updates the mixture's pointer and then calls `setValue()` on each component. The mixture constructor already calls `setOwnsValue(false)` on each component, so there is no need for `setValue()` to try to do it again, since the components stay non-owners for their entire lifetime.

4. [ ] `setOwnsValue(false)` deletes the pointer and leaves it dangling. This currently doesn't cause problems because we immediately patch it with a call to `setValue()`, but the design is still unsafe: if anything were to run in between those two calls, we'd have a use-after-free. Even conceptually, `setOwnsValue(false)` should mean "from now on I cannot delete this object" rather than "I am deleting it now". Arguably, the `delete` belongs in the `setValue()` call, which should come first; the `setOwnsValue(false)` call should come second and only change the flag.

5. [ ] `AbstractRootedTreeDistribution::redrawValue()` now uses a more lenient default simulation condition (`MCMC` rather than `VALIDATION`), but the corresponding comment was not updated, so it's now inaccurate.

PR #1072, opened 2026-07-30
Fix an irreversible proposal in mvCollapseExpandFossilBranch

milliescient:
## PR Type
- 🐛 Bug Fix

## Summary
`mvCollapseExpandFossilBranch` starts its proposal window at the taxon's maximum age, which might not contain the parent node's age. This means a collapse move can move to a state that can't be reversed by an expand move. Using the tip age instead makes the move reversible. I regenerated the `test_FBD` and `test_BDSTP` output because the old ones are based on the irreversible proposal.

## Context
A fossil with an uncertain age isn't restricted to it's maxium age. The tip age is sampled anywhere in `[min, max]` and the parent can be any age older than the tip. With a degenerate range (`min == max == tip age`) the bounds coincide, which is why this is invisible for fossils with fixed ages.

## Testing
I added a counter to `collapseBranch` on an otherwise unmodified build of `development`, recording how often the stored parent age fell below `min_age`, and ran the existing suite. Those are the collapses expand cannot reverse.

| script | collapses | unreachable | worst gap |
|---|---|---|---|
| `FBD_dating` | 236 | 69 (29%) | 4.45 |
| `FBD_dating_ex3` | 201 | 59 (29%) | 4.65 |
| `FBDP` | 51 | 27 (53%) | 8.32 |
| `FBDP_fossil_tip_age_uncertainty` | 33 | 7 (21%) | 7.08 |
| `mvFossilTimeSlideUniform_multiple_runs` | 17 | 8 (47%) | 12.05 |
| `mvFossilTimeSlideUniform_checkpointing` | 26 | 4 (15%) | 3.53 |
| `BDSTP` | 36 | 7 (19%) | 3.98 |
| `FBD_no_degree_3_nodes` | 1812 | **0** | |
| `FBD_one_parameter_skyline` | 696 | **0** | |
| `test1` | 672 | **0** | |

The three reporting zero are the scripts whose fossils have min_age=max_age.

## Checklist
  
  The existing FBD and BDSTP tests already use this move on tip-age-uncertainty data. Their output is regenerated here.

### AI/LLM Assistance
- [x] I used an AI/LLM tool to assist with this PR: Claude Code (Opus 5) wrote the counter used for the table above and drafted the two-line fix. I checked the reversibility argument and the regenerated output myself.
      - *I am the primary author of the PR cover letter (e.g. it wasn&#x27;t AI/LLM generated).*


bredelings (approved):
This looks good, thanks for finding this!

It is scary that we had bugs in an operator like this!

PR #1073, opened 2026-07-30
Do not abort when the slice sampler cannot locate the slice

milliescient:
## PR Type
- 🐛 Bug Fix

## Summary
`search_interval` calls `std::abort()` when it can't find a non-empty slice, killing the process.

This falls back to `x0` instead. The function asserts `g(x0) >= logy` on entry, so `x0` is by definition inside the slice, and returning it simply rejects the move. The loop leaves the variable at the last rejected trial point, so `g(x0)` restores it first. A warning is issued once per run. The step limit is down from 200 to 100 per review.

The other `std::abort()` in this file is in a debug-only check and I left it alone.

## Testing
Full suite passes locally (72/72) and in CI.

## Checklist
  I've added a test that demonstrates the abort failure. I filed it under `test_FBDR` because a forthcoming update to the FBDR code will supersede it.

### AI/LLM Assistance
- [x] I used an AI/LLM tool to assist with this PR: Claude Code (Opus 5) wrote the fallback. I checked that the assertion it relies on actually holds.


bredelings:
This seems correct.  Two questions:
* did you actually hit this in practice?  I've seen it occur when the initial point is really really close to one boundary, like the interval is [0,1], and x0=1e-200.  This can be an indication that the move is operating on the wrong scale -- I changed the case I ran into to operate on the log scale.
* should we set the maximum number of iterations to 50 or 100, instead of 200, since we no longer have to worry about aborting? I suspect yes.
* should we only print this message when option "logMCMC" is set to 1 or higher? If we set the number of iterations to a lower number, the warning would occur more often.  Since this is not a bug, I think we should probably not alarm the user.

milliescient:
> * did you actually hit this in practice?  I've seen it occur when the initial point is really really close to one boundary, like the interval is [0,1], and x0=1e-200.

I ran into this very frequently when anayzing single-taxon data with the piece-wise constant FBDR. I am currently running lots of simulations for calibration/validation that can produce hundreds of single-taxon replicates, so this was blocking my validation pipeline.

But I looked into this more carefully, and I discovered this is actually related to the bug in #1074... I can't reproduce the slice sampler crash after fixing the matrix index bug. Basically, a wrong matrix index was causing the slice sampler to use a stale cached value for g(x0), so the sampled height y could sometimes land above the true maximum, yielding an empty slice.

In this case, alerting the user is correct, but the uninformative std::abort() was not very helpful to me tracking down the bug. I still think this fix is probably the robust answer, since there might be numerical edge cases that aren't as easily resolved... but I could see an argument against silencing what is probably an indication of a real error, if not one located in the slice sampler itself. Hence the warning.


> * should we set the maximum number of iterations to 50 or 100, instead of 200, since we no longer have to worry about aborting? I suspect yes.

Sure that is fine. I cut it down to 100. An empty slice never succeeds at any cap, so a lower bound just gets there cheaper.

> * should we only print this message when option "logMCMC" is set to 1 or higher? If we set the number of iterations to a lower number, the warning would occur more often.  Since this is not a bug, I think we should probably not alarm the user.

Hmm. Not sure about this one. The warning would have surfaced a real likelihood-caching bug in my case.

bredelings:
So, it used to be that the slice sampler was the main way to catch incorrect-likelihood-caching bugs :sweat_smile:.  And we had so many of these bugs that the slice sampler was hard to use.  I added the logMCMC=1/2/3 and debugMCMC=1 options mainly to introduce other ways to find the bugs!

Maybe for now we should leave the message visible without specifying any logging, but add an additional line saying that this MIGHT be an indication that there is a likelihood-caching bug?

The case where I was this error was where the starting interval was like [0,10], and the initial x was like 1.0e-200.  The slice sampling shrinks the search interval by a factor of 2 about every 4 iterations, but 200 iterations wasn't enough to shrink the search interval from [0,10] to [0,1.0e-200].  So in this case (i) the slice sampler failed to move and (ii) it was NOT a bug.

Hmm... let me think about whether we can modify the message to distinguish the its-a-bug and the not-a-bug cases.  If think if f(x0) is large enough, then it is not a bug.

bredelings:
OK, I pushed a change to show different messages depending on whether there is a real bug, or its just inefficient.  This is a quick stab at an improved error message, feel free to improve further.

src/core/moves/SliceSamplingMove.cpp
```diff
@@ -379,7 +380,42 @@ double search_interval(double x0,double& L, double& R, slice_function& g,double
...
+    // In a debug build the asserts above will fire, but in a release build we can get here.
+    // The slice as defined as the set of xs where g(x) >= y = g(x0)*U, where U~Uniform(0,1).
+    // If gx0 < logy, then g(x0) has changed since we first evaluated it, which means that
+    // (i) there is a likelihood caching bug.
+    // (ii) no matter how small we shrink the slice, we will never find anything, and
+    static int warned_err = 3;
+    if (gx0 < logy and warned_err > 0)
+    {
+        RBOUT("Warning: slice sampled failed to locate the slice because of a");
+        RBOUT("         because of a likelihood caching bug in a different move.");
```
davidcerny: It looks like the words "because of a" are accidentally repeated here

tests/test_mvSlice/scripts/slice_degenerate.Rev
```diff
@@ -0,0 +1,15 @@
+# A conditional far narrower than the slice sampler's starting window: shrinking from the
+# window down to the slice takes many more than the 200 steps the search allows. The sampler
+# should keep the current value and carry on rather than abort the process.
```
davidcerny, line 3: This is a nice test, but I wouldn't place it in a subdirectory called "test_fbdr", since the script doesn't actually include `dnFBDRP` / `dnFBDSP`. We already have a "[test_mvSlice](https://github.com/revbayes/revbayes/tree/development/tests/test_mvSlice)" subdirectory, which might be a better place for this test.

davidcerny (changes requested):

davidcerny (approved):

davidcerny:
One last question: why does the line "Future warnings like this are suppressed" only get printed after the third warning, rather than the first one?

bredelings:
I'm not completely sure I follow.  Do you mean, why does it print three warnings before suppressing future warnings?

davidcerny:
Yes, exactly. I was referring to the test script but didn't make that clear; apologies.

I could understand it if each warning was generated by a different instance of `mvSlice` working on a different parameter, but that's not the case here.

bredelings:
If the three warnings all occur in the first iteration, it tells you that this is happening very frequently.  But if they occur in iterations 10, 100, and 1000, then it tells you that it is not happening so frequently.

Also, if you enable the logMCMC options, then the location of the three messages should tell you which parameter is being altered.  It wasn't clear how to access that information from inside the move, but maybe it could be done.

We could also change 3 -> 10, which would give a clearer picture.  I presume that continual warnings would seem more like spamming the log.

:man_shrugging: I don't have strong feelings about it.  We could also disable the limiting if logMCMC or debugMCMC is turned on.

davidcerny:
Ah – that all makes sense; thanks for the explanation. The test script doesn't include a screen monitor, but a real-world analysis presumably would, and this would allow drawing the distinction you describe (the three warnings all popping up in the same iteration vs. spread over a large number of iterations).

I don't really think any changes are required, so let's merge this!

PR #1074, opened 2026-07-30
Fix and extend the matrix element moves

milliescient:
## PR Type
- 🐛 Bug Fix
- 🆕 New Feature

## Summary
`mvMatrixElementScale` and `mvMatrixElementSlide` flatten the touched element index by the row count instead of the column count, so the wrong element is reported as touched on a non-square matrix. This fixes that, adds optional `row` and `col` arguments to confine either move to one line.

## Context
The moves report which element changed through `addTouchedElementIndex` so downstream nodes can recompute them. `indexa` is a row index and `indexb` a column index, so the element index is `row * nColumns + column`, but both moves multiplied by the row count. On a square matrix the counts agree and so the math is correct but only by accident. This covers most current usage such as rate and correlation matrices. The non-square case arises for `dnFBDRP`, whose value is an `ntaxa` by 2 matrix of speciation and extinction times.

## Approach
Multiply by the column count. Separately, `row` and `col` are optional 1-based arguments on Scale and Slide. Omitting them gives the default behavior. The two `dnFBDRP` columns represent speciation and extinction times, and cannot be clamped separately. This gives us a way to sample them independently.

## Testing
Full suite passes locally (72/72) and in CI.

## Checklist
  No existing distribution uses a non-square martrix except `dnPhyloBrownianMultivariate` and `dnFBDRP`. The former seems to be dead code, and the latter will be revived in a forthcoming PR that depends on this one. So the index fix is not reproducible in the current test suite.

### AI/LLM Assistance
- [x] I used an AI/LLM tool to assist with this PR: Claude Code (Opus 5) spotted the row/column mix-up and drafted the swap move. I verified the index arithmetic and exercised the Rev interface myself.


bredelings:
This seems good overall.  I'd recommend changing the restrict_row and restrict_column entries to `std::optional<int>`, instead of using `0` as a special value indicating that they are unset.

bredelings (approved):
Thanks, looks good.

PR #1075, opened 2026-07-30
Reject unparseable numeric fields in taxon files

milliescient:
## PR Type
- 🐛 Bug Fix

## Summary
`TaxonReader` converts the numeric taxon fields with `stringstream >> value` and never checks whether it succeeded, so a garbage value silently produces a wrong number. This bug fix parses them with `strtod` instead, requiring the whole field to be consumed, and if it fails it reports the field and line number. It also now accepts `Inf`, which is a natural way to write an occurrence whose age has no upper bound.

## Context
A typo like `3.O`, with a letter O instead of a zero, extracts `3.` as 3.0 and stops at the `O`. The taxon silently gets a minimum age of 3.0 and the analysis runs on it. A field that fails to extract at all yields 0 instead, and `min_age` and `max_age` are declared uninitialized, so the value is indeterminate if nothing is ever assigned.

## Additional notes
Also adds a warning when a row gives `count = 0` together with `max_age > 0`, which says the taxon has no occurrences while also placing one in the past. Happy to split it out if you would rather keep this to the parsing change.

## Testing
Full suite passes locally (73/73). This is the change most likely to break an existing test, since it turns previously-silent parse failures into hard errors, so a clean run means nothing relied on the old behaviour.

## Checklist
  I've added `tests/test_io/scripts/taxon_reader_err.Rev`, using the `## continue-on-error` mechanism. It covers both cases from the Context section, a field that silently parsed as the wrong number and one that silently parsed as 0, plus `Inf` as an upper bound.

### AI/LLM Assistance
- [x] I used an AI/LLM tool to assist with this PR: Claude Code (Opus 5) drafted the `strtod` parsing. I checked the failure cases against the old behaviour myself.

### DIFF
--- a/src/core/io/TaxonReader.cpp
+++ b/src/core/io/TaxonReader.cpp
@@ -20,38 +20,34 @@
 using namespace RevBayesCore;
 
 
-namespace {
+/**
+ * Parse a numeric field, requiring the whole field to be consumed.
+ *
+ * Deliberately not `stringstream >> double`, which yields 0 for a field that does not parse
+ * and reports it only through failbit, and which does not recognize "Inf" -- a taxon file may
+ * use that for an occurrence whose age has no upper bound.
+ *
+ * \param[in]    s        The raw field.
+ * \param[in]    field    Field name, for the error message.
+ * \param[in]    line     1-based line number in the file, for the error message.
+ */
+double RevBayesCore::parseNumericField(const std::string &s, const std::string &field, size_t line)
+{
+    const char *begin = s.c_str();
+    char *end = NULL;
 
-    /**
-     * Parse a numeric field, requiring the whole field to be consumed.
-     *
-     * Deliberately not `stringstream >> double`, which yields 0 for a field that does not parse
-     * and reports it only through failbit, and which does not recognize "Inf" -- a taxon file may
-     * use that for an occurrence whose age has no upper bound.
-     *
-     * \param[in]    s        The raw field.
-     * \param[in]    field    Field name, for the error message.
-     * \param[in]    line     1-based line number in the file, for the error message.
-     */
-    double parseNumericField(const std::string &s, const std::string &field, size_t line)
+    double v = std::strtod(begin, &end);
+
+    // strtod skips leading whitespace; skip trailing before checking for leftovers
+    while ( *end == ' ' || *end == '\t' || *end == '\r' || *end == '\n' ) ++end;
+
+    if ( end == begin || *end != '\0' )
     {
-        const char *begin = s.c_str();
-        char *end = NULL;
-
-        double v = std::strtod(begin, &end);
-
-        // strtod skips leading whitespace; skip trailing before checking for leftovers
-        while ( *end == ' ' || *end == '\t' || *end == '\r' || *end == '\n' ) ++end;
-
-        if ( end == begin || *end != '\0' )
-        {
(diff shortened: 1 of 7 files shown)


davidcerny (commented):
Thank you for this and all your other recent contributions, @milliescient! This looks like a fairly straightforward fix, so I'd be happy to approve it.

Note that we now have a mechanism for tests that are expected to throw an error: you just need to put `## continue-on-error` in the first line of the script. (See [here](https://github.com/revbayes/revbayes/blob/development/tests/test_newick/scripts/newick_err.Rev) for an example.) It might be nice to add such a test, which silently produces an incorrect number on development but throws an error on this branch. The [I/O subdirectory](https://github.com/revbayes/revbayes/tree/development/tests/test_io) might be a natural place for it.

src/core/io/TaxonReader.cpp
```diff
@@ -16,6 +20,41 @@
...
+        if ( end == begin || *end != '\0' )
+        {
+            throw RbException() << "Could not parse \'" << s << "\' as a number in the \"" << field
+                                << "\" field on line " << line << " of the taxon definition file.";
+        }
+
+        return v;
+    }
+
+}
```
davidcerny: I noticed that AI agents often prefer to enclose helper functions in their own namespaces, as opposed to declaring them in the header file... is there any reason for this? I think declaring the function in `TaxonReader.h` might be cleaner.

bredelings: Yeah, I see this too, and it looks odd.  I think this puts the function in a namespace with a secret name so that it is only observable from within this file.  Presumably this is because the AI agents have a strong tendency to prefer "preserve current behavior", just in case something goes wrong if current behavior changes?  Even though the existing codebase has no other calls to parseNumericField...

But often if the function is worth introducing, it is worth sharing.  So putting the function in TaxonReader.h seems like a good idea to me too.

milliescient: Thank you very much guys! That's a good catch on the namespace quirk that I hadn't noticed. I can take care of that. I appreciate all the work you've put into making the PR-template for AI-assisted revisions and everything. It's very important as these tools are not leaving my workflow anytime soon. I tried to conform as closely as possible, but I'm sure there'll be an eensy bit of a learning curve. I figured a few easy PR bug fixes would be a good way to get my feet wet and reintroduce myself here. I'm working on a big overhaul/revival of the FBDR code and I want to make sure everything is squeaky clean going into that.

milliescient: I asked about the namespace and here was Claude's explanation: "An unnamed namespace gives internal linkage, which is the standard C++ idiom for a translation-unit-local helper — it keeps the symbol out of the object file and lets the compiler inline freely."

davidcerny (approved):
Hmm, I might prefer to make `parseNumericField()` a member function of `TaxonReader`, rather than let it freely float in the `RevBayesCore` namespace. Alternatively, if the idea is that classes other than `TaxonReader` might want to make use of this function too, I'd prefer to put it in a more logical place – e.g., `StringUtilities`.

Other than that, this looks good to me.

src/core/io/TaxonReader.cpp
bredelings: In other words, Claude can make identically named helper functions in multiple different files without causing a linker errors :joy:

PR #1076, opened 2026-07-31
MCMC fixes

### LINKED ISSUE #123: Feature request: Generalized distancs with dnPhyloBrownianREML

bredelings:
## Summary
This PR fixes some MCMC issues of two different types.
* in the first type, we create nonsensical trees, such as trees where a node has two sampled-ancestor tips, or where a parent is younger than a child.  These proposals are rejected, because the resulting trees have probability zero, but we really shouldn't create them in the first place.
* in the second type, the moves are not reversible.

## Context 
* In CollapseExpandFossilBranchProposal, don't collapse a node whose sibling is already a sampled ancestor.
* In FNPR and GibbsPR, don't try to attach to a branch where the parent is younger than the moved node.
* In FNPR and GibbsPr, don't prune a subtree whose sibling is a sampled-ancestor, because the reverse move requires attaching below a tip, which is not supported.
* In FossilTipTimeSlideUniform, skip the proposal if the interval has a max_age < min_age
* In NarrowExchange and NNI, don't create a node with two sampled-ancestor-tips as children.
* In NarrowExchange and NNI, when one of the two swapped nodes is a sampled-ancestor tip, set its age to the age of the parent when we propose and when we restore.
* In SubtreeScaleProposal, we need to avoid scaling a node below the age of its oldest descendant tip.

## TODO before merging
I have not regenerated the test output yet.

## Checklist
- [ ] I have added or updated tests for this change, or explained why they are not needed.
- [ ] I have added or updated documentation where necessary.


davidcerny:
Note that [`FossilSiteTimeSlideUniformProposal.cpp`](https://github.com/revbayes/revbayes/blob/development/src/core/moves/proposal/tree/FossilSiteTimeSlideUniformProposal.cpp) might also benefit from the changes made to [`FossilTipTimeSlideUniformProposal.cpp`](https://github.com/revbayes/revbayes/blob/development/src/core/moves/proposal/tree/FossilTipTimeSlideUniformProposal.cpp).

davidcerny:
Ah, so it turns out that `mvFossilSiteSlideUniform` was already throwing an exception in that situation – but a different one that didn't allow the move to be skipped:

https://github.com/revbayes/revbayes/blob/662380636aea7fba0303fb99dc5f2ca756acd53c/src/core/moves/proposal/tree/FossilSiteTimeSlideUniformProposal.cpp#L213-L216

I've modified it to unify the behavior across all three fossil tip age moves.

davidcerny (approved):
I went over these changes, and they all make sense to me. It looks like maintaining the reversibility of the `mvCollapseExpandFossilBranch` move in particular is harder than it seems, as the same concern is also addressed by PR #1072 in what is – as far as I can tell – an additional and unrelated fix.

PR #1077, opened 2026-07-31
Parse lines with more than 8192 characters

### LINKED ISSUE #162: [BUG] Error loading topological constraints above a certain number of species

davidcerny:
## PR Type
- Bug Fix

## Summary
In [`lex.yy.cpp`](https://github.com/revbayes/revbayes/blob/development/src/revlanguage/parser/lex.yy.cpp), we specify `YY_READ_BUF_SIZE` to be 8192, which currently places a hard limit on the length of single-line Rev commands we can parse. Although we can arbitrarily increase this limit, the real solution is to place no upper bound on the length of parsable lines at all – while also preserving existing functionality such as parsing commands distributed across multiple lines. This PR accomplishes this goal by converting `YY_READ_BUF_SIZE` to a chunk size cap rather than line length cap.

## Approach
When a Rev line exceeds the buffer length, we no longer let the parser append a synthetic newline character, which previously caused the lexical analyzer to think that the command ended mid-token. Instead, we return the chunk and call `YY_INPUT` again to grab the rest of the command.

## Additional notes
@bredelings noted in #162 that he had found a way to handle unlimited Rev line lengths, but that it had broken multiline commands. The solution proposed here has been carefully checked to avoid this pitfall.

## Testing
I added a new test: [`test_parser/scripts/Test_ultra_long_line.Rev`](https://github.com/revbayes/revbayes/blob/ultra-long-lines/tests/test_parser/scripts/Test_ultra_long_line.Rev).

## Relevant issues
Fixes #162.

## AI/LLM Assistance
I used Cursor Grok 4.5 to fix the issue. I personally understand the submitted changes and take responsibility for their correctness, style, and maintainability.


basantakhakurel (approved):
Looks good. Thanks!

PR #1079, opened 2026-08-05
Calculate mean and variance of continuous character data in the presence of missing data

prilau:
Currently, some member functions of `ContinuousCharacterData`, e.g., `meanValue()`, `var()`, returns nan when the character contains missing values `'?'`. I made some small modifications such that samples with missing values for the character specified are excluded in the calculation.

I wrote a test to compare the results of, among other functions, `meanValue()` and `var()` of (1) a `ContinuousCharacerData` object where some samples have missing values, and (2) a `ContinuousCharacerData` object with the same data except that samples with missing values are removed. The test can be found under `test_cont_char/`

Also, I removed `tests/output` because they are not supposed to be there, I assume.


davidcerny (commented):
This looks great – thanks for the fix! Two quick comments:

1. An alternative solution would be to provide the relevant methods with an extra boolean argument similar to R's `na.rm`. But I'm not sure whether that's necessary – in practice, would we ever _want_ these methods to return `nan`, rather than a number obtained by suppressing the missing values? If not, then the current solution certainly has the advantage of being simpler.

2. Presumably, guarding the calculations with `RbMath::isFinite()` excludes not only missing data (`?`) but also inapplicables (`-`). If so, it might be useful for the test to demonstrate this.

> Also, I removed `tests/output` because they are not supposed to be there, I assume.

Absolutely; that folder accidentally made it in with PR #1021, due to my oversight as the reviewer. Thanks for deleting it!

davidcerny (approved):

PR #1080, opened 2026-08-07
Add missing delimiter between characters

bredelings:
## PR Type
- 🐛 Bug Fix

## Summary
`writeDelimitedCharacterData( )` has a separator argument for separating the values on each line.  
If  understand correctly, the whole point of "delimited" character data is that there is indeed a delimiter between the characters.
However, it does not actually print the delimiter between the values.

## Approach
Simply print the delimiter after each character.  

## Relevant issues
Fixes #1078

## TODO before merging
Add a test.

## Checklist


basantakhakurel (approved):
This is exactly what I had but didn't have a chance to create a PR. Thanks Ben!

davidcerny (dismissed):
Let's add a quick test to `test_io`, maybe?

bredelings:
Tests added.

It turned out that this code had some bugs, and was adjacent to some other bugs:
* don't print a separator at the end of the line.  (For example, there should be no final `,`)
* the message said that an empty delimiter was treated as `\whitespace\`, but that only makes sense when SPLITTING fields for reading.  For writing, you need a specific string to print, not a regex to split fields.
* stochastic maps were using a empty delimiter (the default) and therefore printed lines like `61021.95379000NAno_change354NANA`.  The file is called `solitariness_terrestrially_corr.events.tsv`, so I made the default delimiter "\t".  I had to update the expected output.

davidcerny (approved):
The changes all look great to me, and the fact that one of our existing tests was producing unreadable output shows that they were long overdue. Thanks!

PR #1083, opened 2026-08-14
Add a Rev function to get the number of MPI processes

### LINKED ISSUE #1082: [Feature Request] Expose the number of MPI processes to the user
In [one of his comments](https://github.com/revbayes/revbayes/pull/991#issuecomment-4840709387) under the now-merged PR #991, @ms609 pointed out that it might be useful for a Rev script to be able to tell how many MPI processes it's being executed on. The original use case was to simplify a workaround for power posterior checkpointing that ultimately proved unnecessary, as it was necessitated by a bug that got fixed before the PR was merged. However, one can imagine, for example, adjusting the number of MCMC replicates according to the number of processes available:

```r
NUM_RUNS = getNumProcesses()
mymcmc = mcmc(mymodel, monitors, moves, nruns = NUM_RUNS)
```

The same logic applies to selecting the number of heated chains for MC^3, or the number of stones for a power posterior analysis. Of course, there are other considerations that affect these choices, but we can account for those too. E.g.,

```r
NUM_STONES = max( [getNumProcesses(), 32] )
pow_p = powerPosterior(mymodel, monitors, moves, "marginal.out",
                       cats=NUM_STONES, sampleFreq=50)
```

davidcerny:
## PR Type
- New Feature

## Summary
This PR adds a new function called `getNumProcesses()`, which returns the number of MPI ranks. The returned number is always equal to 1 for non-MPI builds, and to `N` when an MPI build is executed as `mpirun -np N ./rb-mpi`.

## Approach
The function is just a thin Rev wrapper around `MPI_Comm_size(MPI_COMM_WORLD, size)`.

## Testing
I _could_ add a test that simply prints 1, if people really want me to. But I think a test would be more useful when/if we bring back the option of running the test suite using an MPI build (including comparisons against MPI-specific expected output).

## Relevant issues
Closes #1082.

## AI/LLM Assistance
I used Cursor Grok 4.5 to implement this; I wrote the help file myself. I personally understand the submitted changes and take responsibility for their correctness, style, and maintainability.

### DIFF
--- a/help/md/getNumProcesses.md
+++ b/help/md/getNumProcesses.md
@@ -16,13 +16,11 @@ David Černý
     # For `mpirun -np 16 ./rb-mpi`, this prints 16:
     getNumProcesses()
     
-    # Set the number of MCMC replicates to the number of available processes,
-    # but at least to 4:
-    N_RUNS = max( [getNumProcesses(), 4] )
+    # Record the process count for reproducibility bookkeeping
+    print("Running on " + getNumProcesses() + " process(es).")
     
-    # Create a simple model (unclamped), and run the number of replicates
-    # specified above:
-    x ~ dnExp(10)
-    mymcmc = mcmc( model(x), [mvSlide(x, delta=0.1)], [mnScreen()], nruns=N_RUNS )
-    mymcmc.run(100)
+    # Guard-check the launch configuration
+    if (getNumProcesses() != 32) {
+        stop("Expected 32 processes, got " + getNumProcesses() + ".")
+    }
 ## references
--- a/help/md/getNumProcesses.md
+++ b/help/md/getNumProcesses.md
@@ -1,0 +1,28 @@
+## name
+getNumProcesses
+## title
+Get the number of MPI processes
+## description
+Returns the number of processes used by the current RevBayes session.
+## details
+In an MPI-enabled build (i.e., `rb-mpi`, launched as `mpirun -np N ./rb-mpi`),
+the function returns `N` (i.e., the size of the `MPI_COMM_WORLD` communicator).
+In a non-MPI build, it always returns 1. The value is the number of MPI ranks,
+not necessarily the number of hardware CPU cores.
+## authors
+David Černý
+## see_also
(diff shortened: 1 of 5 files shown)


basantakhakurel (commented):
Thanks for adding this. The implementation itself looks good to me. 

One thing worth flagging before we merge: this PR is supposed to close #1082, but I don't think we reached a consensus there. 
@bredelings's concern is valid and suggests using CLI args (`--nruns=8`) as an alternative. 
I think it matters in this specific case, as the new help file's featured example is exactly the pattern he was objecting to:

```
N_RUNS = max( [getNumProcesses(), 4] )
mymcmc = mcmc( model(x), [mvSlide(x, delta=0.1)], [mnScreen()], nruns=N_RUNS )
```
I think driving nruns itself from `getNumProcesses()` can go slightly astray, the same script run with `-np 4` vs `-np 32` doesn't just finish faster/slower or get redistributed differently, it runs a different number of replicates and produces different combined posterior and different convergence diagnostics. 
That, in my view, can be a real hazard to reproducibility (same seed, same script, but different cluster allocations can produce different statistical output), and the help file cites that as the primary example. 

Two suggestions, either of which I think would resolve this cleanly:
- Lead the example with one of the two use cases from the issue thread (recording the process count for reproducibility bookkeeping):
```
print("Running on " + getNumProcesses() + " process(es).")
```
or guard-checking the launch config (per @ms609's suggestion):
```
if (getNumProcesses() != 32) {
    print("ERROR: expected 32 processes, got " + getNumProcesses())
}
```

- If you want to keep the `nruns=` example, add a one-line caveat noting that it makes the analysis's statistical output depend on how many processes it's launched with.

With those changes, I'd be glad to approve.
Also, happy to be told the issue thread's silence means implicit agreement, thought to flag it out.

davidcerny:
That's a fair point. I've removed the objectionable pattern from the example section.

basantakhakurel:
lgtm, thanks!

basantakhakurel (approved):

Issue #1013, opened 2026-05-05
[Help Wanted] Quadrillion hour model runs when using MCMCMC


joelh251:
I'm trying to run some FBD models using an MCMCMC. These runs are of about 130 taxa. I'm getting time estimates in the quadrillions of hours and my generation time appears to increase exponentially. This issue is reproduced on a Windows 11 laptop and a Linux computing cluster both running RevBayes version 1.3.2.

```
taxa <- readTaxonData("combined_evidence_model/data/reduced_taxa_data.tsv")

moves = VectorMoves()

#Speciation and extinction rates
lambda <- ln(100)
speciation_rate ~ dnExponential(lambda)
extinction_rate ~ dnExponential(lambda)

moves.append( mvScale(extinction_rate, lambda = 0.01, weight=1) )
moves.append( mvScale(extinction_rate, lambda = 0.1, weight=1) )
moves.append( mvScale(extinction_rate, lambda = 1, weight=1) )

moves.append( mvScale(extinction_rate, lambda = 0.01, weight=1) )
moves.append( mvScale(extinction_rate, lambda = 0.1, weight=1) )
moves.append( mvScale(extinction_rate, lambda = 1, weight=1) )

diversification := speciation_rate - extinction_rate
turnover := extinction_rate/speciation_rate

#Probability of sampling extant taxa (rho)
rho <- 1.0

#Fossil sampling rate (psi)
psi ~ dnExponential(1000000) 

moves.append( mvScale(psi, lambda=0.01, weight=1) )
moves.append( mvScale(psi, lambda=0.1,  weight=1) )
moves.append( mvScale(psi, lambda=1,    weight=1) )

#Origin time
origin_time ~ dnUnif(2876, 3000)

moves.append( mvSlide(origin_time, delta=0.01, weight=5.0) )
moves.append( mvSlide(origin_time, delta=0.1,  weight=5.0) )
moves.append( mvSlide(origin_time, delta=1,    weight=5.0) )

#Define FBD distribution
fbd_tree ~ dnFBDP(origin=origin_time, lambda=speciation_rate, mu=extinction_rate, psi=psi, rho=rho, taxa=taxa)

moves.append( mvFNPR(fbd_tree, weight=15.0) )
moves.append( mvCollapseExpandFossilBranch(fbd_tree, origin_time, weight=6.0) )

moves.append( mvNodeTimeSlideUniform(fbd_tree, weight=40.0) )
[...]

PR #1085, opened 2026-08-19
Fix time-to-completion estimates for analyses with stopping rules

### LINKED ISSUE #542: [BUG] ETA displays enormous number

davidcerny:
## PR Type
-  Bug Fix

## Summary, Approach, & Additional Notes

As a brief review, we currently have six stopping rules that we can use to determine when to terminate an MCMC(MC) run: `srMaxIteration`, `srMaxTime`, and four convergence-based rules. All of these can be freely combined with one another, as well as with the `generations` argument of the `.run()` method.

On two different occasions (#542, #1013), users have tried to use a stopping-rule vector containing `srMaxIteration` plus one other rule, and got absurdly large time-to-completion estimates (ETA) in the screen monitor as a result.

I cannot reproduce this problem with my setup – I'm getting `00:00:00` instead, which goes to show that this is yet another problem where Clang manages to optimize away a division by zero (see #794, #1049). However, the exact failure mode is not of interest here. The larger problem is that we can get reasonable ETA values by specifying the `generations` argument but not by supplying an `srMaxIteration` stopping rule, even though these should be strictly equivalent. In fact, `generations` even gets converted to `srMaxIteration` under the hood:

https://github.com/revbayes/revbayes/blob/ace3b793a4564e478c819e83af245a653f0eda5c/src/revlanguage/analysis/mcmc/RlMonteCarloAnalysis.cpp#L121-L126

This problem also made me think about how we should estimate the ETA in other scenarios where `generations` is not set – a problem that basically reduces to exposing the information from the stopping rules to the analysis, which can then pass it on to its monitors. I ended up implementing the following system:

- `srMaxIteration` (and possibly some convergence rules) specified: we extract its value and set `kIterations`  to it, which renders it equivalent to `generations` for all intents and purposes.

- `srMaxTime` (and possibly some convergence rules) specified: we extract its value and pass it to the monitors via a new argument added to the latter. The ETA is then obtained simply by subtracting the elapsed time from the time specified by the stopping rule, exactly as one would expect.

- `srMaxIteration` and `srMaxTime` both specified (possibly along with some convergence rules): the screen monitor compares the two time-to-completion estimates, and prints the smaller one to the ETA column. E.g., if your analysis takes 2 hours to complete 50,000 generations, and your stopping-rule vector looks like this:

    ```r
    stopping_rules = [ srMaxIteration(50000), srMaxTime(5, "hours") ]
    ```
    
    then your ETA will be about 2 hours, but if your stopping-rule vector instead looks like this:
    
    ```r
    stopping_rules = [ srMaxIteration(50000), srMaxTime(1, "hours") ]
    ```
    
    your ETA will be 1 hour.
    
- Only convergence rules specified: we just print `??:??:??` in the ETA column, and explicitly alert the user to the fact that no estimates are available. There are two problems with estimating a time to completion in this scenario. First, we'd have to evaluate the convergence rules not just at their own `frequency`, but also at the `printgen` frequency of the screen monitor. Second, while we could naïvely estimate the time needed to attain a specified ESS target (if it took time _t_ to attain an ESS of _x_, then reaching the target _y_ will require _ty_ / _x_), it's much less clear what the calculation should look like for the other rules (PSRF, Geweke, stationarity).

## Testing
Unfortunately, the ETA values produced by `mnScreen` are non-reproducible, so I couldn't add a test for these changes.

## Relevant issues
Fixes #542.

## AI/LLM Assistance
- I used Cursor Grok 4.5 for this PR. I personally understand the submitted changes and take responsibility for their correctness, style, and maintainability.

### DIFF
--- a/src/core/analysis/HillClimber.cpp
+++ b/src/core/analysis/HillClimber.cpp
@@ -652,7 +652,7 @@ void HillClimber::setScheduleType(const std::string &s)
 /**
  * Start the monitors which will open the output streams.
  */
-void HillClimber::startMonitors( size_t num_cycles, bool reopen )
+void HillClimber::startMonitors( size_t num_cycles, bool reopen, double max_seconds )
 {
 
     // Open the output file and print headers
@@ -660,7 +660,7 @@ void HillClimber::startMonitors( size_t num_cycles, bool reopen )
     {
 
         // reset the monitor
-        monitors[i].reset( num_cycles );
+        monitors[i].reset( num_cycles, max_seconds );
 
         // if this chain is active, print the header
         if ( process_active == true )
--- a/src/core/analysis/HillClimber.h
+++ b/src/core/analysis/HillClimber.h
@@ -25,9 +25,9 @@ namespace RevBayesCore {
     public:
         HillClimber(const Model &m, const RbVector<Move> &moves, const RbVector<Monitor> &mons);
         HillClimber(const HillClimber &m);
-        virtual                                            ~HillClimber(void);                                                                             //!< Virtual destructor
+        virtual                                            ~HillClimber(void);                                                                      //!< Virtual destructor
         
-        HillClimber&                                        operator=(const HillClimber &m);                                                               //!< Overloaded assignment operator
+        HillClimber&                                        operator=(const HillClimber &m);                                                        //!< Overloaded assignment operator
         
         // public methods
 //        void                                                addFileMonitorExtension(const std::string &s, bool dir);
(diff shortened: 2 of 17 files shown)


ms609 (approved):
Makes sense to me.  It feels like it'd be nice to include a rough ETA based on an ESS stopping rule, even if the linear model is imprecise.  But then perhaps the complexities outweigh the pros: I can imagine a user being confused/disheartened when ESS shoots downwards (e.g. where a run discovers a new mode).  A `??:??:??` is clear and consistent; falling back to the maxIter-based ETA when an ESS stopping rule is also specified seems more useful than just printing `?????` when a convergence rule exists.  

Whatever the verdict on the above – fixing estimation from gens/time is useful, and generating an ETA based on rate towards convergence could happily wait until a future PR, so I think this could be merged now.

davidcerny:
> It feels like it'd be nice to include a rough ETA based on an ESS stopping rule, even if the linear model is imprecise. But then perhaps the complexities outweigh the pros: I can imagine a user being confused/disheartened when ESS shoots downwards (e.g. where a run discovers a new mode).

Yes, certainly, that's one potential complication. It would also require a lot of complexity even on the purely technical side of things. Each convergence rule would need to have a new function added to it to estimate the progress fraction, and we'd need to open a new path into the screen monitor (separate from the existing `numCycles` and `maxSeconds` arguments) so that it can figure out the remaining time from that fraction.

PR #1086, opened 2026-08-21
Fix TreeScale proposal

bredelings:
## PR Type
- 🐛 Bug Fix

## Summary
The TreeScale proposal had a bug in swapNodeInternal where it made the `vector_variable` node point to the `rootAge`.  This seems to affect only `FBD_one_parameter_skyline.Rev`.

I've also included a small change so that we continue to run the integration tests if one of the tutorial tests failed.  Stopping the tests early made it hard to see the full extent of failures if one of the tutorial tests failed.


davidcerny (approved):
> I've also included a small change so that we continue to run the integration tests if one of the tutorial tests failed.

This was actually recently requested by @hoehna as well. Thanks!

PR #1088, opened 2026-08-27
Extract non-template DAG node implementation bases

bredelings:
## PR Type
- 💅 Refactor
- 🚦 Infrastructure

## Summary
Move member functions that do not depend on `valueType` from `DeterministicNode<valueType>` -> `DeterministicNodeBase`, from `StochasticNode<valueType>` to StochasticNodeBase, and from `DynamicNode<valueType>` to DynamicNodeBase.  A lot of these functions work on the graph, but are independent of `valueType`.  This has two immediate benefits:
* we can call these functions in places where we have no idea what `valueType` is.
* the body of these functions can be moved out of the header file and into a `*.cpp` file.  So, when we change them, it doesn't recompile all of RevBayes.

This change should make it faster to compile the branch that refactors and untangles invalidation (touch) and snapshotting (touch, keep,restore).
I also have a longer-term change in mind that further separates value-access from graph-walking.

## Context 
Currently the `DeterministicNode<valueType>` and `StochasticNode<valueType>`  depend on the value type `valueType` because they hold a pointer to a value.  This causes trouble because we need to know `valueType` in order to dynamic_cast from `DagNode*` to `StochasticNode<valueType>*` and call them.  So, in order to allow clamping and unclamping nodes, I had already added a StochasticNodeBase that does *not* depend on `valueType`.  

## Approach
For each `XNode<valueType>` class, add a XNodeBase class and inherit from that.  The XNodeBase class currently does NOT inherit from DagNode.  However, future changes could allow it to do so.

## Checklist
This is a basic infrastructure change, so I think compiling and passing the existing test suite is fine.  No new permanent tests are needed, because every existing test already exercises the DAG.

### AI/LLM Assistance
- [x] I used an AI/LLM tool to assist with this PR

### DIFF
--- a/src/core/dag/DynamicNodeBase.cpp
+++ b/src/core/dag/DynamicNodeBase.cpp
@@ -1,0 +1,143 @@
+#include "DynamicNodeBase.h"
+
+#include "DagNode.h"
+#include "DagNodeMap.h"
+#include "RbException.h"
+
+#include <iostream>
+
+using namespace RevBayesCore;
+
+/** Initialize a new dynamic node in the touched state. */
+DynamicNodeBase::DynamicNodeBase(void) :
+    touched( true )
+{
+}
+
+
+/** Copies intentionally begin touched, matching the historical dynamic-node copy behavior. */
+DynamicNodeBase::DynamicNodeBase(const DynamicNodeBase & /*n*/) :
+    touched( true )
+{
+}
+
+
+/** Clone the entire graph: clone children, swap parents. */
+DagNode* DynamicNodeBase::cloneDAG(const DagNode &owner, DagNodeMap &new_nodes, std::map<std::string, const DagNode*> &names) const
+{
+    // Return our clone if we have already been cloned
+    if ( new_nodes.find( &owner ) != new_nodes.end() )
+    {
+        return new_nodes[ &owner ];
+    }
+
+    // just for self checking purposes we keep track of the names for the variables we already cloned
+    if ( owner.getName() != "" )
+    {
+        // check if we already added a variable with this name
+        std::map<std::string, const DagNode* >::const_iterator n = names.find( owner.getName() );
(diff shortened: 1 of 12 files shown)


src/core/dag/DynamicNodeBase.cpp
```diff
@@ -0,0 +1,143 @@
...
+            // no, we haven't cloned a variable with this name before
+            names[ owner.getName() ] = &owner;
+        }
+        else
+        {
+#ifdef DEBUG_SEBASTIAN
+            const DagNode *orgCopy = n->second;
+            std::cerr << "Ptr to org:\t" << orgCopy << "\t\t --- \t\t Ptr to desc:\t" << &owner << std::endl;
+            std::cerr << "Cloning a DAG node with name '" << owner.getName() << "' again, doh! Please tell this to Sebastian because it's most likely a bug." << std::endl;
+#endif
```
davidcerny, line 50: I really wonder if we need to retain this, but I also don't think it's the responsibility of this PR to get rid of it. Just thinking out loud, I guess...

src/core/dag/StochasticNodeBase.h
```diff
@@ -0,0 +1,82 @@
...
+         * The caller needs to be responsible enough to know that we will assume
+         * that the current value is the observed value.
+         * We could use instead as well a call: clamp( getValue() );
+         */
+        void                                               setClamped(bool tf);
+
+        /**
+         * Unclamp this node. If I understand this correctly, we
+         * can just set the clamped flag to false if we keep the
+         * current value. We do not need to tell anyone that we
```
davidcerny, line 36: I know this info can always be extracted using blame, but it would make it more convenient for people reading the file to clarify who the "I" is here. Maybe add something like "Ben Redelings, 2026-08-27" underneath the comment?

davidcerny (approved):

Issue #953, opened 2026-02-28
Compiling v1.3.2 with boost 1.90.0


plewis:
RevBayes 1.3.2 did not compile for me on Red Hat Enterprise Linux release 8.10 (Ootpa) using gcc 15.1.0 and Boost 1.90.0 due to a bug in the Boost Dynamic Bitset library. This bug is easy to fix, however, as detailed in patch 101 (https://github.com/boostorg/dynamic_bitset/pull/101/files): you just need to add this friend declaration

    template< typename Iter >
    friend BOOST_DYNAMIC_BITSET_CONSTEXPR20 difference_type operator-( const bit_iterator_base< Iter > & lhs, const bit_iterator_base< Iter > & rhs );

at line 1443 of boost_1_90_0/boost/dynamic_bitset/dynamic_bitset.hpp. This bug has already been fixed and thus no patch will be necessary in later versions of boost.

davidcerny:
Indeed – see also PR #924. Many thanks for not only reporting the issue, but also showing how it can be circumvented!

bredelings:
What do you think about leaving this open until boost 1.91 is out so that anyone else with this problem can find this solution?  I don't know if people search CLOSED issues, but if so then I suppose its cleaner to leave it closed, so 🤷‍♂️ 

The alternative solution from #924 is to use boost 1.89:

```
            curl -O -L https://archives.boost.io/release/1.89.0/source/boost_1_89_0.tar.gz
            tar -xzvf boost_1_89_0.tar.gz
            cd boost_1_89_0
            ./bootstrap.sh --with-libraries=atomic,chrono,system,regex,thread,date_time,math,serialization --prefix=../installed-boost-1.89>
            ./b2 link=static install
            echo -e "\n    BOOST root is at $(cd ../installed-boost-1.89.0; pwd)\n"
            export BOOST_ROOT=$(cd ../installed-boost-1.89.0; pwd)
            export CMAKE_PREFIX_PATH=$(cd ../installed-boost-1.89.0; pwd)
```

davidcerny:
That's fair – happy to re-open.

bredelings:
Debian seems to have moved from boost-1.83 to boost 1.90 🤦‍♂️.  I will ask them to apply the patch.  However, the fact that boost does not do bugfix releases could make this challenging.

bredelings:
Paul's solution works for me.  I also tested boost 1.91-beta1 and it works without introducing any more problems.

[...]

PR #1093, opened 2026-09-03
No special-casing of boost installation for CI on macOS

davidcerny:
## PR Type
- Infrastructure

Until recently, the latest version of the boost library available from the homebrew package manager was v1.90, which is known to be problematic (see #953). We therefore special-cased the [`build.yml`](https://github.com/revbayes/revbayes/blob/development/.github/workflows/build.yml) workflow script to install the older v1.89. However, since homebrew [now ships](https://github.com/Homebrew/homebrew-core/pull/298472) v1.92, we can just go for the latest version available. I also took the opportunity and changed the [`release.yml`](https://github.com/revbayes/revbayes/blob/development/.github/workflows/release.yml) script to be more in line with the build script: in particular, it no longer contains a hardcoded requirement for one specific version of boost (v1.88).

This PR effectively undoes PR #924.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -160,20 +160,6 @@ jobs:
       if: runner.os == 'macOS'
       run: |
           brew install boost ccache coreutils meson
-          
-          # Boost 1.90 has a bug causing a compilation error.
-          # We'll use a static boost until the version in homebrew is fixed.
-          ( cd
-            curl -O -L https://archives.boost.io/release/1.89.0/source/boost_1_89_0.tar.gz
-            tar -xzvf boost_1_89_0.tar.gz
-            cd boost_1_89_0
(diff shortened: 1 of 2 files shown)


bredelings:
I think we actually compile our own boost in the release script for a reason -- by statically linking the boost libs, users can still run `rb` without installing homebrew for the boost dylibs.

I guess in theory homebrew should ship static libs too.  Possibly I wrote `This is a hack` because homebrew didn't include static libs?  I can't remember.

So maybe leave the release action unchanged, except (i) optionally update to 1.92 and (ii) change `# This is a hack.` to `# Compile static boost libs so that released executables don't depend on homebrew`.

davidcerny:
@bredelings Thank you for the clarification; that makes sense! Changed as suggested.

basantakhakurel (approved):
lgtm

PR #1098, opened 2026-09-21
Standard output formatting for matrix types

### LINKED ISSUE #1094: [BUG] MatrixReal and MatrixRealSymmetric cannot be printed with `print()` function
**Describe the bug**
When using the `print()` function on a variable of type `MatrixRealSymmetric` and `MatrixReal`, it prints out unstructured output instead of the nicely formatted matrix you get when auto-printing a variable.

**To Reproduce**
```
a <- matrix([[1,2,3],[4,5,6],[7,8,9]])
a
print(a)

b <- fnDecompVarCovar([1.0,1.0,1.0], matrix([[1.0,0.4,0.2],[0.4,1.0,0.3],[0.2,0.3,1.0]]))
b
print(b)
```

**Expected behavior**
`print(a)` should show the matrix in a readable, delimited form just like auto-printing (just typing `a`) does. 

**Computer info**
RevBayes version 1.4.2-preview
OS: Linux x86_64 (Ubuntu 26.10)

basantakhakurel:
## PR Type
- 🐛 Bug Fix

## Summary
The `print()` function used to print `MatrixReal` and `MatrixRealSymmetric` values as a string of digits in the standard output instead of a nicely formatted output. 
The matrices mentioned above (and two cladogenetic matrix types) now use the same format as auto-print. 

## Approach

`printValue` in `RlMatrixReal` (inherited by `MatrixRealSymmetric` and `MatrixRealPos`) and in the two cladogenetic matrices now always calls `printForUser`, so the format is the same as auto-print. 

## Additional notes

- While looking for a fix for this, I also noticed that the non-negativity check in `fnCladogeneticSpeciationRateMatrix` had a bug in the `tmp_events.size()` loop, which I corrected.
- In the integration test directory, I had previously added `test_Nexus` as its own folder without noticing that `test_io` existed (see PR #975), which I now moved to `test_io`.

## Testing

Added a test (`tests/test_io/print_matrix.Rev`) which checks for plain vector printing, `MatrixReal`, `MatrixRealSymmetric`, `write()` to a file, and cladogenetic speciation rate and probability matrices.

## Relevant issues

Fixes #1094 

## Checklist


src/revlanguage/datatypes/math/RlCladogeneticProbabilityMatrix.cpp
```diff
@@ -89,11 +89,19 @@ const TypeSpec& CladogeneticProbabilityMatrix::getClassTypeSpec(void) {
...
     static TypeSpec type_spec = getClassTypeSpec();
     return type_spec;
 }
 
+
+/* Print the value of the matrix. */
+void CladogeneticProbabilityMatrix::printValue(std::ostream& o, bool user) const
+{
+    this->getDagNode()->printValue( o, ",", -1, true, true, true );
+}
```
davidcerny: This function, as well as the corresponding functions in `RlCladogeneticSpeciationRateMatrix` and `RlMatrixReal`, takes a `user` argument but doesn't do anything with it – it doesn't propagate it into the `->getDagNode()->printValue()` call inside its body. There are good reasons for this, of course, but it might be worth indicating that it's intentional, just so that someone in the future doesn't attempt to "fix" it. A comment would be one way to do that; changing the signature to read `bool \* user *\` would be another.

tests/test_io/scripts/print_matrix.Rev
```diff
@@ -0,0 +1,45 @@
+################################################################################
+#
+# RevBayes Test-Script: print()/write() formatting for matrices and
+# cladogenetic event matrices
```
davidcerny: Sounds a bit redundant 😉

tests/test_io/scripts/print_matrix.Rev
```diff
@@ -0,0 +1,45 @@
...
+# 
+# authors: Basanta Khakurel
+################################################################################
+
+seed(200456)
```
davidcerny: You don't really need this

tests/test_io/scripts/print_matrix.Rev
```diff
@@ -0,0 +1,45 @@
...
+
+# Numeric Vector
+v <- [1.0, 2.0, 3.0]
+v
+print(v)
+
+# MatrixReal
```
davidcerny: This is extremely nitpicky, since the output of integration tests doesn't really _have to_ be human-readable after all, but it would be easier to figure out what's going on in the output file if these lines indicating which type is being printed were printed out rather than only included as comments. (A liberal use of newlines could help as well.)

davidcerny:
Looks good! I'd recommend rebasing the first 5 commits on development; after that I'd be happy to approve.

davidcerny (approved):

prilau (approved):

